PR #2, opened 2016-08-14
Add tests for measuring distance from random for key exchange output and PRNG

dstebila:


smashra:
I think in kex_test_correctness_wrapper(...) everywhere there is OQS_KEX_new should be changed to new_method.

dstebila:
Thanks Shravan, fixed and merged.

PR #3, opened 2016-08-24
Simplify Makefile and allow for better incremental builds.

aparent:
Also add -std=gnu11 option since some older compilers default to c99 which causes compilation to fail.

I also turned on some warnings.


Makefile
```diff
@@ -23,29 +23,36 @@ links:
...
-	$(CC) -c $(CFLAGS) $(INCLUDES) src/kex/kex.c -o objs/kex.o
-	$(AR) liboqs.a objs/kex.o
-
-lib:
+# CHACHA
+
+CHACHA_OBJS :=  $(addprefix objs/rand_urandom_chacha20/, \
+	            rand_urandom_chacha20.o chacha20.o)
+
+$(CHACHAOBJS): src/rand_urandom_chacha20/rand_urandom_chacha20.h
```
dstebila: I'm not too familiar with Makefile syntax.  Do you intentionally have CHACHA_OBJS in one spot and CHACHAOBJS in another?

aparent:
Good catch I fixed the commit.  

That line is declaring the dependencies of the chacha object files.  Unfortunately make assumes variables which are not set are empty and doesn't respond with an error or warning :(.

PR #4, opened 2016-08-25
Fixes a memory leak.

aparent:
Occurs when memory allocation fails. Also using calloc instead of malloc
since it zeros allocated memory which ensures that freeing all allocated
array indices will not fail.


aparent:
Actually is there a reason these are allocated as pointers using malloc rather then just defining:

```
struct oqs_kex_rlwe_bcns15_fft_ctx {
    uint32_t **x1;
    uint32_t **y1;
    uint32_t **z1;
    uint32_t *t1;
};
```

as

```
struct oqs_kex_rlwe_bcns15_fft_ctx {
    uint32_t x1[64][64];
    uint32_t y1[64][64];
    uint32_t z1[64][64];
    uint32_t t1[64];
};
```

Do they ever have different sizes?

dstebila:
Good point. For BCNS15 they will always have the same size.

aparent:
Ok will close this pull request fix the issue that way.

PR #6, opened 2016-09-14
Add named parameters to OQS_KEX constructor.

dstebila:
This will allow key exchange modules to support a variety of security levels using the same interface, specified via an optional named_parameters argument.


smashra:
The changes look good to me.

PR #14, opened 2016-09-29
Signature API

tlepoint:
**== NOT READY TO MERGE ==**

I wanted to start the discussion on the signature API. 

I used a structure very similar to the key exchange structure (kex.h). Could you have a look and make comments / edits (users with write access to this directory can modify this branch) on this structure?

One of the doubt I have is about `priv` and (the absence of) `priv_len`. I am not sure why `priv` has a different type `void *` than the public key in the kex; do we want the same here?

Other thing about the message and hashing: do we want to make anything explicit on that matter?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,12 @@ CURL=curl
 RANLIB=ranlib
 LN=ln -s
 
-DEFAULTS= -std=gnu11 -Wpedantic -Wall -Wextra -DOQS_RAND_DEFAULT_URANDOM_CHACHA20 -DOQS_KEX_DEFAULT_BCNS15
-CFLAGS=$(DEFAULTS) -DCONSTANT_TIME
+# Defaults algorithms
+DEFAULTS_ALGORITHMS= -DOQS_RAND_DEFAULT_URANDOM_CHACHA20 -DOQS_KEX_DEFAULT_BCNS15
+# Defaults flags
+DEFAULTS= -std=gnu11 -Wpedantic -Wall -Wextra
+
+CFLAGS=$(DEFAULTS) $(DEFAULTS_ALGORITHMS) -DCONSTANT_TIME
 LDFLAGS=-lm
 INCLUDES=-Iinclude
 
@@ -22,6 +26,7 @@ links:
 	$(LN) ../../src/kex_rlwe_bcns15/kex_rlwe_bcns15.h include/oqs
 	$(LN) ../../src/rand/rand.h include/oqs
 	$(LN) ../../src/rand_urandom_chacha20/rand_urandom_chacha20.h include/oqs
+	$(LN) ../../src/sign/sign.h include/oqs
 
 # RAND_URANDOM_CHACHA
 
@@ -45,9 +50,13 @@ $(KEX_RLWE_BCNS15_OBJS): $(KEX_RLWE_BCNS15_HEADERS)
 
 objs/kex/kex.o: src/kex/kex.h
 
+# SIGN
+
+objs/sign/sign.o: src/sign/sign.h
+
 # LIB
 
-lib: $(RAND_URANDOM_CHACHA_OBJS) $(KEX_RLWE_BCNS15_OBJS) objs/rand/rand.o objs/kex/kex.o
+lib: $(RAND_URANDOM_CHACHA_OBJS) $(KEX_RLWE_BCNS15_OBJS) objs/rand/rand.o objs/kex/kex.o # objs/sign/sign.o
 	rm -f liboqs.a
 	$(AR) liboqs.a $^
 	$(RANLIB) liboqs.a
--- a/src/sign/sign.c
+++ b/src/sign/sign.c
@@ -1,0 +1,50 @@
+#include <oqs/rand.h>
+#include <oqs/sign.h>
+
+#if defined(OQS_SIGN_DEFAULT_BLISSB)
+#include <oqs/kex_blissb.h>
+#endif
+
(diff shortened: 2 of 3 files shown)


dstebila:
Now that we have a contribution of a signature scheme #79, we'll need to revisit the development of our signature API.

dstebila:
I was talking with Dan Bernstein at Real World Crypto, and he had some suggestions about our KEX API, some of which carry forward into the design of a signature API.

Some of his comments (my wording) were:

1. the API should not require malloc's (in other words, callers pass in byte arrays of appropriate size, rather than the callee allocating it)
2. make the API as close to Supercop / NIST as possible to make it easier for people to submit
3. having the OQS_KEX or OQS_SIGN data structure is an extra hurdle for people wanting to submit
4. avoid passing the OQS_KEX or OQS_SIGN data structures down multiple levels of the stack
5. every input should be byte strings (i.e., don't use void * for private keys)
6. don't pass a random number generator down, instead rely on a global random number generator function like random_bytes
7. don't have functions representing "families of schemes" parameterized by different values, instead have each function correspond to one scheme

My opinions:

1. okay
2. okay
3. I would like to have the OQS_KEX or OQS_SIGN data structure for a few reasons: still maintaining a generic high level API that is parameterized by the scheme, and to include metadata such as key sizes and security levels
4. okay
5. For ephemeral/passive key exchange where the private key is only used once, I think void * is okay, but for schemes with long-term keys, each scheme would have to provide an explicit serialization/deserialization that the caller has to make use of; so I think for signatures byte strings is simpler (and matches Supercop / NIST)
6. It's not clear to me how to do transforms based on determinism for KEMs if you don't pass an RNG down.  But then again it's also not clear to me how to do such transforms even if you do pass an RNG down.  Thoughts?
7. Probably would simplify the API.

I'm going to try writing a signature API and post it as a comment here shortly.  Hope to get feedback from Tancrede and Christian and anyone else who wants to chime in.

dstebila:
Here's a candidate API and example of how to instantiate one scheme using it:

~~~C
#include <stddef.h>
#include <stdint.h>

#include <oqs/rand.h>

typedef struct OQS_SIGN {
	char *method_name;
	uint16_t estimated_classical_security;
	uint16_t estimated_quantum_security;
	size_t priv_len;
	size_t pub_len;
	size_t sig_len;
	int (*keygen)(OQS_RAND *rand, uint8_t *priv, uint8_t *pub);
	int (*sign)(OQS_RAND *rand, const uint8_t *priv, const uint8_t *msg, const size_t msg_len, uint8_t *sig, size_t *sig_len);
	int (*verify)(OQS_RAND *rand, const uint8_t *pub, const uint8_t *msg, const size_t msg_len, const uint8_t *sig, const size_t sig_len);
} OQS_SIGN;

int OQS_SIGN_douglas_awesome_keygen(OQS_RAND *rand, uint8_t *priv, uint8_t *pub) {
	return 1;
}
int OQS_SIGN_douglas_awesome_sign(OQS_RAND *rand, const uint8_t *priv, const uint8_t *msg, const size_t msg_len, uint8_t *sig, size_t *sig_len) {
	return 1;
}
int OQS_SIGN_douglas_awesome_verify(OQS_RAND *rand, const uint8_t *pub, const uint8_t *msg, const size_t msg_len, const uint8_t *sig, const size_t sig_len) {
	return 1;
}
OQS_SIGN OQS_SIGN_douglas_awesome = {
	"Douglas' signature scheme, awesome security level",
	512, // estimated_classical_security
	256, // estimated_quantum_security
	1024, // private key length
	2048, // public key length
	3072, // max signature length
	&OQS_SIGN_douglas_awesome_keygen,
	&OQS_SIGN_douglas_awesome_sign,
	&OQS_SIGN_douglas_awesome_verify
};

int OQS_SIGN_keygen(const OQS_SIGN *scheme, OQS_RAND *rand, uint8_t *priv, uint8_t *pub) {
	return scheme->keygen(rand, priv, pub);
}
int OQS_SIGN_sign(const OQS_SIGN *scheme, OQS_RAND *rand, const uint8_t *priv, const uint8_t *msg, const size_t msg_len, uint8_t *sig, size_t *sig_len) {
	return scheme->sign(rand, priv, msg, msg_len, sig, sig_len);
}
int OQS_SIGN_verify(const OQS_SIGN *scheme, OQS_RAND *rand, const uint8_t *pub, const uint8_t *msg, const size_t msg_len, const uint8_t *sig, const size_t sig_len) {
	return scheme->verify(rand, pub, msg, msg_len, sig, sig_len);
}

int main() {
	return 0;
}
~~~

dstebila:
This is still different from Supercop (http://bench.cr.yp.to/call-sign.html) and NIST (http://csrc.nist.gov/groups/ST/post-quantum-crypto/documents/example-files/api-notes.pdf) in several ways:

1. Supercop/NIST use the term "open" rather than "verify".
2. Our lower level algorithms (e.g., OQS_SIGN_douglas_awesome_keygen) take an RNG in.
3. The order of parameters is different.
4. We have the convention of returning 1 on success, 0 on failure; Supercop/NIST return 0 on success and negative on failure.
5. Supercop/NIST don't use sig and sig_len, instead they use sm and sm_len ("signed message"), which includes a copy of the original message.

Thoughts?

christianpaquin:
I think your proposed signing API looks good, @dstebila. I might have further comments when I sit down to integrate a signature scheme, but it's a good starting point. Keeping it consistent with the KEX API makes sense.

christianpaquin:
See #92 proposal, following this thread discussion.

PR #16, opened 2016-09-30
Make random number generation and key exchange generation algorithms …

smashra:
…choosable at runtime


smashra:
Line greater than 80 columns are very hard to read so I'm adding following to the pretty print option:
--max-code-length=60.

dstebila:
I'd prefer to keep the source code formatted as is, and leave it to developers to use soft-wrapping in their editor if they prefer a different line length.

PR #20, opened 2016-10-07
Enable autotools

smashra:
Autotools have been enabled for building liboqs.


tlepoint:
I don't know how to add autotools to a project, but the current **+12,568 −85** seems really surprising to me. How come we do not **only** need the usual `configure.in` and `Makefile.am` file (the current project code being small and self-contented)?

Also, I don't know if using autotools, cmake (or other?) is really required for this project; if yes, maybe it would be worth discussing which option is the best.

smashra:
We only need configure.ac and  Makefile.am files for building. But the
assumption is that the system has complete auto build system installed in
one's environment.

What I'm trying to do is generate all the files created by running
autoreconf -i such that user has to only do ./configure and make.

If we agree about the assumption that whoever downloads liboqs is going to
have auto build system then I'll be very happy to remove all those files.

Currently I'm facing this problem of confgure re triggering  when make is
called which will then require auto build system to be there since it is
looking for aclocal defeating the purpose of what I'm trying to achieve.

dstebila:
Hi Tancrède: I'd originally thought that we should have it so that users can immediately run `./configure` when they download the source without having to have autotools installed.  Apparently that requires generating these many files and storing them in the repository.  Shravan agrees with you that it would be cleaner if just `configure.ac` and `Makefile.am` were in the repository, so we'll give that a try.

src/kex/kex.h
```diff
@@ -9,7 +9,12 @@
 #include <stddef.h>
 #include <stdint.h>
 
-#include <oqs/rand.h>
+#include "oqs/rand.h"
+
+enum KEX_ALGO_NAMES {
```
tlepoint: What is that used for?

.travis.yml
```diff
@@ -1,49 +1,73 @@
...
+      -  brew install autoconf
+      -  brew install automake
   
-
 script:
-  - make
-  - make check
\ No newline at end of file
+  - autoreconf -i 
+  - ./configure
```
tlepoint: Are you sure the ./configure is applied with the right compiler everytime? Can we do something like 
`./configure CC=... `
?

.travis.yml
```diff
@@ -1,49 +1,73 @@
...
         - gcc-6
+        - autotools-dev
+        - autoconf
   - os: osx
     compiler: clang
-    env: CC_OQS=clang
+    env: CC=clang
+    before_install:
+      -  brew install autoconf
+      -  brew install automake
```
tlepoint: I don't know why, but on my computer I need
```
brew install libtool
```
too.

tlepoint:
Don't forget to rebase `master`, otherwise the merge is not possible.

.travis.yml
tlepoint: For the first point, see [how it is done in fplll](https://github.com/fplll/fplll/blob/master/.travis.yml).

dstebila:
New pull request coming.

PR #26, opened 2016-10-14
Adding the Niederreiter cryptosystem (using QC-MDPC codes).

sebastianv89:
Before merging this into the main branch, consider that this code
is very slow and at a pre-quantum security level of just 80 bits.
Some serious improvements are required before anyone should consider
using this implementation.


dstebila:
Closing this PR since development continues on in #47.

PR #27, opened 2016-10-16
Adding prettyprint check to Travis

tlepoint:
I added a check on travis to verify that 
```
make prettyprint
```
was run before pushing in the PRs. The tests successfully catched the fact that `make prettyprint` in `master` [should have been run](https://travis-ci.org/tlepoint/liboqs/builds/168086027), and the second commit makes the [checks all green](https://travis-ci.org/tlepoint/liboqs/builds/168086535).

Note that I could not use the astyle package in Ubuntu 12.04 (that is used in Travis): it is an old version of astyle which (1) does not have the same output behavior, and (2) does not behave the same.

### DIFF
--- a/.travis-style-check.sh
+++ b/.travis-style-check.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+if [[ $(make prettyprint | grep Formatted) ]]; 
+then 
+	tput setaf 1;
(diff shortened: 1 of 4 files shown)


dstebila:
Very nice, thanks Tancrède!

PR #28, opened 2016-10-18
WIP - Add AES using AES-NI

aparent:
I still need to document the API.


dstebila:
Other than documentation, is this ready for code review / merging?

aparent:
Sure. I am not sure how portable it currently is since the header includes `wmmintrin.h` (aes intrinsics) for the   `__m128i` type. I can make it a bit more opaque by just requiring a byte array of size 16*20 for the key schedule which I can just cast internally to `__m128i` since the user should not be doing anything with the key schedule except passing it to `OQS_AES128_enc` or `OQS_AES128_dec`.  That way we could have another implementation in a .c file for systems that don't have the AES and SSE instructions which can use the same header.

dstebila:
Yes, if we could have it so that the functions that Frodo (or whatever) needs to call (OQS_AES128_enc, OQS_AES128_dec, OQS_AES128_load_schedule) use opaque byte array types, and then we can have a C implementation for all platforms as well as the ASM implementation when it is supported, that would be great.  I will work with Shravan to get conditional compilation of the ASM implementation in the configure/make system.

aparent:
Updated with some documentation and tests.  Also switched to an opaque byte array for the key schedule.

aparent:
It looks like the tests are showing some portability problems with regards to compiling on osx with clang. I will investigate.

dstebila:
Closed by #32.

PR #32, opened 2016-10-22
Add AES using AES-NI.

dstebila:
Based on #28.


aparent:
I noticed you dropped `monte_test` functions.  The equivalence test is probably ok.  I added the `monte_test` functions because they were specified in http://csrc.nist.gov/groups/STM/cavp/documents/aes/AESAVS.pdf.

aparent:
I can add in the validation file and a check that the test generates it exactly.

dstebila:
That would be great, thanks Alex!

dstebila:
I didn't realize that there were tests like that specified by the standard, so it would be good to have them in as well.

PR #42, opened 2016-10-28
Fix memory leak

mofojed:
Cleaned up a couple memory leaks.

The first is pretty straight forward; the named_parameters weren't being freed properly.

The second fix is for the running the speed tests; TIME_OPERATION_SECONDS was being used to time running some functions, but when those functions allocated memory, it was never cleaned up.
To fix this, I added a TIME_OPERATION_SECONDS_CLEANUP method. It is just the same as TIME_OPERATION_SECONDS, but takes an optional `cleanop` parameter that runs after each iteration.

Verified using valgrind:
valgrind --leak-check=full ./test_kex
valgrind --leak-check=full ./test_aes
valgrind --leak-check=full ./test_rand


dstebila:
Thanks Mike -- was on my list and appreciate the help.  Changed the benchmark macro slightly so we can do it with the existing macro rather than needing a new macro.

PR #45, opened 2016-11-01
Alternate Frodo optimization

dstebila:
Attempting to address issue #41.  Compared with pull request #43, this uses pre-processor macros to expand out some functions for each parameter set (for now, just recommended) which after pre-processing results in code with actual numbers in loops, so the compiler can provide whatever optimizations it likes.  This is much more like how the original Frodo works in terms of preprocessing.

Performance with master using gcc-6:
~~~
Operation                           Iterations  Total time (s)  Time(us): mean      pop. stdev    Cycles: mean      pop. stdev
RLWE BCNS15
alice 0                                   1314           1.000         761.067          76.582         1974182          198596
bob                                        818           1.001        1223.256         114.963         3173123          298168
alice 1                                   6301           1.000         158.714          22.813          411644           59135
LWE Frodo recommended
alice 0                                    970           1.000        1031.047          81.935         2674505          212467
bob                                        700           1.000        1428.890          99.240         3706550          257409
alice 1                                   6325           1.000         158.118          19.310          410084           50048
~~~

Performance with #43 using gcc-6:
~~~
Operation                           Iterations  Total time (s)  Time(us): mean      pop. stdev    Cycles: mean      pop. stdev
RLWE BCNS15
alice 0                                   1291           1.000         774.885          80.883         2009941          209716
bob                                        812           1.001        1232.835         106.505         3197854          276185
alice 1                                   6232           1.000         160.470          23.491          416172           60885
LWE Frodo recommended
alice 0                                    939           1.001        1065.592         128.305         2764074          332772
bob                                        572           1.001        1750.320         184.491         4540200          478551
alice 1                                   9521           1.000         105.038          15.495          272381           40138
~~~

Performance with this using gcc-6:
~~~
Operation                           Iterations  Total time (s)  Time(us): mean      pop. stdev    Cycles: mean      pop. stdev
RLWE BCNS15
alice 0                                   1271           1.001         787.311         109.281         2042218          283407
bob                                        769           1.001        1301.646         192.674         3376377          499702
alice 1                                   6470           1.000         154.579          17.835          400907           46213
LWE Frodo recommended
alice 0                                    975           1.000        1025.790         108.017         2660864          280118
bob                                        817           1.001        1225.311         128.029         3178425          332021
alice 1                                   9750           1.000         102.573          13.652          265995           35359
~~~

Performance with master using clang 8:
~~~
Operation                           Iterations  Total time (s)  Time(us): mean      pop. stdev    Cycles: mean      pop. stdev
[...]

PR #46, opened 2016-11-02
Frodo optimization - macros everywhere

mofojed:
A different spin on #45 , this defines the frodo functions as macros, and allows a new set of frodo functions with different parameters to be defined easily. For example, defining all the recommended functions is just:
```
OQS_KEX_LWE_FRODO_SET(recommended, 752, 8, 15, (1 << 15), 4, 256, 8, 8)
```

Instead of (from #45 ):
```
#define PARAMS_N 752
#define PARAMS_NBAR 8
#define PARAMS_LOG2Q 15
#define PARAMS_Q (1 << PARAMS_LOG2Q)
#define PARAMS_EXTRACTED_BITS 4
#define PARAMS_KEY_BITS 256
#define PARAMS_STRIPE_STEP 8
#define PARAMS_REC_HINT_LENGTH LWE_DIV_ROUNDUP(PARAMS_NBAR * PARAMS_NBAR, 8)
// pre-process code to obtain "recommended" functions
#define MACRIFY(NAME) NAME ## _recommended
#include "lwe_macrify.c"
```

I believe this reduces possible errors when defining a new set of parameters, as an error will be thrown if the values provided to `OQS_KEX_LWE_FRODO_SET` are incorrect. That being said, that's a pretty marginal use case.

Performance appears to be about the same as #45 in my tests, possibly 1-2% faster.

Sharing so Douglas can take a gander, though I personally haven't decided what style I like better - I like being able to define the new parameter set with one macro, but I don't like how large some of the macros are. Just another spin on it.


dstebila:
I don't feel good about the idea of all these functions being only defined via gigantic multi-line macros.  When making changes in the code, having to remember to put a backslash at the end of every line will get annoying real quick.  I would think compiler warnings will be more obscure and debugging might be trickier.  

The only way I could think of getting something like what you achieved but without multi-line macros was to define the parameters one at a time, and then include the inchoate lwe_macrify.c as a form of "executing" your OQS_KEX_LWE_FRODO_SET macro.  Do you think there's any other way?

mofojed:
Yea I agree, I don't feel good about the gigantic macros at all. The largest chance for error there is when defining a new parameter set, which probably won't happen too often and should be okay... should probably just #undef the defined parameters after an #include in yours, eg.
```
// Define recommened parameters
#define PARAMS_N 752
....
// pre-process code to obtain "recommended" functions
#define MACRIFY(NAME) NAME ## _recommended
#include "lwe_macrify.c"
#undef MACRIFY
#undef PARAMS_N

// Define paranoid parameters
#define PARAMS_N 1504
...
// pre-process code to obtain "paranoid" functions
#define MACRIFY(NAME) NAME ## _paranoid
#include "lwe_macrify.c"
#undef MACRIFY
#undef PARAMS_N
```

Then it's clear each block goes together, and you'll get an error if you miss defining a parameter when trying to create a new set.

I'm going to kill this change before I think about it any more :)

dstebila:
Undefining after is a good idea.  I've added that and am going to merge #45.

PR #47, opened 2016-11-02
Rebase niederreiter and factor out sha3

aparent:
This is basically the original pull request from Sebastian with some small changes rebased so that there are no merge conflicts.  I also factored out the sha3 which is used here and in newhope.

There is still one remaining warning due to passing a `const void *` as `par_ch_t[POLY_COUNT]`.  One way I can see to get rid of this is to change `typedef sp_poly_t par_ch_t[POLY_COUNT];` to `typedef sp_poly_t *par_ch_t;`.


dstebila:
Sebastian, in OQS_KEX_code_nrqcmdpc_new, there's a comment: 
> security needs to be re-evaluated, see: https://eprint.iacr.org/2016/858
Does this code use the parameters that were successfully attacked in this paper?  Or does it use new parameters as a consequence of the attack, but for which you don't know the estimated security?

dstebila:
Our newly merged #58 also uses Keccak -- the factoring out of SHA-3 should also be updated to affect that directory.

PR #49, opened 2016-11-10
Fix aesctr on windows

christianpaquin:
Added AES-CTR files to Visual Studio project and fixed compilation errors on Windows.


dstebila:
Thanks Christian!  

It's too bad we can't do continuous integration on Windows too to catch these things at commit time.

I just realized that we may occasionally be regressing changes you've made -- things like avoiding use of bzero.  I've added a continuous integration check for that (.travis-banned-functions-check.sh), at least, but if you have suggestions on other functions to avoid, please let me know or feel free to add.

christianpaquin:
No problem; I'm happy to test for that. It's hard to make sure that committed code works for all platforms at all time during development. As long as we test them all for important "milestones".

dstebila:
Thanks.  Speaking of "milestones", we should probably set some.  At this point we're just kind of going along as we do, but maybe setting a target would be a good idea.

aparent:
We can probably use the projects feature on github to organize that?  Maybe
create a project for each milestone and fill issues in?

PR #51, opened 2016-11-11
Remove statements which have no affect.

aparent:
Doing a scan-build of the source I discovered these two statements that appear to be unnecessary (since b is set to zero at the beginning of the loop).  I was wondering if they should have some affect and there is an implementation error or if they can be safely removed?

I also included some minor code cleanup where variable scope is reduced.


dstebila:
I'm thinking I'd prefer not to merge the NewHope changes -- in case we have to resync with future updates to NewHope, it might be easiest to have as few source code changes with their code.  Make sense?

Frodo changes seem fine, I can merge them.

PR #52, opened 2016-11-14
Speedup Travis and reduce build errors by installing astyle as a package.

aparent:


tlepoint:
I wouldn't have merged that. 

Last time I checked, the version of astyle installed by the package manager was not the 2.05 version, and the behavior **and** the output changed compared to the installed version.

tlepoint:
And from the [build log](https://travis-ci.org/open-quantum-safe/liboqs/builds/175771775):
 
```
$ bash .travis-style-check.sh
Invalid command line options: 
align-reference=name
For help on options, type 'astyle -h' 
Artistic Style has terminated!
make: *** [prettyprint] Error 1
Code adheres to the project standards.
```

aparent:
Ah I should have looked at the build output more carefully

I reverted and added a check for astyle failing in #53.
 
It would be nice if there was a more reliable way to get astyle since the sourceforge download seems to fail pretty regularly. Will see if this is easy to fix with caching on Travis.

PR #53, opened 2016-11-14
Revert the previous change in 250681196f

aparent:
Make style check fail when astyle fails.

### DIFF
--- a/.travis-global-namespace-check.sh
+++ b/.travis-global-namespace-check.sh
@@ -9,7 +9,7 @@ then
 	exit 1;
 else 
 	tput setaf 2;
-	echo "Code adheres to the project standards.";
+	echo "Code adheres to the project standards (global namespace).";
 	tput sgr 0
 	exit 0;
 fi;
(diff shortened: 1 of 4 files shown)


aparent:
Looking at the build output more closely it also seems like the error "find: unknown predicate `-E'" has been happening since e937a6a664df755.

aparent:
Will merge this later today if there are no objections.

.travis-style-check.sh
```diff
@@ -1,12 +1,21 @@
...
+	tput sgr 0;
+	exit 1;
+elif [ ${STATUS[0]} != 0 ];  then
+	tput setaf 1;
+	echo "prettyprint failed.";
+	tput sgr 0;
 	exit 1;
 else 
 	tput setaf 2;
 	echo "Code adheres to the project standards.";
```
tlepoint: Maybe we should add `Code adheres to the project standards (prettyprint)` here.

.travis-global-namespace-check.sh
```diff
@@ -4,10 +4,12 @@ if [[ $(nm -g liboqs.a | grep ' T ' | grep -E -v -i ' T [_]?OQS') ]];
 then 
 	tput setaf 1;
 	echo "Code contains the following non-namespaced global symbols; see https://github.com/open-quantum-safe/liboqs/wiki/Coding-conventions for function naming conventions.";
+	tput sgr 0
 	nm -g liboqs.a | grep ' T ' | grep -E -v -i ' T [_]?OQS'
 	exit 1;
 else 
 	tput setaf 2;
 	echo "Code adheres to the project standards.";
```
tlepoint: Maybe we should add `Code adheres to the project standards (global namespace)` here.

aparent:
Good idea, Added.

Issue #48, opened 2016-11-03
Zeroing Memory


aparent:
We should be zeroing memory that is used to store secret data in our free functions.  It turns out that it is pretty difficult to do this in a way that is portable and guaranteed not to be optimized away.  We could do something a bit ugly as show below (from libsodium): 

```
#ifdef HAVE_WEAK_SYMBOLS
__attribute__ ((weak)) void
_sodium_memzero_as_a_weak_symbol_to_prevent_lto(void * const pnt, const size_t len)
{
    unsigned char *pnt_ = (unsigned char *) pnt;;
    size_t         i = (size_t) 0U;

    while (i < len) {
        pnt_[i++] = 0U;
    }
}
#endif

void
sodium_memzero(void * const pnt, const size_t len)
{
#ifdef _WIN32
    SecureZeroMemory(pnt, len);
#elif defined(HAVE_MEMSET_S)
    if (len > 0U && memset_s(pnt, (rsize_t) len, 0, (rsize_t) len) != 0) {
        abort(); /* LCOV_EXCL_LINE */
    }
#elif defined(HAVE_EXPLICIT_BZERO)
    explicit_bzero(pnt, len);
#elif HAVE_WEAK_SYMBOLS
    _sodium_memzero_as_a_weak_symbol_to_prevent_lto(pnt, len);
#else
    volatile unsigned char *volatile pnt_ =
        (volatile unsigned char * volatile) pnt;
    size_t i = (size_t) 0U;

    while (i < len) {
        pnt_[i++] = 0U;
    }
#endif
}
```

mofojed:
OPENSSL just makes the pointer to the memset function volatile to avoid compiler optimization:

```
/*
 * Pointer to memset is volatile so that compiler must de-reference
 * the pointer and can't assume that it points to any function in
 * particular (such as memset, which it then might further "optimize")
 */
typedef void *(*memset_t)(void *, int, size_t);

static volatile memset_t memset_func = memset;

void OPENSSL_cleanse(void *ptr, size_t len)
{
    memset_func(ptr, 0, len);
}
```
Reference: https://github.com/openssl/openssl/blob/master/crypto/mem_clr.c

Would that be sufficient?

OpenSSL then has further optimizations as well, they have platform specific assembler that overrides the OPENSSL_cleanse function for optimizing further, eg: https://github.com/openssl/openssl/blob/master/crypto/x86_64cpuid.pl#L196

aparent:
[...]

PR #54, opened 2016-11-15
Add function for zeroing memory.

aparent:
Adds a function to zero memory as discussed in #48. Will add more add some more commits to this pull request using it in various places.

### DIFF
--- a/src/common/common.c
+++ b/src/common/common.c
@@ -10,7 +10,6 @@ void OQS_MEM_cleanse(void *ptr, size_t len) {
 #if defined(WINDOWS)
 	SecureZeroMemory(pnt, len);
 #elif defined(HAVE_MEMSET_S)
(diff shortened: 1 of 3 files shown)


src/common/common.c
```diff
@@ -0,0 +1,30 @@
...
+
+#if defined(WINDOWS)
+#include<windows.h>
+#endif
+
+void OQS_MEM_cleanse(void *ptr, size_t len) {
+#if defined(WINDOWS)
+	SecureZeroMemory(pnt, len);
+#elif defined(HAVE_MEMSET_S)
+	memset_s(pnt, (rsize_t) len, 0, (rsize_t) len);
```
mofojed: This line should be deleted, you're doing a memset_s on the next line in the if condition

Issue #63, opened 2016-11-28
Invalid read in sidh code


aparent:
Running valgrind over `text_kex` currently returns this error (among other similar errors):

```
==21169== Invalid read of size 8
==21169==    at 0x41442E: oqs_sidh_cln16_fpcopy751 (fpx.c:38)
==21169==    by 0x41387D: oqs_sidh_cln16_ladder_3_pt (ec_isogeny.c:345)
==21169==    by 0x40C196: oqs_sidh_cln16_SecretAgreement_B (sidh_kex.c:352)
==21169==    by 0x40A20D: OQS_KEX_sidh_cln16_bob (kex_sidh_cln16.c:116)
==21169==    by 0x401E42: kex_test_correctness (test_kex.c:81)
==21169==    by 0x400E48: kex_test_correctness_wrapper (test_kex.c:152)
==21169==    by 0x400E48: main (test_kex.c:289)
==21169==  Address 0x5bff880 is 0 bytes after a block of size 48 alloc'd
==21169==    at 0x4C2BC0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==21169==    by 0x40A1AA: OQS_KEX_sidh_cln16_bob (kex_sidh_cln16.c:100)
==21169==    by 0x401E42: kex_test_correctness (test_kex.c:81)
==21169==    by 0x400E48: kex_test_correctness_wrapper (test_kex.c:152)
==21169==    by 0x400E48: main (test_kex.c:289)
```

I investigated the problem and noticed the following:

A block of 48 bytes is malloc'd [here](https://github.com/open-quantum-safe/liboqs/blob/d23f8d3eddf3e72a5971983562721ba9b9c63df5/src/kex_sidh_cln16/kex_sidh_cln16.c#L100
), then passed [here](https://github.com/open-quantum-safe/liboqs/blob/d23f8d3eddf3e72a5971983562721ba9b9c63df5/src/kex_sidh_cln16/kex_sidh_cln16.c#L116), then passed [here](https://github.com/open-quantum-safe/liboqs/blob/d23f8d3eddf3e72a5971983562721ba9b9c63df5/src/kex_sidh_cln16/sidh_kex.c#L352) (as the forth parameter), finally it is passed [here](https://github.com/open-quantum-safe/liboqs/blob/d23f8d3eddf3e72a5971983562721ba9b9c63df5/src/kex_sidh_cln16/ec_isogeny.c#L345).

[...]

PR #70, opened 2016-12-13
Fix SIDH memory allocation

christianpaquin:
Fixed memory allocation in SIDH KEX. Should fix [issue 63](https://github.com/open-quantum-safe/liboqs/issues/63).


aparent:
The issue appears to be fixed with this pr.

Issue #56, opened 2016-11-17
astyle -> clang-format?


tlepoint:
In light of #53, I think it would be interesting to switch from astyle to clang-format. [Clang-format](http://clang.llvm.org/docs/ClangFormat.html) is both a library and a stand-alone tool aiming at automatically reformatting C sources files, and continus to be developed while astyle has not been updated since December 2014.

aparent:
I agree. As a side benefit it is likely included by default on the build images so there would be no need to compile it foe each build.

dstebila:
Okay, seems like a good idea.  Can you come up with some command line options?  Not a big deal if they don't reproduce the exact format, but I'd like to keep the main formatting principles.

tlepoint:
I'll do it this week! Is there a PR I should wait to be merged for before doing it (bitroting)? I'll also refactor slightly the travis files by creating a `.travis` directory.

dstebila:
I'd like to wait until #58 lands, which should be this Tuesday or Wednesday.  The other big one is #47, but I'm not sure what the status of it is from a cryptographic perspective, so you can go ahead without it.

dstebila:
#58 has landed, so reformatting can go ahead.

dstebila:
I've tried to do this on a branch, but discovered that the version of Ubuntu in Travis (12.04 "precise") has a very old version of clang-format (3.4), which is missing a bunch of options to produce the desired format.  Anyone know how to install a newer version of clang-format on an old Ubuntu?  Otherwise we might have to stay with astyle.

tlepoint:
Sorry very hectic weeks recently, I didn't have time to look into that. I did it for fplll [here](https://github.com/fplll/fplll/blob/master/.travis.yml).

PR #80, opened 2016-12-28
Travis revisited (still with astyle) and nicer makefile output

tlepoint:
Minor modifications:
- moving all the travis scripts in the same repository
- producing a nicer makefile output

For the second point, it will be easier if eventually autotools is used (#20) as we can use `AM_SILENT_RULES`. What is the plan wrt to that PR? 

I didn't feel confortable to change from astyle to clang-format (#56) because of the many possible options there. Should it be an unilateral choice, or should we discuss those?


tlepoint:
@dstebila: I see that you did clang-format in [https://github.com/open-quantum-safe/liboqs/commit/4f76861a794fae6da4b0c4b1bb766cb8594d02fc](https://github.com/open-quantum-safe/liboqs/commit/4f76861a794fae6da4b0c4b1bb766cb8594d02fc). Feel free to add it to this PR if you want, or to create a new one (I admit that the silent makefile can be a moot point).

PR #84, opened 2017-01-07
Clang format instead of astyle

tlepoint:
Using clang-format instead of astyle - cf. #56


dstebila:
This build failed after commit (see https://travis-ci.org/open-quantum-safe/liboqs/builds/190978898) with an error about adding the Ubuntu PPA.  Any ideas why?

tlepoint:
It just happens once in a while apparently. Can you relaunch the errored jobs?

dstebila:
Okay now, phew.

PR #85, opened 2017-01-07
Add option to build with NTRU 

aparent:
Also added script to automatically download and build `libntruencrypt`.

For some reason the prettyprint check fails on OSX only (I have run `make prettyprint` and it looks ok on my end).  The clang format change will likely fix this issue.


dstebila:
Alex, your comment says you added a script to automatically download and build `libntruencrypt`, but I can't find that in the repository.  Did you miss it in the commit?

tlepoint:
Small comment: the prettyprint test fails here because `git-status -s` returns something because of the download of NTRU. To be fixed.

aparent:
@dstebila the file is called install-ntru and it is in the root directory of the project.  

Also install might not be a great name now that I think of it since it just builds it in a sub directory and links against it statically rather then doing a system wide install.  Maybe it could be renamed `download-and-build-ntru` to be more specific.

PR #79, opened 2016-12-23
Add XMSS signature scheme

dstebila:
Code contributed by Denis Butin and Stefan Gazdag

PR #92, opened 2017-01-25
Add signature API (proposal)

christianpaquin:
=== NOT YET READY FOR MERGE ===

Following discussion on pull request #14, here is a suggested API for signing. No scheme has been integrated yet. If the team agrees, I'll start work on integrating one.

One open question is what to do with stateful schemes, such as XMSS, that do not adhere with a conventional signature API. Since the main goal of OQS is to compare PQ schemes, it is useful to include them (like XMSS in pull request #79) even if there are limitations on usage of the private key. (NIST for example [suggests](http://csrc.nist.gov/groups/ST/post-quantum-crypto/faq.html#Q2) to define specific use cases for stateful schemes). One option is to not expose the state through the API, and simply return an error if the key is now unusable (this could be sufficient for comparison testing, but insecure for a real-application). The alternative would be to expose the state through the API (and that might still not be sufficient enough for security, to deal with state reset attacks). Thoughts?


src/sig/sig.h
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * OQS signature object
+ */
+typedef struct OQS_SIG {
+
+	/**
+	 * PRNG
+	 */
+	OQS_RAND *rand;
+
```
dstebila: Remove rand from OQS_SIG

src/sig/sig.h
```diff
@@ -0,0 +1,138 @@
...
+	/**
+	 *  Maximum signature length.
+	 */
+	uint16_t max_sig_len;
+
+	/**
+	 * Named parameters for this key exchange method instance, if any.
+	 */
+	char *named_parameters;
```
dstebila: Remove named_parameters from OQS_SIG.

src/sig/sig.h
```diff
@@ -0,0 +1,138 @@
...
+
+	/**
+	 * Opaque pointer for passing around instance-specific data
+	 */
+	void *params;
```
dstebila: Remove params from OQS_SIG.

src/sig/sig.h
```diff
@@ -0,0 +1,138 @@
...
+
+	/**
+	 * Opaque pointer for passing around any computation context
+	 */
+	void *ctx;
```
dstebila: Remove ctx from OQS_SIG.

src/sig/sig.h
```diff
@@ -0,0 +1,138 @@
...
+
+	/**
+	 * Pointer to a function for public and private signature key generation.
+	 *
+	 * @param s                Signature structure
+	 * @param priv             The private key
+	 * @param pub              The public key
+	 * @return                 1 on success, or 0 on failure
+	 */
+	int (*keygen)(const OQS_SIG *s, uint8_t **priv, uint8_t **pub);
```
dstebila: - Pass in OQS_RAND, not OQS_SIG.
- priv and pub should be uint8_t *, not uint8_t ** (i.e., up to the caller to provided memory, rather than callee to do a malloc)

src/sig/sig.h
```diff
@@ -0,0 +1,138 @@
...
+	 *
+	 * @param s                Signature structure
+	 * @param priv             The signer's private key
+	 * @param msg              The message to sign
+	 * @param msg_len          Length of the message to sign
+	 * @param sig              The generated signature
+	 * @param sig_len          Length of the generated signature
+	 * @return                 1 on success, or 0 on failure
+	 */
+	int (*sign)(const OQS_SIG *s, const uint8_t *priv, const uint8_t *msg, const size_t msg_len, uint8_t **sig, size_t *sig_len);
```
dstebila: - Pass in OQS_RAND, not OQS_SIG.
- sig should be uint8_t *, not uint8_t ** (i.e., up to the caller to provided memory, rather than callee to do a malloc)
- remove sig_len argument
- return signature length

src/sig/sig.h
```diff
@@ -0,0 +1,138 @@
...
+	 * @return                 1 on success, or 0 on failure
+	 */
+	int (*verify)(const OQS_SIG *s, const uint8_t *pub, const uint8_t *msg, const size_t msg_len, const uint8_t *sig, const size_t sig_len);
+
+	/**
+	 * Pointer to a function for freeing the allocated signature structure
+	 *
+	 * @param s                Signature structure
+	 */
+	void (*free)(OQS_SIG *s);
```
dstebila: No longer needed.

src/sig/sig.h
```diff
@@ -0,0 +1,138 @@
...
+ * @param named_parameters   Name or description of method-specific parameters
+ *                           to use for this instance (as a NULL-terminated C string),
+ *                           if any, or NULL.
+ * @return                   The object on success, or NULL on failure.
+ */
+OQS_SIG *OQS_SIG_new(OQS_RAND *rand, enum OQS_SIG_alg_name alg_name, const char *named_parameters);
+int OQS_SIG_keygen(const OQS_SIG *s, uint8_t **priv, uint8_t **pub);
+int OQS_SIG_sign(const OQS_SIG *s, const uint8_t *priv, const uint8_t *msg, const size_t msg_len, uint8_t **sig, size_t *sig_len);
+int OQS_SIG_verify(const OQS_SIG *s, const uint8_t *pub, const uint8_t *msg, const size_t msg_len, const uint8_t *sig, const size_t sig_len);
+void OQS_SIG_free(OQS_SIG *s);
```
dstebila: OQS_SIG_free not needed.

src/sig/sig.h
```diff
@@ -0,0 +1,138 @@
...
+ * @param alg_name           Algorithm to be instantiated
+ * @param named_parameters   Name or description of method-specific parameters
+ *                           to use for this instance (as a NULL-terminated C string),
+ *                           if any, or NULL.
+ * @return                   The object on success, or NULL on failure.
+ */
+OQS_SIG *OQS_SIG_new(OQS_RAND *rand, enum OQS_SIG_alg_name alg_name, const char *named_parameters);
+int OQS_SIG_keygen(const OQS_SIG *s, uint8_t **priv, uint8_t **pub);
+int OQS_SIG_sign(const OQS_SIG *s, const uint8_t *priv, const uint8_t *msg, const size_t msg_len, uint8_t **sig, size_t *sig_len);
+int OQS_SIG_verify(const OQS_SIG *s, const uint8_t *pub, const uint8_t *msg, const size_t msg_len, const uint8_t *sig, const size_t sig_len);
```
dstebila: OQS_SIG_keygen, OQS_SIG_sign, and OQS_SIG_verify should be updated in light of the comments above.  They should all take OQS_RAND as an input. These ones *should* take OQS_SIG as an input, since these ones will call the corresponding algorithm from the OQS_SIG structure.

src/sig/sig.h
```diff
@@ -0,0 +1,138 @@
...
+ * Allocate a new signature object.
+ *
+ * @param rand               Random number generator.
+ * @param alg_name           Algorithm to be instantiated
+ * @param named_parameters   Name or description of method-specific parameters
+ *                           to use for this instance (as a NULL-terminated C string),
+ *                           if any, or NULL.
+ * @return                   The object on success, or NULL on failure.
+ */
+OQS_SIG *OQS_SIG_new(OQS_RAND *rand, enum OQS_SIG_alg_name alg_name, const char *named_parameters);
```
dstebila: OQS_SIG_new not needed any more.

christianpaquin:
@dstebila, wrt your comments: that means that we would have a different design for the kex and sig API. Are you suggesting we refactor kex to match these design guidelines? I think uniformity between the two is desirable. Thoughts?

src/sig/sig.h
zaverucha: sig_len should be used by the caller to specify the length of sig, so that the sign function can ensure not to overflow sig. Keeping sig_len as an in-out parameter (size_t*) allows the signature function to return the number of bytes written to sig, keeping the return value of the sign function consistent with the other functions in this API (booleans). 

Aside -- in the documentation of sig_len, reference s.max_sig_len. Do you expect callers to always provide a buffer of length max_sig_len bytes?

zaverucha: sig_len should be used by the caller to specify the length of sig, so that the sign function can ensure not to overflow sig. Keeping sig_len as an in-out parameter (size_t*) allows the signature function to return the number of bytes written to sig, keeping the return value of the sign function consistent with the other functions in this API (booleans). 

Aside -- in the documentation of sig_len, reference s.max_sig_len. Do you expect callers to always provide a buffer of length max_sig_len bytes?

src/sig/sig.h
```diff
@@ -0,0 +1,138 @@
...
+	/**
+	 * Pointer to a function for signature verification.
+	 *
+	 * @param s                Signature structure
+	 * @param pub              The signer's public key
+	 * @param msg              The signed message
+	 * @param msg_len          Length of the signed message
+	 * @param sig              The signature to verify
+	 * @param sig_len          Length of the signature to verify
+	 * @return                 1 on success, or 0 on failure
```
zaverucha: Clarify return value meaning.  I assume that 1 means that the signature is valid? (in addition to the verification code running without error)

src/sig/sig.h
```diff
@@ -0,0 +1,138 @@
...
+	/**
+	 *  Equivalent quantum security in terms of the number of bits provided by the
+	 *  signature method.
+	 */
+	uint16_t estimated_quantum_security;
+
+	/**
+	 *  Private key length.
+	 */
+	uint16_t priv_key_len;
```
zaverucha: Use size_t for lengths of buffers.  I'd also use it for the estimated_*_security members, but that's just preference.

dstebila:
@christianpaquin Yes, my plan is to refactor the KEX API to be simpler and match as well.

src/sig/sig.h
christianpaquin: method_name is typically allocated with strdup and must be freed through OQS_SIG_free.

dstebila: If we're allocating OQS_SIG statically, can't we just say 
~~~
OQS_SIGN OQS_SIGN_douglas_awesome = {
	"Douglas' signature scheme, awesome security level",
	512, // estimated_classical_security
	256, // estimated_quantum_security
	1024, // private key length
	2048, // public key length
	3072, // max signature length
	&OQS_SIGN_douglas_awesome_keygen,
	&OQS_SIGN_douglas_awesome_sign,
	&OQS_SIGN_douglas_awesome_verify
};
~~~

This wouldn't have any dynamic allocation.

src/sig/sig.h
dstebila: Supercop and NIST APIs expect the caller to always provide a buffer of max_sig_len bytes, and then return how many bytes of that were actually used.

src/sig/sig.h
christianpaquin: Sure, we can do that. By "typically", I meant what is done in the kex schemes, so we can go refactor these later.

christianpaquin: How are users supposed to get a scheme with a specific ID? Surely, you don't want to always instantiate a OQS_SIG struct with all that info. A factory method that returns the instantiated struct would be useful, even if we don't call it OQS_SIG_new (to make it clear you don't have to free it later).

christianpaquin:
I've updated the code to reflect this discussion. Comments are welcome. Not yet ready for merge, because without an actual scheme implemented, the code doesn't run.

README.md
```diff
@@ -49,6 +49,7 @@ This will generate:
 - `test_rand`: A simple test harness for the random number generator.  This will test the distance of PRNG output from uniform using statistical distance.
 - `test_aes`: A simple test harness for AES.  This will test the correctness of the C implementation (and of the AES-NI implementation, if not disabled) of AES, and will compare the speed of these implementations against OpenSSL's AES implementation.
 - `test_kex`: A simple test harness for the default key exchange algorithm.  This will output key exchange messages; indicate whether the parties agree on the session key or not over a large number of trials; and measure the distance of the sessions keys from uniform using statistical distance.
+- `test_sig`: A simple test harness for the signature algorithms. FIXME: contiune description.
```
marcmanzano: typo -> 'continue'

christianpaquin:
Surpeceded by [PR 120](https://github.com/open-quantum-safe/liboqs/pull/120)

PR #97, opened 2017-01-30
Call OQS entropy collection api in NTRU

christianpaquin:
Replaced direct call to dev/urandom with OQS' new system entropy collection api.


src/kex_ntru/kex_ntru.c
```diff
@@ -53,25 +53,10 @@ static uint8_t get_entropy_from_dev_urandom(ENTROPY_CMD cmd, uint8_t *out) {
...
-		}
-#else
-		int fd = open("/dev/urandom", O_RDONLY);
-		if (fd <= 0)
-			return 0;
-		int r = read(fd, out, 1);
-		close(fd);
-#endif
-		if (r != 1)
+		// TODO: why is this called to get entropy bytes one by one?
```
dstebila: I was following some example code and that's what they did.  I'm not sure why.  If it can be changed without causing errors, then we can do that.  I'll accept your PR as it is though.

christianpaquin: Yeah, I didn't know if was done for some security reason. We can ask the original authors, or just change it.

PR #99, opened 2017-02-08
Autotools

dstebila:
Build using autotools.


tlepoint:
Why do we have a fail on macOS not caught by Travis?

```
./test_kex
dyld: Library not loaded: /usr/local/lib/libntruencrypt.0.dylib
  Referenced from: /Users/travis/build/open-quantum-safe/liboqs/./test_kex
  Reason: image not found
make: *** [test] Trace/BPT trap: 5
```

PR #119, opened 2017-04-21
Add sidh2.0

christianpaquin:
Updated SIDH CLN16 library to v2.0 (https://github.com/Microsoft/PQCrypto-SIDH). This version of the library adds a variant of the protocol to compress the public keys from 564 to 330 bytes).


dconnolly:
🎉

PR #124, opened 2017-05-20
Update Readme with Ubuntu commands

markszabo:
I've built the library on Ubuntu 16.04 and here are the commands to install the dependencies.

I've also found a small mistake in the Readme: to build with kex_code_mcbits enabled the command is not `./configure --enable-libsodium` but `./configure --enable-mcbits`.


dstebila:
Thanks very much!

PR #131, opened 2017-06-28
Mlwe kyber

tlepoint:
Now that [Kyber](https://eprint.iacr.org/2017/634.pdf)'s implementation is public (on [github.com/pq-crystals/kyber](https://github.com/pq-crystals/kyber), here is the integration with liboqs.

I had to revert to a previous commit because master is broken.


.travis.yml
```diff
@@ -51,14 +48,20 @@ matrix:
...
+      - CC_OQS=clang 
+      - AES_NI=0 
+      - USE_OPENSSL=1 
+      - OPENSSL_DIR=/usr/local/opt/openssl
+      - ENABLE_CODE_MCBITS=1 
+      - ENABLE_NTRU=1 
+      - ENABLE_SIDH_IQC_REF=1
 before_install:
-      - brew install clang-format openssl libsodium libm4ri-dev
+      - brew install clang-format libsodium
```
dstebila: Shouldn't we keep brew's openssl package since we're referencing it in the OPENSSL_DIR=/usr/local/opt/openssl above?

tlepoint: Openssl is [already installed by default](https://blog.travis-ci.com/page-2)

Issue #129, opened 2017-06-28
Need to do at least two macOS in travis


tlepoint:
We need at least two macOS in Travis, otherwise we cannot catch bugs like #127 (current macOS install all dependencies to run the full project)

dstebila:
Fixed by #132.

Issue #128, opened 2017-06-28
Travis is failing


tlepoint:
Because `libm4ri-dev` does not exist, introduced in [3aaa26aacee5e332a517f2de6436fe1c665e0785](https://github.com/open-quantum-safe/liboqs/commit/3aaa26aacee5e332a517f2de6436fe1c665e0785)

@smashra It is important to only do PR and to not directly commit to master, otherwise we do not catch this.
@dstebila: is it possible to forbid commits to master and force to go through PR?

dstebila:
Good suggest Tancrede.  I didn't know this was an option but I've now found it in the Github configuration and turned it on.  If I interpret the option correctly, master now prevent "force pushes" and instead requires code to come on a pull request.  I think it might also demand at least one review on every PR before it can be merged, but I am not certain about that option yet -- we shall see!

tlepoint:
Fixed by #134.

PR #132, opened 2017-06-28
Add 2 OSX to travis

tlepoint:
As said in #129, I think we should have two OSX to Travis: one with and one without all the options. That would help debugging #128.


.travis.yml
```diff
@@ -6,16 +6,9 @@ matrix:
...
     compiler: gcc
-    env: CC_OQS=gcc-4.8 ENABLE_SIDH_IQC_REF=1
-    addons:
-      apt:
-        sources:
-        - ubuntu-toolchain-r-test
-        packages:
-        - gcc-4.8
-        - libgmp-dev
+    env: ENABLE_SIDH_IQC_REF=1
```
dstebila: Is there a reason you took out CC=gcc-4.8?  Is the idea to have this target have building with the default compiler?

tlepoint: Yes: since we use `trusty` now, the default version of gcc is 4.8.4.

dstebila:
Fixes #132.

PR #120, opened 2017-05-11
Add sig api with picnic

christianpaquin:
This pull request proposes a new signature API modeled after the kex api, and adds the [Picnic](https://github.com/Microsoft/Picnic) signature scheme.

Picnic must currently be downloaded and set up (to generate parameters) using a script (download-and-setup-picnic.sh). We can migrate to a git submodule at a later time, after this PR is reviewed. 

The Visual Studio build has not yet been modified to support Picnic; this will be done after this PR is reviewed.

PR #138, opened 2017-07-04
Christianpaquin add sig api with picnic

smashra:


dstebila:
@smashra Why is there a new PR instead of using PR #120?

smashra:
Sorry, I'll delete the new one.

PR #144, opened 2017-07-10
Revert "Enabling NTRU by default, minor cleaning of configure.ac"

dstebila:
Reverts open-quantum-safe/liboqs#142

clang-format in brew has been updated very recently and apparently this has changed how it formats things **again**.  This is not the first time we've run into problems with clang-format inconsistencies.  I am not sure how to ensure reliable formatting behaviour across platforms and clang-format versions.


tlepoint:
Sigh... Let's authorize osx to fail?

tlepoint:
Or rather, let's remove the clang-format test from macOS?

dstebila:
We could do that. I think I've figured out how to force brew to install a specific version of clang-format, so we could force it to install clang-format-3.9 and then theoretically we're using the same version on both macOS and Linux.

dstebila:
Not going to revert - trying to fix via #146

Issue #145, opened 2017-07-10
macOS Illegal instruction: 4 (seen on Travis)


tlepoint:
From [https://travis-ci.org/open-quantum-safe/liboqs/jobs/251993063](https://travis-ci.org/open-quantum-safe/liboqs/jobs/251993063), before the fails due to clang-format, we get the following error:

```
./test_kex
================================================================================
Sample computation for key exchange method LWE Frodo recommended
================================================================================
make: *** [test] Illegal instruction: 4
```

I have no idea where it comes from, I never got it on my mac. Anyone? Maybe we should `valgrind`.

tlepoint:
Idem in [https://travis-ci.org/open-quantum-safe/liboqs/jobs/252002293](https://travis-ci.org/open-quantum-safe/liboqs/jobs/252002293) that reverted the last commit to a previous version of master.

dstebila:
We just discussed this in our weekly call, @smashra will be looking into it.

vsoftco:
Fixed it by replacing .global in all asm files with .globl (that's because macOS uses some ancient "as" tool)

dstebila:
Fixed in #152

dstebila:
Still borken.

Issue #130, opened 2017-06-28
Non-namespaced global symbols: Travis should fail even more...


tlepoint:
In addition to #128, current travis builds that are flagged as passing output what is below, and therefore should fail.

```
Code contains the following non-namespaced global symbols; see https://github.com/open-quantum-safe/liboqs/wiki/Coding-conventions for function naming conventions.
00000000000000c0 T allocateProof
0000000000000060 T allocateRandomTape
0000000000000130 T allocateSignature
0000000000000190 T allocateUnruhPermutations
0000000000000000 T allocateView
00000000000003b0 T cleanup_EVP
0000000000000520 T Commit
0000000000002450 T computeRoundKey
0000000000001460 T computeRoundKey2
0000000000000f50 T deserializeSignature
0000000000000110 T freeProof
00000000000000a0 T freeRandomTape
0000000000000160 T freeSignature
00000000000001c0 T freeUnruhPermutations
0000000000000040 T freeView
0000000000000660 T G
00000000000005c0 T generateAesIv
00000000000002b0 T getAllRandomness
0000000000000230 T getBit
00000000000001e0 T getBitFromWordArray
00000000000008b0 T getChallenge
0000000000000280 T getRandom32
0000000000000950 T H3
0000000000000390 T init_EVP
0000000000002320 T mpc_AND
00000000000011c0 T mpc_AND_verify
0000000000002660 T mpc_LowMC
0000000000002250 T mpc_multiplyWithKeyMatrix
0000000000001390 T mpc_multiplyWithKeyMatrix2
0000000000002130 T mpc_multiplyWithMatrix
0000000000001280 T mpc_multiplyWithMatrix2
0000000000002460 T mpc_substitution
0000000000001470 T mpc_substitution2
00000000000003e0 T mpc_xor
00000000000004e0 T mpc_xor2
0000000000000480 T mpc_xor4_2
0000000000000420 T mpc_xor4_3
0000000000000290 T numBytes
00000000000008f0 T print_hex
0000000000000bd0 T prove
0000000000002890 T random_bytes_default_openssl
0000000000002850 T runMPC
0000000000000d30 T serializeSignature
0000000000000250 T setBit
0000000000000200 T setBitInWordArray
0000000000000850 T setChallenge
00000000000028b0 T sign
0000000000001aa0 T verify
0000000000001680 T verifyLowMC
0000000000001840 T verifyProof
0000000000000620 T freeLookupTables
[...]

PR #146, opened 2017-07-10
Fix clang-format on macOS at version 3.9.

dstebila:


dstebila:
I'm aware that Travis job https://travis-ci.org/open-quantum-safe/liboqs/jobs/252014918 has the illegal instruction bug #145 but is showing green because of #130.  For now I'm planning to merge this with #145 unresolved.

PR #151, opened 2017-07-19
Fix windows build july2017

christianpaquin:
* Repaired Windows builds broken in the last few commits
  - Re-disabled NTRU by default on Windows; needs manual steps
  - Re-enabled SIDH by including malloc.h
* Enabled Kyber
* Updated project to VS2017


dstebila:
@christianpaquin, macOS doesn't have <malloc.h> and according to this link (https://stackoverflow.com/questions/12973311/difference-between-stdlib-h-and-malloc-h#12973361) the preferred approach is to use <stdlib.h>.  Can you check if that is okay on Windows?

christianpaquin:
@dstebila, yes <stdlib.h> works on Windows.

PR #152, opened 2017-07-20
Fix unknown pseudo-op: .global under macOS

vsoftco:
replaced .global with .globl in all .S files due to ancient “as” tool
in macOS

### DIFF
--- a/src/kex_rlwe_msrln16/AMD64/error_asm.S
+++ b/src/kex_rlwe_msrln16/AMD64/error_asm.S
@@ -24,7 +24,7 @@
 //  Error sampling from psi_12
 //  Operation: c [reg_p2] <- sampling(a) [reg_p1]
 //*********************************************************************** 
-.global oqs_rlwe_msrln16_error_sampling_asm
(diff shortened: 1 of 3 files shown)


dstebila:
Thanks Vlad!

vsoftco:
@dstebila That's something odd going on, all test seem to pass, however Travis Build #498.5 spits out an Illegal Instruction on line 1789. It seems that `./test_rand --quiet` is spitting out this. I compiled and checked on my side (osX clang), everything works fine

dstebila:
Very weird.  Not sure why that makes a difference.

Google didn't help me much, the only partial explanation I could find was here, but it doesn't seem to offer a clear solution: https://stackoverflow.com/questions/14268887/what-is-the-illegal-instruction-4-error-and-why-does-mmacosx-version-min-10

I've submitted a new PR #154 which switches the Xcode version to the same one on my desktop.  If that works, that seems like a bit of a cheat, but I don't really know how else to diagnose since we don't have old versions of macOS sitting around.

dstebila:
I'm trying to install an older version of OS X in a virtual machine to try to diagnose as well.

PR #153, opened 2017-07-20
minor update in the README.md re: openssl install

vsoftco:


README.md
```diff
@@ -44,11 +44,17 @@ You need to install autoconf, automake and libtool:
 
 	brew install autoconf automake libtool
 	
+In case OpenSSL is not installed:
+    
+	brew install openssl
+	cd /usr/local/include
+	ln -s ../opt/openssl/include/openssl .
```
dstebila: We should not require the user to link brew's OpenSSL into /usr/local, as this might override other preferences they have.  (Also, the command above only links in the includes, but not the libraries, so it could lead to inconsistencies.)  

Instead, we want to get the following environment variables added.  (See the instructions in `brew info openssl`.)

```
LDFLAGS:  -L/usr/local/opt/openssl/lib
CPPFLAGS: -I/usr/local/opt/openssl/include
```

So we need to find some way of telling configure to look there when on macOS, but I don't know autoconf well enough to know how to tell it to do that.  @smashra? @tlepoint?

vsoftco: Most out there recommend to use the soft link solution, I agree it's not the most elegant, but macOS has some issues code compliance... Basically Apple removed support for OpenSSL on OS X El Captain (after the famous OpenSSL bug). See e.g. this proposed solution, which is frequently cited: https://solitum.net/openssl-os-x-el-capitan-and-brew/

vsoftco: I'll take a look on how to automate the libSSL detection on macOS

tlepoint: I also would recommend not using soft link. I tailor my installations of homebrew and I would not want to have to investigate a solution that forces to edit files in case the default solution would error. What was wrong with specifying the openssl path for configure?

Also, how come we need OpenSSL everywhere now?

PR #154, opened 2017-07-20
Fix illegal instruction on macOS, redux

dstebila:


dstebila:
This is still failing with a different version of macOS and a different version of Xcode.  I have no idea why and am not able to reproduce.  Any thoughts?

vsoftco:
@dstebila No idea! I cannot reproduce it at all, I have 3 versions of clang on my mac and all are running it fine. Can you perhaps add some of the params to "env" so we can see when it doesn't break anymore on Travis?

dstebila:
It looks like it might be the AES NI implementation causing problems.  Commit b273093 had AES_NI=0 in the first macOS build, which had previously been failing but didn't in that commit.  

So I propose we use AES_NI=0 for both our macOS builds for now -- that's commit caede46.  This would mean our test coverage on Travis doesn't cover the case of how must users will probably build on macOS.  But for Vlad and I it builds on our Macs with AES_NI=1.  And we probably won't be changing AES code much, so it seems like this small gap in test coverage is not likely to be too problematic.

PR #158, opened 2017-07-25
Enable or disable each algorithm

boson-lepton:


dstebila:
@smashra Sadly, this patch will need to be updated because of conflicts from the Picnic merge.

smashra:
I'll update it.

PR #170, opened 2017-08-24
Arm compilation

vlad-sensei:
Enable liboqs to compile on ARM64 platforms and Android

Summary:
- Compilation on ARM64 platforms
- Assembly optimizations for SIDH key exchange algorithms from official SIDH 2.0 library
- Configure wrapper script for Android cross-compilation

Sponsored by [Sectra Communications](http://communications.sectra.com/)


dstebila:
Hi Vladimir!  Thanks for this -- it will be great to have our library building on ARM too.  I am travelling over the next few days but we will look at it next week and try to get it merged.

vlad-sensei:
Sounds good! Hope that everything is satisfactory

dstebila:
Thanks very much Vladimir!  I was wondering if you have any idea how we can add this Android build to our continuous integration system.  We use Travis for Linux and macOS build testing and AppVeyor for Windows build testing.  Do you know if there's anything similar for Android?  Or maybe Travis Linux builds can cross compile for Android?  Something like that will help us check that we don't break the Android builds going forward.

vlad-sensei:
No problem Douglas! This is pull request enables two somewhat separate features (ARMv8 (cross)compilation and Android compilation). Both parts are a bit tricky to test properly in this case. The ARM part requires ARM environment to run, which can be [problematic](https://www.tomaz.me/2013/12/02/running-travis-ci-tests-on-arm.html) due absence of proper support. As for the Android part, it looks like Travis [requires gradle](https://docs.travis-ci.com/user/languages/android/) and there is no apparent support for NDK.

However both can be (and were) cross-compiled from a normal Linux box. Android may require NDK toolchain though. The only real difference that I came across when compiling C for Android is the lack of `stderr` and the [mandatory](https://source.android.com/security/enhancements/enhancements50) `-pie` flag to run the executable.

PR #169, opened 2017-08-22
Updated Windows build (added sig, fixed warnings, 2017 update)

christianpaquin:
Updated Visual Studio solution to 2017 version, fixed build warnings (issue #159), added signature and Picnic support.

PR #175, opened 2017-09-19
Vsoftco fix windows warnings

vsoftco:


dstebila:
Will be back-ported to #169.

PR #176, opened 2017-09-24
Patch-based approach for adding NewHope AVX2

smashra:
AVX2 build by applying git patch for taking care of travis tests.


patches/kex_rlwe_newhope_avx2/patch1.patch
```diff
@@ -0,0 +1,1073 @@
...
+ 
+-void shake128_absorb(uint64_t *s, const unsigned char *input, unsigned int inputByteLen)
++static void shake128_absorb(uint64_t *s, const unsigned char *input, unsigned int inputByteLen)
+ {
+   keccak_absorb(s, SHAKE128_RATE, input, inputByteLen, 0x1F);
+ }
+ 
+ 
+-void shake128_squeezeblocks(unsigned char *output, unsigned long long nblocks, uint64_t *s)
++static void shake128_squeezeblocks(unsigned char *output, unsigned long long nblocks, uint64_t *s)
```
tlepoint: I'm confused, some functions are `static` while others are renamed with `OQS_`. Is there a reason for this?

patches/kex_rlwe_newhope_avx2/patch1.patch
```diff
@@ -0,0 +1,1073 @@
...
++#endif
++
++#include <oqs/kex.h>
++#include <oqs/rand.h>
++
++#include "kex_rlwe_newhope_avx2.h"
++#include "newhope.c"
++#include "params.h"
++
++OQS_KEX *OQS_KEX_rlwe_newhope_avx2_new(OQS_RAND *rand) {
```
tlepoint: Should this be in the patch?

patches/kex_rlwe_newhope_avx2/patch1.patch
```diff
@@ -0,0 +1,1073 @@
...
++static void poly_uniform(poly *a, const unsigned char *seed)
+ {
+   unsigned int pos=0, ctr=0;
+   uint16_t val;
+@@ -99,9 +99,9 @@ void poly_uniform(poly *a, const unsigned char *seed)
+ }
+ 
+ 
+-extern void cbd(poly *r, unsigned char *b);
++extern void OQS_cbd(poly *r, unsigned char *b);
```
tlepoint: Should that be `static` (after including the c file rather than renaming the function)?

tlepoint: Oh, is it because it is in assembly?

patches/kex_rlwe_newhope_avx2/patch1.patch
tlepoint: I meant, should these two files be in the patch?

patches/kex_rlwe_newhope_avx2/patch0.patch
```diff
@@ -0,0 +1,18 @@
+diff --git a/src/kex_rlwe_newhope/avx2/Makefile.am b/src/kex_rlwe_newhope/avx2/Makefile.am
+new file mode 100644
+index 0000000..baced33
+--- /dev/null
++++ b/src/kex_rlwe_newhope/avx2/Makefile.am
```
tlepoint: Same here, should that be in the patch?

patches/kex_rlwe_newhope_avx2/patch1.patch
s24mishra: Removed the binary files from under test directory.
All the files and changes in the patch should be there because the aim is to keep the external original code untouched as much as possible.

patches/kex_rlwe_newhope_avx2/patch1.patch
s24mishra: All the function which are public and part of API and of course implemented in assembly have to be OQS_ prefixed.

patches/kex_rlwe_newhope_avx2/patch1.patch
s24mishra: Yes.

patches/kex_rlwe_newhope_avx2/patch0.patch
s24mishra: Yes, because autoconf requires the presence of Makefile.am whether or not the algorithm is enabled or not.

PR #177, opened 2017-09-24
Flags for configured algorithms generated in config.h

smashra:


dstebila:
This defines the algorithm flags.  But then we don't do anything based on them.  We should subsequently update test_kex etc. so that it does stuff based on which algorithms are defined  And also the config.h file would need to be part of the install so that relying libraries can also see those macros during their compilation.

s24mishra:
Including config.h under include/oqs and updating kex.h to include config.h.

PR #179, opened 2017-10-16
Enabled and documented building on ARM32 (Raspberry Pi).

christianpaquin:
Enabled and documented building on ARM32 (tested on Raspberry Pi).


dstebila:
@smashra Can autoconf detect which platform it's being run and set an appropriate compiler flag?

README.md
```diff
@@ -223,6 +223,12 @@ Run it from your Android device:
 
 Tested on SM-930F
 
+Building for ARM
+----------------
+
+To build on ARM (such as a Raspberry Pi), certain ciphers must be disabled when running configure:
+
+   	 ./configure --disable-kex-lwe-frodo --disable-kex-mlwe-kyber --disable-kex-ntru --disable-aes-ni
```
dstebila: If autoconf can detect which platform configure is running on, then we could disable these automatically.

dstebila:
It looks like the preprocessor macros `__arm__` (for ARM 32) and `__aarch64__` are defined in both the GNU preprocessor (https://sourceforge.net/p/predef/wiki/Architectures/) and clang (https://stackoverflow.com/questions/23934862/what-predefined-macro-can-i-use-to-detect-the-target-architecture-in-clang#23969845).  That should be reliable enough for the relevant part of ds_benchmark.h.

As for disabling certain ciphers on ARM, do you think we should continue to demand that they be turned off via configure, or should we wrap the entire implementation of those algorithms in platform-specific ifdefs using the macros above?

christianpaquin:
Ok, I'll look into using these macros. As for the disabled ciphers; it would be nice if we could define the --disable-* parameters automatically; I feel like adding ifdefs pollutes the code unnecessarily; is that possible? Note also that I greedily disabled the schemes that failed to compile. Maybe they can also be fixed to support arm; I'll take a closer look.

dstebila:
I agree that adding ifdefs pollutes the code.  We already should have macros for each algorithm, which correspond to the --disable-*/--enable-* configure-time options.  What if we had a header file somewhere which mapped platform pre-processor macros like __arm__ onto our macros for each algorithm?  Then the main source code files would still only have a single algorithm-specific ifdef around them, but we'd still get automatic platform selection as well.

christianpaquin:
The suggested `__arm__` macro works. I also re-tested the disabled schemes, and they work if aes-ni is disabled. I'm happy to merge this as-is, and maybe create another work item to clean up the build system as you propose, @dstebila.

christianpaquin:
Just noticed the build failures are due to prettyprint; which don't work on my machine. Can someone fix? Thanks!

PR #181, opened 2017-10-18
Properly separate SIDH CLN16 from SIDH CLN16 compressed

mvd-ows:
This change will allow a user of the library to specify a SIDH CLN16 variant without worrying about parameters. In the previous code you had to get the algorithm and parameter combination right, and you could even specify the compressed variant and (accidentally) override it by not pairing with the right parameter.


src/kex_sidh_cln16/kex_sidh_cln16.c
```diff
@@ -17,18 +17,15 @@
...
 static int isCompressed(const char *named_parameters) {
-	int compressed = 0; // defaults to non-compressed
-	if (named_parameters == NULL ||
-	    strncmp(P751, named_parameters, strlen(P751)) == 0) {
-		compressed = 0;
-	} else if (strncmp(CompressedP751, named_parameters, strlen(CompressedP751)) == 0) {
-		compressed = 1;
-	}
-	return compressed;
+	if (named_parameters != NULL && strcmp(named_parameters, CompressedP751) == 0) {
```
christianpaquin: Can you use strncmp like in the original code?

mvd-ows: I'd be happy to change it to use strncmp, but could you please elaborate a bit on the reason for checking only the beginning of the parameter string? IMO expected behavior for similar checks is to do exact matching. In the original code, if you try to use "p751compressed" (which is an invalid parameter), it would match on "p751" and use that.

christianpaquin: You are right that the original code had this quirk. I prefer in general to use the strncmp, but since we use strcmp elsewhere in OQS, I'm fine with its use here too.

christianpaquin:
@dstebila, this is ready to merge. I think the update to the Travis script you did last week will fix the Travis failures here. Can the test be re-run?

dstebila:
I'm now getting an error in AppVeyor for this.  Does anyone know what this means?

```
Package source with Name: nuget.org removed successfully.
nuget : The name specified has already been added to the list of available package sources. Please provide a unique name.
At line:1 char:104
+ ...  nuget.org; nuget sources add -name nuget.org -source https://www.nug ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (The name specif... a unique name.:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
git clone -q https://github.com/open-quantum-safe/liboqs.git C:\projects\liboqs

```

christianpaquin:
Do we get this error on all new PRs or just here?

dstebila:
I just re-ran the build and the "nuget" error went away.  The build is now failing with a compiler warning/error, which I think is actually legit:

```
kex_sidh_cln16.c
c:\projects\liboqs\src\kex_sidh_cln16\kex_sidh_cln16.c(63): error C2220: warning treated as error - no 'object' file generated [C:\projects\liboqs\VisualStudio\oqs\oqs.vcxproj]
c:\projects\liboqs\src\kex_sidh_cln16\kex_sidh_cln16.c(63): warning C4090: '=': different 'const' qualifiers [C:\projects\liboqs\VisualStudio\oqs\oqs.vcxproj]
```

mvd-ows:
Thanks, I will look into it.

christianpaquin:
You can fix the warning on Windows by adding 4090 to the pragma directive at the top of the kex_sidh_cln16.c file: `#pragma warning(disable : 4047 4090)`. This will remove the warning-turned-into-an-error on line `k->named_parameters = compressed ? CompressedP751 : P751;` (Visual Studio doesn't like this assignment of const values).

mvd-ows:
Thank you, that may be the quickest fix for now. I realize const correctness isn't a thing in C, but I would have preferred to change the definition of `OQS_KEX` in `kex.h`. Right now it's saying "I will happily point to algorithm names and parameters but only if they reside in a writable page."

mvd-ows:
I see you have already disabled a warning for the whole of this file. I usually disable warnings only for the section of code that needs it. Would that be acceptable or should I add to the list at the top?

christianpaquin:
You can certainly isolate only the affected code.

mvd-ows:
I ended up using your suggested fix. Isolating the code became too bloated for pragma-warning-push-disable-pop since that required two if-def sections. And pragma-warning-suppress would suppress only for the #endif tag so that also wasn't an option.

dstebila:
I'll merge this.  Odd, what did you think we should change in OQS_KEX in kex.h?

mvd-ows:
In the OQS_KEX definition there are two members `method_name` and `named_parameters` of type `char *`. It would be slightly better to use the type `const char *` and get rid of a few `strdup` at the same time. At least the algorithm name will always be known and can/should therefore reside in a read-only segment of the executable.

I'm a little surprised that GCC did not warn when assigning a pointer of type `const char *` to a different pointer of type `char *`.

PR #182, opened 2017-10-20
Added VisualStudio DLL build configurations

nthuemmel:
As in the current master branch, the Visual Studio build files only allow the creation of static libraries and statically linked executables. Since I needed bindings for other languages (such as Java), which in turn required dynamic linking libraries (DLLs), I added DLL build configurations to the Visual Studio solution.

All build configurations (DLL and non-DLL builds, x86 and x64) have been tested with the standard settings using Visual Studio 2013 on a 64-bit Windows 7. I have updated my changes to be on-par with the latest Visual Studio build file changes of commit 40ffb4eb38bc80ecd6d1e42848ed7145a2e49946 and squashed them for better readabilty.

Additional changes which were required to get the DLL builds working:

* Added inline preprocessor definition to VisualStudio project file - this allows compilation with older versions of VisualStudio (e.g. 2013) which do not support the C99 inline keyword
* Removed newhope.c from VisualStudio build files, as it is already included in kex_rlwe_newhope.c and will lead to multiple symbol definition errors in shared library builds
* Added ENABLE_CODE_MCBITS guard to kex_code_mcbits.c to protect against missing symbol definition errors in VisualStudio shared library builds
* Added VisualStudio shared library exports file dll.def, including the necessary core functions to use oqs from applications
* Also export free when building Windows DLLs - this is important as the correct runtime library has to be used to free memory allocated by the OQS library. Applications may be using multiple runtime libraries or memory managers, calling the wrong free will result in memory corruption or segfault


dstebila:
Thanks Nikolaus!  We will try to look at this over the next few days and get some feedback on the pull request.

christianpaquin:
I'll take a look next week.

nthuemmel:
Thanks for the feedback and the newline fix!

PR #183, opened 2017-10-24
Initialize out-parameters to NULL

mvd-ows:
This changeset mainly adds code to consistently initialize out-parameters to NULL. This is important so we don't call free() on any invalid pointers in the failure path.

I also added code that clears out-parameters on failure, where such code was not already in place. It's generally considered good manners to not leak invalid pointers, even when returning an error code.

SIDH IQC REF has not been reviewed for this changeset since it doesn't consider failures at all, and would need a larger restructuring to get up to par.

### DIFF
--- a/src/kex_lwe_frodo/kex_lwe_frodo_macrify.c
+++ b/src/kex_lwe_frodo/kex_lwe_frodo_macrify.c
@@ -13,7 +13,7 @@ OQS_KEX *MACRIFY(OQS_KEX_lwe_frodo_new)(OQS_RAND *rand, const uint8_t *seed, con
 
 	k = malloc(sizeof(OQS_KEX));
 	if (k == NULL) {
-		goto err;
+		return NULL;
 	}
 	k->named_parameters = NULL;
 	k->method_name = NULL;
--- a/src/kex_lwe_frodo/kex_lwe_frodo_macrify.c
+++ b/src/kex_lwe_frodo/kex_lwe_frodo_macrify.c
@@ -13,7 +13,7 @@ OQS_KEX *MACRIFY(OQS_KEX_lwe_frodo_new)(OQS_RAND *rand, const uint8_t *seed, con
 
 	k = malloc(sizeof(OQS_KEX));
 	if (k == NULL) {
-		return NULL;
+		goto err;
 	}
 	k->named_parameters = NULL;
 	k->method_name = NULL;
--- a/src/kex_ntru/kex_ntru.c
+++ b/src/kex_ntru/kex_ntru.c
@@ -90,6 +90,7 @@ int OQS_KEX_ntru_alice_0(UNUSED OQS_KEX *k, void **alice_priv, uint8_t **alice_m
 	ntru_alice_priv = malloc(sizeof(OQS_KEX_ntru_alice_priv));
 	if (ntru_alice_priv == NULL)
 		goto err;
+	ntru_alice_priv->priv_key = NULL;
 	*alice_priv = ntru_alice_priv;
 
 	/* calculate length of public/private keys */
@@ -119,7 +120,7 @@ int OQS_KEX_ntru_alice_0(UNUSED OQS_KEX *k, void **alice_priv, uint8_t **alice_m
 
 err:
 	ret = 0;
-	if (ntru_alice_priv != NULL && ntru_alice_priv->priv_key != NULL)
+	if (ntru_alice_priv != NULL)
 		free(ntru_alice_priv->priv_key);
 	free(ntru_alice_priv);
 	*alice_priv = NULL;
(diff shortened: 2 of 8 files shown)


src/kex_ntru/kex_ntru.c
```diff
@@ -78,10 +78,13 @@ int OQS_KEX_ntru_alice_0(UNUSED OQS_KEX *k, void **alice_priv, uint8_t **alice_m
 	DRBG_HANDLE drbg;
 	OQS_KEX_ntru_alice_priv *ntru_alice_priv = NULL;
 
+	*alice_priv = NULL;
+	*alice_msg = NULL;
+
 	/* initialize NTRU DRBG */
 	rc = ntru_crypto_drbg_instantiate(256, (uint8_t *) "OQS Alice", strlen("OQS Alice"), (ENTROPY_FN) &get_entropy_from_dev_urandom, &drbg);
```
mvd-ows: NTRU doesn't define a value to represent an invalid DRBG_HANDLE. Thus, `ntru_crypto_drbg_uninstantiate()` can only safely be called if `ntru_crypto_drbg_instantiate()` was successful.

src/kex_lwe_frodo/kex_lwe_frodo_macrify.c
```diff
@@ -13,7 +13,7 @@ OQS_KEX *MACRIFY(OQS_KEX_lwe_frodo_new)(OQS_RAND *rand, const uint8_t *seed, con
 
 	k = malloc(sizeof(OQS_KEX));
 	if (k == NULL) {
-		goto err;
+		return NULL;
```
mvd-ows: This change isn't essential. Both goto/return would work either well.

src/kex_sidh_cln16/kex_sidh_cln16.c
```diff
@@ -82,8 +82,9 @@ int OQS_KEX_sidh_cln16_alice_0(OQS_KEX *k, void **alice_priv, uint8_t **alice_ms
 	uint8_t *alice_tmp_pub = NULL;
 
 	if (!k || !alice_priv || !alice_msg || !alice_msg_len) {
-		return NULL;
+		return 0;
```
mvd-ows: NULL happens to be represented the same as 0, but it's not the same type.

src/kex_sidh_cln16/kex_sidh_cln16.c
```diff
@@ -126,34 +127,33 @@ int OQS_KEX_sidh_cln16_alice_0(OQS_KEX *k, void **alice_priv, uint8_t **alice_ms
 
 err:
 	ret = 0;
-	if (alice_msg && *alice_msg) {
```
mvd-ows: All contemporary implementations of `free()` can be called with a NULL pointer. The majority of code in liboqs does not check to see if the pointer is NULL before calling `free()`.

src/kex_lwe_frodo/kex_lwe_frodo_macrify.c
dstebila: I'd rather keep this as goto err.  While it's currently equivalent and no cleanup is required, if some later edit adds code above this line, they may no longer be equivalent, and changing it back to goto err might get forgotten.

src/kex_sidh_cln16/kex_sidh_cln16.c
dstebila: Good point, thanks.

src/kex_ntru/kex_ntru.c
```diff
@@ -116,10 +119,12 @@ int OQS_KEX_ntru_alice_0(UNUSED OQS_KEX *k, void **alice_priv, uint8_t **alice_m
 
 err:
 	ret = 0;
-	if (ntru_alice_priv != NULL)
+	if (ntru_alice_priv != NULL && ntru_alice_priv->priv_key != NULL)
```
dstebila: Below you wrote "All contemporary implementations of `free()` can be called with a NULL pointer. The majority of code in liboqs does not check to see if the pointer is NULL before calling `free()`.".  Does that also apply here for `ntru_alice_priv->priv_key`?

src/kex_lwe_frodo/kex_lwe_frodo_macrify.c
mvd-ows: I agree. I can't remember exactly what I was thinking but it must have been something along the lines of unifying all OQS_KEX_xxx_new functions. It would be much better to update all such non-trivial functions to use an err label.

src/kex_ntru/kex_ntru.c
mvd-ows: Absolutely, good catch. That must have been a knee-jerk reaction to try and deal with `ntru_alice_priv->priv_key` not being initialized. The real problem is of course if `ntru_crypto_ntru_encrypt_keygen` fails we end up here with an uninitialized pointer in `ntru_alice_priv->priv_key`. I will update the code.

PR #186, opened 2017-10-26
Add checks to verify length of input data for McBits

mvd-ows:
This change adds the bare minimum checks to ensure input data is valid. In the specific case of McBits providing too little data will result in a memcpy operation reading beyond the buffer. A similar read will typically return garbage data but could also induce a crash if it crosses into a guard page or uncommitted page.


mvd-ows:
That's an interesting failure. Is it part of the SIDH CLN16 tests or is there a missing separator between that and the next test?

PR #190, opened 2017-11-08
Added optimized version of Picnic

christianpaquin:
Implementation from https://github.com/IAIK/Picnic.git, replacing current one.

Some notes:
 - Picnic is now a git submodule, built separately (build_picnic.sh) and linked into OQS. cmake is required to build it.
 - The params are part of the Picnic library, no need to generate them separately.
 - The Picnic params have been renamed.
 - Picnic has been disabled in the Visual Studio build (until the library supports VS).


christianpaquin:
I added the picnic build instructions to the configure script, so this is ready to merge.

christianpaquin:
Travis is failing because build_picnic can't be run:
   ./.travis-tests.sh: line 61: ./build_picnic.sh: Permission denied
I'm not sure how to give proper permission to the script; @dstebila, who would know how to fix this?

vsoftco:
@christianpaquin Can you try doing "sh ./build_picnic.sh" instead?

dstebila:
The output in the Travis build logs is quite weird.  For example, take a look at https://travis-ci.org/open-quantum-safe/liboqs/jobs/312708238.  You'll see that Picnic looks like it's being built 4 or 5 times.  The first time it's built there are a bunch of compiler warnings (lines 656-787), then the patch is applied again and whole bunch of patch failures (lines 1018-1275) (probably because the patch was already applied), then more patch failures, then more compiler warnings, and so on and so forth.

The multiple patch failures suggests to me we're building more times than we need to.  The fact that there are compiler warnings that don't cause a Travis failure is also unclear.

christianpaquin:
I'll take a look

christianpaquin:
I updated the external Picnic lib, removed the duplicated picnic build script invocation, and changed a few things to avoid re-patching if it was already done. Picnic is only built in the configure step. However, it seems that part of the configure script is called during the make process, therefore trying to rebuild Picnic. @dstebila, can this be avoided?

christianpaquin:
I tried this PR from a fresh repo and I didn't see my double-building issue. This should be good to go.

Issue #194, opened 2017-11-23
scan-build errors


ETeissonniere:
Hey,
I have just runned `scan-build` from `clang` via `scan-build -analyze-headers --use-cc /usr/bin/clang --use-c++ /usr/bin/clang++ make -j4`, it found a few errors and I thought it may interests you.
This includes:
 -  2 double free
 -  1 memory leak
 -  2 dead assignments

![capture du 2017-11-23 15-27-32](https://user-images.githubusercontent.com/10683430/33177435-df5f98ac-d062-11e7-9af9-fce09bf49b67.png)

xvzcf:
Having re-run this at [this commit](https://github.com/open-quantum-safe/liboqs/commit/25c9f8cf745c833fdac7536fa0765d4911435526) on master, there are 14 reported bugs. I'll investigate their validity as I move the build system to CMake.

dstebila:
Noting that an old PR, #195, contained some code related to doing this in our old build system.  May want to replicate that work when working on this issue.

baentsch:
Also closed by #1240

PR #195, opened 2017-11-23
Add scan-build to travis

ETeissonniere:
To help prevent other issues like #194 I decided to add the scan-build checks on travis.

Note: it is normal that that the travis won't pass here, since scan-build's problems aren't solved yet.


dstebila:
We won't be applying this PR as is, so I'm closing it.  But the overall goal of running scan-build is not dead, and #194 continues to track that goal.

PR #202, opened 2018-01-02
Add new SHA3 implementation from John Underhill

dstebila:


dstebila:
For some reason this breaks Kyber.  I created another branch (debugging-replace-sha3) of this pull request which contains a copy of the old SHA3 code we had, and is instrumented to compare every call made to any SHA3 function by Kyber to see where there are differences.  The macros at the top of that branch's src/kex_mlwe_kyber/old_sha3.c file show which lines of code create different behaviour in old SHA3 and the new SHA3 code, and which lines cause Kyber to break -- in particular, the actual SHA3-256 computation is wrong in the old code, and when it is made correct, Kyber no longer works.  But having identified that information, I don't actually now how to fix it.  @tlepoint, any thoughts?

Steppenwolfe65:
Hi Douglas,
I noticed the break in Kyber as well (but wasn't sure if it was due to my own tinkering, or part of the original problem). I don't think it has anything to do with SHA3, but rather, that the coefficient arrays of *a* were all identical in the 'working' version (is the public key being truncated/misaligned somehow?). I did implement the other version of Kyber, using a correct implementation of SHAKE, and that appears to work properly.

Steppenwolfe65:
OK.. I made a standalone version using the pq-crystals implementation as a base, and integrated your random provider and sha3, and it's working fine now. As to the current problem, I believe it is in polyvec.c . 
Compare your version with that one: https://github.com/pq-crystals/kyber/blob/master/ref/polyvec.c
You will notice that the KYBER_K constant which sets the outer bound for loops in various functions in that class, has been replaced with the KYDER_D constant. This should work properly if D=K=3, but in your implementation, K=4 ('paranoid mode'), which causes the polynomials to be under-processed.
There are a number of other changes that I am unsure of (cbd, poly_getnoise, gen_matrix..), enough that I felt it was better to begin again, using that version as a starting point.
I'll create a new project (Misc C) tomorrow, use it house the sha3 and the Kyber implementation.
I still have to go over it thoroughly (another 'tune-up'), and make it a drop-in replacement, but I should have this done by Sunday.
I plan to add the cSHAKE change (but working this time ;) because it is actually an improvement to the original.

dstebila:
Thanks for the analysis, John.  @tlepoint, does that match what you know about the Kyber code?  I know @tlepoint and I are planning to eventually replace the Kyber implementation currently in liboqs with the version submitted to NIST, so there may be diminishing returns on trying to diagnose this particular Kyber code.

Steppenwolfe65:
I'll write it anyways, as a stand-alone implementation based on the Crystals branch, nothing fancy, just unify standard int, and remove any obvious secure coding standards violations. It's really not a big deal, since I am implementing ModuleLWE in C++ anyways, and it is part of my process (picking it apart, and trying to reconcile the paper to what I see in code, before transitioning it to C++). 
One thing to watch out for, in the gen_matrix loop, where shake_absorb is called, you must zero the state just before that call (inside the loop). This was a change they made to fips202.c, adding that zero-ing loop directly to keccak_absorb, which is what broke cSHAKE, and kicked this whole thing off.

Steppenwolfe65:
Hi Douglas,
I gave the pq-crystals branch of Kyber a tune-up, and it is now posted in the new repo, Misc-C:
https://github.com/Steppenwolfe65/Misc-C/tree/master/Kyber

I also made a couple of small changes to the sha3 class to make it Misra compliant, (and branched it to a second standalone implementation with a namespace truncation), your files have moved to SHA3/OQS.

I ran a PC-Lint/VisualLint scan and this version of Kyber is now Misra C 2004 compliant (I'll use a Coverity scan as well, finish that tomorrow, and give it one more look..).
Many small changes; unified standard int, doxygen documentation, fixed a few small defects (wrong int size, negating unsigned integers, etc). Removed unused headers, added missing brackets and braces.. all fixed now.
Also, I got the cSHAKE option implemented, it can be turned on or off via a def in params.h (reverts to SHAKE).

I'll use this as a starting point for my C++ implementation, and now there is a version out there that works in Visual Studio..

@dstebila If you want to use it, let me know, and I can download a new copy of your library, integrate it, and put a [temporary] zip in the Kyber directory..

Cheers,
John

dstebila:
Replaced by #244.

PR #204, opened 2018-01-02
Add a NIST-compliant thread-safe randombytes based on OQS_RAND.

dstebila:


dstebila:
Grrr... now we have a namespace collision since libsodium includes randombytes and McBits includes libsodium.  Should we instead go with OQS_randombytes?

Steppenwolfe65:
Hi Douglas.. I can update this system provider to suit your library:
https://github.com/Steppenwolfe65/Misc-C/blob/master/Kyber/Kyber/sysrand.c

I can also add an RDRAND provider, a system-entropy collector- SHAKE(timers | statistics | handles), and what I would recommend; an auto provider that combines all three. I'll look at NIST recommendations for TRNGs and create them to that specification, and could have it done by next weekend. Let me know if that would be helpful.

PR #221, opened 2018-01-31
vsoftco refactor #ifdef WINDOWS

vsoftco:
Summary:

1. Got rid of code that was using `#if defined(WINDOWS)` as it's not portable, used `#if defined(_WIN32)` instead. Before refactoring, we manually added `WINDOWS` as a pre-processor define in VS build. There's no need, `_WIN32` is automatically defined under VS (on both 32bit and 64bit platforms) and this refactoring relies on the later.

2. Use an `enum` for `OQS_STATUS` instead of #define's, much more type-safe.

3. Also, changed signatures from `int main()` to `int main(void)` to make it C-standard compliant.

4. make prettyprint

### DIFF
--- a/VisualStudio/oqs/oqs.vcxproj
+++ b/VisualStudio/oqs/oqs.vcxproj
@@ -226,7 +226,7 @@
       </PrecompiledHeader>
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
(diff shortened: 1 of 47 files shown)


christianpaquin:
I'll take a look later today.

vsoftco:
@christianpaquin Thanks. Build is fine now (make prettyprint was missing).

vsoftco:
@christianpaquin Thanks Christian!

PR #222, opened 2018-01-31
added const to uint8_t *str pointer in OQS_print

vsoftco:
code safety


vsoftco:
minimal change, added `const` for const correctness in `OQS_print(_part)_hex_string`

PR #226, opened 2018-02-07
Added post-build step to merge the oqs and picnic library on Windows

christianpaquin:
Added post-build step to merge the oqs and picnic library in x64 static VS projects, to simplify downstream integration (mirroring the Linux build). This is needed for cleaner OpenSSL integration.

Also removed forced rebuild params of picnic in oqs.


christianpaquin:
@vsoftco, can you take a look. I just tested with my OpenSSL Windows project, and it solves a linking issue.

vsoftco:
@christianpaquin I'll take a look later today, thanks!

PR #231, opened 2018-02-16
Included picnic directly (vs. as cmake-built 3rd party lib).

christianpaquin:
Now uses OQS's rand; modified to avoid pedantic warnning; modified build and CI scripts (removed cmake dependency).

This not only simplifies OQS, but the downstream projects which won't need to deal with the picnic dependency.


christianpaquin:
@dstebila, @vsoftco, can one of you try this PR on the mac? I tried with clang on Ubuntu and it works, so the Travis failures must be mac-specific. Can you share with me the content of src/sig_picnic/external/build/CMakeFiles/picnic.dir/flags.make after building the master branch of OQS? Perhaps some extra flags need to be set on mac for this PR.

vsoftco:
@christianpaquin Will try today

christianpaquin:
Awesome, thanks @vsoftco .

PR #233, opened 2018-03-02
Fix build on ARM

christianpaquin:
Fix build on ARM (that has been failing since Picnic was merged directly into OQS).
 - added logic to configure.ac to detect presence on ARM
 - added logic to picnic's Makefile.am to properly set flags
 - removed arm-specific restriction from the readme

Welcoming suggestions on better configure/makefile logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,13 +203,6 @@ Run it from your Android device:
 
 Tested on SM-930F
 
-Building for ARM
-----------------
-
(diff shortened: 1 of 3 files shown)


christianpaquin:
I re-ran the stalled tests; all green now. Once this gets merged, I'll update the OpenSSL fork so it can build on ARM.

PR #244, opened 2018-03-06
Replace SHA-3 implementation, take 2

dstebila:
This aims to do what #202 did, but is being redone against the current version of master.

Note that this pull request is not yet complete, as @Steppenwolfe65 has since pushed some updates to his SHA-3 implementation that we should pull in. (https://github.com/Steppenwolfe65/Misc-C/tree/master/SHA3/SHA3)


Steppenwolfe65:
Please note that in the newest version of sha3, the position of the state variable has changed in some of the api. Whereas previous iterations had the state positioned as the last array in the call (ex. shake128_squeezeblocks((unsigned char*)buf, nblocks, state)), now all api in that file has the state always in the first position. This was changed to provide a consistent interface across all of the keccak function calls.

README.md
```diff
@@ -235,15 +236,14 @@ We are also interested in assistance from code reviewers.
...
-
-
 License
 -------
 
 liboqs is licensed under the MIT License; see [LICENSE.txt](https://github.com/open-quantum-safe/liboqs/blob/master/LICENSE.txt) for details.  liboqs includes some third party libraries or modules that are licensed differently; the corresponding subfolder contains the license that applies in that case.  In particular:
 
 - `src/crypto/aes/aes_c.c`: public domain
 - `src/crypto/rand_urandom_chacha20/external`: public domain
+- `src/crypto/sha3`: public domain
```
christianpaquin: I haven't seen a statement on https://github.com/Steppenwolfe65/Misc-C that the code is in the public domain. Do we have authorization to use this code?

Steppenwolfe65:
Everything in Misc-C is public domain, (though I may mark it as CC0 for the free Coverity scan, essentially the same thing), or defaults to the original license. Feel free to use whatever you like..

README.md
christianpaquin: Acknowledging @Steppenwolfe65 comment below; approving review.

PR #245, opened 2018-03-08
Adding BIG_QUAKE implementation from NIST

s24mishra:


dstebila:
Thanks Shravan. 

I factored out your build-archive script into several individual scripts under `scripts/` so that it was easier to write the relevant tasks on both Linux and macOS (47f36a1 and ea810c8).  Also factored out the doxygen stuff into a separate commit (4c26986).  

This PR fails with Appveyor because we're not trying to build this on Windows; 55c382c8 has fixed it so any branches with the word `nist` in them won't trigger Appveyor, but for this PR we'll just ignore the Appveyor failure.

The final blocker to merging is licensing.  I can't see a license in the BIG QUAKE source or website; I've emailed the authors to find out what if any open source license it's under.

dstebila:
There's an updated version of BIG QUAKE from the authors at https://bigquake.inria.fr/source-code/ which contains licensing information (it's BSD).  We should update to include that.

PR #257, opened 2018-03-26
Try changing algorithm identifiers to strings

dstebila:
I was using an enum for algorithm identifiers, and some code assumed that these were sequential.  Assuming sequential means that if we want to remove an algorithm, we can't have gaps. So we'd have to reassign numbers.  But then this would mean binary incompatibility between different versions of liboqs nist-branch.  We also could not have binary compatibility with liboqs master (once master gets the new KEM API) because a different order of KEMs.  So I am trying strings for algorithm identifiers.


vsoftco:
Code is OK after all, tricky thing with the string. See https://stackoverflow.com/questions/49634832/memory-leak-by-returning-pointer-to-local-c-style-array#49634907 for more details.

PR #270, opened 2018-04-11
BIKE

drucker-nir:
Hi,

This update includes the reference code (in C) of BIKE1 and BIKE3 in levels L1/3/5. It also includes BIKE2 but it is disabled for now.


dstebila:
I've reverted the line wrapping changes. Can you check if your `make prettyprint CLANGFORMAT=clang-format-3.9` munges them again?  If not then perhaps it was an overly zealous text editor that rewrapped.

dstebila:
I also tried running `./test_kem` on my Mac and it also hung.  I turned on VERBOSE=3 and here are the first 10k lines of the log file; I'm not really sure how to interpret them, but maybe the algorithm experts can.  I let it run for about 10 seconds and it produce 6M lines of debug messages totally 198MB, I can send those to you if you want.  :)
[bike.log](https://github.com/open-quantum-safe/liboqs/files/1900951/bike.log)

PR #274, opened 2018-05-02
Nist ledakem

smashra:


dstebila:
Shravan, why did you remove the LEDAkem KAT checks in scripts/check_kats.sh?  If they aren't working then we should fix it, not remove the check.

smashra:
Sorry, that might have been by mistake.

dstebila:
Okay, so can you check on why the KAT tests are failing for LEDAkem?

smashra:
I'll take a look again.

dstebila:
Unless we get a resolution within the next day, this will not be part of the 2018-05 snapshot release of nist-branch.

dstebila:
Thanks for looking at this, Shravan. Are these new KATs from the original LEDAkem source code, or did you generate them afresh?

smashra:
I generated them afresh.

Sincerely
Shravan

dstebila:
We should be trying to get the code to generate the same KATs as in the original submission -- that way we know it's actually matching the original implementation correctly, and that we have not introduced any errors.

dstebila:
If you just remove the rng.c file from the upstream source directory, will it build without you having to do the weird strip function which is causing the failure?

smashra:
Rng.c contains other functions used by the code.

I have commented the random bytes portion of the code.

I’ll push the updated code it doesn’t contain strip function now.

But it still doesn’t work with - j8 .

I’ll investigate further.

dstebila:
The reason it doesn't work with make -j8 is as follows. make -j8 will try to build multiple targets in parallel, such as the `kem_ledakem_C1_N04` and `kem_ledakem_C1_N03` targets. Both of those targets involve cd'ing into the same directory and running make with different parameters.  So there may be two or more parallel processes building object files in the same directory, which leads to confuse.  FrodoKEM had a similar problem: I used a technique where the upstream directory was duplicated for each target prior to building, take a look at src/kem/froodokem/Makefile for details.

smashra:
Thanks, will try that.

smashra:
Ready to merge.

smashra:
Ready for merge

PR #275, opened 2018-05-15
Try to support multi-threaded make

dstebila:


crockeea:
Works for me. Thanks!

PR #277, opened 2018-05-15
Use OQS_MEM_cleanse to zero-out memory

dstebila:


dstebila:
Is this what you wanted, @vsoftco?

Issue #280, opened 2018-05-15
Add SABER


dstebila:
https://github.com/Angshumank/SABER

License confirmed to be public domain by Sujoy Sinha Roy

PR #282, opened 2018-05-16
Add SABER KEMs to nist-branch

dstebila:
See issue #280


dstebila:
@christianpaquin would you be able to give this a quick review/test before I merge it?

PR #287, opened 2018-05-22
Adding BIKE additional implementations (AVX2/AVX512)

drucker-nir:


dstebila:
When BIKE is built in Travis, is it building the reference version or one of the AVX versions?  Ideally our continuous integration platform would exercise both code paths.

PR #295, opened 2018-06-05
Add new KEM API and SIKE

dstebila:


Makefile.am
```diff
@@ -164,6 +188,7 @@ test_sha3_CPPFLAGS += $(AM_CPPFLAGS)
 
 test: clean-tests
 	make
+	./test_kem
```
christianpaquin: Doesn't need a quiet option like the others?

Issue #297, opened 2018-06-26
Non-fixed key sizes for signature schemes


manvari:
In `sig/sig.h`, the `OQS_SIG` structure expects fixed sizes for public and private keys (`priv_key_len` and `pub_key_len`). These sizes are to be set when an instance of the signature algorithm is created via a call to `OQS_new`, and are then used to allocate sufficient memory for keys before calling the algorithm's keygen method. However, some QS signature schemes have varying key sizes (e.g. NIST candidate [Falcon](https://falcon-sign.info/), which features optional compression for private keys).

Considering this, it could be better to let the algorithm's keygen method modify the key size fields appropriately. This would also require adding a `max_key_len` field to `OQS_SIG` so as to enable the caller to allocate memory before the keygen method is called.

dstebila:
Hi Maxime! Thanks for the feedback. We're actually working on a revision of our signature API to simplify the OQS_SIG structure and to bring it closer to the NIST/SUPERCOP API.  We should have a candidate version of the new API for feedback in the next week or so -- @christianpaquin has been working on it.  I'll post a comment here once we've got something for you to look at.

manvari:
Sure, I will look at the new API when it rolls out. In the meantime, I'll be checking for updates to the `master-new-api` branch. :)

dstebila:
@manvari does the current status of #298 satisfy your concerns in this issue?

manvari:
@dstebila I have mixed feelings about this one!

The signature API changes introduced by @christianpaquin in #298 help OQS stick to the NIST API. This definitely is the way to go, considering all NIST submissions conform to this API. This makes integrating these submissions into OQS very easy; only a thin wrapper, with no superfluous parameters added, is required.

On the other hand, by fully adhering to the NIST API, we get to miss out on some of the optimizations that submission authors may have made. In the case of this issue, that would be forgoing shorter public/private keys even though the underlying algorithm may offer them. I have two possible solutions in mind:

1. Add a `void *options` argument to the algorithm's `keypair` function in `OQS_SIG`, where the data to pass is defined in the algorithm's OQS integration (`src/sig/sig_{algname}.h`, presumably).\[\*\] Then, a first call to the `keypair` with `public_key` and `secret_key` set to `NULL` along with a valid `options` argument would modify `options` to indicate actual key sizes. A second call with the modified `options` argument and non-`NULL` values for `public_key` and `secret_key` will then generate the keypair.<br>
[...]

PR #298, opened 2018-07-03
Adding sig to nist-branch

christianpaquin:
Added sig API with Picnic as first scheme


christianpaquin:
@dstebila; you can look at this PR. As always, the prettyprint doesn't work on my machine, and it's blocking the first Travis test; could you "make prettyprint" for me, please?

manvari:
Looking at the API changes in `src/sig/sig.h`, it appears that the `OQS_SIG` structure still uses a maximum signature size parameter (namely, `length_signature`).

Similarly to #297, this is a limitation when dealing with variable length signatures. In order to facilitate the integration of schemes with variable length signatures into liboqs, I suggest replacing this with something similar to the NIST API's `CRYPTO_BYTES` parameter. Since it quantifies the maximal amount of "bytes of overhead in a signed message compared to the original message", it makes it that much easier to deal with variable length signatures.

Should we do this, we also get to stick closer to the NIST API, which -- from my understanding -- is @dstebila's intention.

christianpaquin:
@manvari, the `length_signature` value actually uses the `CRYPTO_BYTES` value directly from the upstream implementations, so it is meant to achieve what you say. The name is not very clear, so perhaps I should rename it to `length_signature_overhead` (I find "crypto bytes" to be also confusing :)).

As you noticed, I didn't have time to work on #297; still in the queue...

manvari:
@christianpaquin Thanks for pointing this out! I initially thought `length_signature` matched NIST's `CRYPTO_BYTES`, but the comment above the declaration `size_t signature length` (in `src/sig/sig.h`) reads: 
> /** The (maximum) length, in bytes, of signatures for this signature scheme. */

Along with your suggestion of renaming the value, this could be swapped out for something along the line of:

> /** The length, in bytes, of the maximum overhead in a signed message compared to the original message for this signature scheme. */

I could have looked at Picnic further to see that `#define OQS_SIG_picnic_L1_FS_length_signature 34004` (`src/sig/picnic/sig_picnic.h`) does match `#define CRYPTO_BYTES (4 + 34000)` (`src/sig/picnic/.../picnicl1fs/api.h`). Alas :)

christianpaquin:
Marked no ready for merged. I'm adding qTesla on this PR.

christianpaquin:
Now ready for merged; modulo the Travis checks. Added qTESLA, fixed a few things. @dstebila, I don't know why the first Travis test fails; I don't get an error locally when running make pre-push. I also don't know why the third Travis test fails (on macOS). Perhaps you can identify the problem while reviewing; if not, I can pull out my macOS machine and investigate.

dstebila:
I've got the build fixed for Ubuntu but the build fails on my Mac and on Travis' Mac build, with the same error about `aligned_alloc`.  I see that there are some macros in Picnic's macros.h files that is trying to auto-detected `aligned_alloc`, but they seem to not be working with clang on macOS.  I played around for a bit but couldn't figure out how to get it to work.

christianpaquin:
OK, thanks, I'll investigate.

christianpaquin:
The original Picnic code compiles on macOS (as integrated in the master branch), but the NIST package doesn't out of the box. We can either disable Picnic (and other such packages) in Travis, or integrate more code from Picnic (although this wouldn't be the package as submitted to NIST).

dstebila:
It's fine if we use the authors' preferred version, even if that differs slightly from the NIST submission.  For several of the KEMs on nist-branch already, we're using versions from the authors' websites rather than the exact NIST submission.

christianpaquin:
@dstebila, I disabled picnic and made qtesla the new default, in order to pass merge this PR. Picnic is still available and works on Linux. I'll fix the macOS build in a different PR, in order not to block further progress on sig API.

christianpaquin:
I got an update from the picnic team and integrated the new bits to fix the failing macOS build. One last issue however: aligned_alloc and aligned_free are internal functions that only appear in the macOS build (they are ifdef'ed out in the Linux build). If I had them to symbols_local.txt the build works on macOS and fails on Linux; if I omit them, we have the opposite situation (see last two commits).

@dstebila, do you have a suggestions to deal with that situation?

dstebila:
Oh, that's annoying.  Next week I'll take a look at the scripts that use symbols_local.txt and see if there is way to make them so that they don't fail if they are applied with name that isn't present, should be solvable.

christianpaquin:
I can disable Picnic once more if you want to merge this PR now; and we can fix this on a separate one.

dstebila:
Well maybe let's disable Picnic on macOS for now so we can land this and resolve the issue later.

christianpaquin:
Done; I disabled Picnic and left a comment in the Makefile. Ready to merge again.

dstebila:
BTW @manvari will you be able to resubmit your PR for Falcon against the landed version now on nist-branch?

PR #305, opened 2018-07-10
Makefile: automatic detection of binutils version via ld

### LINKED ISSUE #290: Automatic detection of binutils version

manvari:
Fixes #290

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@ ifeq ($(arch), "x64")
   CFLAGS+= -arch x86_64
 endif
 
+BINUTILS_VER=$(shell ld -v | grep -o "[0-9][0-9]*.[0-9][0-9]*")
 ifneq (,$(BINUTILS_VER))
   ifeq ($(shell expr $(BINUTILS_VER) \>= 2.26), 1)
     SUPPORTED_BINUTILS=1


Makefile
```diff
@@ -49,6 +49,7 @@ ifeq ($(arch), "x64")
   CFLAGS+= -arch x86_64
 endif
 
+BINUTILS_VER=$(shell ld -v | grep -o "[0-9][0-9]*.[0-9][0-9]*")
```
dstebila: This doesn't give meaningful output with the ld present on macOS.  `ld -v` on my Mac outputs:

~~~
@(#)PROGRAM:ld  PROJECT:ld64-351.8
configured to support archs: armv6 armv7 armv7s arm64 i386 x86_64 x86_64h armv6m armv7k armv7m armv7em (tvOS)
LTO support using: LLVM version 9.1.0, (clang-902.0.39.2) (static support for 21, runtime is 21)
TAPI support using: Apple TAPI version 9.1.0 (tapi-902.0.9)
~~~

While this happens to result in `SUPPORTED_BINUTILS` not getting set to 1, is there a more robust way we can do this?

manvari: I don't have access to a Mac, but we could try checking either `ar --version` (which might result in a similar output to `ld -v`) or `ls /usr/lib/libbfd*` (Binary File Descriptor library, which is part of binutils).

On Arch Linux, `ls /usr/lib/libbfd*` returns `/usr/lib/libbfd-2.30.so  /usr/lib/libbfd.a  /usr/lib/libbfd.so`.

dstebila: @drucker-nir Should BIKE build on macOS with some appropriate version of GNU binutils or the llvm toolchain?  If so then we need to find a way to detect this appropriately on macOS.  If not then we can just focus on detecting it in GNU binutils on Linux.

drucker-nir: BIKE AVX2/AVX512 implementations were built for Linux, and were tested on Linux. We will consider adaptation for MAC OS in the future.

PR #319, opened 2018-07-24
Adding CRYSTALS-Dilithium to the NIST branch

tlepoint:
Adding CRYSTALS-Dilithium to the NIST branch. 

Two things:

1. I had to modify the `Makefile`'s of BIKE. They do not work if the path has a whitespace.

2. I have a surprising bug during the cleaning of the signature.
```
================================================================================
Sample computation for SIG Dilithium_II_medium
================================================================================
verification passes as expected
test_sig_shared(84761,0x7fffb73783c0) malloc: *** error for object 0x7faacee00228: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
.travis/all-tests.sh: line 52: 84761 Abort trap: 6           ./test_sig_shared
```

I have yet to pin-point whether it comes from the Dilithium implementation or the OQS wrapper.


tlepoint:
I found the problem! In Dilithium's `crypto_sign_open`, we expect `message` to be of size `signed_message_length` and not of size `message_length`. I added the fixaround in the `test_sig.c`; I'll discuss internally to see whether we want to change our code.

tlepoint:
I asked Peter, this is not a mistake; cf. the [nacl API](http://nacl.cr.yp.to/sign.html):
> The maximum possible length `mlen` is `smlen`. The caller must allocate at least `smlen` bytes for `m`. 

The current PR fixes this.

dstebila:
Thanks Tancrède!  I may not be able to review this today but should be able to get to it tomorrow.

src/sig/example_sig.c
```diff
@@ -30,7 +30,7 @@ static OQS_STATUS example_stack() {
 
 	uint8_t public_key[OQS_SIG_qTESLA_I_length_public_key];
 	uint8_t secret_key[OQS_SIG_qTESLA_I_length_secret_key];
-	uint8_t message[MESSAGE_LEN];
+	uint8_t message[MESSAGE_LEN + OQS_SIG_qTESLA_I_length_sig_overhead];
```
dstebila: Would this be a clearer example if we had separate `message` and `opened_message` variables, where `message` remains of length `MESSAGE_LEN` and is only used as input to `sign`; `opened_message` would be of length `MESSAGE_LEN` + `...overhead` and would only be used as the output to `sign_open`?

PR #324, opened 2018-07-26
Fix opened message length to match NIST/SUPERCOP/Nacl conventions

dstebila:
Identified by @tlepoint in #319 but doing as a separate pull request so it gets into the July snapshot release of nist-branch.


tlepoint:
Looks all good to me!

PR #327, opened 2018-07-28
-march=native compiler optimization added for qTesla, NewHope and Kyber

erdemalkim:
I have added -march=native compiler flag to Makefiles in qTesla, NewHope, and Kyber subdirectories. It has noticable effect on qTesla and NewHope, but it seems that Kyber has similar results with or without the flag.


dstebila:
Thanks Erdem!

PR #338, opened 2018-08-15
Refactor BIKE renaming macros

dstebila:
As discussed, I moved the renaming macros into a common OQS header. 

I found that when I did this, the `-include functions_renaming.h` line from the Makefile caused a bunch of problems, which could be resolved by removing that from the Makefile and including it deliberately in each file.  Maybe you can find a better way.


dstebila:
@drucker-nir this pull is now obsoleted by your work on #347, right?

dstebila:
Obsolete by #347

PR #352, opened 2018-08-29
Fixed Visual Studio build after master's API refactoring.

christianpaquin:
Simply fixed VS solution without adding new content, and removed test_rand project from the solution. Next, I'll add new KEM api and algorithms.


christianpaquin:
@dstebila, I repaired the VS solution; appveyor is now passing. I suggest merging this, and I'll start working on integrating the KEM api and updating the algorithms on a separate PR.

dstebila:
Thanks so much, Christian!

Issue #349, opened 2018-08-29
Use CryptReleaseContext on Windows


kareldonk:
Hi, you should probably use `CryptReleaseContext` in the function `OQS_randombytes_system` in the file `https://github.com/open-quantum-safe/liboqs/blob/master-new-api/src/common/rand.c` (line 82) to release resources before returning.

Issue #350, opened 2018-08-29
Surface SIDH as a KEM in master branch


dstebila:
SIKE is already exposed as a KEM, but SIDH is not exposed as a KEM (even though the code is probably already there).  Once this is done, we can remove kex\_sidh\_msr.

dstebila:
Resolved by #368

PR #354, opened 2018-08-30
Adds KEM API and frodokem to Visual Studio

christianpaquin:
Added the KEM API and frodokem to the Visual Studio build. Also fixed the call the Window's system randomness, and released the crypto context (fixing issue #349).

### DIFF
--- a/VisualStudio/winconfig.h
+++ b/VisualStudio/winconfig.h
@@ -6,4 +6,8 @@
 #define ENABLE_KEX_RLWE_NEWHOPE
 #define ENABLE_KEX_SIDH_MSR
 #define ENABLE_SIG_PICNIC
+#define OQS_ENABLE_KEM_frodokem_640_aes
+#define OQS_ENABLE_KEM_frodokem_976_aes
+#define OQS_ENABLE_KEM_frodokem_640_cshake
+#define OQS_ENABLE_KEM_frodokem_976_cshake
 #endif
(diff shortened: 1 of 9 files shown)


VisualStudio/winconfig.h
```diff
@@ -1,9 +1,13 @@
...
+
+/* Enable schemes supported on Windows */
+#define ENABLE_KEX_NTRU
+#define ENABLE_KEX_RLWE_NEWHOPE
+#define ENABLE_KEX_SIDH_MSR
+#define ENABLE_SIG_PICNIC
+#define OQS_ENABLE_KEM_frodokem_640_aes
+#define OQS_ENABLE_KEM_frodokem_976_aes
+#define OQS_ENABLE_KEM_frodokem_640_cshake
+#define OQS_ENABLE_KEM_frodokem_976_cshake
```
dstebila: Do you want to add `OQS_ENABLE_KEM_sike_p503` and `OQS_ENABLE_KEM_sike_p751`?  The `KEX_` versions will be going away in the not-too-distant future.

christianpaquin: I'll do that as part of issue #350, while I harmonize the SIDH and SIKE code. I just wanted to add at least one KEM with the API to enable the test_kem program to work. All the remaining ones will need to be added one by one, and may take some adjusting to work on Windows; I didn't want to block integrating the KEM API in VS.

dstebila: Understood, thanks.

PR #358, opened 2018-09-05
Remove kex_rlwe_newhope

dstebila:
As we have fewer KEX algorithms left, minimal\_kex\_oqs has to be updated (and also becomes less necessary?).  It is currently causing the Windows build to fail.  @christianpaquin Do you think it's worth investigating, or should we just remove minimal\_kex\_oqs from the Windows build since all the KEXes will go away over the next month or so?


dstebila:
This could be resolved by point minimal\_kex\_oqs to use SIDH rather than NTRU because SIDH is available on Windows.

dstebila:
I'll do this today, never mind @christianpaquin.

dstebila:
@smashra Can you please review this?

PR #367, opened 2018-09-10
Add LOTUS to nist-branch

truth-quark:
This needs a minor amount of work:
* The data sheet has some values as TODO as I'm not sure of the correct answers.
* Licence is missing, currently tracking that down.
* The nist levels etc need to be updated in the kem_lotus.c file.


truth-quark:
NB: The patch has not been applied to nist-branch-lotus branch.

Problem: the patch applies only to the Reference implementation & is invalid for Optimised. This branch only includes optimised version under the assumption that will be the most commonly used. Options to proceed include:
* Include and patch the reference implementation only (on a new branch)
* Ask authors to fix the optimised implementation & provide another patch for this PR
* Include both implementations on this branch/PR (assuming a patch for the optimised fix is provided)

@dstebila do you have any preferences for what should be included?

dstebila:
Will not be resolved since LOTUS did not make it to round 2.

Issue #374, opened 2018-09-17
Remove rand object from sig API


christianpaquin:
Remove the rand object from the sig API, to harmonize with the KEM API.

dstebila:
Resolved by #379.

PR #375, opened 2018-09-17
Removed the OQS_RAND object from the sig API

christianpaquin:
Removed the OQS_RAND object from the sig API to harmonize with the KEM API (c.f., issue #374)

### DIFF
--- a/src/sig/sig.c
+++ b/src/sig/sig.c
@@ -9,16 +9,11 @@
 #include <oqs/sig_qtesla.h>
 #endif
 
-OQS_API OQS_SIG *OQS_SIG_new(OQS_RAND *rand, enum OQS_SIG_algid algid) {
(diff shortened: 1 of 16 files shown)


dstebila:
While we're at it, should we also change the algorithm identifier passed to `OQS_SIG_new` to be a string, just like in nist-branch?

christianpaquin:
Yes; it would be good to have uniformity. I quickly push that out to enable moving the OpenSSL fork to the KEM API (where the rand object is not needed anymore). I got that working locally, so we can either do the refactoring in two parts (this PR and a following one), or do all the work on this PR and wait to push the OpenSSL update (the latter has fewer PRs, so perhaps cleaner)

christianpaquin:
Branch created on open-quantum-safe repo by mistake. Moving to my own fork.

PR #379, opened 2018-09-18
Refactored sig API

christianpaquin:
Removed OQS_RAND object; harmonized params names with nist-branch. The sign/verify functions however remain. Fixes issue #374. Supersedes PR #375.

### DIFF
--- a/src/sig/sig.c
+++ b/src/sig/sig.c
@@ -9,16 +9,11 @@
 #include <oqs/sig_qtesla.h>
 #endif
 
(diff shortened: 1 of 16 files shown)


dstebila:
Let's do the change of the SIG algorithm identifier to a string on a separate PR.

Also, should we worry that the \_sign function in master branch has a different connotation than \_sign in nist-branch?  \_sign in master outputs just the digest, \_sign in nist-branch outputs the message and digest combined.  It's true that the verification functions are named differently (\_verify in master, \_sign\_open in nist-branch), but the signing functions have the same name.  Perhaps we should rename one of them so they are distinct (maybe call nist-branch \_sign\_combined?).

PR #382, opened 2018-09-19
Master add kem newhopenist

smashra:


dstebila:
Needs to be merged with master and building successfully on AppVeyor.

christianpaquin:
Would you like me to take a look at the Windows build, @smashra ?

smashra:
Yes please.

Thanks

christianpaquin:
I’ll take a look tomorrow.

On Oct 2, 2018, at 5:14 PM, Shravan Mishra <notifications@github.com<mailto:notifications@github.com>> wrote:

Yes please.

Thanks

christianpaquin:
Updated the Visual Studio project to support newhope. Since the C files are shared between the 512 and 1024 parameters, and macros are used to create function names, we can't simply include the newhope files directly into the VS project. Instead, I created windows-only 512 and 1024 source files that define the right macros and include the appropriate C files, to have separate compilation units for each params set.

src/kem/newhope/optimized/rng.h
```diff
@@ -0,0 +1,49 @@
+//
+//  rng.h
```
dstebila: Is this file required for the build?  I think it is a left-over from the NIST submission.

christianpaquin: @smashra; I'll let you update this PR. Visual Studio doesn't refer to this file, so the Windows build should be ok.

christianpaquin:
I updated the Makefile to copy the kem_newhope.h file when running "make install", and I merged with master. I think this is ready now. I'll let @dstebila update his review, and @smashra see if you want to add anything.

PR #386, opened 2018-09-19
Master kat openssl independence

smashra:


smashra:
Ready to merge to master.

Issue #380, opened 2018-09-18
Change SIG algorithm identifier on master to a string


dstebila:
Following approach on nist-branch.

Also, should we worry that the _sign function in master branch has a different connotation than _sign in nist-branch? _sign in master outputs just the digest, _sign in nist-branch outputs the message and digest combined. It's true that the verification functions are named differently (_verify in master, _sign_open in nist-branch), but the signing functions have the same name. Perhaps we should rename one of them so they are distinct (maybe call nist-branch _sign_combined?).

dstebila:
Perhaps instead call the master branch version \_sign\_digest and leave nist branch as \_sign?

christianpaquin:
You are not forced to sign a digest with this API, so this could be confusing.

christianpaquin:
This issue needs more work than simply changing the algid enum to a string. If the goal is more harmonization with the nist-branch, then we should as well: change the directory structure (/src/sig__ALG_ to /src/sig/_ALG_), change the scheme constructors (from one handler functions to individual OQS_SIG__ALG__new functions wrapped with ifdef OQS_ENABLE_SIG__ALG_).

I'll plan to do all that in this issue.

dstebila:
I might be able to find some time to work on this over the weekend.  Are you in the middle of this?

christianpaquin:
Yes, I'm almost done with a big refactoring; essentially migrating a lot of code from the nist-branch. I still have some finishing touches to complete. I'm dealing with a sick kid today, so I'll be delayed a bit, but I'm hoping to work on this tonight and over the weekend. Depending on how it's going, I might hand off the baton to you if you have free time over the weekend...

dstebila:
Okay, if you have something in a state to hand off then please feel free. But I know that getting to a hand-off state can sometimes be more work, so don't feel obliged, and don't interrupt family time to do that, it can also wait to next week.

christianpaquin:
Ready in PR #391

PR #391, opened 2018-09-23
Refactored sig api following nist-branch (issue380)

christianpaquin:
Refactored sig api following nist-branch, to harmonize with new KEM api; keeping however the sign/verify api since it's easier to use in applications (goal of the master branch). Fixes issue #380.


christianpaquin:
Worked done, ready to merge. @dstebila you can take a look.

src/sig/picnic/sig_picnic.h
```diff
@@ -0,0 +1,90 @@
...
+
+#ifdef OQS_ENABLE_SIG_picnic_L1_FS
+
+#define OQS_SIG_picnic_L1_FS_length_public_key 33
+#define OQS_SIG_picnic_L1_FS_length_secret_key 49
+#define OQS_SIG_picnic_L1_FS_length_signature 34004
+
+OQS_SIG *OQS_SIG_picnic_L1_FS_new();
+
+extern OQS_STATUS OQS_SIG_picnic_L1_FS_keypair(uint8_t *public_key, uint8_t *secret_key);
```
dstebila: I don't think we need the `extern` keyword here or anywhere else in this file.

src/sig/qtesla/sig_qtesla.h
```diff
@@ -0,0 +1,48 @@
...
+
+#ifdef OQS_ENABLE_SIG_qTESLA_I
+
+#define OQS_SIG_qTESLA_I_length_public_key 1504
+#define OQS_SIG_qTESLA_I_length_secret_key 2112
+#define OQS_SIG_qTESLA_I_length_signature 1376
+
+OQS_SIG *OQS_SIG_qTESLA_I_new();
+
+extern OQS_STATUS OQS_SIG_qTESLA_I_keypair(uint8_t *public_key, uint8_t *secret_key);
```
dstebila: I don't think we need the `extern` keyword here or anywhere else in this file.

src/sig/sig.c
```diff
@@ -1,79 +1,120 @@
...
 		return NULL;
 	}
-
-	return s;
 }
 
-OQS_API OQS_STATUS OQS_SIG_keygen(const OQS_SIG *s, uint8_t *priv, uint8_t *pub) {
-	if (s == NULL) {
+OQS_STATUS OQS_SIG_keypair(const OQS_SIG *sig, uint8_t *public_key, uint8_t *secret_key) {
+	if (sig == NULL || sig->keypair(public_key, secret_key) != 0) {
```
dstebila: Why are you comparing the output of `sig->keypair` with 0 as opposed to `OQS_SUCCESS`?

src/sig/sig.c
```diff
@@ -1,79 +1,120 @@
...
 	} else {
-		return s->keygen(s, priv, pub);
+		return OQS_SUCCESS;
 	}
 }
 
-OQS_API OQS_STATUS OQS_SIG_sign(const OQS_SIG *s, const uint8_t *priv, const uint8_t *msg, const size_t msg_len, uint8_t *sig, size_t *sig_len) {
-	if (s == NULL) {
+OQS_STATUS OQS_SIG_sign(const OQS_SIG *sig, uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *secret_key) {
+	if (sig == NULL || sig->sign(signature, signature_len, message, message_len, secret_key) != 0) {
```
dstebila: Why are you comparing the output of `sig->sign` with 0 as opposed to `OQS_SUCCESS`?

src/sig/sig.c
```diff
@@ -1,79 +1,120 @@
...
 	} else {
-		return s->sign(s, priv, msg, msg_len, sig, sig_len);
+		return OQS_SUCCESS;
 	}
 }
 
-OQS_API OQS_STATUS OQS_SIG_verify(const OQS_SIG *s, const uint8_t *pub, const uint8_t *msg, const size_t msg_len, const uint8_t *sig, const size_t sig_len) {
-	if (s == NULL) {
+OQS_STATUS OQS_SIG_verify(const OQS_SIG *sig, uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *public_key) {
+	if (sig == NULL || sig->verify(message, message_len, signature, signature_len, public_key) != 0) {
```
dstebila: Why are you comparing the output of `sig->verify` with 0 as opposed to `OQS_SUCCESS`?

src/sig/sig.h
```diff
@@ -63,108 +84,123 @@ struct OQS_SIG {
 	size_t length_public_key;
 	/** The (maximum) length, in bytes, of secret keys for this signature scheme. */
 	size_t length_secret_key;
-	/** The (maximum) length, in bytes, of a signature for this signature scheme. */
-	size_t max_length_signature;
+	/** The (maximum) length, in bytes, of signatures for this signature scheme. */
```
dstebila: Let's take the (maximum) out of parentheses.

src/sig/sig.h
```diff
@@ -63,108 +84,123 @@ struct OQS_SIG {
...
+	size_t length_signature;
```
dstebila: I wonder if we should call this `length_signature_max` instead of `length_signature`, since it's not necessarily the length of the signature but the maximum length of the signature.

src/sig/sig.h
```diff
@@ -63,108 +84,123 @@ struct OQS_SIG {
...
+	 * Signature generation algorithm.
 	 *
-	 * @param s                The signature structure.
-	 * @param priv             The signer's private key.
-	 * @param msg              The message to sign.
-	 * @param msg_len          Length of the message to sign.
-	 * @param sig              The generated signature. Must be allocated by the caller, or NULL to learn how much space is needed, as returned in sig_len.
-	 * @param sig_len          In: length of sig, out: length of the generated signature.
-	 * @return                 OQS_SUCCESS on success, or OQS_ERROR on failure.
+	 * Caller is responsible for allocating sufficient memory for `signed_message`,
```
dstebila: Change documentation to remove `signed_message`.

Also, should we have a comment explicitly explaining that in this branch `signature` contains just the digest, whereas in nist-branch `signature` contains both the digest and the message?

src/sig/sig.h
```diff
@@ -63,108 +84,123 @@ struct OQS_SIG {
...
+	 * @return OQS_SUCCESS or OQS_ERROR
 	 */
-	OQS_STATUS(*sign)(const OQS_SIG *s, const uint8_t *priv, const uint8_t *msg, const size_t msg_len, uint8_t *sig, size_t *sig_len);
+	OQS_STATUS (*sign)(uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *secret_key);
 
 	/**
-	 * Pointer to a function for signature verification.
+	 * Signature verification algorithm.
+	 *
+	 * Caller is responsible for allocating sufficient memory for `message`, based
```
dstebila: This comment is no longer relevant since `message` is input-only here.

src/sig/sig.h
```diff
@@ -63,108 +84,123 @@ struct OQS_SIG {
...
-	 * @return                 OQS_SUCCESS on success, or OQS_ERROR on failure.
+	 * @param[in] message The message represented as a byte string.
+	 * @param[in] message_len The length of the message.
+	 * @param[in] signature The signature on the message represented as a byte string.
+	 * @param[in] signature_len The length of the signature.
+	 * @param[in] public_key The public key represented as a byte string.
+	 * @return OQS_SUCCESS or OQS_ERROR
 	 */
-	OQS_STATUS(*verify)(const OQS_SIG *s, const uint8_t *pub, const uint8_t *msg, const size_t msg_len, const uint8_t *sig, const size_t sig_len);
+	OQS_STATUS (*verify)(uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *public_key);
```
dstebila: `message`, `message_len`, and `signature_len` should be marked as `const`.

src/sig/sig.h
```diff
@@ -63,108 +84,123 @@ struct OQS_SIG {
...
-	 *
-	 * @param s                Signature structure
-	 */
-	void (*free)(OQS_SIG *s);
-};
+} OQS_SIG;
 
 /**
- * Instantiate a new signature object.
+ * Consturcts an OQS_SIG object for a particular algorithm.
```
dstebila: Typo: constructs

src/sig/sig.h
```diff
@@ -63,108 +84,123 @@ struct OQS_SIG {
...
+ *
+ * Callers should always check whether the return value is `NULL`, which indicates either than an
+ * invalid algorithm name was provided, or that the requested algorithm was disabled at compile-time.
  *
- * @param algid              The id of the signature algorithm to be instantiated.
- * @return                   A new signature object on success, or NULL on failure.
+ * @param[in] method_name Name of the desired algorithm; one of the names in `OQS_SIG_algs`.
+ * @return An OQS_SIG for the particular algorithm, or `NULL` if the algorithm has been disabled at compile-time.
```
dstebila: ... if the algorithm has been disabled at compile-time or is not recognized.

src/sig/sig.h
```diff
@@ -63,108 +84,123 @@ struct OQS_SIG {
...
- * @param s         Pointer to the signature object.
- * @param priv      Pointer to the signer's private key, of expected length `s->priv_key_len` bytes.
- * @param msg       Pointer to the message to sign.
- * @param msg_len   Length of the message to sign `msg`.
- * @param sig       Pointer where the generated signature will be stored. Caller must have allocated `s->max_sig_len` bytes.
- * @param sig_len   Pointer to the length of the generated signature.
- * @return          OQS_SUCCESS on success, or OQS_ERROR on failure
+ * Signature generation algorithm.
+ *
+ * Caller is responsible for allocating sufficient memory for `signed_message`,
```
dstebila: Same comment as above.

src/sig/sig.h
```diff
@@ -63,108 +84,123 @@ struct OQS_SIG {
...
+	 *
+	 * @param[out] signature The signature on the message represented as a byte string.
+	 * @param[out] signature_len The length of the signature.
+	 * @param[in] message The message to sign represented as a byte string.
+	 * @param[in] message_len The length of the message to sign.
+	 * @param[in] secret_key The secret key represented as a byte string.
+	 * @return OQS_SUCCESS or OQS_ERROR
 	 */
-	OQS_STATUS(*sign)(const OQS_SIG *s, const uint8_t *priv, const uint8_t *msg, const size_t msg_len, uint8_t *sig, size_t *sig_len);
+	OQS_STATUS (*sign)(uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *secret_key);
```
dstebila: `message_len` should be marked as const

src/sig/sig.h
```diff
@@ -63,108 +84,123 @@ struct OQS_SIG {
...
+ * @param[in] sig The OQS_SIG object representing the signature scheme.
+ * @param[out] signature The signature on the message represented as a byte string.
+ * @param[out] signature_len The length of the signature.
+ * @param[in] message The message to sign represented as a byte string.
+ * @param[in] message_len The length of the message to sign.
+ * @param[in] secret_key The secret key represented as a byte string.
+ * @return OQS_SUCCESS or OQS_ERROR
  */
-OQS_API OQS_STATUS OQS_SIG_sign(const OQS_SIG *s, const uint8_t *priv, const uint8_t *msg, const size_t msg_len, uint8_t *sig, size_t *sig_len);
+OQS_STATUS OQS_SIG_sign(const OQS_SIG *sig, uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *secret_key);
```
dstebila: `message_len` should be marked as const

src/sig/sig.h
```diff
@@ -63,108 +84,123 @@ struct OQS_SIG {
...
+ * @param[in] sig The OQS_SIG object representing the signature scheme.
+ * @param[in] message The message represented as a byte string.
+ * @param[in] message_len The length of the message.
+ * @param[in] signature The signature on the message represented as a byte string.
+ * @param[in] signature_len The length of the signature.
+ * @param[in] public_key The public key represented as a byte string.
+ * @return OQS_SUCCESS or OQS_ERROR
  */
-OQS_API OQS_STATUS OQS_SIG_verify(const OQS_SIG *s, const uint8_t *pub, const uint8_t *msg, const size_t msg_len, const uint8_t *sig, const size_t sig_len);
+OQS_STATUS OQS_SIG_verify(const OQS_SIG *sig, uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *public_key);
```
dstebila: `message, `message_len`, and `signature_len` should be marked as `const`

tests/Makefile.am
```diff
@@ -1,18 +1,13 @@
 AUTOMAKE_OPTIONS = foreign
 check_PROGRAMS = example_kem speed_kem test_kem \
                       minimal_kex_oqs test_kex \
-                      test_sig \
+                      minimal_sig_oqs test_sig \
```
dstebila: I'd prefer to use the name `example_sig` instead of `minimal_sig_oqs`, to mirror what we've done for KEMs and as on nist-branch.

tests/test_kem.c
```diff
@@ -1,4 +1,3 @@
-#include <stdbool.h>
```
dstebila: Out of curiosity, why did you delete this?

src/sig/picnic/sig_picnic.h
christianpaquin: This came from the nist-branch; should we remove it on both branches?

src/sig/qtesla/sig_qtesla.h
christianpaquin: This came from the nist-branch; should we remove it on both branches?

src/sig/sig.c
christianpaquin: Copy-pasted from nist-branch; fixed here; should fix it there too...

src/sig/sig.c
christianpaquin: Copy-pasted from nist-branch; fixed here; should fix it there too...

src/sig/sig.c
christianpaquin: Copy-pasted from nist-branch; fixed here; should fix it there too...

src/sig/sig.h
christianpaquin: Yes; leftovers from copy-pasting from nist-branch

src/sig/sig.h
christianpaquin: Yes; leftovers from copy-pasting from nist-branch

src/sig/sig.h
christianpaquin: Yes; leftovers from copy-pasting from nist-branch

tests/Makefile.am
christianpaquin: Ah yes, I was wondering if there would be a minimal_kem_oqs to replace the kex one. I can rename that one.

tests/test_kem.c
christianpaquin: Wasn't needed in test_sig.c, so after comparing with test_kem.c, I figured it wasn't needed there either.

src/sig/sig.c
christianpaquin: Also created PR #393 on nist-branch to fix this

src/sig/sig.c
christianpaquin: Also created PR #393 on nist-branch to fix this

src/sig/sig.c
christianpaquin: Also created PR #393 on nist-branch to fix this

src/sig/sig.h
christianpaquin: I copied that from kem.h; we should keep or change both...

src/sig/sig.h
christianpaquin: given that maximum was implied in the comment, and that applied to the keys as well, I didn't want to single the signature field with a special "max" marker in its name

src/sig/sig.h
christianpaquin: Changed. We need to explain it somewhere, but I'm not sure where. For folks using the master branch for integration, it might not make sense to talk about a "nist-branch"; and the signature/verify paradigm used by the master branch is the classic API for signatures.

src/sig/sig.h
christianpaquin: Fixed; also fixed on nist-branch (PR #393)

src/sig/sig.h
christianpaquin: Fixed.

src/sig/sig.h
christianpaquin: We should update kem.h as well...

src/sig/picnic/sig_picnic.h
dstebila: There was a recent PR on nist-branch #387 that removed a bunch of superfluous extern's, but perhaps they missed some.  Please remove it here, and I'll ask Ben to do the same on nist-branch.

src/sig/qtesla/sig_qtesla.h
dstebila: There was a recent PR on nist-branch #387 that removed a bunch of superfluous extern's, but perhaps they missed some. Please remove it here, and I'll ask Ben to do the same on nist-branch.

src/sig/sig.h
dstebila: Okay, makes sense.

src/sig/sig.h
dstebila: Will do.

src/sig/sig.h
dstebila: Well we do it for uint8\_t which can't be changed by the function either, so why one and not the other?

src/sig/sig.h
dstebila: Well we do it for uint8_t which can't be changed by the function either, so why one and not the other?

src/sig/sig.h
dstebila: Well we do it for uint8_t which can't be changed by the function either, so why one and not the other?

dstebila:
I noticed that the Travis build was failing so I worked through fixing it, but it turned out there were a multitude of flaws in the Travis scripts that meant a bunch of things weren't working properly, so the solution got bigger than I would have liked.  Sorry to pollute your PR with all that extra stuff.

tests/Makefile.am
christianpaquin: I migrated the example_sig from the nist-branch.

christianpaquin:
Let me know when you are done editing the PR, and I'll take back the baton. I also need to add speed_sig to master.

dstebila:
Seems to be working now, so baton is yours.

src/sig/picnic/sig_picnic.h
truth-quark: I think a few algorithms didn't get cleaned of externs as there was active work on them. Chances are I forgot to add it in the PR while avoiding potential merge conflicts/jumbled history.

src/sig/sig.h
christianpaquin: we do it for uint8_t _pointers_, the const being a promise to the caller that the function won't modify the content. I don't see why that would be useful for a size parameter.

src/sig/sig.h
christianpaquin: we do it for uint8_t _pointers_, the const being a promise to the caller that the function won't modify the content. I don't see why that would be useful for a size parameter.

christianpaquin:
I think I've addressed all the comments on this PR, with the exception of the "const size_t" comment which I suggest not to do. @dstebila; you can take another look, or provide more comments if I missed something.

PR #397, opened 2018-10-02
Sync common files on master with nist-branch

dstebila:


dstebila:
This will (probably) have to be updated once #403 lands.

PR #398, opened 2018-10-02
Sync common files on nist with master

dstebila:


dstebila:
This will (probably) have to be updated once #403 lands.

Issue #389, opened 2018-09-21
Remove OQS_RAND object


dstebila:

dstebila:
Depends on #377

dstebila:
Resolved by #403

PR #403, opened 2018-10-04
Remove old OQS_RAND object and API

dstebila:
Also had to do some tweaking of building shared libraries and the things that are considered part of the public API.


dstebila:
Aims to resolve #389

PR #407, opened 2018-10-05
Use getentropy() if available

oittaa:


oittaa:
I don't have a mac and don't really care why Apple products are always so broken. This worked on other systems that HAVE_GETENTROPY....

dstebila:
Thanks @oittaa, good improvement.

Looks like macOS needs `#include <sys/random.h>` for getentropy() to be available.  (The Linux manpage says that getentropy() can be imported from either unistd.h or sys/random.h.)

dstebila:
... Except that sys/random.h isn't present on older versions of Linux, in which case we seem to have to use unistd.h.  Sigh.

PR #411, opened 2018-10-10
Set version in configure, library, and algorithms

dstebila:
This adds an `alg_version` field to the `OQS_KEM` and `OQS_SIG` objects.  The intention for this field is as described in README.md:

> **Algorithm compatibility:** Unlike existing standardization cryptographic algorithms (SHA-2, SHA-3, PKCS\#1v1.5, nistp256, ...), post-quantum algorithms are under active development, and the mathematical algorithm of a cryptographic scheme may change: a particular name (e.g., "FrodoKEM-AES-640") may refer to different mathematical algorithms over time.  liboqs may update implementations as algorithms evolve.  Using our semantic versioning, versions with the same first two components ($X.Y.*$) will be interoperable.  But version $X.Y.Z$ and version $X.(Y+1).Z'$ may not be interoperable.  liboqs-reliant applications can check the `alg_version` member of the `OQS_KEM` data structure for each algorithm to obtain an identifier of the algorithm version used in a particular implementation; implementations returning the same `alg_version` for an algorithm will be interoperable.

This version is also recorded in the corresponding algorithm datasheet.

Where possible, a GitHub commit URL will serve as the version identifier for a particular algorithm.


dstebila:
@christianpaquin Can you provide the exact GitHub commit as the version for qTESLA and Picnic?  Insert at the `TODO` in:
- `docs/algorithms/sig_picnic.md`
- `docs/algorithms/sig_qtesla.md`
- `src/sig/picnic/sig_picnic.c`
- `src/sig/qtesla/sig_qtesla.c`

dstebila:
@crockeea or @drucker-nir Can you provide some exact version as the version for BIKE?  I am not sure which version from http://bikesuite.org/#implementation is the one in use here.  Insert at the `TODO` in:
- `docs/algorithms/kem_bike.md`
- `src/kem/bike/kem_bike.c`

christianpaquin:
I updated the version for qtesla and picnic

dstebila:
The way I understand semantic version is that 1.Y.Z and 1.A.B should be *binary* compatible, meaning they have the same binary API.  My sentence above was meant to say that cryptographic algorithms in 1.Y.Z and 1.A.B may not be algorithmically interoperable, meaning that Frodo in 1.Y.Z may produce incompatible ciphertexts with Frodo in 1.A.B.  But Frodo in 1.Y.Z would produce compatible ciphertexts with Frodo in 1.Y.W.

smashra:
I think we still want to maintain the existing semantics of semantic
versioning.

Because the way we are doing is hacking semantic versioning to accommodate
algorithm versioning.

Probably we should augment the scheme to introduce algorithm version such
that semantic versioning works as such.

For e.g.

X.Y.Z.A

X - algorithm version number
The rest is the default semantic versioning.
Once every thing stabilized we discard X and still be semantic version
compatible.

dstebila:
I see what you're saying, Shravan, but I'm still not sure I agree it's the right approach.  It is quite likely that various algorithms will be changing over the next 0-3 years, so we could rather quickly be incrementing major version numbers with that approach.  That could get us up to liboqs 12 fairly quickly, which might convey to someone the message that "liboqs is mature and stable", when in fact it's because there's been a lot of changes.  

I did some reading about various versioning systems.  Here are some of the things I read:

- Debate about how semantic versioning should deal with the same API but different behaviour of that API: https://github.com/semver/semver/issues/311
- An alternative to semantic versioning called romantic versioning, which I think is more like what I envisioned: https://gist.github.com/jashkenas/cbd2b088e20279ae2c8e, http://blog.legacyteam.info/2015/12/romver-romantic-versioning/
- A few more opinion pieces: https://blog.scottlogic.com/2016/12/15/semantic-versioning-is-not-enough.html, http://blog.hypesystem.dk/on-versioning

Apparently in semantic versioning, while you're in 0.y.z, you're allowed to do whatever you want because you're still in "early development".  I'm inclined to think of ourselves as being in the 0.y.z phase for at least the next year.  But even still, we'd want to have a reasonable number strategy within that, in which case I'd agree to incrementing y whenever we make an API or algorithm-incompatible change.  What do you think?  

I've updated the README.md with a discussion of this approach.  I'm going to merge the PR now so that I can do an rc1 as I promised, but we can continue discussions about version numbers of the next week and finalize before the rc2 next Friday.

Issue #413, opened 2018-10-10
Add alg_version fields on nist-branch


dstebila:
Following #411 on master branch.

PR #417, opened 2018-10-13
Sync nist-branch with master and sort algorithms alphabetically

dstebila:
Still need to add alg\_version fields for all KEM and SIG algorithms as per #413

### DIFF
--- a/src/common/common.c
+++ b/src/common/common.c
@@ -32,25 +32,3 @@ OQS_API void OQS_MEM_secure_free(void *ptr, size_t len) {
 OQS_API void OQS_MEM_insecure_free(void *ptr) {
 	free(ptr); // IGNORE free-check
 }
-
-/* Displays hexadecimal strings */
-void OQS_print_hex_string(const char *label, const uint8_t *str, size_t len) {
-	printf("%-20s (%4zu bytes):  ", label, len);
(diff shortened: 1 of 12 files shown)


truth-quark:
Should contributing.md be updated with a dot point like "add new algorithms in alphabetic order to the kem/sig source"?

src/common/common.c
```diff
@@ -32,25 +32,3 @@ OQS_API void OQS_MEM_secure_free(void *ptr, size_t len) {
 OQS_API void OQS_MEM_insecure_free(void *ptr) {
 	free(ptr); // IGNORE free-check
 }
-
-/* Displays hexadecimal strings */
-void OQS_print_hex_string(const char *label, const uint8_t *str, size_t len) {
```
truth-quark: OQS_print_hex_string() is also duplicated:
```
./src/kem/kat_kem.c:12:void OQS_print_hex_string(const char *label, const uint8_t *str, size_t len) {
./src/kem/test_kem.c:9:void OQS_print_hex_string(const char *label, const uint8_t *str, size_t len) {
```
Should test _kem just include from a common one?

dstebila: We had a common one before and I moved it out of the common files intentionally.  Otherwise it would be part of our public API, and that seemed unnecessary.

PR #419, opened 2018-10-17
Fix Visual Studio DLL projects

christianpaquin:
Updated the dll def file after OQS's API refactoring, fixed test_kem's DLL projects config, and added ReleaseDLL config to appveyor.


VisualStudio/oqs/dll.def
```diff
@@ -1,19 +1,33 @@
...
-	OQS_MEM_secure_free
-	OQS_MEM_insecure_free
+	OQS_MEM_cleanse
+	OQS_MEM_secure_free
+	OQS_MEM_insecure_free
+	OQS_randombytes_switch_algorithm
+	OQS_randombytes_custom_algorithm
+	OQS_randombytes
+	OQS_randombytes_nist_kat_init
+	OQS_KEM_frodokem_640_aes_keypair
```
dstebila: Why are the FrodoKEM functions listed explicitly here, but no others?

christianpaquin: Only Frodo and qTESLA have OQS_API-exported function, which is what I used to populate the DLL def file (but qTESLA isn't supported on Windows).

christianpaquin: Yes indeed; we should open a work item to do this on the Linux build, and I can update this PR once the OQS_API tag is properly set everywhere.

PR #429, opened 2018-10-26
Added version and branch info to config

christianpaquin:
Added version and branch info to config.h. Needed by OpenSSL fork to enable branch-dependent functionality.


dstebila:
OpenSSL has a numeric version field as well so that one can do comparisons `<=`, `>=`, etc. on the version.

```c
/*-
 * Numeric release version identifier:
 * MNNFFPPS: major minor fix patch status
 * The status nibble has one of the values 0 for development, 1 to e for betas
 * 1 to 14, and f for release.  The patch level is exactly that.
 * For example:
 * 0.9.3-dev      0x00903000
 * 0.9.3-beta1    0x00903001
 * 0.9.3-beta2-dev 0x00903002
 * 0.9.3-beta2    0x00903002 (same as ...beta2-dev)
 * 0.9.3          0x0090300f
 * 0.9.3a         0x0090301f
 * 0.9.4          0x0090400f
 * 1.2.3z         0x102031af
 *
 * For continuity reasons (because 0.9.5 is already out, and is coded
 * 0x00905100), between 0.9.5 and 0.9.6 the coding of the patch level
 * part is slightly different, by setting the highest bit.  This means
 * that 0.9.5a looks like this: 0x0090581f.  At 0.9.6, we can start
 * with 0x0090600S...
 *
 * (Prior to 0.9.3-dev a different scheme was used: 0.9.2b is 0x0922.)
 * (Prior to 0.9.5a beta1, a different scheme was used: MMNNFFRBB for
 *  major minor fix final patch/beta)
 */
# define OPENSSL_VERSION_NUMBER  0x10101010L
```

Should we take the same approach?

christianpaquin:
Would make sense. Although I currently don't need the OQS version number in the OpenSSL fork (I only need the branch info macro), so we can optionally differ that to the next version.

christianpaquin:
I tested with OpenSSL's [PR46](https://github.com/open-quantum-safe/openssl/pull/46); works. We can merge this PR now to enable the functionality. Still outstanding: add the OQS_MASTER_BRANCH macro to the master branch (I wasn't sure how add this value in the autotool configure file), and optionally add the version number (although we could create a separate PR for this because it isn't needed to enable using OpenSSL with OQS's nist-branch).

christianpaquin:
Followed @dstebila's suggestion and made the version number numeric following OpenSSL's model. Tested with OpenSSL's [PR46](https://github.com/open-quantum-safe/openssl/pull/46); works.

This PR is ready, but we need to add another work item to replicate this on the master branch. I looked at it, but I wasn't sure how to deal with the automake config files...

dstebila:
I added a text representation of the version as well in case someone ever wants to print it out

dstebila:
In terms of doing this on master branch, I think we would edit configure.ac and use `AC_DEFINE` (https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Defining-Symbols.html) which will output the specified `#define` to config.h.

christianpaquin:
Ok, I'll look into adding the version to master as well.

PR #438, opened 2018-11-06
Update kem_saber.md

vsoftco:
typo

### DIFF
--- a/docs/algorithms/kem_saber.md
+++ b/docs/algorithms/kem_saber.md
@@ -6,7 +6,7 @@ Summary
 
 - **Name**: SABER
 - **Algorithm type**: key encapsulation mechanism
-- **Main cryptographic assumption**: module learning with rounding (MLWR)
+- **Main cryptographic assumption**: module learning with rounding (MLWE)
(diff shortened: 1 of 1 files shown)


dstebila:
@christianpaquin Any idea what this build failure in AppVeyor means?  I'm at a loss on how changing a text file could have triggered a build failure...

christianpaquin:
appveyor shouldn't run on the nist-branch, since there are no VS projects...

dstebila:
Oh right, of course.  appveyor.yml disables it for all branches that contain the word "nist", but this patch's branch doesn't.

dstebila:
@vsoftco why do you think this is a typo?  The assumption is Module Learning with Rounding, MLWR seems like the right abbreviation.  (In the NIST submission documentation, they use Module-LWR.)

vsoftco:
> @vsoftco why do you think this is a typo? The assumption is Module Learning with Rounding, MLWR seems like the right abbreviation. (In the NIST submission documentation, they use Module-LWR.)

My bad, sorry, I read it as "module learning with errors"... Please ignore.

PR #442, opened 2018-11-08
Fix mac error

smashra:


dstebila:
Can you explain briefly?

dstebila:
> Do what twice ?

It's what I wrote in my original comment: 

> Line 75 seems to imply that all platforms get `-Wl,--enable-new-dtags` in LD\_FLAGS, and then lines 76-78 seem to imply that Linux gets an additional copy of `-Wl,--enable-new-dtags` appended to LD\_FLAGS. 

Linux would get TWO copies of `-Wl,--enable-new-dtags` in LD\_FLAGS, all other platforms would get ONE copy of ``-Wl,--enable-new-dtags`` in LD\_FLAGS.

smashra:
Fixed.

PR #448, opened 2018-12-07
Kyber

cryptojedi:
Added Kyber to the Master branch, trying to follow all the coding guidelines (and using NewHope as an example).


christianpaquin:
I've modified the oqs project file to copy the kyber header file to the Visual Studio dir, which should make appveyor tests pass. I'll look into adding kyber Windows support in a different PR, when this one lands.

PR #457, opened 2019-01-16
Updated Picnic to v1.3.1

christianpaquin:
Integrated the latest version of the [optimized Picnic implementation (v1.3.1)](https://github.com/IAIK/Picnic/tree/v1.3.1).

The code (still) contains an optimized implementation of Keccak; I haven't yet replaced it with the internal OQS one. It would be desirable to also offer optimized versions os SHA3.


christianpaquin:
@dstebila; I fixed the issue on macOS. You can have a look. As mentioned above, the Picnic code (still) has its own optimized sha3 implementation. Choices are: 1) keep it there, 2) remove and used oqs's unoptimized one. It'd be desirable for oqs to provide an optimized version of sha3; the question is in which order should we do things (migrate picnic to oqs's sha3 first, then add an optimized version later, or vice versa). What do you think?

PR #466, opened 2019-02-27
Replace the Kyber768 implementation in liboqs master with PQClean's implementation

dstebila:
First attempt at using an algorithm from PQClean in liboqs master.

Goal is to drop in a directory of source code from PQClean without any changes to that source code, and then just wire it into liboqs.

In order to get this working, I had to make a few changes in liboqs that shouldn't have to be made again:

- Exempt PQClean symbols from namespace checks
- Exempt PQClean subdirectories from prettyprint
- Add shims that translate PQClean's SHA-3 / randombytes provider to liboqs' functions

Then it's a matter of wiring the particular implementation into the appropriate OQS_KEM data structure and Makefile.

Finally while doing this work I uncovered a bug in our SHA-3 implementation, which I fixed.


christianpaquin:
I'll take a look

dstebila:
> I looked at the diff and tried the PR on both Linux and Windows; works well. I compared the copied code from the PQClean project; in the upstream project methods are prepended with `PQCLEAN_KYBER768_CLEAN_`, but in OQS `PQCLEAN_KYBER768_` is used (missing the `CLEAN_` suffix); why is that? A clean code diff would make it easier to verify proper import.

They changed their prefixing from `PQCLEAN_KYBER768_` to `PQCLEAN_KYBER768_CLEAN_` earlier today, after I'd already copied the code.  :(  The intention would be for OQS to use exactly the prefixing from PQClean, and I would use the new names in the next sync.  (Which I can do before merging.)

christianpaquin:
> The intention would be for OQS to use exactly the prefixing from PQClean, and I 
> would use the new names in the next sync.  (Which I can do before merging.)

Up to you if you want to do it before the merge; might be a good idea for a clean v1 (especially if you want to also update the documentation folder)

dstebila:
Updated and ready for review again @christianpaquin

PR #471, opened 2019-03-26
Add support for building on OpenBSD

tobhe:
See title. I added some additional instructions in the Readme that explain the OpenBSD specific steps in the build process.
The changes to picnic were necessary to prevent a double definition error.

One thing that would be nice to have, that i haven't figured out how to do right (since I'm not too familiar with automake) would be setting the default compiler to clang for OpenBSD.
OpenBSD's gcc is based on an old version that does not support several of the newer compilation flags used in the build process and thus can not build this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,19 @@ You need to install the following packages using brew (or a package manager of y
 
 	brew install autoconf automake libtool openssl wget
 
+### Install dependencies for OpenBSD
+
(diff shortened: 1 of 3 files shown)


dstebila:
Thanks Tobias!  Do you happen to know of any continuous integration service that supports OpenBSD that we could use to make sure the OpenBSD build keeps working?

PR #475, opened 2019-04-14
Add FrodoKEM Round 2

dstebila:
Based on PQClean implementation


dstebila:
@christianpaquin Would you be able to create update the VisualStudio project with the Frodo additions/changes?

christianpaquin:
I'll take a look.

christianpaquin:
@dstebila, I removed Frodo round2 from the Visual Studio build. It will take some time to properly integrate it, due to conflictingly-named files in the output directory (the duplicated files for each param set).

christianpaquin:
So the undefined Frodo algs makes example_kem fail and therefore AppVeyor fails as well; we can either switch to another default alg, or wait to next week to give me more time to properly integrate the code in VS.

dstebila:
> It will take some time to properly integrate it, due to conflictingly-named files in the output directory (the duplicated files for each param set).

Is this a problem going forward? In general PQClean implementations are going to have duplicated files for each param set. Will you have to do something manual and annoying each time?  If so we should rethink our integration strategy, for example renaming files to have unique names?

christianpaquin:
Currently, and by default, all the object files are created in the same directory. I'll research the options to see if I can change that.

christianpaquin:
I fixed the obj file collision issue; pushed PR #478 separately to see if it passes all tests before coming back to this PR.

dstebila:
I've been investigating the cause of the current build failure. It seems to be an algorithm problem: different KAT values are generated for FrodoKEM-640-AES when using Configuration=Debug versus Configuration=Release.  In particular, the `S` value generated during sampling `PQCLEAN_FRODOKEM640AES_CLEAN_sample_n` is different.

This suggests a problem in the upstream PQClean code or even the upstream LWEKE code from @patricklonga.  @christianpaquin, do you know what the difference is between Configuration=Debug and Configuration=Release?  Is there a way to replicate that directly in the command-line when compiling with cl.exe (which is how we do tests on Windows in PQClean, see e.g. [Makefile.Microsoft\_nmake](https://github.com/PQClean/PQClean/blob/master/crypto_kem/frodokem640aes/clean/Makefile.Microsoft_nmake)).  If so, then I'd be curious to first replicate Configuration=Release onto the PQClean Makefile.Microsoft\_nmake and see whether it produces different KAT values there (which would suggest the problem is in the Frodo code) or not (which would suggest the problem is in the liboqs AES or SHA-3 code).

christianpaquin:
Looking into it...

patricklonga:
Is the code up-to-date? This might be related to one issue I solved recently for FrodoKEM-640 on Windows. See commit [8a24c4a](https://github.com/Microsoft/PQCrypto-LWEKE/commit/8a24c4af5e901925228fc79a0d03b9b5615e6c73) from LWEKE.

dstebila:
Thanks @patricklonga, that seems to have fixed it!

PR #476, opened 2019-04-16
Add implementation of NTRU LPRime

### LINKED ISSUE #465: Proposal: add NTRU LPRime implementation
Would there be interest in the submission of an [NTRU LPRime](https://ntruprime.cr.yp.to/) implementation to the master branch? NTRU LPRime is a round two KEM candidate in the NIST competition.

Assuming I can get a clear answer from the authors (Bernstein et al) on the license that applies to their reference implementations, would you be interested in such a contribution? I'm thinking this would be a drop-in of the reference implementation, adjusted as necessary to meet the requirements for your master branch.

The motivation for this is our desire to use NTRU LPRime (in hybrid mode) with TLS for some research work we are doing. Step one would be contributing NTRU LPRime to liboqs. Step two would be making contributions to your OpenSSL fork.

I'm proposing NTRU LPRime rather than Streamlined NTRU Prime, since the metrics on their home page suggest NTRU LPRime offers faster performance for a key gen, followed by encaps and decaps. Given our use case is TLS, this is the primary "round trip" we'd like to optimise for.

dmjones:
This implementation is based on the NTRU LPRime reference code, which is
public domain.

The notes in src/kem/ntrulprime/ref/README should be heeded, for they
apply to this modified variant. Modifications were made only to detach
the reference code from SUPERCOP.

Fixes #465.


dmjones:
Ah, test failures. I will look at those.

dmjones:
Also, I missed the round 2 tweaks discussed at https://ntruprime.cr.yp.to/nist/ntruprime-20190330/Supporting_Documentation/mods.pdf. I'll update this PR and will comment again when it's ready for consideration.

PR #483, opened 2019-05-13
Add SPHINCS+ from PQClean

### LINKED ISSUE #482: Add SPHINCS+ from PQClean

thomwiggers:
This PR adds all SPHINCS+ signature schemes from PQClean.

I have some scripts that generated the major files, I don't suggest maintaining any of these files by hand. (Where) should I store those?

Also adds an incremental hashing API to FIPS202 and adds SHA2 + the accompanying pqclean shims.

Closes #482


thomwiggers:
Someone not being me will have to update the Windows build files.

dstebila:
Thanks Thom.  Over the weekend I was also working on automating bringing schemes from PQClean to liboqs, and in PR #484 have built some scripts and templates for automatically adding KEMs.  My inclination would be to try to extend that to deal with signature schemes.  Give me a few days to get the KEM stuff sorted out, and then I'll take a look at this PR and signature schemes.

thomwiggers:
The SHA256 and Haraka implementations in this PR are not thread-safe. If this is a concern, we would need to fix that in PQClean.

dstebila:
Yes, we do need thread-safe because our consuming applications (OpenSSL, OpenSSH) are multi-threaded.

Seems to me that "clean" in PQClean should include "thread-safe"...

thomwiggers:
Agreed, it's super annoying because it's also a very nasty side effect for my Rust library (needed to disable multi-threaded tests, which are enabled by default).

dstebila:
#521 replaces this PR.

Issue #480, opened 2019-05-08
Add NTRU round 2 from PQClean


dstebila:

dstebila:
Resolved by #484

PR #484, opened 2019-05-13
Add Kyber and NTRU round 2

dstebila:


dstebila:
Fixes #480

dstebila:
@christianpaquin I've started created some automated techniques for adding round 2 schemes -- generating parts of the various files as required.  It seems to be working well so far.  But I haven't tried using this technique to generate the VisualStudio files as well.  Do you think that's worth trying?  The XML looks pretty straightforward and I think it should be possible.  But on the flip side I also remember that when I manually edited the XML it sometimes didn't work.

christianpaquin:
I'll take a look.

christianpaquin:
As we discussed on the dev call, @dstebila , we have three options:

1. use scripts to modify the VS projects, adding the new PQClean algs
2. use scripts to modify the VS projects, disabling the new PQClean algs (at a minimum, the respective include file `kem_<scheme>h` must be copied to the VS include dir), and adding them manually.
3. (in the future), use cmake to generate both the nix and windows builds.

You can take a shot at 1 if you have time. I looked at your script, but it would be simpler if you try it first. Otherwise I can plan to do option 2.

dstebila:
> 1. use scripts to modify the VS projects, adding the new PQClean algs
> You can take a shot at 1 if you have time. I looked at your script, but it would be simpler if you try it first. Otherwise I can plan to do option 2.

I've give number 1 a shot over the next couple of days when I get a chance

dstebila:
@christianpaquin I've used my code generation templates to generate the VS files.  Looks like they're working (although not all builds finished at the time I write this comment).  I opened up the generated project in Visual Studio, and it looks like the filters aren't working.  I used the same UUIDs for ones that already existed, and generated new UUIDs for new ones, and the XML looks analogous, but the filters aren't showing up in Visual Studio Community for me.  How does it look for you?

christianpaquin:
Back at the computer, I'll take a look.

christianpaquin:
I'll need more time to investigate this issue. I've opened a separate issue #487 to avoid blocking this PR.

PR #486, opened 2019-05-16
Link main liboqs library to OpenSSL when enabled

### LINKED ISSUE #485: liboqs built with --enable-openssl does not depend on libcrypto
Perhaps this is intentional, but building with `--enable-openssl` doesn't result in the liboqs library depending upon libcrypto.

After building with `--enable-openssl`:

```
$ ldd /usr/local/liboqs/lib/liboqs.so
	linux-vdso.so.1 (0x00007ffe195c2000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2907601000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f2907d11000)
```

This makes it slightly complicated to use liboqs, since I need to manually link against `lcrypto` and also need to specify `-Wl,--no-as-needed` to avoid undefined symbol errors.

dmjones:
This adds the correct linker instructions when OpenSSL is enabled. It also removes the now-redundant linker instructions when building the tests.

Fixes #485.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -42,6 +42,9 @@ endif
 if ENABLE_KEM_KYBER
 liboqs_la_LIBADD += src/kem/kyber/libkemkyber.la
 endif
+if USE_OPENSSL
+liboqs_la_LIBADD += -L${OPENSSL_DIR}/lib -lcrypto
+endif
 
 oqsconfigh:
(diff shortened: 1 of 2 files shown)


dstebila:
Thanks Duncan!

PR #495, opened 2019-06-19
Add LEDAcrypt-KEM-LT-12 from PQClean

dstebila:


dstebila:
Currently paused because the AES key schedule API from PQClean is incompatible with the AES key schedule API from OQS.  In particular the PQClean AES ECB API assumes that ECB can be called multiple times with the same key schedule.  Whereas the OQS API releases the key schedule after it's used once.  Static allocation on the stack can't work because OpenSSL's EVP API doesn't permit this.  Currently thinking that PQClean's API will have to be changed to have a release method.

dstebila:
Currently blocked because liboqs doesn't have SHA3-384.  Will possibly be resolved by #502.

dstebila:
#515 replaces this PR.

PR #500, opened 2019-06-25
Switch Frodo to use MSR codebase directly

dstebila:
@christianpaquin As discussed here is a start at switching liboqs to use the Frodo code more directly from MSR rather than via PQClean.  

In the MSR codebase, there are multiple places where the various configuration macros get set -- including config.h, frodoxxx.c, and the Makefile.  I've consolidated all of that into one configuration file for each parameter set -- frodo640aes.h, etc. -- and then use gcc's `-include` option to force that configuration file to get loaded before compiling that group of files.  I'm not sure if there's a way to do the analogous thing in VisualStudio.  Do you know?  If not, then I can change it so that there's a frodo640aes.c file which includes the kem.c, noise.c, util.c files just like in Patrick's code base.


christianpaquin:
@dstebila, I'll take a look.

christianpaquin:
@dstebila, here is what I found. The [`/FI` compiler option](https://docs.microsoft.com/en-us/cpp/build/reference/fi-name-forced-include-file) (force include) should work. The `ForcedIncludeFiles` property can be added to a `ClCompile` element, specifying the include file. For example, 
`<ClCompile Include="..\..\src\kem\frodokem\external\kem_frodokem640aes.c">`
would become
`<ClCompile Include="..\..\src\kem\frodokem\external\kem_frodokem640aes.c">`
`   <ForcedIncludeFiles>..\..\src\kem\frodokem\external\frodo640aes.h</ForcedIncludeFiles>`
`</ClCompile>`

dstebila:
Thanks Christian, although now I realize that this may not quite work.  What should happen is that, for example, the file src/kem/frodokem/external/util.c should be compiled 6 times, each time with a different force-include header.  Do you think there's anyway to do that?  If not I know how I can rearchitect the code.

christianpaquin:
Ah, I see. No, I don't know how to do this in one VS project. In the SIKE codebase I do use the original include strategy, so perhaps that's the way to go here as well.

PR #509, opened 2019-07-01
Use AVX instructions on Frodo

### LINKED ISSUE #504: Add AVX2 implementation of Frodo
There's AVX2 code in https://github.com/microsoft/PQCrypto-LWEKE/blob/master/src/frodo_macrify.c.  Need to copy it in and make sure our compile-time macros are triggering it.  Need to figure out how to handle SHAKE 4x in liboqs, or maybe just leave that in Frodo.

dstebila:
I noticed that CPU feature detection isn't working, so this PR enables it first.

Fixes #504


dstebila:
Replaced by #519

PR #515, opened 2019-07-05
Add LEDAcrypt KEMs from PQClean

dstebila:
Obsoletes #495.


dstebila:
Currently missing:

- [ ] algorithm datasheet
- [ ] KATs (see https://github.com/PQClean/PQClean/pull/186#issuecomment-503513868)

Issue #474, opened 2019-04-09
Differentiate unsupported KEM/sigs and allocation errors


christianpaquin:
`OQS_[KEM|SIG]_new` returns `NULL` if the requested scheme is unsupported or if there is an allocation error. We should differentiate both cases to better report the error in applications.

PR #520, opened 2019-07-11
bug fix on return value

vsoftco:


dstebila:
In fact we have long-standing issue #474 about potentially changing this behaviour.  Thoughts on how to proceed?  We could add a function like `bool OQS_KEM_is_enabled(const char *method_name)`, and keep `OQS_KEM_new` returning NULL both when the KEM is disabled or when an allocation error occurs.

vsoftco:
> In fact we have long-standing issue #474 about potentially changing this behaviour. Thoughts on how to proceed? We could add a function like `bool OQS_KEM_is_enabled(const char *method_name)`, and keep `OQS_KEM_new` returning NULL both when the KEM is disabled or when an allocation error occurs.

But definitely we shouldn't `return OQS_SUCCESS;`, right?

christianpaquin:
A different return value for "error" and "not enabled" would make sense. Perhaps even add a 3rd option "not supported", like we do in the wrappers.

vsoftco:
@christianpaquin Agree, I'll think about it and propose a fix soon

dstebila:
#530 is meant to solve the same issue. Closing this PR to focus work on that one.

PR #523, opened 2019-07-16
Add Rainbow from PQClean

dstebila:


dstebila:
Replaced by #532 to sidestep conflicts

PR #530, opened 2019-07-31
Add functions to test if algorithms are enabled

### LINKED ISSUE #474: Differentiate unsupported KEM/sigs and allocation errors

dstebila:
Fixes #474.  Replaces #520.

### DIFF
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -54,7 +54,7 @@ typedef enum {
 	OQS_SUCCESS = 0,
 	/** Used to indicate that algorithm is not enabled */
 	OQS_NOT_ENABLED,
-    /** Used to indicate that algorithm is not supported */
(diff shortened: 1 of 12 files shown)


src/common/common.h
```diff
@@ -52,6 +52,10 @@ typedef enum {
 	OQS_ERROR = -1,
 	/** Used to indicate successful return from function. */
 	OQS_SUCCESS = 0,
+	/** Used to indicate that algorithm is not enabled */
+	OQS_NOT_ENABLED,
```
dstebila: I don't see anywhere that `OQS_NOT_ENABLED` or `OQS_NOT_SUPPORTED` are actually used.

PR #532, opened 2019-08-01
Add Rainbow from PQClean

dstebila:


thomwiggers:
FYI, Rainbow contains UB: https://github.com/PQClean/PQClean/issues/220

PR #537, opened 2019-08-07
Add integration tests on ARM platforms

dstebila:


dstebila:
@thomwiggers if you have a chance to take a look if I've configured the emulated builds correctly, that would be great. I followed what you did in PQClean, and it seems to be working.

thomwiggers:
It appears your appveyor tests have been disabled for the last run, but otherwise I think it looks okay.

dstebila:
Thanks @thomwiggers. I put `[skip ci]` on the last commit since it wasn't changing anything I wanted to have the tests re-run for, so that's why most of them were skipped.

PR #540, opened 2019-08-13
Fix build failure on Visual Studio x86

christianpaquin:
This fixes a build failure on Visual Studio x86 (SIKE wasn't recognizing this as a valid platform). Also adds x86 as an appveyor test platform.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,9 @@ version: 1.0.{build}
 
 image: Visual Studio 2017
 
-platform: x64
+platform:
+  - x86
+  - x64
 
 build:
   verbosity: minimal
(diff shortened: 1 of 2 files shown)


dstebila:
CircleCI and AppVeyor builds didn't get triggered.  I just changed a setting which I think should cause CircleCI to get properly triggered for PRs from forks.  Not sure what went wrong with AppVeyor.  Any thoughts, @vsoftco?

vsoftco:
@dstebila I'll take a look asap

christianpaquin:
I changed the appveyor script to add x86. Did I screw it up? Anyway to check locally, @vsoftco?

vsoftco:
@christianpaquin No idea what's going on that appveyor is not triggered. That's the way to add more platforms, so nothing wrong with the appveyor.yml

dstebila:
I backed this out because it was causing build failures.  We'll need to revisit.

PR #550, opened 2019-09-23
Re-enabled picnic in arm test

christianpaquin:
Re-enabled picnic in arm test

### DIFF
--- a/config/ac_checks.m4
+++ b/config/ac_checks.m4
@@ -18,7 +18,6 @@ AC_TYPE_UINT64_T
 AC_TYPE_UINT8_T
 
 # Checks for library functions.
-AC_FUNC_MALLOC
 AC_CHECK_FUNCS([gettimeofday memmove memset pow sqrt strdup])
 AC_CHECK_SIZEOF([size_t])
 AC_CHECK_FUNCS(getentropy)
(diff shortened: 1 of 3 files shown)


christianpaquin:
I tested the build on a raspberry pi (armv7l) and picnic worked. I'm not sure why picnic fails on the travis arm test.

dstebila:
Compilation is succeeding, now we're getting segmentation faults for Picnic.

christianpaquin:
Hmm. The cross-compiling script still doesn't work in my env, I'm getting some compilation errors. I'll try to sort this out to investigate the segfault.

sebastinas:
We have seen SIGBUS errors on some ARM platforms. This is caused by the opt64 SHAKE implementation assuming its okay to load uint64_t from unaligned memory addresses. If the target ARM platform does not support unaligned memory loads (or the fix up in the Linux kernel is not enabled), this part of the opt64 SHAKE implementation can be disabled by defining  `NO_MISALIGNED_ACCESSES` during the build.

If there are other issues left, let me know and I'll try to integrate fixes in the Picnic implementation.

dstebila:
Since switching to CMake, Picnic has been enabled on all platforms. The armel and armhf emulated CI tests have been passing with Picnic, and a cross compilation I did to Raspberry Pi also passed tests.  This appears to no longer be needed. Closing.

PR #553, opened 2019-10-04
Adds avx2 implementation for qtesla

christianpaquin:
Adds avx2 implementation for qtesla.

Two things to fix before merge:

- KAT are currently different between portable and avx2, resulting in test failures.
- Some optimized sha3 code is present in the implementation; need to abstract it out into OQS.


src/crypto/sha3/sha3.h
```diff
@@ -519,10 +519,32 @@ void OQS_SHA3_cshake256_simple(uint8_t *output, size_t outlen, uint16_t cstm, co
...
+* \param output3 The fourth output byte array
+* \param outlen The number of output bytes to generate in every output array
+* \param cstm0 The first 16bit customization integer
+* \param cstm1 The second 16bit customization integer
+* \param cstm2 The third 16bit customization integer
+* \param cstm3 The fourth 16bit customization integer
+* \param input The input seed byte array
+* \param inplen The number of seed bytes to process
+*/
+void OQS_SHA3_cshake128_simple4x(uint8_t *output0, uint8_t *output1, uint8_t *output2, uint8_t *output3, size_t outlen, uint16_t cstm0, uint16_t cstm1, uint16_t cstm2, uint16_t cstm3, const uint8_t *in, size_t inlen);
```
dstebila: Given the cSHAKE "simple" is not a standardized primitive, I'd be inclined to keep this out of the common files, unless multiple implementations are using it.  However I realize that might be tricky, because the implementation wants to call keccak\_4x which is not exported outside of the compilation unit.  Any thoughts?

christianpaquin:
Ready for merge (@dstebila, I'm not sure if you were done with your review)

christianpaquin:
Ready for merge (@dstebila, I'm not sure if you were done with your review)

christianpaquin:
Ready for merge (@dstebila, I'm not sure if you were done with your review)

PR #554, opened 2019-10-07
BIKE Round 2

drucker-nir:
This PR includes: 

1. Removing BIKE-2/3
2. Removing BIKE-1 Level-5
3. Renaming BIKE-1 Level-1/3 to BIKE-1-CPA
4. Adding BIKE-1 Level-1/3 with FO transformation and with negligible DFR.


src/kem/bike/additional/defs.h
```diff
@@ -0,0 +1,149 @@
...
+ * The license is detailed in the file LICENSE.md, and applies to this file.
+ *
+ * Written by Nir Drucker and Shay Gueron
+ * AWS Cryptographic Algorithms Group.
+ * (ndrucker@amazon.com, gueron@amazon.com)
+ */
+
+#pragma once
+
+#include "../../../../config.h"
```
xvzcf: Is it possible to remove this dependency?

drucker-nir: Done

src/kem/bike/additional/measurements.h
```diff
@@ -0,0 +1,103 @@
+/*
```
dstebila: Do we need this file?

src/kem/bike/kem_bike.c
```diff
@@ -2,290 +2,102 @@
...
-	kem->method_name = OQS_KEM_alg_bike3_l1;
-#if defined(OQS_KEM_BIKE_ADDITIONAL_IMPLEMENTATION)
+	kem->method_name = OQS_KEM_alg_bike1_l1_fo;
 	kem->alg_version = "Additional - 05/23/2018";
-#else
-	kem->alg_version = "Reference - 06/29/2018";
-#endif
 
 	kem->claimed_nist_level = 1;
 	kem->ind_cca = false;
```
dstebila: Should `kem->ind_cca = true` for the `_fo` variants?

src/kem/bike/kem_bike.c
```diff
@@ -2,290 +2,102 @@
...
-	kem->method_name = OQS_KEM_alg_bike3_l3;
-#if defined(OQS_KEM_BIKE_ADDITIONAL_IMPLEMENTATION)
+	kem->method_name = OQS_KEM_alg_bike1_l3_fo;
 	kem->alg_version = "Additional - 05/23/2018";
-#else
-	kem->alg_version = "Reference - 06/29/2018";
-#endif
 
 	kem->claimed_nist_level = 3;
 	kem->ind_cca = false;
```
dstebila: Should `kem->ind_cca = true` for the `_fo` variants?

src/kem/kem.h
```diff
@@ -31,24 +31,14 @@ extern "C" {
...
-/** Algorithm identifier for BIKE3-L5 KEM. */
-#define OQS_KEM_alg_bike3_l5 "BIKE3-L5"
+/** Algorithm identifier for BIKE1-L1-CPA KEM. */
+#define OQS_KEM_alg_bike1_l1_cpa "BIKE1-L1-CPA"
+/** Algorithm identifier for BIKE1-L3-CPA KEM. */
+#define OQS_KEM_alg_bike1_l3_cpa "BIKE1-L3-CPA"
+/** Algorithm identifier for BIKE1-L1-FO KEM. */
+#define OQS_KEM_alg_bike1_l1_fo "BIKE1-L1-FO"
+/** Algorithm identifier for BIKE1-L3-FO KEM. */
+#define OQS_KEM_alg_bike1_l3_fo "BIKE1-L3-FO"
```
dstebila: Also need to edit `OQS_KEM_algs_length` below.

dstebila:
Note to self: before merging, update `OQS_KEM_algs_length` in templating and re-run copy-from-pqclean to ensure subsequent applications of copy-from-pqclean are clean.

dstebila:
@drucker-nir To confirm my understanding, do the reference and additional implementations now have the same KATs?  If memory serves, with the Round 1 code there were different key sizes and KATs for the different implementations.

src/kem/bike/additional/measurements.h
drucker-nir: No, will remove.

drucker-nir:
BIKE has wo implementations: the Reference implementation and the Additional implementation that have different (but equivalent) KATs. Previously, LibOQS used the Additional implementation for platforms that support AVX2 and the Reference implementation for other platforms. In Round-2 the Additional implementation supports all platforms. Therefore, we now need only one implementation and one set of KATs (i.e., this PR).

src/kem/bike/kem_bike.c
Shay-Gueron: No. 
The _fo variants provide te FO flows and negligible DFR. 
Currently, we prefer not to claim IND-CCA security. 
In any case, the intention is to use this KEM only with ephemeral keys (for forward secrecy), and in such cases, an IND-CPA suffices. 
A paper with the relevant analysis is going to be posted soon on ePrint.

PR #555, opened 2019-10-07
Update README.md

vsoftco:
put links on the CI icons instead of CI names


dstebila:
This is incompatible with how Doxygen parses Markdown files, which is why we did it in the unusual way.

PR #561, opened 2019-10-19
Add Kyber90s variants

dstebila:


xvzcf:
Could you add `void` to all the function prototypes taking no arguments?

dstebila:
Ready for review

PR #576, opened 2019-12-08
Updating BIKE's code

drucker-nir:
Updating BIKE's code according to:
- Drucker, Nir, Shay Gueron, and Dusan Kostic. 2019. “On Constant-Time QC-MDPC Decoding with Negligible Failure Rate.” Cryptology ePrint Archive, 2019. https://eprint.iacr.org/2019/1289.
- Drucker, Nir, Shay Gueron, and Dusan Kostic. 2019. “QC-MDPC decoders with several shades of gray.” Cryptology ePrint Archive, 2019.


dstebila:
@drucker-nir I pushed an update to the algorithm data sheet (`docs/algorithms/kem_bike.md`), which is the final step before merging.  Can you confirm that I've recorded the right version in that file?

PR #577, opened 2019-12-11
Harmonized example_sig with example_kem.

christianpaquin:
Ported latest example_kex changes to example_sig. In particular, example_sig now returns failure correctly.


christianpaquin:
I expect the changes from PR #573 to fix the travis failure.

PR #579, opened 2020-01-10
OSX testing, cron testing limitation

baentsch:
This includes two things I suggested to @dstebila: An OSX-build and a "test-necessity" check feature. Please take a look at let me know whether we want to take this forward in this or another form (test failure if no test is necessary isn't quite ideal).


.circleci/config.yml
```diff
@@ -106,16 +112,46 @@ jobs:
...
+            name: Prep env
+            command: brew update; brew unlink python@2; brew install autoconf automake libtool wget doxygen graphviz clang-format; pip3 install pytest
+        - run:
+            name: Configure
+            command: autoreconf -i && ./configure ${CONFIGURE_ARGS}
+        - run:
+            name: Build
+            command: make && make check
+        - run:
+            name: Recon
```
dstebila: Is this temporary for testing or will it always be there?

.circleci/config.yml
```diff
@@ -5,6 +5,9 @@ version: 2
     - image: ${IMAGE}
   steps:
     - checkout
+    - run:
+        name: Testing needed?
+        command: r=`git log --name-only --since="1 day ago" -n 2`; if [ "x$r" == "x" ]; then echo "No. Exiting."; exit -1; else echo "Yes."; fi
```
dstebila: -1 when no test needed is indeed suboptimal, since we'll got all kinds of build failures as a result.  

A little bit of Googling did not find any way of terminating the build without failures.  Only thing I can think of is to wrap every subsequent command in the same check as a way to skip them.

baentsch: Agree. The "successful" failures also annoyed me over the weekend. Either wrapping this code into a convenience script or shaming the CI provider into "CO2 friendliness" (there _should_ be a way to do nightly builds only if something truly changed). What does the OSX job do when no OSX is configured in the account?

dstebila: > What does the OSX job do when no OSX is configured in the account?

Not sure I really understand the question. We haven't used Mac OS X jobs in CircleCI before.

baentsch:
Closed in favour of #597

PR #581, opened 2020-01-14
Centos testing added

baentsch:
Added testing as per [this discussion](https://github.com/baentsch/testing/wiki/Docker-image-goals) for Centos7 & 8. Arguably testing should only be nightly (iff some code changed) --> Moving new targets `centos-7` and `centos-8` down to the nightly section suggested (as would be testing target configs only either at check-in/`build` _or_ `nightly`).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,11 +10,13 @@ version: 2
         command: autoreconf -i && ./configure --enable-silent-rules ${CONFIGURE_ARGS}
     - run: 
         name: Build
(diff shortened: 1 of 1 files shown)


dstebila:
CircleCI does not seem to be building on forks.  Can you push to a branch on open-quantum-safe/liboqs which should then trigger a build?

dstebila:
The Centos 7 and Centos 8 images pass on my `circleci local`.

PR #584, opened 2020-01-22
Sync with PQClean as of January 22, 2020

### LINKED ISSUE #566: memory leaks in OpenSSL based SPHINCS+-SHA256 versions
For SPHINCS+-SHA256-* using the OpenSSL hash there are memory leaks because the hash context created by `PREFIX_initialize_hash_function` callee `sha256_inc_init` is not cleaned up.

Ideally there'd be some hook that allows proper destruction of `hash_state_seeded` at the end of sign.c functions

- `PREFIX_crypto_sign_seed_keypair`
- `PREFIX_crypto_sign_signature`
- `PREFIX_crypto_sign_verify`

dstebila:
Fixes #566


scripts/copy_from_pqclean/src/kem/family/kem_family.h
```diff
@@ -9,7 +9,7 @@
 #define OQS_KEM_{{ family }}_{{ scheme['scheme'] }}_length_secret_key {{ scheme['metadata']['length-secret-key'] }}
 #define OQS_KEM_{{ family }}_{{ scheme['scheme'] }}_length_ciphertext {{ scheme['metadata']['length-ciphertext'] }}
 #define OQS_KEM_{{ family }}_{{ scheme['scheme'] }}_length_shared_secret {{ scheme['metadata']['length-shared-secret'] }}
-OQS_KEM *OQS_KEM_{{ family }}_{{ scheme['scheme'] }}_new();
+OQS_KEM *OQS_KEM_{{ family }}_{{ scheme['scheme'] }}_new(void);
```
xvzcf: Can you remove the double underscores from `__OQS_KEM` in the first and second lines and re-run the templating?

scripts/copy_from_pqclean/src/sig/family/sig_family.h
```diff
@@ -9,7 +9,7 @@
 #define OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_length_secret_key {{ scheme['metadata']['length-secret-key'] }}
 #define OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_length_signature {{ scheme['metadata']['length-signature'] }}
 
-OQS_SIG *OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_new();
+OQS_SIG *OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_new(void);
```
xvzcf: Can you remove the double underscores from `__OQS_SIG` in the first and second lines and re-run the templating?

src/kem/kyber/kem_kyber_512.c
```diff
@@ -28,9 +28,9 @@ OQS_KEM *OQS_KEM_kyber_512_new() {
 	return kem;
 }
 
-OQS_STATUS PQCLEAN_KYBER512_CLEAN_crypto_kem_keypair(unsigned char *pk, unsigned char *sk);
-OQS_STATUS PQCLEAN_KYBER512_CLEAN_crypto_kem_enc(unsigned char *ct, unsigned char *ss, const unsigned char *pk);
-OQS_STATUS PQCLEAN_KYBER512_CLEAN_crypto_kem_dec(unsigned char *ss, const unsigned char *ct, const unsigned char *sk);
+int PQCLEAN_KYBER512_CLEAN_crypto_kem_keypair(unsigned char *pk, unsigned char *sk);
```
xvzcf: What is the reason for this change? Is it just that `copy_from_pqclean/src/kem/family/kem_scheme.c` was not updated?

dstebila: Yeah, seems to have been some previous inconsistency.  The `PQCLEAN_` functions from PQClean do actually return an int, so we should have that be their interface when we reference their prototype, and then cast to `OQS_STATUS` later when we actually call into them, which is I believe what this PR now accomplishes correctly.

PR #591, opened 2020-01-23
Use OpenSSL's EVP_MD_CTX_new and _free

### LINKED ISSUE #565: Rename EVP_MD_CTX_{create|destroy} to EVP_MD_CTX_{new|free}
The `sha2_ossl.c` and `sha3_ossl.c` files make references to the `EVP_MD_CTX_{create|destroy}` functions, which were redefined in OpenSSL 1.1.1 to be macros pointing to `EVP_MD_CTX_{new|free}`. 

For various reasons, liboqs has kept the original names and modifications were made to OQS-OpenSSL-1.1.1 to retain the original `create/destroy` functions. Given that these modifications can be problematic in some situations, and OpenSSL 1.0.2 is being deprecated at the end of 2019, it is perhaps time to rename the functions to be consistent with the OpenSSL 1.1.1 API and drop liboqs support for 1.0.2 altogether.

dstebila:
Fixes #565.

Does so by requiring OpenSSL version 1.1.0 or higher.  But one of our target platforms (Ubuntu Xenial 16.04) only comes with OpenSSL 1.0.2.  So what should we do?  Require users on that platform to build OpenSSL 1.0.2 (and update our Docker images accordingly)?  Or remove that as a target platform?  Thoughts, @baentsch and @xvzcf?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,50 +66,28 @@ jobs:
     environment:
       ARCH: armel
       CONFIGURE_ARGS: --disable-sig-sphincs --disable-sig-qtesla # sig-sphincs exhausts memory on CircleCI servers; qTesla fails on armel
-  ubuntu-xenial-x86_64-gcc49:
-    <<: *oqsjob
-    environment:
-      IMAGE: dstebila/liboqs:ubuntu-xenial-x86_64-0.1.0
-      CC: gcc-4.9
-      CONFIGURE_ARGS: --disable-sig-picnic 
-  ubuntu-xenial-x86_64-gcc5:
-    <<: *oqsjob
-    environment:
-      IMAGE: dstebila/liboqs:ubuntu-xenial-x86_64-0.1.0
-      CC: gcc-5
-  ubuntu-xenial-x86_64-gcc6:
-    <<: *oqsjob
-    environment:
-      IMAGE: dstebila/liboqs:ubuntu-xenial-x86_64-0.1.0
-      CC: gcc-6
-  ubuntu-xenial-x86_64-gcc7:
-    <<: *oqsjob
-    environment:
-      IMAGE: dstebila/liboqs:ubuntu-xenial-x86_64-0.1.0
-      CC: gcc-7
-  ubuntu-xenial-x86_64-gcc8:
-    <<: *oqsjob
-    environment:
-      IMAGE: dstebila/liboqs:ubuntu-xenial-x86_64-0.1.0
-      CC: gcc-8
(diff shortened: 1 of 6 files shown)


baentsch:
I'd vote for removing support for Ubuntu 16 and providing documentation how someone can help himself if he's "bound" to it (build OpenSSL 1.1.1 from source first). The same is true for OSX, if I'm not mistaken: Default OpenSSL is 1.0.2 and you have to brew install 1.1.1 first before working with OQS-OpenSSL. 

I'm not sure how a Docker image update can help, though: Which image do you suggest updating in which way? If you think about the dev image, I'm not sure that would solve a user's issue: They then may have a ready-made build environment, but I don't think they would keep working in it (for their "main project" that "keeps them bound" to Ubuntu 16).

Net-net: I'd remove support for Xenial, document as suggested above and wait whether there's howls of protest.

dstebila:
> The same is true for OSX, if I'm not mistaken: Default OpenSSL is 1.0.2 and you have to brew install 1.1.1 first before working with OQS-OpenSSL.

Yes, I modified the instructions to require `brew install openssl@1.1` and `./configure` looks in the place that brew would put it.

> I'm not sure how a Docker image update can help, though: Which image do you suggest updating in which way? 

I was meaning installing OpenSSL in our Xenial image.

> Net-net: I'd remove support for Xenial, document as suggested above and wait whether there's howls of protest.

Fine with me.  Let's give it a couple of days on here for any such preliminary howls.

xvzcf:
I agree with @baentsch; we could have suggestions/instructions for upgrading to 1.1.1, but we shouldn't worry about support for 1.0.2.

dstebila:
Xenial is currently where we do all our testing of old versions of gcc, so if we want to keep testing how this compiles under old versions of gcc, we'd have to move those tests over to another image, e.g., Ubuntu bionic. Any thoughts on how many versions of gcc back we should go?

baentsch:
In the absence of concrete needs for more (do we have/know of any?), I'd vote for "as few as possible". 

Actually, I wouldn't explicitly go any one back from the baseline CI images we decide to support: Wouldn't it be enough to be sure OQS compiles & runs OK with LTS versions of Ubuntu, Debian, Centos (plus OSX + Windows + any cross-build images (ARM?) we decide to support)?

xvzcf:
I also agree, we could just not go beyond the baseline images barring specific requests.

dstebila:
- [ ] Before merging make sure https://github.com/open-quantum-safe/openssl/pull/139 is also working

Issue #590, opened 2020-01-23
BIKE code failing on AVX512f


dstebila:
@drucker-nir, the new BIKE code from #576 is failing on our continuous integration system in some AVX512f-related code; see https://circleci.com/gh/open-quantum-safe/openssl/1192 which contains error message:

```  CPPAS    additional/libkembike1_l1_cpa_la-gf_mul.lo
In file included from /usr/lib/gcc/x86_64-linux-gnu/5/include/immintrin.h:45:0,
                 from additional/secure_decode_avx512.c:35:
/usr/lib/gcc/x86_64-linux-gnu/5/include/avx512fintrin.h: In function 'OQS_KEM_bike1_l1_cpa_rotate_right':
/usr/lib/gcc/x86_64-linux-gnu/5/include/avx512fintrin.h:997:10: error: the last argument must be an 8-bit immediate
   return (__m512i) __builtin_ia32_psrlqi512_mask ((__v8di) __A, __B,
          ^
/usr/lib/gcc/x86_64-linux-gnu/5/include/avx512fintrin.h:918:10: error: the last argument must be an 8-bit immediate
   return (__m512i) __builtin_ia32_psllqi512_mask ((__v8di) __A, __B,
          ^
```

Are you able to assist?  I don't have a machine with AVX512f on it.

dstebila:
Another build failure as well in https://circleci.com/gh/open-quantum-safe/liboqs/3189:

```
src/kem/bike/.libs/libkembike.a(libkembike1_l3_cpa_la-secure_decode_avx2.o): In function `rotate256_big':
/root/project/src/kem/bike/additional/secure_decode_avx2.c:40: multiple definition of `rotate256_big'
src/kem/bike/.libs/libkembike.a(libkembike1_l1_cpa_la-secure_decode_avx2.o):/root/project/src/kem/bike/additional/secure_decode_avx2.c:40: first defined here
src/kem/bike/.libs/libkembike.a(libkembike1_l1_fo_la-secure_decode_avx2.o): In function `rotate256_big':
/root/project/src/kem/bike/additional/secure_decode_avx2.c:40: multiple definition of `rotate256_big'
src/kem/bike/.libs/libkembike.a(libkembike1_l1_cpa_la-secure_decode_avx2.o):/root/project/src/kem/bike/additional/secure_decode_avx2.c:40: first defined here
[...]

PR #592, opened 2020-01-23
Partial fix for failing BIKE code

dstebila:
Partial fix to #590 as suggested by @franziskuskiefer in https://github.com/open-quantum-safe/liboqs/issues/590#issuecomment-577667909.


dstebila:
Probably obsoleted by #593.

PR #595, opened 2020-01-27
OSX testing

baentsch:
This adds OSX testing to CircleCI testing. To limit use of OSX test cycles, tests only run at midnight on master and only if something has been checked in to master during the last day.

### DIFF
--- a/.build-osx.sh
+++ b/.build-osx.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -o xtrace
+
+r=`git log --name-only --since="1 day ago" -n 2`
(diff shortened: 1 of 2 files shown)


dstebila:
@baentsch I pulled your code onto a branch inside open-quantum-safe/liboqs so that it would actually trigger CircleCI; see https://github.com/open-quantum-safe/liboqs/commits/osx-circle.  Also I wanted to have it use the CircleCI run steps, so I created a script that checks for recent commits and then wrapped each command in that; please take a look.

baentsch:
Closing in favour of PR #597

PR #600, opened 2020-01-28
fix kyber clean leaks

franziskuskiefer:
This should probably go into PQClean as well. cc @thomwiggers


thomwiggers:
PQClean is not affected since the current API for AES and FIPS202 still completely relies on the full state being on the stack. This is not a very friendly assumption when you want to allow swapping out the implementation for these primitives, and may be something we want to fix, but it's there for now.

Solving this would require adding `_copy` and `_free` functions to all primitives exposed by the PQClean common API. We should then also somehow make sure that these `_free` functions are used...

I'm crossposting this as an issue on the PQClean repo.

franziskuskiefer:
Ok, makes sense. So I'd recommend taking this as a temporary fix here for liboqs and pull in a new version once https://github.com/PQClean/PQClean/issues/264 got fixed.

dstebila:
Closing this PR since it will be fixed in PQClean. See https://github.com/PQClean/PQClean/pull/272.

PR #601, opened 2020-01-28
Clean up heap on success

franziskuskiefer:
Using `goto` isn't great but at least you won't leak this way.


dstebila:
Thanks @franziskuskiefer. I want to avoid the goto's to keep the style consistent with our other example program. So I made a simpler version in #603, and checked it with valgrind. Does that look okay to you?

PR #605, opened 2020-01-30
Add sanitizer builds

franziskuskiefer:
A very simple `--with-sanitizer` to enable ASAN and UBSAN. This will produce a couple errors that will need to get fixed before this should be enabled on CI. Note that this should always go along with  `CC=clang`.


franziskuskiefer:
Oh, looks like that flag didn't work. That's what happens when you enable the sanitizers :wink: 
My autoconf skills are a little limited :neutral_face:

dstebila:
Thanks for trying. :)  We're also not great at autoconf so we're in the process of switching to CMake.  I'll see what I can do to get this working.

dstebila:
Replaced by #617

PR #609, opened 2020-02-09
Update BIKE (FO) Round-2 to BIKE's v3.2 specification

drucker-nir:
Updating BIKE code to use the latest additional code and also to match the latest BIKE v3.2 specification.


dstebila:
Thanks @drucker-nir!  We've got a temporary code freeze while we finish moving the build system from autoconf to CMake (#556) but after that (probably later this week) we'll get your updates merged.

dstebila:
Replaced by #623.

PR #616, opened 2020-02-16
Sync with PQClean and update hashing API

dstebila:


thomwiggers:
This commit also pulls in the heap-based versions of (at least) the FIPS202 primitives. They're only on the heap to help detect problems, it's much more efficient to use stack-based APIs.

dstebila:
We pull in OpenSSL's implementations when available which would be more performant, so I don't think it's too much of a problem that the base implementation is slower.

thomwiggers:
Right.

PR #622, opened 2020-02-18
Testing macOS release build.

xvzcf:
I basically kept the CPU extension detection code as is, so that the `USE_*_INSTRUCTIONS` variables are still set the same way, but removed the part adding individual flags and used `march=native` instead. 

The macOS release build passes on CircleCI with this. I'm a bit surprised it works, but I think we can make do with this until we have the `cpu_features` library.


vsoftco:
Works fine for me

xvzcf:
I'll merge it in once the AppVeyor tests pass.

PR #626, opened 2020-02-22
Update CMakeLists.txt C flags for Android

nicholasfulton:
Run options which break Android build only if the target system is not Android


dstebila:
Just curious why you closed #625 and this one?

nicholasfulton:
I realized they should be in the same pull request, because this one is necessary for the Android one to work. I'm creating a new pull request with both commits.

nicholasfulton:
Opened another pull request with all of the necessary changes.

PR #627, opened 2020-02-24
Enable cross-compilation for Android

nicholasfulton:
Add build script and modify CMakeLists.txt to fix pieces which break Android build.


dstebila:
Do you think it would be possible to create a Docker image for a continuous integration test that exercises your new Android build script?  Just thinking that it would be preferable if we could ensure this new build doesn't suffer from code rot over time.

nicholasfulton:
I'll look into that. I'll need to figure out how to so it may take a bit but I'll get it done.

PR #628, opened 2020-02-24
Fixing oqsconfig.h.cmake

xvzcf:
@baentsch encountered this error while building HAProxy with liboqs:

```
src/backend.c:1483:17: error: operator '&&' has no left operand
```
 due to this line:
```
1483 | #if USE_OPENSSL && (defined(OPENSSL_IS_BORINGSSL) || (HA_OPENSSL_VERSION_NUMBER >= 0x10101000L))
```

i.e `USE_OPENSSL`, is defined and used, but not assigned a value, by liboqs, thereby causing an error in HAProxy, which also uses the same macro.

This PR therefore adds an `OQS_` prefix to all the `#define`s in `oqsconfig.h.cmake` and assigns them a value.


baentsch:
Thanks!

PR #610, opened 2020-02-12
Mingw-w64 support

gadoofou87:
Builds successfully on Arch Linux and passes tests in Wine

PR #630, opened 2020-02-25
Reworked search for executable files in tests

gadoofou87:
I'm sorry, this patch was supposed to be included in my last pull request https://github.com/open-quantum-safe/liboqs/pull/610. Without it, tests do not pass when cross-compiling on Linux. Mingw-w64 builds executables with the ".EXE" extension, which is not expected by the script that runs the tests.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -112,16 +112,18 @@ def filtered_test(func):
     return wrapper
 
 def path_to_executable(program_name):
+    path = "."
     if sys.platform.startswith("win"):
         if 'APPVEYOR_BUILD_FOLDER' not in os.environ: os.environ['APPVEYOR_BUILD_FOLDER'] = "."
-        return os.path.join(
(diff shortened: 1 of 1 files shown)


tests/helpers.py
```diff
@@ -112,16 +112,19 @@ def wrapper(*args, **kwargs):
...
-        )
+        path = os.path.join(path, os.environ['APPVEYOR_BUILD_FOLDER'])
+    path = os.path.join(path, "build", "tests")
+    for executable in [
+        os.path.join(path, program_name),
+        os.path.join(path, program_name + ".EXE"),
+        os.path.join(path, program_name + ".exe")]:
+            if os.path.isfile(executable):
+                return executable
     else:
```
dstebila: Is the indentation of this `else` weird? In fact I think maybe the else should be eliminated and then the function flows into the `assert False` if there was no `return executable` completed...

gadoofou87: @dstebila Fixed it

PR #632, opened 2020-02-26
Updated SIKE implementation with latest changes.

christianpaquin:
This PR includes:
- Optimizations changes from PR #558, adapted to cmake
- New changes from the [SIDH library](https://github.com/microsoft/PQCrypto-SIDH), moving from commit [ebd1c8](https://github.com/microsoft/PQCrypto-SIDH/commit/ebd1c80a8ac35e9ca2ef9680291a8a43b95a3bfa) to [fde210](https://github.com/microsoft/PQCrypto-SIDH/commit/fde210a7cf03e835cceb1d5ff34ccf58625d0311).

PR #633, opened 2020-02-28
Clear compiler warnings

### LINKED ISSUE #564: Address all warnings in the CMake build.
Compiling SIKE with the CMake build and the `-Wall -Wextra -Werror` flags set results in several errors, and different ones at that for Clang on macOS and gcc on Ubuntu.  

[Here](https://github.com/open-quantum-safe/liboqs/files/3764244/errors.txt) are the GCC-Ubuntu errors. Their validity has to be investigated.

dstebila:
Fixes #564.  Turns `-Werror` back on.

I discovered three different types of warnings:

1. Warnings that we should actually fixed.  These have been fixed.
2. Warnings that are technically things that should be fixed if we're being really pedantic, but aren't worth changing.  (For example, an issue with how const is interpreted on typedef'd typedef'd datatypes in SIKE, or unused functions in the way SIKE includes files.)  These have been worked around by disabling those warnings in the relevant subdirectory (e.g., in SIKE's CMakeLists.txt, but not in the top-level CMakeLists.txt).
3. Warnings that are bugs in the compiler.  (For example, how `{0}` is interpreted by gcc-4.9.)  These have been worked around by disabling those warnings in the relevant subdirectory.  For gcc-4.9, in fact one type of warning can't be disabled, so I have disabled `-Werror` on that subdirectory when running gcc-4.9.

The summary is that a small number of warnings on some versions of gcc have been disabled or do not trigger errors in a limited number of directories.  All warnings on clang should trigger build failures.


xvzcf:
mac_warnings

I get these warnings on my older MacBook Air, though I'm not sure if it's worth addressing them.

xvzcf:
MSVC 19.24.28316.0 on Windows also emits some warnings, the build passes there since none of the equivalent compiler options have been set yet.

dstebila:
What does the warning on macOS mean?  That the program won't run on older versions?

xvzcf:
Although, for example, `speed_sig picnic_L1_FS` runs fine on my laptop, I'm guessing that's what it means. Googling does not seem to yield much useful information.

dstebila:
> Although, for example, `speed_sig picnic_L1_FS` runs fine on my laptop, I'm guessing that's what it means. Googling does not seem to yield much useful information.

I don't have a Mac with that version of macOS, but I could [install one in a virtual machine](https://github.com/myspaghetti/macos-guest-virtualbox) if needed, or we could [configure a CircleCI instance to run an older version of macOS](https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions).  The question is whether we feel obligated to support older versions of macOS.

I will need to rebase this after #632 lands, so we have a couple of days to decide what to do.

christianpaquin:
I'm getting build errors in bike on my Ubuntu 18.04 on WSL, after building with `cmake -Gninja .. && ninja`

```
../src/kem/bike/additional/secure_decode_avx2.c: In function ‘rotate256_big’:
../src/kem/bike/additional/secure_decode_avx2.c:46:48: error: overflow in implicit constant conversion [-Werror=overflow]
  const __m256i all_one_mask = _mm256_set1_epi8(0xff);
                                                ^~~~
../src/kem/bike/additional/secure_decode_avx2.c: In function ‘rotate256_small’:
../src/kem/bike/additional/secure_decode_avx2.c:76:48: error: overflow in implicit constant conversion [-Werror=overflow]
  const __m256i all_one_mask = _mm256_set1_epi8(0xff);
                                                ^~~~
../src/kem/bike/additional/secure_decode_avx2.c:80:27: error: overflow in implicit constant conversion [-Werror=overflow]
                           0x86, 0x86, 0x86, 0x86, 0x86, 0x86, 0x86, 0x86, 0x84, 0x84, 0x84, 0x84,
                           ^~~~
```

dstebila:
> I'm getting build errors in bike on my Ubuntu 18.04 on WSL, after building with `cmake -Gninja .. && ninja`

Do we consider Ubuntu 18.04 on Windows Subsystem for Linux to be a target platform?

As for fixing, can you see if the warning is cleared by explicitly casting to `(char)`?

christianpaquin:
> Do we consider Ubuntu 18.04 on Windows Subsystem for Linux to be a target platform?

I would certainly hope so as this has been my dev environment for the last 2 years!

> As for fixing, can you see if the warning is cleared by explicitly casting to `(char)`?

Yes, it does. I've added a `(char)` cast in front of the arguments to `_mm256_set1_epi8` on lines 46, 76, and 79.

dstebila:
@xvzcf 

> mac_warnings
> 
> I get these warnings on my older MacBook Air, though I'm not sure if it's worth addressing them.

Looking into, the Picnic code is already checking if aligned\_alloc is properly available on the platform, so think we can ignore this.  I've disabled this particular compiler warning for Picnic in commit 3ec60af, can you see if that clears it, @xvzcf?

dstebila:
> > Do we consider Ubuntu 18.04 on Windows Subsystem for Linux to be a target platform?
> 
> I would certainly hope so as this has been my dev environment for the last 2 years!

And here I thought you were normal.  :)

> > As for fixing, can you see if the warning is cleared by explicitly casting to `(char)`?
> 
> Yes, it does. I've added a `(char)` cast in front of the arguments to `_mm256_set1_epi8` on lines 46, 76, and 79.

Okay, I've pushed that change in commit 519feb8, can you confirm it works?

christianpaquin:
> Okay, I've pushed that change in commit [519feb8](https://github.com/open-quantum-safe/liboqs/commit/519feb8716312bb25ca20a71d063678e5c99a13b), can you confirm it works?

It works! Thanks

src/kem/bike/additional/secure_decode_avx2.c
```diff
@@ -31,7 +31,7 @@ rotate256_big(OUT syndrome_t *out, IN const syndrome_t *in, IN size_t ymm_num) {
 	bike_static_assert(sizeof(*out) > (YMM_SIZE * (R_YMM + (2 * R_YMM_HALF_LOG2))),
 	                   rotr_big_err);
 
-	const __m256i all_one_mask = _mm256_set1_epi8(0xff);
+	const __m256i all_one_mask = _mm256_set1_epi8((char) 0xff);
```
drucker-nir: Why did you add this `(char)` cast everywhere? Which warning/compiler/os triggers this change? 
In general, if BIKE's code is changed without updating me or Shay, we may override the change next time we merge BIKE's repository. It is best to update the code there and then merge.

xvzcf:
> Looking into, the Picnic code is already checking if aligned_alloc is properly available on the platform, so think we can ignore this. I've disabled this particular compiler warning for Picnic in commit [3ec60af](https://github.com/open-quantum-safe/liboqs/commit/3ec60af7202f9147a7b0e2cc8f9902419492e83a), can you see if that clears it, @xvzcf?

It did.

PR #646, opened 2020-03-05
Fix BIKE code to not depend on char

drucker-nir:
This PR partially resolves https://github.com/open-quantum-safe/liboqs/pull/633 and also merges BIKE's latest code. 

BIKE code package is compiled with "-funsigned-char". The reason is (according to [gcc C-Dialect-Options](https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html)) that:
> Each kind of machine has a default for what char should be. It is either like unsigned char by default or like signed char by default. [...]
The type char is always a distinct type from each of signed char or unsigned char, even though its behavior is always just like one of those two.

When this flag is set `char` is always treated as `unsigned char`. 

In order to allow the code to be compiled as part of other libraries (e.g., LibOQS) we removed this flag from BIKE code package and updated the code accordingly.


christianpaquin:
This works on my Ubuntu on WSL.

PR #647, opened 2020-03-05
Added option to make the library portable

gadoofou87:


xvzcf:
Hi @gadoofou87, I don't think this option is necessary. 

Code that makes use of CPU extensions can be introduced into the build process only when the `USE_*_INSTRUCTIONS` options are set, which happens when the `.CMake/cpu-extensions.cmake` file is included, which is only when `CMAKE_BUILD_TYPE` is not `Debug` or `Generic` (and when CMAKE_SYSTEM_NAME is not `Android`).

We can compile portable (and optimized) code by running `cmake .. -DCMAKE_BUILD_TYPE=Generic`.

gadoofou87:
Hi @xvzcf, sorry for my inattention, I'll close the PR.

xvzcf:
No problem.

Issue #606, opened 2020-01-30
SHA-3 AVX2 Tests


xvzcf:
The `*4x` variants [here](https://github.com/open-quantum-safe/liboqs/blob/master/src/crypto/sha3/sha3x4.c) should also be tested in `tests/test_sha3.c`.

We should also ensure we're up to date with the Keccak code package.

dstebila:
Fixed by #651.

PR #651, opened 2020-03-09
Added testing of SHA-3 AVX2 implementations

gadoofou87:
Attempt to solve https://github.com/open-quantum-safe/liboqs/issues/606


dstebila:
Thanks!

PR #652, opened 2020-03-09
CMake refactoring.

xvzcf:
1. Overhauled the propagation of compiler flags

2. Added `Dependency` CMake build type and the `LIBOQS_BUILD_TYPE` variable. 

`LIBOQS_BUILD_TYPE`, when specified, is assigned to `CMAKE_BUILD_TYPE`, and can be set in the CMakeLists.txt build files of other projects that use liboqs as a dependency. 

When `CMAKE_BUILD_TYPE` is `Dependency`, only `liboqs` is built, and the `gen_docs`, `prettyprint`, and `run_tests` targets (along with any corresponding build artifacts) are excluded. Besides reducing build time, this also ensures that a name collision will not occur if, for example, there is a `run_tests` target in the parent project using liboqs.

3. Removed src/common/CMakeLists.txt and src/crypto/CMakeLists.txt


.CMake/compiler-flags.cmake
```diff
@@ -0,0 +1,85 @@
+if(CMAKE_C_COMPILER_ID MATCHES "Clang")
+    add_compile_options(-Werror)
+    add_compile_options(-Wall)
+    add_compile_options(-Wextra)
+    add_compile_options(-Wpedantic)
+    #add_compile_options(-Wreserved-id-macro)
+    #add_compile_options(-Wbad-function-cast)
+    # valid BIKE code sets this off
+    add_compile_options(-Wno-missing-braces)
```
dstebila: For warnings that only affect individual implementations, can we put them in their respective CMakeLists.txt file?

.CMake/compiler-flags.cmake
```diff
@@ -0,0 +1,85 @@
...
+    add_compile_options(-Wstrict-prototypes)
+    add_compile_options(-Wshadow)
+    add_compile_options(-Wformat=2)
+    add_compile_options(-Wfloat-equal)
+    add_compile_options(-Wwrite-strings)
+    add_compile_options(-fvisibility=hidden)
+    if(CMAKE_C_COMPILER_VERSION VERSION_LESS "5")
+        # GCC 4.9 erroneously reports warnings for certain valid code in BIKE
+        # https://stackoverflow.com/questions/13746033/how-to-repair-warning-missing-braces-around-initializer
+        add_compile_options(-Wno-missing-braces)
```
dstebila: Same as above

Issue #624, opened 2020-02-20
Graceful runtime CPU feature handling


baentsch:
Ensure all code is designed such that it can not only be configured at compile time to deal with different CPU feature sets (e.g., presence/absence of AVX2) but also handle those gracefully at runtime, e.g, by automatically falling back to a non-optimised version of the same functionality.

In other words, all code compiled with, e.g., `-mavx2` should also run without error on CPUs that do *not* have support for all features (AVX2 in this example).

To help with this, we will integrate the [CPU features](https://github.com/google/cpu_features) library and for starters will add warnings if the library is running on a CPU with a feature set different from the one encountered during build (reducing the surprise when one encounters an `illegal instruction exception` during execution until this issue is fully addressed :-)

Edited by @dstebila to add:

- [X] Use cpu\_features instead of \_\_builtin\_cpu
- [X] Update copy-from-pqclean schemes to do runtime detection
- [x] Runtime detection for SHA-3 ~using  [XKCP](https://github.com/XKCP/XKCP)~
- [X] Runtime detection for Frodo (#878)
- [ ] Runtime detection for SIKE
- [x] Runtime detection for BIKE (see https://github.com/open-quantum-safe/liboqs/pull/1030#issuecomment-877292940)
~- [ ] Runtime detection for qTesla~
- [x] Runtime detection for Picnic
- [x] Runtime detection for AES

dstebila:
Do we consider this done? @xvzcf @baentsch

xvzcf:
Not yet, but almost there.

baentsch:
[...]

PR #654, opened 2020-03-10
Use cpu_features instead of __builtin_cpu

gadoofou87:
FMA4 and SSE4A detection is not supported by the library. Additionally enables NEON detection in Picnic


dstebila:
We have been having some discussions internally for a while about how to approach this.  @baentsch and @xvzcf have been leading the investigation, maybe they can chip in here about what they've been planning.

baentsch:
This looks like a step in the right direction: Adding cpu_features to the overall build and using it initially to detect compile-time features. When the Linux CI builds also pass (CXX packages to be added to the docker images) I'd say this PR is OK. 

@xvzcf, I'd defer to you as to when to include this: Please also let me know whether you'd add CXX to the Linux CI images or whether I should do it. If you want to do it before pushing the docker add-on, the package targets to be added are `[yum install] gcc-c++` for Centos and `[apt-get install] g++` for Ubuntu/Debian

Next step after this then is using the package for runtime feature detection _within_ the code.

xvzcf:
I can add the C++ compilers to the CI images, which I've been meaning to look at anyway since some packages installed there (like `xsltproc` or `graphviz`) might be unnecessary.

baentsch:
@xvzcf Thanks in advance for the CI image updates! Regarding copying `cpu_features`, I'm not sure: The guys have an open bug (regarding shared lib building) and it kind of would be nice if the fix at some point automatically flows into `liboqs` instead of us having to copy the new fixed version in.

gadoofou87:
Adding a compiler to Linux CI images may solve the problem, but the library requires it by mistake, it is successfully built without it.
```
[archie@archlinux build]$ cmake .. -GNinja
-- The C compiler identification is GNU 9.2.1
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Looking for dlfcn.h
-- Looking for dlfcn.h - found
-- Looking for getauxval
-- Looking for getauxval - found
-- Configuring done
-- Generating done
-- Build files have been written to: /home/archie/Documents/cpu_features/build
[archie@archlinux build]$ ninja -v
[1/9] /usr/bin/cc -DHAVE_DLFCN_H -DHAVE_STRONG_GETAUXVAL -DSTACK_LINE_READER_BUFFER_SIZE=1024 -I../include -I../include/internal -O3 -DNDEBUG   -std=gnu99 -MD -MT CMakeFiles/unix_based_hardware_detection.dir/src/unix_features_aggregator.c.o -MF CMakeFiles/unix_based_hardware_detection.dir/src/unix_features_aggregator.c.o.d -o CMakeFiles/unix_based_hardware_detection.dir/src/unix_features_aggregator.c.o   -c ../src/unix_features_aggregator.c
[2/9] /usr/bin/cc -DSTACK_LINE_READER_BUFFER_SIZE=1024 -I../include -I../include/internal -O3 -DNDEBUG   -std=gnu99 -MD -MT CMakeFiles/utils.dir/src/filesystem.c.o -MF CMakeFiles/utils.dir/src/filesystem.c.o.d -o CMakeFiles/utils.dir/src/filesystem.c.o   -c ../src/filesystem.c
[3/9] /usr/bin/cc -DHAVE_DLFCN_H -DHAVE_STRONG_GETAUXVAL -DSTACK_LINE_READER_BUFFER_SIZE=1024 -I../include -I../include/internal -O3 -DNDEBUG   -std=gnu99 -MD -MT CMakeFiles/unix_based_hardware_detection.dir/src/hwcaps.c.o -MF CMakeFiles/unix_based_hardware_detection.dir/src/hwcaps.c.o.d -o CMakeFiles/unix_based_hardware_detection.dir/src/hwcaps.c.o   -c ../src/hwcaps.c
[4/9] /usr/bin/cc -DSTACK_LINE_READER_BUFFER_SIZE=1024 -I../include -I../include/internal -O3 -DNDEBUG   -std=gnu99 -MD -MT CMakeFiles/utils.dir/src/stack_line_reader.c.o -MF CMakeFiles/utils.dir/src/stack_line_reader.c.o.d -o CMakeFiles/utils.dir/src/stack_line_reader.c.o   -c ../src/stack_line_reader.c
[5/9] /usr/bin/cc -DSTACK_LINE_READER_BUFFER_SIZE=1024 -I../include -I../include/internal -O3 -DNDEBUG   -std=gnu99 -MD -MT CMakeFiles/utils.dir/src/string_view.c.o -MF CMakeFiles/utils.dir/src/string_view.c.o.d -o CMakeFiles/utils.dir/src/string_view.c.o   -c ../src/string_view.c
[6/9] /usr/bin/cc -DSTACK_LINE_READER_BUFFER_SIZE=1024 -I../include -O3 -DNDEBUG   -std=gnu99 -MD -MT CMakeFiles/list_cpu_features.dir/src/utils/list_cpu_features.c.o -MF CMakeFiles/list_cpu_features.dir/src/utils/list_cpu_features.c.o.d -o CMakeFiles/list_cpu_features.dir/src/utils/list_cpu_features.c.o   -c ../src/utils/list_cpu_features.c
[7/9] /usr/bin/cc -DSTACK_LINE_READER_BUFFER_SIZE=1024 -I../include -I../include/internal -O3 -DNDEBUG   -std=gnu99 -MD -MT CMakeFiles/cpu_features.dir/src/cpuinfo_x86.c.o -MF CMakeFiles/cpu_features.dir/src/cpuinfo_x86.c.o.d -o CMakeFiles/cpu_features.dir/src/cpuinfo_x86.c.o   -c ../src/cpuinfo_x86.c
[8/9] : && /usr/bin/cmake -E remove libcpu_features.a && /usr/bin/ar qc libcpu_features.a  CMakeFiles/utils.dir/src/filesystem.c.o CMakeFiles/utils.dir/src/stack_line_reader.c.o CMakeFiles/utils.dir/src/string_view.c.o CMakeFiles/cpu_features.dir/src/cpuinfo_x86.c.o && /usr/bin/ranlib libcpu_features.a && :
[9/9] : && /usr/bin/cc -O3 -DNDEBUG  -rdynamic CMakeFiles/list_cpu_features.dir/src/utils/list_cpu_features.c.o  -o list_cpu_features  libcpu_features.a  -ldl && :
```
I opened the issue https://github.com/google/cpu_features/issues/109. Maybe we should just wait for the fix? I can also try to add FMA4 and SSE4A support to their library.

baentsch:
One way or another, all checks have passed now :-) @gadoofou87 if you'd have time, sure, more CPU feature support is always good. If you know that lib somewhat (?) it'd also be great if you could maybe take a look at [cpu_features issue 108](https://github.com/google/cpu_features/issues/108) which is blocking shared-lib usage of `cpu_features`.

xvzcf:
I think we should use `add_subdirectory` instead of `find_package` or `ExternalProject_Add ` since that's also what's recommended [here](https://github.com/google/cpu_features/blob/master/cmake/README.md).

It also (sadly) seems like the best way to include the project would be to copy the source over verbatim, maybe with a script to fetch a new version every so often.

gadoofou87:
This is recommended, but we cannot use `add_subdirectory` because we need to build `cpu_features` before running `list-cpu-extensions.c`. I also don't understand why we can't use `ExternalProject_Add` if it works correctly?

baentsch:
@gadoofou87 Thank you very much for the cpu_features shared-lib fix! This had held me up making progress on #624 

Regarding the question how to integrate `cpu_features`, I usually also like stuff that "just works", but this statement from [the link quoted above](https://github.com/google/googletest/blob/master/googletest/README.md#incorporating-into-an-existing-cmake-project ) makes me uneasy in this case as we want to ensure our system works (builds/runs) smoothly on all platforms (incl. Windows):

> This has the significant advantage that the same compiler and linker settings are used between gtest and the rest of your project, so issues associated with using incompatible libraries (eg debug/release), etc. are avoided. This is particularly useful on Windows

--> How certain are you that `ExternalProject_Add` has no negative aspects in our case, incl. Windows (with which I have very little experience)?

@xvzcf : At first glance, _I_ don't see any drawbacks as --unlike `gtest` users in the case quoted above-- we don't need to share build parameters with cpu_features from liboqs, so what'd you say: If it works (all build checks pass again) on all platforms, what about giving the `ExternalProject_Add` integration a go?  As `cpu_features` sure looks like a live project with updates/fixes coming online, a "dynamic" integration (either via `ExternalProject_Add` or git subproject) benefits OQS. The stuff I read about [git subproject integration](https://git-scm.com/book/en/v2/Git-Tools-Submodules) makes me think that that approach would be too cumbersome for us.

baentsch:
@gadoofou87 On second thought, I possibly have to retract parts of my statement above: `liboqs` building _will_ (need to) be interdependent with `cpu_features` build: If `liboqs` is build as a shared lib, so will (need to be) `cpu_features` (as we intend to use it internally to select which algorithm variant to run (generic C vs e.g., AVX2-optimized code),right? --> From your experience, (how) does `ExternalProject_Add` cater to this use case (going beyond the compile-time feature detection that this PR addresses)?

gadoofou87:
Updated PR. Returned detection of FMA4 and SSE4A, as support in cpu_features is added. cpu_features now inherits parameter BUILD_SHARED_LIBS.

It also looks like there is a bug on Windows with the build of optimized version of Keccak which is not related to this PR.
```
FAILED: src/CMakeFiles/oqs.dir/common/sha3/sha3x4.c.obj 
C:\PROGRA~2\MICROS~1\2019\COMMUN~1\VC\Tools\MSVC\1424~1.283\bin\Hostx64\x64\cl.exe  /nologo -Doqs_EXPORTS -Iinclude /DWIN32 /D_WINDOWS /W3 /showIncludes /Fosrc\CMakeFiles\oqs.dir\common\sha3\sha3x4.c.obj /Fdsrc\CMakeFiles\oqs.dir\ /FS -c ..\src\common\sha3\sha3x4.c
..\src\common\sha3\sha3x4.c(23): warning C4244: '=': conversion from 'uint64_t' to 'uint8_t', possible loss of data
C:\projects\liboqs\src\common\sha3\keccak4x/KeccakP-1600-times4-SIMD256.c(423): error C2440: 'type cast': cannot convert from '__m256d' to 'V256'
C:\projects\liboqs\src\common\sha3\keccak4x/KeccakP-1600-times4-SIMD256.c(423): error C2168: '_mm256_xor_si256': too few actual parameters for intrinsic function
```

baentsch:
@gadoofou87 Perfect & Thank you again! I'm now all in favour of accepting this PR and ask @xvzcf for final review.

xvzcf:
> This is recommended, but we cannot use `add_subdirectory` because we need to build `cpu_features` before running `list-cpu-extensions.c`. I also don't understand why we can't use `ExternalProject_Add` if it works correctly?

That is true. Let me think about it a bit more before getting back to you.

xvzcf:
Having thought about it a bit more, it seems there's no good way to get around the use of `ExternalProject_Add`. Also,

- I think `CMakeLists.txt.in`should be in a `third_party/cpu_features/` folder, the way it's done [here](https://github.com/opencv/opencv/tree/master/3rdparty/ffmpeg). 
- We should note the `cpu_features` commit we are using, so that each liboqs commit can be associated with a specific `cpu_features` commit that it is known to work with.

There is one more wrinkle. When the time comes when `cpu_features` has to be linked to liboqs, I think it would be best if we retrieved the values of the variables [here](https://github.com/google/cpu_features/blob/master/CMakeLists.txt#L143) (`CPU_FEATURES_HDRS` and `CPU_FEATURES_SRCS`) and directly built them as part of liboqs, rather than having to juggle two different libraries where one depends on the other. I think this is possible, but I haven't convinced myself completely. What do you think @gadoofou87?

gadoofou87:
@xvzcf Unfortunately, we cannot build `cpu_features` as part of `liboqs` because, as I said earlier, we need to build `cpu_features` before running `list_cpu_extensions.c`. In my opinion, the only thing we can do is to completely stop detecting CPU extensions in CMake and go to https://github.com/open-quantum-safe/liboqs/issues/624. After that we can opt out of using `list_cpu_extensions.c` and build `cpu_features` as part of `liboqs`

baentsch:
@gadoofou87 @xvzcf I agree (no big surprise :-) What about this, then: Remove CPU feature detection in initial build ("Debug" BUILD_TYPE doesn't depend on/needs this, right?), build liboqs incl. cpu_features then (say in a `build_bootstrap` folder) and only _then_ use the remaining BUILD_TYPES (then enabled by `cpu_features` having been created in `build_bootstrap` by then)? I have the gut feeling that we'll need non-optimised object files for best portability anyway...

xvzcf:
> @xvzcf Unfortunately, we cannot build `cpu_features` as part of `liboqs` because, as I said earlier, we need to build `cpu_features` before running `list_cpu_extensions.c`. In my opinion, the only thing we can do is to completely stop detecting CPU extensions in CMake and go to #624. After that we can opt out of using `list_cpu_extensions.c` and build `cpu_features` as part of `liboqs`

I understand we have to build `cpu_features` for `list_cpu_extensions.c`, and I do see `list_cpu_extensions.c`as a temporary workaround until we can move all feature detection into the code and out of CMake. Where I was going with this was that we could build `cpu_features` twice for now: the first time as a library to attach to `list_cpu_extensions.c`, and the second time as a set of object files to be compiled into liboqs.

> @gadoofou87 @xvzcf I agree (no big surprise :-) What about this, then: Remove CPU feature detection in initial build ("Debug" BUILD_TYPE doesn't depend on/needs this, right?), build liboqs incl. cpu_features then (say in a `build_bootstrap` folder) and only _then_ use the remaining BUILD_TYPES (then enabled by `cpu_features` having been created in `build_bootstrap` by then)? I have the gut feeling that we'll need non-optimised object files for best portability anyway...

If I understand correctly, we get CPU information from the first Debug build of liboqs+cpu_features, which we then use to set the variables [here](https://github.com/open-quantum-safe/liboqs/blob/master/src/oqsconfig.h.cmake#L12) for the second Optimized `liboqs` build?

baentsch:
@xvzcf Yup - that two-step approach was what I had in mind -- without having done it yet -- but I don't see a fundamental flaw in it.

xvzcf:
After some discussion with @baentsch, I think the best way forward for runtime use is to build `cpu_features` as a static PIC library whose symbols can be copied into a static or shared version of liboqs.

@gadoofou87 could you apply [this patch](https://github.com/open-quantum-safe/liboqs/files/4371131/patch.txt) and rebase with master?

gadoofou87:
@xvzcf It's done!

xvzcf:
One last thing. We maintain a list of contributors [here](https://github.com/open-quantum-safe/liboqs/blob/master/CONTRIBUTORS). If you'd like, you can add your name and/or your affiliation/link to GitHub profile there. The choice is yours.

gadoofou87:
@xvzcf Thank you for the suggestion :)

PR #655, opened 2020-03-10
Fixing BIKE compilation on an AVX512 supporting platform

drucker-nir:
This PR fixes the compilation for BIKE in the following scenarios:
1) When using Clang-3.9
2) When running on an AVX512 platform

It also fixes a "missing new line" warning for a qtesla header.

Note that the compilation of LibOQS (with clang-3.9) still fails due to the flag "-Wno-unguarded-availability-new" (in picnic) that is not recognized by clang-3.9.


dstebila:
Thanks @drucker-nir!

@xvzcf note that this might affect #652.

CMakeLists.txt
```diff
@@ -49,12 +49,13 @@ if(CMAKE_C_COMPILER_ID STREQUAL "GNU" OR
         if (NOT CMAKE_SYSTEM_NAME STREQUAL "Android")
            include(.CMake/cpu-extensions.cmake)
            set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O3 -fomit-frame-pointer -march=native")
```
xvzcf: Why were `-fomit-frame-pointer` and `-O3` removed?

xvzcf:
Removed my comment since I misread a line.

PR #669, opened 2020-03-19
Set sanitizer flags differently

dstebila:
Sanitizer builds fail during linking, because sanitizer flags also need to be passed during linking.  It seems like the approach introduced in #652 using `add_compile_options` doesn't do that.  I tried using `add_link_options` but that didn't work, so I went back to what was working before #652.


xvzcf:
@dstebila could you try the latest commit?

dstebila:
> @dstebila could you try the latest commit?

Works for me.

Issue #608, opened 2020-02-07
LedaKemLT52 cannot be run in a worker thread on macOS


vsoftco:
Due to small stack size of worker threads under macOS (512kb), LEDAKEM52LT emits a BUS ERROR (access violations -Exception: EXC_BAD_ACCESS (code=1, address=0x70000292c4c8) ) when attempted to be run in a non-main (worker) thread under macOS. The first culprit is

https://github.com/open-quantum-safe/liboqs/blob/fb458165698c5850dcf08d7268a2b6c90e47e378/src/kem/ledacrypt/pqclean_ledakemlt52_leaktime/dfr_test.c#L13

where the array `gamma` seems to have 609068 bytes, more than can be stored on the stack of a single non-main thread in macOS (512kb). Allocating `gamma` on the heap still doesn't solve the problem, other violations of the same sort keep appearing (I didn't have time to go over the rest). So we should think whether we want to keep the code as is, or document that it shouldn't be run on a thread under macOS, or change it.

dstebila:
Do you have a sense of how many places we'd have to change to make this work?  A few or dozens and dozens?

vsoftco:
@dstebila Definitely more than 1 place. I tried changing the first stack allocation, and got into another issue of the same flavour down the line. I'll look into it when I get some time. Hopefully not dozens and dozens and places we'd need to change.

dstebila:
https://github.com/open-quantum-safe/liboqs/commit/795c1b7932d8e3d44277c7bb1c6c45d091cde1c7#diff-9d282cb2abf60a78fb263d773e0dabe2R84 fixed a memory leak in LEDAkem, there could be a small chance this fixed it?

dstebila:
WONTFIX for now.

PR #671, opened 2020-03-19
Run the test_kem and test_sig tests on a thread when possible

dstebila:
Algorithms with large stack usage cause problems when running in threads (e.g., #608, https://travis-ci.com/github/open-quantum-safe/liboqs-cpp/jobs/299650240).  This PR moves the tests in `test_kem` and `test_sig` onto a thread when compiling with clang or gcc, so as to detect these failures at the liboqs level, rather than leaving it to downstream sub-projects.

Note that the following schemes are currently failing:
```
Classic-McEliece-348864
Classic-McEliece-348864f
Classic-McEliece-460896
Classic-McEliece-460896f
Classic-McEliece-6688128
Classic-McEliece-6688128f
Classic-McEliece-6960119
Classic-McEliece-6960119f
Classic-McEliece-8192128
Classic-McEliece-8192128f
LEDAcryptKEM-LT52
Rainbow-IIIc-Classic
Rainbow-IIIc-Cyclic
Rainbow-IIIc-Cyclic-Compressed
Rainbow-Vc-Classic
Rainbow-Vc-Cyclic
Rainbow-Vc-Cyclic-Compressed
```

I'm not sure how to approach solving this.  Do we disable these algorithms by default?  Try to rewrite them so they're on the heap, rather than the stack?  Try to enlarge the stack in how all our subprojects are compiled so that they can be run (but then people might encounter this bad behaviour outside of our build)?

Advice welcome, @vsoftco @xvzcf @christianpaquin @thomwiggers @baentsch.


baentsch:
My immediate reaction: Disable them until they "behave"... (... incl. under pthreads). But then again, the great thing about `liboqs` and all libs/apps building on top is the availability to run all these algorithms (in different stages of "implementation hardening") side-by-side.

So what about this compromise: 
- Keep all algorithms in
- Keep adding them to downstream libs/apps like openssl
- Exclude the list above from pthread testing
- Document that exclusion clearly on the OQS website/git repo.
- Make the algorithm implementors aware of this problem so they can fix it (if they want to do that)

--> This way, we (OQS) can focus on what matters to all users: people can keep using any PQC they want courtesy `liboqs` -- until they hit such "snag". Fixing those "snags" can/and should be done upstream (it's not only "us"/`liboqs`-enabled SW stacks failing on this). E.g., wouldn't it be a nice addition to the PQClean project to require such 'pthreads' safety (and test for it) adding to the [prerequisites documented by PQClean](https://github.com/PQClean/PQClean#requirements-on-c-implementations-that-are-automatically-checked).

thomwiggers:
The large stack-usage of some implementations is a long-known problem but I also don't know how to get around it (without large amounts of engineering effort). It sometimes helps to put the public keys/secret keys on the heap themselves; this allows the stack-based implementations that much more room on the stack. However, it's still going to be a problem.

> E.g., wouldn't it be a nice addition to the PQClean project to require such 'pthreads' safety (and test for it) adding to the prerequisites documented by PQClean.

I'm afraid that's not possible. PQClean bans any use of the heap to allow running implementations on platforms that don't offer an allocator. I'd prefer having more reference implementations, even if they have these kinds of problems.

vsoftco:
@dstebila will review by Friday

vsoftco:
@dstebila Is https://github.com/open-quantum-safe/liboqs/pull/671/commits/306d5f974754f1ddde35d362e425636b3a39b630 ok with you?

dstebila:
> @dstebila Is [306d5f9](https://github.com/open-quantum-safe/liboqs/commit/306d5f974754f1ddde35d362e425636b3a39b630) ok with you?

Fancier than we need, but that's fine.

> We most likely need to add `pthread` to the linker for test_kem/sig, @xvzcf ?

I just added it, we'll see if it works.

vsoftco:
@dstebila Seems to be working, I'll merge once AppVeyor passes

PR #673, opened 2020-03-22
Fix typo in picnic.

jyao1:
Fix typo in picnic algorithm.
AATR_CONST should be ATTR_CONST.


jyao1:
Close by mistake. Reopen now.
Please review the change.

PR #674, opened 2020-03-23
Support to cross-compile for Windows amd64 on Linux

kevinmkane:
Add a toolchain file for cross-compiling for Windows amd64 on Linux.

Also make the MinGW/MSYS/Cygwin builds usable by CMake before version 3.13.0, when `add_link_options` was introduced, by settings these linker flags in the older fashion. Ubuntu 18 (Bionic) provides only version 3.10.2, and this seems to be an important build platform for cross-compiling.


dstebila:
Thanks Kevin!  Are there any specific notes we should add to our [cross compilation wiki page](https://github.com/open-quantum-safe/liboqs/wiki/Platform-specific-notes-for-building-liboqs) with respect to this?

kevinmkane:
Following the example to cross-compile on ARM, this should be sufficient:
```
apt install gcc-mingw-w64
mkdir build && cd build
cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=../.CMake/toolchain_windows-amd64.cmake ..
ninja
```

As an aside, the ARM instructions have a small error: it lists the toolchain file as `toolchain-rasppi.cmake` (with a hyphen), whereas the file's name is `toolchain_rasppi.cmake` (with an underscore) in the repo.

dstebila:
Thanks @kevinmkane, I've added your instructions to the wiki.

Issue #666, opened 2020-03-18
Ensure Keccak in picnic has separate namespace


baentsch:
Goal: PQClean algorithms referencing `KeccakP1600times4_PermuteAll_24rounds` should not accidentally "wind up" calling Picnic's implementation thereof

christianpaquin:
Done in PR #676; @baentsch, does that fix your issue?

baentsch:
@christianpaquin Yes, it does: Thanks! Sorry for not closing the issue earlier.

baentsch:
Unfortunately, I didn't test well enough: When now "opening up" (removing the `static` declaration) the `common` symbol `KeccakP1600times4_PermuteAll_24rounds` (in `src/common/sha3`) linking creates a collision again. I'll need to investigate some more. Do we agree that it should be the goal that the `common/sha3`-originating code should be the one made available to `liboqs` algorithms?

christianpaquin:
Yes, all non-namespaced Keccak calls should refer to the code in `common/sha3`. What platform and optimizations (e.g., avx2, etc.) are you using? The Keccak code uses a lot of macros (functions and variants are defined by macros, sometimes implemented in C, sometimes in assembly), and I tried to namespace all of the picnic ones but maybe I missed a few (quite possible if I didn't test on the same platform w/ same opt as you). I'll investigate.

christianpaquin:
PR #705 should fix the picnic namespacing.

baentsch:
Confirmed. Will follow up with a change to open up symbol for (avx2) `src/common/sha3` Keccak

PR #676, opened 2020-03-24
Properly namespace the picnic internal functions

christianpaquin:
Addresses issue #666.


christianpaquin:
Still failing on some platform; not sure why...

christianpaquin:
Finally working. @dstebila, I contacted the picnic team. They confirmed their package is the unmodified keccak implementation, so in a 2nd phase we could migrate this to the common folder, but I suggest merging this now to unblock other dependent PRs.

PR #677, opened 2020-03-24
As per suggestion

baentsch:
Thanks @xvzcf for the reminder


CONTRIBUTORS
```diff
@@ -17,3 +17,4 @@ Douglas Stebila (University of Waterloo)
 Goutam Tamvada (University of Waterloo)
 John Underhill
 Sebastian Verschoor (University of Waterloo)
+Michael Baentsch (IBM Research-Zurich)
```
xvzcf: Could you place this before "Eric Crockett"? The list is alphabetically sorted by last name.

PR #678, opened 2020-03-25
build-type Optimized reactivated

baentsch:
Without this, `CMAKE_BUILD_TYPE=Optimized` doesn't execute (correctly, i.e., CPU feature defines are not set correctly). 

Going forward, might it be an idea to add to the CI test suite validation for different build types? Possibly performance checks (`Optimized`-built test results should be better than `Debug`-built test runs)? I admit I don't know how/whether this is possible (tests between job results), though.


gadoofou87:
@baentsch I've seen this error before, and I've written about it https://github.com/open-quantum-safe/liboqs/pull/654#issuecomment-598131843. It looks like the optimized implementation of keccak does not work on Windows.

xvzcf:
CircleCI's absence is strange indeed. Could you apply this [patch.txt](https://github.com/open-quantum-safe/liboqs/files/4382084/patch.txt)? It should fix set the `OQS_USE_*` variables if `CMAKE_BUILD_TYPE` is `Optimized` or quit with a fatal error otherwise. I've also modified the top-level CMakeLists.txt to include `cpu_features` only if the build type is `Optimized`.

Maybe another commit will kickstart CircleCI.

baentsch:
@gadoofou87 Thanks for this explanation/reminder. @xvzcf Would you want to add to your patch also some disablement of Windows-build of keccak (if you know how to easily do it)? Or shall I apply  and commit as-is?

dstebila:
> Hmm - question: Why is this PR not triggering CircleCI tests, too? I must admit I'm not well-versed in AppVeyor and MSVC to understand what's happening here...

I do all my stuff in branches in branches in open-quantum-safe/liboqs rather than in my dstebila/liboqs fork.  Maybe you having it in baentsch/liboqs is why the PR is not triggering CircleCI builds?

xvzcf:
CircleCI was triggered [here](https://github.com/open-quantum-safe/liboqs/pull/676), where the branch to be merged in is in the fork.

baentsch:
@dstebila I don't think so -- I'm working like this in all repos and normally, all OQS-CI's are triggered upon a PR... I only recall an instance where it didn't work in `oqs-demos` but Goutam did "something" to remedy this (do you recall, @xvzcf ?)

xvzcf:
> @gadoofou87 Thanks for this explanation/reminder. @xvzcf Would you want to add to your patch also some disablement of Windows-build of keccak (if you know how to easily do it)? Or shall I apply and commit as-is?

This is trickier, we have to remove `march=native` and assign specific CPU extension flags for each algorithm, so maybe this can be done as part of the runtime-feature detection effort.

xvzcf:
> @dstebila I don't think so -- I'm working like this in all repos and normally, all OQS-CI's are triggered upon a PR... I only recall an instance where it didn't work in `oqs-demos` but Goutam did "something" to remedy this (do you recall, @xvzcf ?)

In that case, I just had to give you write access to `oqs-demos`. You already have that for `liboqs`

baentsch:
Strange, this -- indeed, a CCI test build was triggered in my account -- but that's normal as I checked in something to my account. That a check-in is part of a PR in OQS should trigger a build there too, no? Whatever -- the CCI tests in my account all passed, so I'd say this is good to merge :-) And thanks for the sanity check for the `list_cpu_features` build in the patch, @xvzcf

xvzcf:
Could you remove `-DCMAKE_BUILD_TYPE=Optimized` from `appveyor_build.bat` for MSVC and add a TODO note? Could you also fix the comment [here](https://github.com/open-quantum-safe/liboqs/blob/master/.CMake/toolchain_rasppi.cmake#L2) by changing `USE_OPENSSL` to `OQS_USE_OPENSSL`?

PR #683, opened 2020-03-26
Opt build2

baentsch:
msvc Optimized off again. May replace PR #682


dstebila:
What's the status of this one?  I'm a bit confused on its relationship with other open PRs and what needs to happen to merge this.

baentsch:
@dstebila This (failure) was the reason for the work on PR #690 (to see whether the same problems appear in Windows on CircleCI). I personally think this could and should be merged as it fixes a bug. The problem though is that that "activates" other bugs in different algorithms when running them under Windows (@xvzcf has more insight there).

gadoofou87:
@baentsch Can you apply this [patch](https://github.com/open-quantum-safe/liboqs/files/4413066/patch.txt)? It should fix the Windows build and avoid having to disable the Optimized build in Appveyor. This is a temporary solution until the second stage is completed (https://github.com/open-quantum-safe/liboqs/pull/676#issuecomment-604792529)

gadoofou87:
`..\src\common\common.c(11): fatal error C1083: Cannot open include file: 'cpu_features_macros.h': No such file or directory`
@baentsch It looks like we need to build `cpu_features` regardless of the `CMAKE_BUILD_TYPE` parameter value.

baentsch:
@gadoofou87 This doesn't sound right: I actually wanted to add a #define for OQS_RT_DETECTION to en/disable `cpu_feature` use/RT feature detection as not everyone needs that. In general it is beyond me why MSVC doesn't "like" building this -- this is a cmake construct, right? Maybe @xvzcf has an idea given more experience with cmake in Windows (?)

baentsch:
Ahhh -- I think I just understood the latest failure: There's indeed a #define guard missing in `src/common/common.c`: I need to add that before this goes through (as `cpu_feature` use does not have to be active under all [BUILD_TYPE] circumstances). I just wonder which branch to add this to -- quite a lot going on in parallel... I'd say it's best added after PR #692 is merged... @dstebila : Are you OK I go ahead and merge (and squash) that or would you like to take a look (you're also listed as Reviewer on that one)?

gadoofou87:
@baentsch But what do you think about enabling the Optimized build back? My patch was supposed to fix it. I tested it on Visual Studio 2019 and everything was successfully compiled.

baentsch:
@gadoofou87 Done that. Might pass, but does not solve the problem with Debug builds. But OK, one problem after the other...

gadoofou87:
@baentsch Thank you! If you want, I can try to make a patch that solves this problem.

baentsch:
OK, it's now passing & I asked @xvzcf for Review (and merge, ideally). @gadoofou87 : Thanks for the offer and support so far. I have submitted PR #696 to deal with non-optimised builds. Let's see whether that goes through AppVeyor...

vsoftco:
@baentsch @xvzcf Now there seem to be an issue with Linux failing under CI

xvzcf:
That should be fixed by https://github.com/open-quantum-safe/liboqs/pull/696. I'm just waiting for the builds to pass there before merging ...

PR #689, opened 2020-03-30
Runtime feature detection

baentsch:
This PR creates static flags to test for presence/absence of the same CPU features that we presently make available with #defines. This way, the presence of the `cpu_features` library can be completely hidden from external users. For efficiency reasons I also added an `OQS_init` method which sets these flags once so that the code for this does not have to be in all places where feature detection may be needed.

Please have a look at this proposal and let me know what you think.

If you'd be OK, integrating RT feature detection then is trivial indeed: Where we currently only have `OQS_USE_AVX2_INSTRUCTIONS` compile time switches we can augment them now with RT switches
```
if (OQS_RT_AVX2_INSTRUCTIONS) { 
   // call AVX2-enabled functions here
} 
else {
  // call non-AVX2 code here
}
```


dstebila:
Can you think of a clever way of getting OQS\_init to run implicitly, rather than having every calling program run it?

dstebila:
> If you'd be OK, integrating RT feature detection then is trivial indeed: Where we currently only have `OQS_USE_AVX2_INSTRUCTIONS` compile time switches we can augment them now with RT switches
> 
> ```
> if (OQS_RT_AVX2_INSTRUCTIONS) { 
>    // call AVX2-enabled functions here
> } 
> else {
>   // call non-AVX2 code here
> }
> ```

That looks like a good approach to me.

baentsch:
> Can you think of a clever way of getting OQS_init to run implicitly, rather than having every calling program run it?

I'm afraid I don't see an "automatic" way. Maybe for a shared lib but certainly not for a static version. The only way to hide `OQS_init` would be to put an "initialisation-asserted" call to it into all `liboqs` entry points. But if I'm not mistaken, there are many (which, admittedly, are most generated by `copy_from_pqclean`). Such a "hidden call" also would cause a performance deterioration to all liboqs invocations -- not large but visible. What about this proposal: We merge this as-is, wire up all PRs and do performance tests throughout the SW stack all the way down at OQS-OpenSSL apps, possibly with a version containing "initialisation assertion calls" to see the impact of them? If we'd do automatic initialisation, we could also make RT feature detection a #define'able feature such that users unwilling to pay the performance penalty for this (i.e., that only build a system for a fixed platform) won't have to do so.

Let's decide in tomorrow's call how to proceed (maybe create an issue?).

dstebila:
Wiring up all the entry points would be annoying, but doable.  Presumably the check we'd do at the beginning of each entry point would be to check if a single global flag has been set, and if not, then run the CPU features detection which also sets the flag.  It's just checking one flag, the performance hit would come from cache misses or branch prediction failures.  It seems like there is a GCC attribute that one can use (`__builtin_expect`) to provide some branch prediction hints.  Rather than doing full stack performance testing down to OQS-OpenSSL, I'd start with just seeing what the performance impact is with liboqs' speed\_kem and speed\_sig is, and if it seems to be negligible (which I think it will be), then just go with it.

baentsch:
Agree. I'll put something to facilitate that right into the `copy_from_pqclean` templates.

dstebila:
This seems to be causing a build failure in liboqs-cpp (https://travis-ci.com/github/open-quantum-safe/liboqs-cpp/jobs/310866948).

baentsch:
Yes, it does :-( but there's a fix with PR #692 (Thanks @vsoftco !). I'd probably extend it with an (OQS_) define for RT detection so that this whole feature can be toggled on or off as not everyone needs it and we can selectively activate it as we work through the whole stack.

PR #692, opened 2020-04-01
fixing multiple defs linker errors

vsoftco:
We need those flag variables to be declared `extern` in the header file `common.h`, and have them defined in `common.cpp`, otherwise we get "multiple definition error" when linking object files that each include "common.h" (that's why the liboqs-cpp wrapper is emitting a linker error in unit testing, see https://travis-ci.com/github/open-quantum-safe/liboqs-cpp/jobs/310866948). Do we want the flags to be visible in the oqs object file? That's what I assumed, and used `OQS_API` for the definitions in `common.c`.

We may also consider grouping all flags in a `struct`...


baentsch:
@vsoftco Thanks very much: Looking at the changes, this is a logically correct fix to the downstream errors observed. I'm out of my depth though why Windows doesn't "like" the definitions (see AppVeyor build).

vsoftco:
@baentsch weird... maybe because of `OQS_API` missing in the header?

vsoftco:
@baentsch btw, how does this last commit looks? I packed all those flags into a struct, and added a `OQS_API OQS_RT OQS_RT_cpu_flags(void)` that retrieves them. So there should be no more initialization issues and MSVC incompatibilities.

baentsch:
@vsoftco Looks good to me. Let's see what Windows-CI says (btw, is there any way to cancel AppVeyor runs that are still executing already stale checkins?)

vsoftco:
> @vsoftco Looks good to me. Let's see what Windows-CI says (btw, is there any way to cancel AppVeyor runs that are still executing already stale checkins?)

I was looking at the exact same thing... the AppVeyor is on Douglas' account, and I believe only he can cancel on-going builds

PR #704, opened 2020-04-06
Expose common Keccak symbol

baentsch:
As discussed with @xvzcf. This exposes the `src/common` based Keccak used by Dilithium, Sphincs and Kyber in their AVX2 implementations and hides the Picnic variant.


baentsch:
@christianpaquin This proposal goes beyond what you'd already done. However, the previous edits didn't "hide away" everything. This PR in particular does this _only_ with the symbol `KeccakP1600times4_PermuteAll_24rounds`. If we should do this for all "Picnic-based" Keccak symbols, that'd be good: Either I do it or you(r team) can do: Please just let me know. Without this change, the three PQClean algorithms mentioned above call into Picnic's Keccak -- which isn't ideal.

baentsch:
Superseded by PR #705. Common Keccak will be separately exposed.

Issue #695, opened 2020-04-01
"Optimized" CMAKE_BUILD_TYPE should possibly be renamed to Release


thomwiggers:
The cmake profile `Optimized` to build with `-O2` etc, seems to break CMake convention. The Rust ``cmake`` crate has a [automatic profile selection](https://docs.rs/cmake/0.1.42/cmake/struct.Config.html#method.profile) that expects ``Release`` instead of ``Optimized``.

This seems to match https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html, which also suggests that #652 was abusing ``CMAKE_BUILD_TYPE`` a little bit...

dstebila:
@xvzcf What do you think?

xvzcf:
I broadly agree. There's just a few small things I want to think about a bit more, and I was planning to get to this after the cpu features work.

baentsch:
As per suggestion by @xvzcf that I support:
>My initial thoughts are that we should add another #define/CMake option to differentiate between portable and non-portable optimized builds

--> It would be good to decide what shall be 'Release' then: An optimised version running on all platforms OR an optimised version (possibly crashing on platforms without suitable CPU extensions). In any case it would be good to have CMAKE_BUILD_TYPES specifying either as a variant without dynamic CPU feature runtime checks is markedly faster.

baentsch:
I correct myself: I just repeated the performance tests with and without run-time status tests on HEAD and the performance drop due to those bit-status tests is much smaller than I had seen before. So I rest my case: "Optimised" and "Portable" could well be the same thing: @xvzcf Please do more relevant things before looking into this micro-improvement.

thomwiggers:
This is no longer up-to-date then? https://github.com/open-quantum-safe/liboqs/wiki/Customizing-liboqs#cmake_build_type

baentsch:
Right, the statement is not quite accurate anymore: The catch-all `-mnative` has been replaced by CPU-specific optimisation flags at each target and those targets (aim to) test for run-time presence of those capabilities. Until this has been added to all algorithms, the statement that the resulting binary (built on a platform supporting all optimisation features like AVX2) may not be portable (i.e., crash on platforms without AVX2 in the same example) still is correct, though. Ultimately, that sentence should go, too, i.e., `Optimised` build binaries should run the fastest code possible on the platform it happens to run on.

xvzcf:
I'll update the wiki once the changes are settled.

[...]

PR #713, opened 2020-04-13
Overhauled CPU-extension detection.

xvzcf:
Refactored .CMake/alg_support.cmake, and kem/ and sig/ CMakeLists.txt files.

Refactored copy_from_pqclean/ templating.

Added custom x64 CPU extension detection code and removed cpu_features.

Removed duplicate Kyber source directories.

In another PR, I will add: Detection for MSVC intrinsics, a fix for https://github.com/open-quantum-safe/liboqs/issues/695, more refactoring of kem/ and sig/ CMakeLists.txt files.


baentsch:
Heavens - what a huge effort -- I hope you don't expect me to check all 295 changed files? :-) 

What I like very much is the introduction of the 'default' implementation: This deals with those algorithms who have a "better-than-clean", no-CPU-features-requiring implementation.

I also have to (blushingly in some areas) acknowledge that mastery of `cmake` really helps with this job: Thanks for this big clean-up!

What I'm not sure I find is the best solution is that you apparently have combined runtime and compile time checking into one: In other words, an optimised implementation always has to pay the execution cost of the runtime capabilities check. In my tests, this cost about 10-15% raw performance and was the reason for introducing a "disable-RT-check-completely" flag, i.e., have a library executing at full throttle (and dying quickly if run on another processor with less capabilities) --> Would it be possible to re-introduce this (and avoid the cost of RT checking for those who want "just speed")?

I'll check this out some more tomorrow and give you more of a feedback if I can think of something more constructive.

xvzcf:
> Heavens - what a huge effort -- I hope you don't expect me to check all 295 changed files? :-)

Only the non-autogenerated ones. A lot of the diffs are just deleted files, so hopefully it shouldn't be too bad. I could walk you through the changes if you'd like.

> What I'm not sure I find is the best solution is that you apparently have combined runtime and compile time checking into one: In other words, an optimised implementation always has to pay the execution cost of the runtime capabilities check. 

Apologies, it looks like I missed the purpose of that code. 

> In my tests, this cost about 10-15% raw performance and was the reason for introducing a "disable-RT-check-completely" flag, i.e., have a library executing at full throttle (and dying quickly if run on another processor with less capabilities) --> Would it be possible to re-introduce this (and avoid the cost of RT checking for those who want "just speed")?

Could I do it in another PR? My initial thoughts are that we should add another `#define`/CMake option to differentiate between portable and non-portable optimized builds, but I'd like to think about it some more and have it go hand in hand with the fix for #695.

baentsch:
@xvzcf Yes on both items: 1) see my comment in issue #695 and 2) picking out which files are most important, can we have a quick call today to close this out most efficiently? Just ping me via Skype when you have time.

PR #717, opened 2020-04-15
Frodo runtime detection

dstebila:


dstebila:
Any ideas @xvzcf? It runs on my mac when compiling with gcc; currently I'm not able to compile with clang. Are the AVX flags getting set for clang on macOS?

gadoofou87:
src/kem/frodokem/CMakeLists.txt
```
if(OQS_USE_AES_INSTRUCTIONS AND OQS_USE_AVX2_INSTRUCTIONS)
    target_compile_options(frodokem PRIVATE -maes -mavx2)
endif()
```
Tests are successful only on machines that support AVX2. I suppose you should remove this check for instructions.

xvzcf:
Looking at it now.

baentsch:
>Tests are successful only on machines that support AVX2. I suppose you should remove this check for instructions.

@gadoofou87 : *Eventually* you are right. However, until all algorithms are supporting run-time feature checks according to #624 (of which this PR is an implementation) these checks cannot be removed as otherwise the resultant binary would immediately crash upon execution on the machine where they were built and this would be undesirable.

dstebila:
@xvzcf Should I close this PR?

PR #722, opened 2020-04-15
Add AVX2 and AESNI versions of SPHINCS+

### LINKED ISSUE #721: Add AVX2/AESNI versions of SPHINCS+
Currently, only the portable implementations are available in liboqs. The sped-up versions are much faster.

I might take a stab at this myself as I'll need those implementations anyway.

thomwiggers:
Also remove all of the dead code that was included (probably due to a name format change sometime).

Closes #721 

I'm not entirely sure how `copy_from_pqclean.py` works, so I could probably use some direction there. Specifically, the use of `sources` is unclear to me.


thomwiggers:
Unclear why Appveyor failed.

.CMake/alg_support.cmake
```diff
@@ -161,39 +161,147 @@ cmake_dependent_option(OQS_ENABLE_SIG_rainbow_Vc_cyclic_compressed "" ON "OQS_EN
 
 option(OQS_ENABLE_SIG_SPHINCS "" ON)
 cmake_dependent_option(OQS_ENABLE_SIG_sphincs_haraka_128f_robust "" ON "OQS_ENABLE_SIG_SPHINCS" OFF)
+if(ARCH STREQUAL "x86_64" AND CMAKE_SYSTEM_NAME MATCHES "" AND OQS_USE_AES_INSTRUCTIONS)
```
xvzcf, line 164: Could you update the templating code [here](https://github.com/open-quantum-safe/liboqs/blob/master/scripts/copy_from_pqclean/.CMake/alg_support.cmake/add_enable_by_alg.fragment#L7) and [here](https://github.com/open-quantum-safe/liboqs/blob/master/scripts/copy_from_pqclean/.CMake/alg_support.cmake/add_enable_by_alg.fragment#L21) to not check for `CMAKE_SYSTEM_NAME` if `platform['operating_systems']` is not present, and then re-run `copy_from_pqclean.py`?

xvzcf, line 164: The templating scripts will have to be updated to remove the check for `CMAKE_SYSTEM_NAME` where there are no names to check against. I'll take a look at this, and I might clean up `copy_from_pqclean.py` some more.

thomwiggers: I've pushed an attempt at fixing this.

xvzcf:
Sorry, I accidentally re-ran the tests, which I then cancelled, hence all the red crosses.

dstebila:
Replaced by #736 since I couldn't push my rebase to Thom's repo.

PR #724, opened 2020-04-15
Adding separate AVX2 runtime-checked common/sha3

baentsch:
In pursuit of #624 provide clean and RT-protected AVX2 optimised variant(s) of SHA3.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,4 +1,11 @@
 add_subdirectory(common)
+set(_COMMON_OBJS $<TARGET_OBJECTS:common>)
+if(OQS_USE_AVX2_INSTRUCTIONS)
+add_subdirectory(common_avx2)
+set(_COMMON_OBJS ${_COMMON_OBJS} $<TARGET_OBJECTS:common_avx2>)
+endif()
+set(COMMON_OBJS ${_COMMON_OBJS} PARENT_SCOPE)
+
 
 if(OQS_ENABLE_KEM_BIKE)
     add_subdirectory(kem/bike)
@@ -75,7 +82,7 @@ add_library(oqs kem/kem.c
                 ${KEM_OBJS}
                 sig/sig.c
                 ${SIG_OBJS}
-                $<TARGET_OBJECTS:common>)
+                ${_COMMON_OBJS})
 if(DEFINED SANITIZER_LD_FLAGS)
     target_link_libraries(oqs PUBLIC ${SANITIZER_LD_FLAGS})
 endif()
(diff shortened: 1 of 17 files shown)


dstebila:
A few of the files look like they are duplicated in both src/common and src/common\_avx2, such as sha3\_ossl.c and sha3\_c.c.  Is that true? (If so, why?) Or are there differences? (If so, then we should make a note about those differences so they can be preserved if the different versions of the files get updated.)

baentsch:
Indeed, `sha3_ossl.c` is an oversight and could be deleted (as for all I know it has no AVX2 dependency). `sha3_c.c` in the AVX2 case does carry a few stubs necessary/worth retaining (the redundant stuff I deleted). What is true, though, is that the core "functionality" files are exact copies and get different "functionality" only by #defines putting them into a different symbol namespace. I'm just not sure whether symlinks are possible across platforms (Windows?)

xvzcf:
I'll take a look at this once the fix for https://github.com/open-quantum-safe/liboqs/issues/695 goes in.

baentsch:
> I'll take a look at this once the fix for #695 goes in.

Thanks in advance. If you don't get around to it this week, please let me know -- I'll then update this based on the current defines incl. `OQS_PORTABLE_BUILD` (which I like, btw). I'd probably try to copy the `common/sha3/* files into the `common_avx2` tree to avoid the actual duplication. But if you have a better idea, @xvzcf please post a comment (or just do a PR obviating this one -- whatever is easier for you).

xvzcf:
I'm not sure I understand how the runtime feature detection and dynamic dispatch is happening here.

baentsch:
I once learned from a wise man "If they didn't hear it, you didn't say it". In this case, I need to rework this to make it clear(er). The base idea was to create 'avx2'-tagged SHA3 entry points via #defines and compile those separately from the "normal" (clean/non-avx2) code base. Runtime-switching occurred in `sha3_c.c`. Let me take another stab at this, hopefully reducing the amount of files involved.

xvzcf:
Hmm I see. I was looking into the possibility of having a `sha3_portable.c` file, in which, for example, there is a `OQS_SHA3_cshake128_optimized_portable` that switches between `OQS_SHA3_cshake128_simple4x` and `OQS_SHA3_cshake128_simple` much the same way as the `kem_{{ scheme }}.c` files for the PQClean algorithms. 

There are only 2 public `4x` functions, so the work shouldn't be that much.

baentsch:
Well, it's not only the `4x` functions -- many other parts of what (seems to be) "clean" C code benefits from (& get compiled into) AVX2 instructions, too.

baentsch:
Withdrawing in favour of integrating XKCP

Issue #698, opened 2020-04-02
Reduce check-in testing on AppVeyor


baentsch:
As it might be so 
- that we cannot use CircleCI for Windows testing (see [PR #690](https://github.com/open-quantum-safe/liboqs/pull/690)), and 
- the AppVeyor account is rather slow regarding Windows testing (@vsoftco and @baentsch at least were waiting for hours to see whether some trivial changes were OK), and
- we decided yesterday to curtail support for "problematic" algorithms on Windows
can I suggest we reduce Windows/AppVeyor-based checkin testing considerably, e.g., with a blacklist of algorithms we don't want to run --at least on Windows--at every check-in?

Can I ask for your comments on this proposal, @dstebila @xvzcf @christianpaquin ? Anyone else?

christianpaquin:
I support reducing the number of tested algs at check it, test all of them on a nightly basis.

dstebila:
Makes sense.  

In PQClean, all of the tests are parameterized by a [`filtered_test` command](https://github.com/PQClean/PQClean/blob/master/test/helpers.py) which then pulls various environment variables to decide what to skip.  These environment variables can then be set in the various CI test configurations.

christianpaquin:
I'll remove the slow-running algs from the per-commit appveyor run; I'm waiting for @xvzcf to finish work to properly disable the algorithm.

xvzcf:
Just noting that it seems to be the SPHINCS and Rainbow variants that impose a big slow down on the Appveyor tests.

xvzcf:
Closed by https://github.com/open-quantum-safe/liboqs/pull/726

PR #726, opened 2020-04-16
Removed non-standard CMake build types and other changes

xvzcf:
Addresses https://github.com/open-quantum-safe/liboqs/issues/695, https://github.com/open-quantum-safe/liboqs/issues/698, and other issues.


scripts/copy_from_pqclean/copy_from_pqclean.py
```diff
@@ -67,6 +67,13 @@ def load_instructions():
         family['family'] = family['name']
         for scheme in family['schemes']:
             scheme['metadata'] = yaml.safe_load(file_get_contents(os.path.join(os.environ['PQCLEAN_DIR'], 'crypto_sign', scheme['pqclean_scheme'], 'META.yml')))
+            # This is a temporary hack to work around the fact that
```
baentsch, line 70: @xvzcf When implementing a direct upstream-copy of Dilithium, I only found this now :-( Do I read this right in that it removes OSX support for Dilithium? Why is this so? Wouldn't it make sense to mark this up as a ToDo? Wouldn't it be worth while to correct this in PQClean if you think it's wrong there? In any case, this is one more reason to work on importing straight from upstream pqcrystals as OSX is supported and working there...

thomwiggers: PQClean uses the flags here as specified by `/proc/cpuinfo`. It uses those for the detection of support for each scheme. The flags used by the compiler, which is what OQS expects, unfortunately don't match up. Perhaps this would better be solved more generically by a function like below.

```
# Maps /proc/cpuinfo flags as used in `META.yml` to `-mFLAG` compiler flags.
MAPPING = {
    'bmi1': 'bmi',
}
def cpuinfo_flags_to_compiler_flags(flags):
        return [MAPPING.get(flag, flag) for flag in flags]
```

With regards to the upstream versions of Dilithium and Kyber in PQClean: Dilithium in PQClean already has OSX support, for Kyber we still need to apply the same set of patches. We would definitely welcome help there, as we obviously only have so much time and our contributions are mostly driven by our direct research needs — we need to write papers after all.

thomwiggers: Ah I see [below](https://github.com/open-quantum-safe/liboqs/pull/726/files/eab0e5492d254c7b98be464cf2afc5776b1b0b2f#diff-92fe1cb95aca55f702b1f28d5220c5caR74) disables OSX support. The reason for that is probably that there is something going on which makes life slightly complicated.

There are unaligned operations being done using `_mm256_loadu_si256` or one of its brethren. These of course need vector types — but `__m256i*` is 8-byte aligned in newer versions of GCC and Clang! This leads to compiler warnings / errors when casting array pointers to `__m256i*`. The unaligned variant, `_m256i_u` is only available in fairly new versions of the compilers, though...

baentsch: > With regards to the upstream versions of Dilithium and Kyber in PQClean: Dilithium in PQClean already has OSX support, for Kyber we still need to apply the same set of patches. We would definitely welcome help there, as we obviously only have so much time and our contributions are mostly driven by our direct research needs — we need to write papers after all. 

@thomwiggers Can you share some information what this support would entail? How much effort is a `pqcrystals` to `PQClean` import usually? (Sorry for the possibly trivial question, I've never looked at PQClean internals so far). Do you see any chance (for anyone) to get an automated import (like from `PQClean` to `liboqs`) realised?

thomwiggers: Probably not worth the time, the changes are so infrequent that it's probably lower effort to just port them manually each time.

What would save time is doing the work to bring PQCRYSTALS upstream in line with PQClean's test suite, and then merging those changes back. Once that is done, I imagine you could more directly import the changes. Unfortunately, such back-PRs are often quite messy to review.

thomwiggers: Updating Kyber to support OSX is mainly making sure the symbols work in the assembly files: the C ABI on Linux just translates `void functionname()` to the label `functionname`, on OSX (and Windows) it's prefixed with an underscore: `_functionname`. 

For the most part, you just need labels in the assembly to also exist in the underscore-prefixed versions. If you reference variables/constants defined in C files, it gets a bit more messy and you will need compiler macros (see https://github.com/PQClean/PQClean/blob/master/crypto_sign/dilithium3/avx2/cdecl.inc)

xvzcf: > PQClean uses the flags here as specified by `/proc/cpuinfo`. It uses those for the detection of support for each scheme. The flags used by the compiler, which is what OQS expects, unfortunately don't match up. Perhaps this would better be solved more generically by a function like below.
> 
> ```
> # Maps /proc/cpuinfo flags as used in `META.yml` to `-mFLAG` compiler flags.
> MAPPING = {
>     'bmi1': 'bmi',
> }
> def cpuinfo_flags_to_compiler_flags(flags):
>         return [MAPPING.get(flag, flag) for flag in flags]
> ```

Indeed, my fix was supposed to be temporary. I meant to revisit it once https://github.com/PQClean/PQClean/pull/290 went in, but it slipped my mind. 

> There are unaligned operations being done using _mm256_loadu_si256 or one of its brethren. These of course need vector types — but __m256i* is 8-byte aligned in newer versions of GCC and Clang! This leads to compiler warnings / errors when casting array pointers to __m256i*. The unaligned variant, _m256i_u is only available in fairly new versions of the compilers, though...

Me (and @vsoftco I believe) ran into errors of this nature when trying to build the AVX2 variant on our MacBooks, which is what prompted me to remove macOS support.

PR #744, opened 2020-05-01
default alg setting brought in line with documentation

baentsch:
Without this, cmake-time `-D` settings for default algs have no effect

### DIFF
--- a/.CMake/alg_support.cmake
+++ b/.CMake/alg_support.cmake
@@ -1,10 +1,10 @@
 include(CMakeDependentOption)
 
-if(NOT OQS_KEM_DEFAULT)
-set(OQS_KEM_DEFAULT "OQS_KEM_alg_frodokem_640_aes")
+if(NOT DEFINED OQS_KEM_DEFAULT)
+    set(OQS_KEM_DEFAULT "OQS_KEM_alg_frodokem_640_aes")
 endif()
-if(NOT OQS_SIG_DEFAULT)
-set(OQS_SIG_DEFAULT "OQS_SIG_alg_dilithium_2")
+if(NOT DEFINED OQS_SIG_DEFAULT)
+    set(OQS_SIG_DEFAULT "OQS_SIG_alg_dilithium_2")
 endif()
 
 if(NOT WIN32)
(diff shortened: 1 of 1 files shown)


.CMake/alg_support.cmake
```diff
@@ -1,7 +1,11 @@
 include(CMakeDependentOption)
 
+if(NOT OQS_KEM_DEFAULT)
```
xvzcf, line 3: Could you replace this with `if(NOT DEFINED OQS_KEM_DEFAULT)`?

.CMake/alg_support.cmake
```diff
@@ -1,7 +1,11 @@
...
 set(OQS_KEM_DEFAULT "OQS_KEM_alg_frodokem_640_aes")
+endif()
+if(NOT OQS_SIG_DEFAULT)
```
xvzcf, line 6: Same here, with `NOT DEFINED OQS_SIG_DEFAULT`

.CMake/alg_support.cmake
```diff
@@ -1,7 +1,11 @@
 include(CMakeDependentOption)
 
+if(NOT OQS_KEM_DEFAULT)
 set(OQS_KEM_DEFAULT "OQS_KEM_alg_frodokem_640_aes")
```
xvzcf, line 4: Could you indent this with 4 spaces?

.CMake/alg_support.cmake
```diff
@@ -1,7 +1,11 @@
...
+endif()
+if(NOT OQS_SIG_DEFAULT)
 set(OQS_SIG_DEFAULT "OQS_SIG_alg_dilithium_2")
```
xvzcf, line 7: Same here, with 4 spaces.

.CMake/alg_support.cmake
baentsch: Done

.CMake/alg_support.cmake
baentsch: Done

.CMake/alg_support.cmake
baentsch: Done

.CMake/alg_support.cmake
baentsch: Done

PR #749, opened 2020-05-11
Add SPDX-License-Identifier headers

### LINKED ISSUE #642: Should we add SPDX license info to files?
In the recently merged #640, @drucker-nir included machine-readable license information in each file, specifically an [SPDX-License-Identifier](https://spdx.org/ids).  Should we add these to all our files?

dstebila:
Fixes #642.

Before merging:

- [x] Update [coding conventions wiki page](https://github.com/open-quantum-safe/liboqs/wiki/Coding-conventions)


src/sig/qtesla/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+# SPDX-License-Identifier: MIT
```
christianpaquin, line 1: It looks you are only marking the files not in `external` directories, correct? qtesla doesn't use an external directory, but instead uses a `portable` and `avx2` one. Perhaps these should be moved down into an `external` folder for consistency, and not have the license applied to each external file.

src/common/common.c
```diff
@@ -18,6 +20,7 @@ static unsigned int available_cpu_extensions_set = 0;
 /* The code here, including x86_64_helpers.h, has been taken from:
  * https://github.com/vectorclass/version2
  * https://github.com/google/cpu_features/blob/master/src/cpuinfo_x86.c
+ * SPDX-License-Identifier: Apache-2.0
```
christianpaquin, line 23: This seem to indicate that all the file is Apache, but the readme says this file "includes portions which are Apache License v2.0". Which one is correct?

dstebila:
> What's the rule for files we write to wrap external code (like our sig_* and kem_* files): should they be MIT or inherit their external license?

The copyright license applies to individual source files, and our wrapper files do not include any of the contents of the external files, so they still get our MIT license.

src/sig/qtesla/CMakeLists.txt
dstebila: Yes, I marked every file not in an external or pqclean directory.

src/common/common.c
dstebila: Good point, we should probably put a start/end indicating which portions are licensed that way. @xvzcf, which portions should be marked?

xvzcf: I pushed a commit clarifying the portions.

christianpaquin: Is that compatible with the [SPDX IDs](https://spdx.org/ids) usage? The page says one ID per file. [This page](https://spdx.org/ids-how) suggests we should use conditionals, e.g.:
`// SPDX-License-Identifier: Apache-2.0 AND MIT`

dstebila: Oh, I hadn't seen that aspect of usage.  Yes, let's change the top of the file to be Apache-2.0 AND MIT.

dstebila:
> Should we mark all of the external files with their respective license, if not, why not? 

I think of 'external' and 'pqclean' files as being added to our repo without being touched by human hands, and I don't want to have to modify all of those files and make them inconsistent with upstream changes.  That's certainly the case for pqclean.  If you think external should be handled differently, let me know.

> What's the purpose of this feature: to facilitate exclusion of files you can't integrate with (if so, can/will our build system be able to facilitate that?)

The purpose of doing this is to very clearly document the license in every file.  I added a unit test that checks that every .c, .h, .s, .S, .py, and various CMake files all have SPDX headers, except for files in the pqclean and external directories.  However, it does not check what license is present, just that the header is present (i.e., it is not automatically checking that no incompatibly licensed files are added); that will still be manual.

> I'd like to see the proposed [coding conventions](https://github.com/open-quantum-safe/liboqs/wiki/Coding-conventions) update before finishing this review.

I've added it to https://github.com/open-quantum-safe/liboqs/wiki/Coding-conventions and https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide.

christianpaquin:
> I think of 'external' and 'pqclean' files as being added to our repo without being touched by human hands

Makes sense. I should move the qtesla `portable` and `avx2` external files into a parent  external folder. I can do that in a separate PR.
 
> I've added it to https://github.com/open-quantum-safe/liboqs/wiki/Coding-conventions and https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide.

Sounds good.

PR #754, opened 2020-05-16
Fixing test behavior

gadoofou87:
`kem_kat` and `sig_kat` never end with an error because the `ret` variable is overwritten with `OQS_SUCCESS`


dstebila:
Thanks!

PR #755, opened 2020-05-20
Clean up FrodoKEM algorithm datasheet

dstebila:


xvzcf:
How about merging the summary and implementation sections into a single not-too-big summary section on top? Here's a stab at merging them (justification comes after)

# FrodoKEM (https://frodokem.org/) 

## Summary

- **Algorithm type**: key encapsulation mechanism
- **Main cryptographic assumption**: learning with errors (LWE)
- **Scheme authors**: Michael Naehrig, Erdem Alkim, Joppe Bos, Léo Ducas, Karen Easterbrook, Brian LaMacchia, Patrick Longa, Ilya Mironov, Valeria Nikolaenko, Christopher Peikert, Ananth Raghunathan, Douglas Stebila
- **Version**: NIST Round 2 submission
- **Implementation:** https://github.com/Microsoft/PQCrypto-LWEKE/commit/d5bbd0417ba111b08a959c0042a1dcc65fb14a89
- **Implementation Constant-time?**: Yes
- **Implementation Notes:** MIT License, Portable C with optional use (selected at compile-time) of AVX2 and AESNI instructions

... followed by the parameter set table.

# Justification

I don't see the need to repeat `kem_frodokem` in the header since the filename is already `kem_frodokem.md`. We can instead use the heading to convey the **Name** and **Authors' website** fields. If we also split the `docs/algorithms` directory into something like `docs/kem_datasheets`/`docs/kex_datasheets` and `docs/sig_datasheets`, we can remove the `algorithm datasheet` text from the heading and the `kem` and `sig` prefixes from the files.

I merged **Source of implementation** and **Implementation version** into one **Implementation** field. This field points to the location of the commit, which can also be the source of the implementation. I've moved everything besides the **Constant-time** field into a single **Implementation Notes** field, as I thought (comparatively speaking) they weren't as important as the preceding two fields.

PR #758, opened 2020-06-01
fixes issue 757

baentsch:
@xvzcf Could you please check whether you can see how (C)CI runs on this change? As discussed, I'm now trying to do a PR from 'within' OQS and not my own fork -- and all kind of weird things happen (not seeing CCI results being one thing). Locally things look OK. Thanks in advance.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -4,7 +4,7 @@ differently; the corresponding subfolder contains the license that applies in
 that case.
 
 
-Copyright (c) 2016-2019 Open Quantum Safe project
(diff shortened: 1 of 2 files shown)


xvzcf:
Strange, it doesn't look like the CircleCI tests have been run at all.

baentsch:
Thanks @dstebila for triggering the re-build. The failing runs are now related to missing Windows credits in 'mb-'tagged commits. Won't do that going forward when doing further "oqs-originating" PRs.

PR #762, opened 2020-06-04
Allowing cmake to export project targets

onavratil-monetplus:
Hello there,
I made a small tweak that allows inclusion of liboqs into other cmake projects using `add_subdirectory` and other funny tools.

So far, liboqs doesnt export any of its targets, and thus makes it more difficult to link against in other projects. This simple fix allows something like this to happen in a custom project:
```cmake
FetchContent_Declare(liboqs
    GIT_REPOSITORY https://github.com/onavratil-monetplus/liboqs
)
message(STATUS  "Fetching liboqs...this may take a while")
FetchContent_MakeAvailable(liboqs)
# ....

target_link_libraries(myCustomTarget PUBLIC OQS::oqs)
```
And boom! There we go, no need to install the lib system-wide prior to project build, allows to use static version of the lib etc etc etc.

Now I am nothing for a cmake professional, but this solution works nicely for me, and hopefully can help someone else as well.

On a footnote, the fix is based on a discussion regarding my desparate attempts to use cmake.
https://stackoverflow.com/questions/61504626/the-cmake-quandary

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "C_Cpp.default.configurationProvider": "vector-of-bool.cmake-tools"
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "C_Cpp.default.configurationProvider": "vector-of-bool.cmake-tools"
+}
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -82,6 +82,8 @@ if(OQS_USE_OPENSSL)
     target_include_directories(oqs PUBLIC ${OPENSSL_INCLUDE_DIR})
 endif()
 
+
+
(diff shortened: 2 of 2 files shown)


GlitchedPolygons:
👍

grassnick:
This does not work when specifying `CMAKE_INSTALL_PREFIX`.

I have to add `include(GNUInstallDirs)`  before installing to `CMAKE_INSTALL_FULL_LIBDIR` for it to work correctly, see  https://cmake.org/cmake/help/v3.17/module/GNUInstallDirs.html and https://cmake.org/cmake/help/v3.17/command/install.html#installing-files.

OS: Arch Linux, CMake Version 3.17.3

dstebila:
@xvzcf Can this be merged?

xvzcf:
Closing in favour of https://github.com/open-quantum-safe/liboqs/pull/840

Issue #737, opened 2020-04-27
SIKE CMakeLists.txt should selectively build files based on ENABLE_KEM_SIDH and ENABLE_KEM_SIKE


xvzcf:
Tracking the discussion [here](https://github.com/open-quantum-safe/liboqs/pull/734).

christianpaquin:
Addressed by PR #769.

PR #769, opened 2020-06-19
Integrated SIKE v3.3

christianpaquin:
New in Version 3.3
- Improved versions of the four parameter sets for compressed SIDH and compressed SIKE
- Optimized implementations of the field arithmetic for 64-bit ARMv8 processors for Linux.
- General optimizations to the field arithmetic.
- Support for Mac OS X for the optimized x64 assembly implementations.
- Support for big endian platforms, specifically IBM s390x processors.

TODO before merge:
- [x] Address all `FIXMEOQS`
- [X] Enabled optimizations on Darwin
- [x] Address issue #737 
- [x] Integrate [latest commits](https://github.com/microsoft/PQCrypto-SIDH/commit/a971e7758c164a4770af8003ca2d16a493b862e7) from SIKE above v3.3


christianpaquin:
@dstebila, @xvzcf; some Linux tests (e.g., [centos 7](https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/808/workflows/045d4643-ae5c-4685-8620-9871c09c34f2/jobs/8268/steps)) are running out of time and therefore failing. What should I do? Can we increase test time in CircleCI?

baentsch:
> It looks like there's an option one can set in the CircleCI config file: no_output_timeout.

Do we understand why there are such long delays without output? Wouldn't it be more sensible to change the tests to be more 'responsive'?

christianpaquin:
> > It looks like there's an option one can set in the CircleCI config file: no_output_timeout.
> 
> Do we understand why there are such long delays without output? Wouldn't it be more sensible to change the tests to be more 'responsive'?

I thought things got a bit slower and needed a longer timeout, but it looks like there is an issue with one of the sike params. I'm investigating. I'll revert the timeout to the required values before merging.

PR #770, opened 2020-06-23
Re-add AES-NI with CPU feature detection

dstebila:


dstebila:
@xvzcf, I am going to re-add our AES-NI implementation so that we have a fast and side-channel-resistant AES implementation when OpenSSL is not used.  Before I get too far, can you take a look at the proposed changes in [src/common/CMakeLists.txt](https://github.com/open-quantum-safe/liboqs/pull/770/files#diff-89526a27954877955acb61f19cd7aec6R16) and [lines 12-25 of src/common/aes/aes.c](https://github.com/open-quantum-safe/liboqs/pull/770/files#diff-801ad41ce7ff56df6f30e1829b0d188bR12) to let me know if I'm doing CPU feature detection correctly in all the cases we need to handle.

src/common/CMakeLists.txt
```diff
@@ -13,7 +13,11 @@ endif()
 if(OQS_USE_AES_OPENSSL)
     set(AES_IMPL aes/aes_ossl.c)
 else()
-    set(AES_IMPL aes/aes_c.c)
+   set(AES_IMPL aes/aes.c aes/aes_c.c)
+    if (OQS_USE_CPU_EXTENSIONS)
```
xvzcf, line 17: This should be changed to `if(OQS_USE_AES_INSTRUCTIONS)`

src/common/CMakeLists.txt
```diff
@@ -13,7 +13,11 @@ endif()
...
+      set(AES_IMPL ${AES_IMPL} aes/aes_ni.c)
+      add_compile_options(-maes)
```
xvzcf, line 19: Can you change this to `set_source_files_properties(aes/aes_ni.c PROPERTIES COMPILE_FLAGS -maes)`?

src/common/aes/aes.c
```diff
@@ -0,0 +1,137 @@
...
+#include <stdio.h>
+#include <string.h>
+
+#include <oqs/common.h>
+
+#include "aes.h"
+#include "aes_local.h"
+
+void OQS_AES128_ECB_load_schedule(const uint8_t *key, void **_schedule, UNUSED int for_encryption) {
+	#if defined(OQS_USE_AES_INSTRUCTIONS)
```
xvzcf, line 13: The logic here should be:

```c
#if defined(OQS_USE_CPU_EXTENSIONS)

#if defined(OQS_PORTABLE_BUILD)
OQS_CPU_EXTENSIONS available_cpu_extensions = OQS_get_available_CPU_extensions();
if (available_cpu_extensions.AES_ENABLED) {
#endif /* OQS_PORTABLE_BUILD */
    oqs_aes128_load_schedule_ni(key, _schedule);
#if defined(OQS_PORTABLE_BUILD)
} else {
    oqs_aes128_load_schedule_c(key, _schedule, for_encryption);
#endif /* OQS_PORTABLE_BUILD */
#else
    oqs_aes128_load_schedule_c(key, _schedule, for_encryption);

#endif /*OQS_USE_CPU_EXTENSIONS */
```

Also, the `OQS_USE_*_INSTRUCTIONS` variables and #define's are indeed confusing. Perhaps it would be better to rename them to `OQS_COMPILE_WITH_*_INSTRUCTIONS`.

dstebila: Thanks.  I've pushed two commits, one that directly applies the logic you gave above, and then one that puts it into a macro: since I know I'm going to need to use the same logic many times in that file, I thought a macro might be cleaner. Can you take a look and see if the macros matches the logic?

xvzcf: Agreed, it is cleaner. And it does match the logic.

dstebila:
@xvzcf @christianpaquin Can either of you help me understand the failure on Windows? It says that `oqs_aes128_load_schedule_ni` is undefined at link time.

As far as I can tell all calls to `oqs_aes128_load_schedule_ni` are wrapped inside an `#if defined(OQS_USE_CPU_EXTENSIONS)`.  So for this error to occur, the compiler on Windows must think that OQS\_USE\_CPU\_EXTENSIONS is true.

The file that provides `oqs_aes128_load_schedule_ni` is in a CMakeLists.txt surrounded by `if (OQS_USE_AES_INSTRUCTIONS)`.  

So I'm left to conclude that `OQS_USE_CPU_EXTENSIONS` and `OQS_USE_AES_INSTRUCTIONS` might be set inconsistently on Windows?

Remind me: are we intending for Windows to use optimized instructions?

xvzcf:
Both `OQS_USE_CPU_EXTENSIONS` and `OQS_PORTABLE_BUILD` are `ON` by default. `OQS_USE_CPU_EXTENSIONS` should only have effect, however, when the compiler is GCC or Clang. 

That means MSVC definitely can't use optimized instructions. What's probably happening is that `OQS_USE_CPU_EXTENSIONS`, being `ON`, causes the corresponding macro to be defined in the C code, which then leads to the AES-NI functions being mentioned in `aes/aes.c`, which ultimately causes the MSVC build to fail with link errors (because `aes128_ni.c` and `aes128_ni.c` are only included when `OQS_USE_AES_INSTRUCTIONS` is defined, which is only possible if `OQS_USE_CPU_EXTENSIONS` is defined and the compiler is GCC/Clang and the `__AES__` macro is defined)

From a cursory glance, the msys2 build seems to fail because AES instructions are supported but `#pragma unroll` isn't.

xvzcf:
I'll think about fixes, one is that perhaps `OQS_USE_CPU_EXTENSIONS` should be set more carefully than [this](https://github.com/open-quantum-safe/liboqs/blob/master/CMakeLists.txt#L39).

PR #772, opened 2020-06-26
Update cross-compiling instructions for Windows in README.md

kevinmkane:
OQS_USE_CPU_EXTENSIONS has to be OFF in the cross-compiling scenario, or CMake will fail.


kevinmkane:
I've opened #775, since that way I can also mention the OQS-OpenSSL cross-compilation wiki should be updated as well, if and when CPU_EXTENSIONS supports cross-compilation.

baentsch:
@kevinmkane Thanks for the doc update+issue.

Issue #776, opened 2020-06-28
Implicit conversion changes signedness error while building in clang for 


greebie:
To replicate: 

(Using clang / Mac Catalina) 

run scripts/build-android.sh /Users/ryandeschamps/Library/Android/sdk/ndk/21.3.6528147/ -a arm64-v8a  -b build-arm64-v8a

Results in:

![image](https://user-images.githubusercontent.com/111091/85956873-c7490980-b956-11ea-944d-eff32ad6dd49.png)

Changing the return type of `USING_TIME_RATHER_THAN_CYCLES` to uint64_t resolves the build. 

PR Upcoming!  

Also possible that `_bench_init_perfcounters` values should be unsigned for consistency.

PR #777, opened 2020-06-28
Change return value of `USING_TIME_RATHER_THAN_CYCLES` to unsigned int.

greebie:
Resolves issue #776 when attempting to build for arm64-v8a in clang.

Would also like advice on whether `_bench_init_perfcounters` parameters and local variables should be unsigned, as it would make sense to fix them here.

### DIFF
--- a/tests/ds_benchmark.h
+++ b/tests/ds_benchmark.h
@@ -139,7 +139,7 @@ static uint64_t _bench_rdtsc(void) {
 #define USING_TIME_RATHER_THAN_CYCLES
 	struct timespec time;
 	clock_gettime(CLOCK_REALTIME, &time);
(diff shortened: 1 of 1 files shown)


xvzcf:
@dstebila should the `int32_t` types [here](https://github.com/open-quantum-safe/liboqs/blob/master/tests/ds_benchmark.h#L147) be changed to `uint32_t`?

dstebila:
> @dstebila should the `int32_t` types [here](https://github.com/open-quantum-safe/liboqs/blob/master/tests/ds_benchmark.h#L147) be changed to `uint32_t`?

I don't think those matter, since they're just being used as flags and bitmasks.

Issue #779, opened 2020-07-01
OSX build warnings


baentsch:
When I just built on OSX for the first time in a long time, these warnings appeared - and they're also in CCI...
```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: lib/liboqs.a(gf2x_portable.c.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: lib/liboqs.a(gf2x_mul.c.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: lib/liboqs.a(gf2x_portable.c.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: lib/liboqs.a(gf2x_mul.c.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: lib/liboqs.a(gf2x_portable.c.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: lib/liboqs.a(gf2x_mul.c.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: lib/liboqs.a(gf2x_portable.c.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: lib/liboqs.a(gf2x_mul.c.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: lib/liboqs.a(cpu.c.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: lib/liboqs.a(gf2x_portable.c.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: lib/liboqs.a(gf2x_mul.c.o) has no symbols
[...]

PR #780, opened 2020-07-02
fixing #779

baentsch:
As discussed in #779 , 
@xvzcf please advise whether you know a way to make compile dependent on `OPENSSL_NO_EC2M` being defined
@christianpaquin please check whether you agree with the Darwin-specific non-compilation of `cpu.c`

CI results accessible [here](https://app.circleci.com/pipelines/github/baentsch/liboqs/290/workflows/69a1517d-c284-4487-b3e8-691a49a8cb64)


christianpaquin:
So, what was the specific issue with picnic? On Darwin, the `cpu.c` filed would be empty creating a warning? Wouldn't that have caused a build failure? This fix looks ok to me in the picnic CMakeFile assuming the above is correct.

I have PR #771 ready, updating picnic, so that should be replicated there as well.

baentsch:
> On Darwin, the cpu.c filed would be empty creating a warning? 

Yes

> Wouldn't that have caused a build failure?  

No: It was/is just a warning.

christianpaquin:
> > On Darwin, the cpu.c filed would be empty creating a warning?
> 
> Yes
> 
> > Wouldn't that have caused a build failure?
> 
> No: It was/is just a warning.

Ok, picnic changes are fine by me, but as I said, perhaps after PR #771 gets merged so it can be applied to the new code.

xvzcf:
Could you try using https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html to detect `OPENSSL_NO_EC2M`?

baentsch:
> Could you try using https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html to detect `OPENSSL_NO_EC2M`?

@xvzcf Thanks very much for the pointer!

baentsch:
Replacing with #782 to ease review.

PR #784, opened 2020-07-07
KAT testing using hashes

baentsch:
As discussed on the developer mailing list, this is to remove the need to keep 37MB of KATs in the repo.

The most relevant changes are in `tests/test_kat.py`. The rest is just deletion of full KAT files and addition of .sha256 files (via `scripts/genkatsha256.sh`).

Tests [ran OK here](https://app.circleci.com/pipelines/github/baentsch/liboqs/302/workflows/9e4c1d54-1b89-4374-8391-1731ca00cc19)


xvzcf:
As a further suggestion, could we store the KAT hashes in a python dictionary indexable by algorithm name (with appropriate allowances for KATs varying with KEM/signature implementation) ?

tests/test_kat.py
```diff
@@ -15,12 +16,13 @@ def test_kem(kem_name):
...
+    h256 = sha256()
+    h256.update(output.encode())
     kats = []
     for filename in os.listdir(os.path.join('tests', 'KATs', 'kem')):
-        if filename.startswith(kem_name + '.') and filename.endswith('.kat'):
+        if filename.startswith(kem_name + '.') and filename.endswith('.sha256'):
             with open(os.path.join('tests', 'KATs', 'kem', filename), 'r') as myfile:
-                kats.append(myfile.read())
-    assert(output in kats)
+                assert(myfile.read() == h256.hexdigest())
```
dstebila, line 25: This logic could be problematic. In particular, if the KAT file is not present, the test doesn't fail, because the `assert` is guarded by an `if`.  Contrast that with what was present before, where the `assert` always happens.

Why not just immediately try to open the file with the required filename (if the file is not present and the open fails, that's an error anyways), rather than looping through all files to find the one you want?

tests/test_kat.py
```diff
@@ -30,24 +32,26 @@ def test_sig(sig_name):
...
-                    append_kat = True
+                    if not(avx2_aes_enabled_on_linux and not helpers.is_build_portable()):
+                        append_kat = True
             else:
                 append_kat = True
             if append_kat:
                 with open(os.path.join('tests', 'KATs', 'sig', filename), 'r') as myfile:
-                    kats.append(myfile.read())
-    assert(output in kats)
+                    assert(myfile.read() == h256.hexdigest())
```
dstebila, line 54: Similar concerns to the comment above. I also can't make sense of what the `append_kat` stuff is actually doing.

baentsch:
> As a further suggestion, could we store the KAT hashes in a python dictionary indexable by algorithm name (with appropriate allowances for KATs varying with KEM/signature implementation) ?

Good idea. So implemented.

tests/test_kat.py
baentsch: Agree - logic simplified and corrected in new, dict-based version. Please check again.

tests/test_kat.py
baentsch: `append_kat` was a remnant of the previous logic creating a structure of all KATs (which the one under test needed to match one of). Now streamlined to _one_ perfect match required.

PR #786, opened 2020-07-08
Add pull request template

dstebila:


dstebila:
> Good guidance. Only nitpick: Would it be worthwhile adding a statement where the version information is to be updated (file(s) and package(s) - also downstream?)? Should this be done _within_ the PR or separately?

Let's do it separately, since another algorithm may have already caused a bump in an earlier PR after which there has not yet been a release.  It's more a checkbox to flag my attention so that I know it has to be done eventually.  [This wiki page](https://github.com/open-quantum-safe/liboqs/wiki/Release-process-(Under-construction)) documents where the version number has to be changed.

PR #787, opened 2020-07-08
Added invalid ciphertext test in test_kem

christianpaquin:
Test now generates a random ciphertext and expect an invalid decapsulation or non-matching shared secret. Also replaced invalid signature with random array vs one with only the first byte modified (to potentially catch more errors).

- [ ] Fix SIKE compressed decapsulation error on bad data before merging this (issue #788)


christianpaquin:
These changes have been incorporated in PR #790.

PR #789, opened 2020-07-09
Enable automated upstream imports

baentsch:
This is adding capabilities to `copy_from_pqclean` to alternatively automatically import upstream code bases.

This PR is an "empty" run (without actual upstream additions) to ease review. When approved, it will be used to import new upstream versions of `kyber` and `dilithium` (setting the corresponding environment variables listed in the `copy_from_pqclean.yml` file to the folders of the upstream repositories along the same lines as is done for a `PQClean` import). 

But by no means this facility is restricted to those algorithms. The basic idea is to have `META.yml` files modelled on the ones used by `PQClean` within the upstream repositories: These contain 
- names of files to import
- possible compile directives
- possibly changed & properly namespace'd algorithm entry points  
- possibly changed KATs. Any such KAT changes will be called out when running the new script variant to ensure library version bumps can be initiated.

A more complete documentation of this will be added to [the Wiki](https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide) when merged. Conceptually, this could be used to automatically introduce any new/changed algorithms provided the upstream code maintainer provides NIST-like APIs and these ".yml-import-drivers".

CCI [ran successfully here](https://app.circleci.com/pipelines/github/baentsch/liboqs/307/workflows/742d5394-1bd1-4acd-b89a-78e365753c40)

All files changed outside of `copy_from_pqclean` have the facility added to provide their own compile directives and to allow for their own versions of otherwise common includes (generated by new `copy_from_pqclean`).

A minor correction is also done to `scripts/copy_from_pqclean/src/sig/sig.h/algs_length.fragment` to cater for the new number of algorithms not imported from PQClean.


scripts/copy_from_pqclean/src/kem/family/kem_scheme.c
```diff
@@ -31,15 +31,52 @@ OQS_KEM *OQS_KEM_{{ family }}_{{ scheme['scheme'] }}_new() {
 	return kem;
 }
 
-extern int PQCLEAN_{{ scheme['pqclean_scheme_c']|upper }}_{{ scheme['default_implementation']|upper }}_crypto_kem_keypair(unsigned char *pk, unsigned char *sk);
-extern int PQCLEAN_{{ scheme['pqclean_scheme_c']|upper }}_{{ scheme['default_implementation']|upper }}_crypto_kem_enc(unsigned char *ct, unsigned char *ss, const unsigned char *pk);
-extern int PQCLEAN_{{ scheme['pqclean_scheme_c']|upper }}_{{ scheme['default_implementation']|upper }}_crypto_kem_dec(unsigned char *ss, const unsigned char *ct, const unsigned char *sk);
+    {%- for impl in scheme['metadata']['implementations'] if impl['name'] == scheme['default_implementation'] %}
+
+        {%- if impl['signature_keypair'] %}
+           {%- set cleankeypair = scheme['metadata'].update({'default_keypair_signature': impl['signature_keypair']}) -%}
```
dstebila, line 37: I'm a bit confused about what's going on here. Are the set variables cleankeypair, cleanenc, cleandec ever used? If so I missed it. And you're updating the scheme['metadata'] data structure here in the template? Shouldn't that logic be back in the Python file that created that data structure in the first place?

scripts/copy_from_pqclean/src/kem/family/CMakeLists.txt
```diff
@@ -8,17 +8,23 @@
...
+           {%- endif -%}
 
         {%- else %}
 
 if(OQS_ENABLE_KEM_{{ family }}_{{ scheme['scheme_c'] }}_{{ impl['name'] }})
     add_library({{ family }}_{{ scheme['scheme'] }}_{{ impl['name'] }} OBJECT {% for source_file in impl['sources']|sort -%}pqclean_{{ scheme['pqclean_scheme'] }}_{{ impl['name'] }}/{{ source_file }}{%- if not loop.last %} {% endif -%}{%- endfor -%})
         {%- endif %}
-    target_include_directories({{ family }}_{{ scheme['scheme'] }}_{{ impl['name'] }} PRIVATE ${PROJECT_SOURCE_DIR}/src/common/pqclean_shims)
     target_include_directories({{ family }}_{{ scheme['scheme'] }}_{{ impl['name'] }} PRIVATE ${CMAKE_CURRENT_LIST_DIR}/pqclean_{{ scheme['pqclean_scheme'] }}_{{ impl['name'] }})
+    target_include_directories({{ family }}_{{ scheme['scheme'] }}_{{ impl['name'] }} PRIVATE ${PROJECT_SOURCE_DIR}/src/common/pqclean_shims)
```
dstebila, line 21: I assume this change is because you want to let other non-PQClean implementations use the PQClean symmetric crypto API?

scripts/copy_from_pqclean/src/kem/family/kem_scheme.c
baentsch: The variables indeed are never used: The purpose of the `set` statement is only to trigger the side-effect of setting the scheme-array-based 'variables' ("default_keypair_signature" in this example) as those 'variables' are needed in later template fields (beyond the reach of jinja's normal, [limited scoping rules](https://jinja.palletsprojects.com/en/2.11.x/templates/)). I also left this code intentionally in this place to ease readability as to what the logic flow is (based on presence/absence of specific `impl` fields). Looks like I failed :-(

scripts/copy_from_pqclean/src/kem/family/CMakeLists.txt
baentsch: Yes. But that's not new: I just changed the include order such as to allow non-PQClean code to override 'common' (PQClean) files with its own implementation(s).

baentsch:
> Seems reasonable to me. I wasn't able to fully reconstruct all the logic in copy_from_pqclean.py, but I trust that it's doing what you want it to do.

It is - albeit I'm sure things can be done more elegantly. So, as the upstream code is not yet ready, we can let this PR linger a bit (also to give @xvzcf a chance to take a look).

baentsch:
@xvzcf I'd like to merge this in tomorrow -- so if you'd be able to spend a few minutes looking this over today that would be most beneficial. Thanks in advance!

xvzcf:
Will do.

scripts/copy_from_pqclean/copy_from_pqclean.yml
```diff
@@ -70,26 +70,32 @@ kems:
       -
         scheme: "512"
         pqclean_scheme: kyber512
+        upstream_location: KYBER
```
xvzcf, line 73: The name `pqclean_scheme` seems confusing if we are able to pull implementations from `pq-crystals` and `PQClean`. Would it be possible to change `pqclean_scheme` to something like `scheme_name` or  `scheme_identifier`?

baentsch: Well, I'd argue it still _is_ the `pqclean_scheme` name (as well as the upstream scheme name). If the environment variable KYBER is not set, is _is_ copied from pqclean. If I were thinking this to the end, one could argue that the whole `copy_from_pqclean` (script _and_ variable naming) would have to be re-thought as it's now becoming more of a general 'copy-from-upstreams' facility. But that would warrant _lots of_ further changes and make the PR pretty large and unwieldy and pretty hard to review. I'd thus like to keep it as-is for now. If you feel this change _is_ warranted (and particularly we could utilize it more generally to motivate other upstreams to ready their code for automated integration to `liboqs` -- and maybe the same mechanism could be used by `PQClean` to do _its_ imports more automatically), I'd suggest to open an issue to "Rename copy_from_pqclean script and variables". OK for you?

xvzcf: Yeah, that sounds good.

Issue #788, opened 2020-07-08
SIKE compressed decapsulation sometimes fails on garbage ciphertext


christianpaquin:
Garbage data sometimes lead to a division by 0 in SIKE compressed, [here](https://github.com/open-quantum-safe/liboqs/blob/master/src/kem/sike/external/fpx.c#L1391).

Need to make this code more robust, and identify other potential and similar issues.

PR #790, opened 2020-07-10
Integrated latest SIKE bug fixes

christianpaquin:
Integrated SIKE upstream commits from 521aab to 50f352, including a fix for a crash when fed garbage data. Added extra test from PR #787. Fixes issue #788.

### DIFF
--- a/docs/algorithms/kem/sike.md
+++ b/docs/algorithms/kem/sike.md
@@ -12,7 +12,7 @@ Implementation
 --------------
 
 - **Source of implementation**: https://github.com/Microsoft/PQCrypto-SIDH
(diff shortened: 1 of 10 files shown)


christianpaquin:
> Do we know if it passes #787? If so, we could just put them both in one PR. If not, then further fixes may be required.

It does locally. It felt like differently scoped PRs, but I see your point. Let me add the extra test code to this PR...

christianpaquin:
> > Do we know if it passes #787? If so, we could just put them both in one PR. If not, then further fixes may be required.
> 
> It does locally. It felt like differently scoped PRs, but I see your point. Let me add the extra test code to this PR...

Done. @dstebila, you can have another look.

Issue #793, opened 2020-07-22
Investigate errors apparently triggered by -DOQS_PORTABLE_BUILD=OFF


baentsch:
Similar issue to #732 : Using gcc8 in Centos-8, Dilithium self-tests and KAT tests fail when running AVX2 code. On no other platform and neither using clang on Centos-8, this problem surfaces.

dstebila:
Dilithium self tests also fail for me on macOS with `env CC=gcc-8 cmake -GNinja -DOQS_PORTABLE_BUILD=OFF ..`.

dstebila:
@baentsch:

Alright, I've modified the Dilithium 2 code to print out a bunch of intermediate values, and compared the output of `tests/kat_sig DILITHIUM_2` when using `CC=gcc-8` versus `CC=clang`.  The first difference seems to occur inside `src/sig/dilithium/pqclean_dilithium2_avx2/rounding.c` function `power2round_avx`, specifically the variable `a0` ends up different when the code is executed using clang versus gcc-8.  

Now here's where it gets trippy.  After I saw the difference by printing out `a0` at [commit `96aba2cf` line 51](https://github.com/open-quantum-safe/liboqs/commit/96aba2cfc45f48bdfd2fbd25b2e451e7636afd59#diff-4bf14b41c9f86fd0fee194aee2240b0cR51), I added some print statements inside `power2round_avx` to print out the values of intermediate values in commit https://github.com/open-quantum-safe/liboqs/commit/603bffa53d6e2780a06ecb86997629de79fcf8b7.  But when I added the print statements inside the for loop for variables `f`, `f1`, and `f0`, suddenly Dilithium2 with gcc-8 output the same values as with clang and passed the  KAT!!!   I don't know enough about intrinsics to know what those operations are doing, but in the past for me, differences in behaviour when print statements are added/removed speaks to memory corruption.  I'll stop here and leave it to you / the Dilithium team to investigate further.

In https://github.com/open-quantum-safe/liboqs/commit/13580f0f147b1ffd6a461db1eda9bc559128fdb8, I copied over the C version of power2round from the "clean" version of Dilithium, and it also causes the KATs to pass (at least for Dilithium 2, I didn't try for 3 and 4).  So using that could be a temporary fix if we wanted to get KATs passing and clear the build failures while the Dilithium team investigates the AVX2 bug.

baentsch:
@dstebila Thanks _very much_ for this deeper investigation. 

[...]

PR #794, opened 2020-07-22
upstream dilithium

baentsch:
This updates Dilithium with current upstream code from pqcrystals: Performance enhancements and added OSX support for AVX2 code. It also adds randomised signing behaviour and with this the KATs.

CI tests ran successfully [here](https://app.circleci.com/pipelines/github/baentsch/liboqs/337/workflows/bbcd59e5-61af-42d3-a284-82d6f7beaaf7)


baentsch:
@dstebila Additional information for your review: The only "manual" changes 1) are in `test_namespace.py` where I added `pqcrystals` as a new namespace - which should be OK, as this code is equally totally namespace-separated (as is `PQClean` code); and 2) pertains to #793, setting clang for Centos-8 CI. All else just is a run of `copy_from_pqclean` with DILITHIUM set to the upstream as per [the updated Wiki guidance on automation](https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide)

baentsch:
> For the subdirectories where the Dilithium code is kept, would you want to rename them from `pqclean_dilithium2_clean` to `pqcrystals_dilithium2_clean` etc.?

I also thought about it, but No: I see direct upstream import only as an 'extension' option: In addition, with the announcement of the round 3 candidates now, I'd think there'll be a renewed interest to get the best and latest into the PQClean code base, so that eventually the folder name pqclean will not just represent the script adding the code to the repo but also the actual upstream repo.

PR #796, opened 2020-07-23
Fix execute failures on Alpine

### LINKED ISSUE #785: Understand & fix alpine execute failures
CCI contains support for testing alpine-based images. Those fail for some algorithms with core dumps (list below). Initial investigations show identical (algorithm) failures regardless of alpine or gcc version used (8.2+9.3) or CPU features being used or not. If build type is switched to `Debug`, more algorithms fail in the same way.

```
FAILED tests/test_cmdline.py::test_kem[BIKE1-L3-FO] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_kem[FrodoKEM-1344-AES] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_kem[SIKE-p751-compressed] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_kem[FrodoKEM-1344-SHAKE] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_kem[FrodoKEM-976-AES] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_kem[SIDH-p610-compressed] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_sig[MQDSS-31-64] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_sig[Rainbow-Ia-Cyclic-Compressed] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_sig[qTesla-p-III] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_sig[Rainbow-Ia-Cyclic] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_sig[MQDSS-31-48] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_kem[SIDH-p751-compressed] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_kem[SIKE-p610-compressed] - AssertionError: Got unexpected return code -11
FAILED tests/test_cmdline.py::test_sig[Rainbow-Ia-Classic] - AssertionError: Got unexpected return code -11
```

dstebila:
Algorithm failures in Alpine were due to the default small stack size of musl libc [[1]](https://www.rookout.com/blog/my-big-fat-alpine-crash), [[2]](https://wiki.musl-libc.org/functional-differences-from-glibc.html#:~:text=128k).  I've increased it using [`-Wl,-z,stack-size=N`](https://wiki.musl-libc.org/functional-differences-from-glibc.html#:~:text=stack-size=N) with N=16MB, but my solution isn't the most general because I'm just increasing it when compiling with GCC regardless of which platform or libc implementation is used; ideally we'd detect that musl libc is being used and only increase it in that case, but I don't know how to do that.

This also fixes two unrelated failures in other CI tests, one related to namespacing (liboqs.so on Alpine has two extra symbols, init and fini) and one related to a return code seemingly being different for xargs/grep.

Fixes #785.


dstebila:
Apparently `add_link_options` was added to cmake in version [3.13](https://cmake.org/cmake/help/v3.13/command/add_link_options.html), but we're using cmake version 3.10 in our Ubuntu images.

dstebila:
> Apparently `add_link_options` was added to cmake in version [3.13](https://cmake.org/cmake/help/v3.13/command/add_link_options.html), but we're using cmake version 3.10 in our Ubuntu images.

Options are manually installing a newer version of cmake or switching to a later version of Ubuntu, e.g. Ubuntu 20.04LTS.

dstebila:
Okay, this should now only do the stack size increase when it's running on Alpine musl, and should be robust against CMake versions.  The Alpine musl detection is not done the proper way -- the proper way would probably be to actually link against libc and then detect what it was, but googling seems to indicate that that's hard, so instead I'm just checking for the presence of the string "alpine-linux-musl" in a certain CMAKE variable.

Troublingly, we get Dilithium self-test and KAT failures on Alpine now (https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/900/workflows/727841fa-1060-4836-ba5e-ee2cf13664af/jobs/9084/tests), which seems related to #793.

At this point, @baentsch you can review and see if this seems to satisfy the main Alpine execute failures, and then we can focus on the Dilithium problems in #793.

PR #800, opened 2020-07-25
AES-NI bug fixes

### LINKED ISSUE #799: Investigate AES-NI failure on Alpine
https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/909/workflows/84067560-d98e-40ad-80f5-28e38c59acf6/jobs/9165 shows that the AES-NI code we just added in #770 fails on Alpine.  I've temporarily worked around in 11246cd4eb544187ec53b5b09166a3ff6b5479e5 by having Alpine use OpenSSL for AES.  Eventually should investigate in more detail.

dstebila:
Fixes a few bugs related to the introduction of AES-NI code.  In particular:

- Makes sure the key schedule is actually being set for decryption.
- Uses the type `__m128i` rather than `__m128i_u` since some compilers choke on the latter.
- Notes in the API documentation that output arrays for AES must now be aligned to a 16-byte boundary, and applies this to BIKE.
  - This was the cause of the BIKE crash in #799.
  - Most AES calls in BIKE seemed to have been aligned, but the `sk_t` `sigma0->raw` and `sigma1->raw` members were not, which was causing the crash.  
  - The solution I took for BIKE was to wrap the AES call in a wrapper that ensures it's always aligned [[aes_wrap.h, line 41]](https://github.com/open-quantum-safe/liboqs/pull/800/files#diff-c3f72798cb0514e552cfcdd92d68a116R41). I tried to keep the scope of the change relatively small, but this does come at a small performance penalty on all BIKE AES calls.  
  - Just putting `ALIGN(16)` around the `r_t` type didn't work, as there are some subsequent routines that call sizeof on that struct, and adding ALIGN(16) apparently changes its size.  
  - @drucker-nir, is my approach okay?  If you have a better solution, please go ahead, I'm not too experienced with alignment issues.

Fixes #799.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ localCheckout: &localCheckout
     - checkout # change this from "checkout" to "*localCheckout" when running CircleCI locally
     - run:
         name: Configure
-        command: scripts/git_no_checkin_in_last_day.sh || (mkdir build && cd build && source ~/.bashrc && cmake -GNinja ${CONFIGURE_ARGS} ..)
+        command: scripts/git_no_checkin_in_last_day.sh || (mkdir build && cd build && source ~/.bashrc && cmake -GNinja ${CONFIGURE_ARGS} .. && cmake -LA ..)
     - run:
         name: Build
         command: ../scripts/git_no_checkin_in_last_day.sh || ninja
@@ -52,6 +52,7 @@ localCheckout: &localCheckout
           mkdir build &&
           (cd build &&
           cmake -GNinja ${CONFIGURE_ARGS} .. &&
+          cmake -LA .. &&
           ninja &&
           ninja run_tests)
           "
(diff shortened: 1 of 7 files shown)


drucker-nir:
Overall, the change looks fine to me, but I wonder whether this is the best possible solution... 
If you use `_mm_loadu_si128`/`_mm_storeu_si128` instead  of `_mm_load_si128`/`_mm_store_si128` in [here](https://github.com/open-quantum-safe/liboqs/blob/34ec4b83a7cff2f53e45b7f536a94fe07d7510c6/src/common/aes/aes256_ni.c#L107) similar to the code in [here](https://github.com/open-quantum-safe/liboqs/blob/34ec4b83a7cff2f53e45b7f536a94fe07d7510c6/src/common/aes/aes256_ni.c#L15). Then, 
1) Because `loadu` is already being used no pattern is broken - consistency.
2) There is no need to limit the functions to only 16 bytes alignment inputs/outputs.
3) Calling these functions is less prone to future mistakes by all schemes.
4) Less lines of documentation, which may burden future users, are needed.

dstebila:
> If you use `_mm_loadu_si128`/`_mm_storeu_si128` instead of `_mm_load_si128`/`_mm_store_si128` in [here](https://github.com/open-quantum-safe/liboqs/blob/34ec4b83a7cff2f53e45b7f536a94fe07d7510c6/src/common/aes/aes256_ni.c#L107) similar to the code in [here](https://github.com/open-quantum-safe/liboqs/blob/34ec4b83a7cff2f53e45b7f536a94fe07d7510c6/src/common/aes/aes256_ni.c#L15). 

Thank you so much @drucker-nir, that's exactly what I needed! I didn't know enough about SSE instructions to know about unaligned versus aligned load/store operations. Indeed it seemed weird that I was forcing a very internal detail (alignment) back on callers.  I've reverted the change in the BIKE AES wrapper.

PR #804, opened 2020-07-31
Add HQC from PQClean

### LINKED ISSUE #709: Add HQC from PQClean

dstebila:
Adds HQC from PQClean.  HQC makes use of nistseedexpander so this adds that as well.

Depends on #803 

Fixes #709.

Note that HQC-256 variants use lots of stack space (similar to Classic McEliece).  Downstream consumers will have to be updated to take this into account.


dstebila:
Replaced by #804

Issue #807, opened 2020-08-06
Undefined behaviour in FireSaber-KEM


claucece:
Hi!

I've been looking at the library and running some things, and this is a really nice work, so thanks for all!

I recently ran some tests, enabling clang's UndefinedBehaviorSanitizer using the build sequence:

```
mkdir build && cd build
cmake -DCMAKE_BUILD_TYPE=Debug -DUSE_SANITIZER=Undefined -GNinja ..
ninja
```

My compiler is clang-11, and arch is x86_64-apple-darwin18.7.0

I ran onto several undefined behaviour, on some algorithms (I'm just pasting only one of each 'family'):

```
tests/test_kem FireSaber-KEM    
Configuration info
==================
Target platform:  x86_64-Darwin-18.7.0
Compiler:         clang (11.0.0 (clang-1100.0.33.17))
OQS version:      0.4.0-rc1
Git commit:       477ef3f60cd04f9b72b499d000a8f77ae99c0d89 (+ local modifications)
OpenSSL enabled:  Yes (OpenSSL 1.1.1g  21 Apr 2020)
AES:              OpenSSL
SHA-2:            OpenSSL
SHA-3:            OpenSSL

================================================================================
Sample computation for KEM FireSaber-KEM
================================================================================
../src/kem/saber/pqclean_firesaber_clean/poly_mul.c:50:44: runtime error: signed integer overflow: 57354 * 40101 cannot be represented in type 'int'
shared secrets are equal
```

```
tests/test_kem FrodoKEM-640-SHAKE
Configuration info
==================
Target platform:  x86_64-Darwin-18.7.0
Compiler:         clang (11.0.0 (clang-1100.0.33.17))
OQS version:      0.4.0-rc1
Git commit:       477ef3f60cd04f9b72b499d000a8f77ae99c0d89 (+ local modifications)
OpenSSL enabled:  Yes (OpenSSL 1.1.1g  21 Apr 2020)
AES:              OpenSSL
SHA-2:            OpenSSL
SHA-3:            OpenSSL

================================================================================
Sample computation for KEM FrodoKEM-640-SHAKE
================================================================================
../src/kem/frodokem/external/frodo_macrify_optimized.c:281:34: runtime error: signed integer overflow: 65534 * 59141 cannot be represented in type 'int'
../src/kem/frodokem/external/frodo_macrify_optimized.c:281:28: runtime error: signed integer overflow: 39658 + 2147450877 cannot be represented in type 'int'
[...]

PR #809, opened 2020-08-06
Fix undefined behaviour in FrodoKEM

dstebila:
Partially addresses #807 


dstebila:
@claucece does this address the Frodo undefined behaviour you noticed?

claucece:
Yes! It fixes it @dstebila

PR #829, opened 2020-09-23
Refactored CI.

xvzcf:
Refactored CI configuration to:

- Use CircleCI 2.1 syntax
- Remove nightly jobs and build all jobs in PRs by default
- Reorganize the execution of jobs
- Trigger downstream CI in OpenSSL, OpenSSH, BoringSSL upon a merge to master


baentsch (approved):
Looks good to me incl. the "win-" removal (relying on AppVeyor instead) and more simple shared lib namespace testing. One question/suggestion for improvement if I may: What about a mechanism (e.g., based on checkin message or branch name) to trigger downstream tests even before merge to master?

dstebila (approved):

xvzcf:
> What about a mechanism (e.g., based on checkin message or branch name) to trigger downstream tests even before merge to master?

This is on my TODO list. The ideal way would've been to differentiate between draft/non-draft PRs in CircleCI, but it seems like that functionality isn't available yet.

baentsch:
> This is on my TODO list. The ideal way would've been to differentiate between draft/non-draft PRs in CircleCI, but it seems like that functionality isn't available yet.

Agree. Until then, what about the ability to trigger this with a PR for a(ny) branch ending with "-rft" (request full test) or the like? This could alleviate manual/local testing efforts and eliminate merge-only "surprises".

xvzcf:
> Agree. Until then, what about the ability to trigger this with a PR for a(ny) branch ending with "-rft" (request full test) or the like? This could alleviate manual/local testing efforts and eliminate merge-only "surprises".

Let me think about this a little bit more.

Issue #833, opened 2020-09-30
Re-run copy from PQClean to pick up HQC revisions


dstebila:

dstebila:
HQC addressed by #835

PR #835, opened 2020-10-01
Sync with PQClean

dstebila:
PQClean commit 383874f4082ab26a234e0e3428ee0f419e008765

Major revisions to HQC algorithm suite.  Minor changes in NTRU and other algorithms.


dstebila:
Failing to do problems with the HQC code from PQClean.  Created https://github.com/PQClean/PQClean/issues/327 to track.

dstebila:
@xvzcf Part of the fix for this problem will probably involve the CPU extensions flag `-mpclmul`.  Are we checking for that already?  Will we need to change something to include that?

xvzcf:
> @xvzcf Part of the fix for this problem will probably involve the CPU extensions flag `-mpclmul`. Are we checking for that already? Will we need to change something to include that?

We aren't. You'd have to change `.CMake/detect_gcc_clang_intrinsics.c`, `.CMake/gcc_clang_intrinsics.cmake`, and the HQC CMakeLists.txt for compile time support.

You'd also have to change https://github.com/open-quantum-safe/liboqs/blob/master/src/common/common.h and https://github.com/open-quantum-safe/liboqs/blob/master/src/common/x86_64_helpers.h for runtime detection, but this might not be necessary since the instruction set came out in 2010.

dstebila:
Thanks @xvzcf.  I think I've figured that out.  But I notice as well that the PQClean META.yml is now specifying `bmi1`, but we only detect `bmi`, so it's not being triggered.  There seems to be something in some of the copy\_from\_pqclean stuff that's remapping bmi1 to bmi, but I don't understand what's going on there.

xvzcf:
> Thanks @xvzcf. I think I've figured that out. But I notice as well that the PQClean META.yml is now specifying `bmi1`, but we only detect `bmi`, so it's not being triggered. There seems to be something in some of the copy_from_pqclean stuff that's remapping bmi1 to bmi, but I don't understand what's going on there.

See https://github.com/PQClean/PQClean/pull/290#discussion_r412231164. Basically, it's reported as `bmi1` by `/proc/cpuinfo` but is referred to by GCC as `bmi`.

dstebila:
Yes I see that, but our copy\_from\_pqclean templating creates macros and code involving the term BMI1, but the CPU features common.c and common.h code involves the term BMI, so there's an inconsistency.

baentsch:
I do recall this, too, and indeed it might have been me who added mapping "BMI1<->BMI" (e.g., [here](https://github.com/open-quantum-safe/liboqs/blob/e1a3c5725f83811b1a0022be146e5e2a45e8ab4e/scripts/copy_from_pqclean/src/sig/family/CMakeLists.txt#L23) ) -- but I guess, also given @xvzcf 's statement above, we must map somewhere as I doubt we can change the way `procinfo` and gcc report/use this feature. But then again, when looking at [this](https://github.com/open-quantum-safe/liboqs/blob/e1a3c5725f83811b1a0022be146e5e2a45e8ab4e/.CMake/detect_gcc_clang_intrinsics.c#L24) we surely could use "BMI1" consistently in our code. But this raises the question how to deal with upstreams that use "BMI" instead of "BMI1" (as seems to be the case)?

dstebila:
> But this raises the question how to deal with upstreams that use "BMI" instead of "BMI1" (as seems to be the case)?

Are there upstreams that use BMI?  I think PQClean is consistent with BMI1.

baentsch:
[This](https://github.com/PQClean/PQClean/pull/290) led me to believe there is -- but at closer look at the current code, you're right: PQClean is, well, clean (only using BMI1). Unfortunately, [there's a "BMI" in Dilithium](https://github.com/pq-crystals/dilithium/blob/master/DILITHIUM_2_META.yml) -- but I know the author of this... :-} So let me try to fix it there, too -- just in case we'd want to pull from there again.

baentsch:
When reviewing the Dilithium import mechanism in general, I stumbled across [this code --labelled as a hack](https://github.com/open-quantum-safe/liboqs/blob/8af52852c46d1b588ef77b609541115ce92b5155/scripts/copy_from_pqclean/copy_from_pqclean.py#L113) ... What speaks against fixing this upstream (@xvzcf ?)

xvzcf:
> When reviewing the Dilithium import mechanism in general, I stumbled across [this code --labelled as a hack](https://github.com/open-quantum-safe/liboqs/blob/8af52852c46d1b588ef77b609541115ce92b5155/scripts/copy_from_pqclean/copy_from_pqclean.py#L113) ... What speaks against fixing this upstream (@xvzcf ?)

If I remember correctly, this was because, in addition to the inconsistency in the required flags, the optimized Dilithium code was not building cleanly on macOS.  The macOS build issue might've been fixed now, but I haven't tried it in a while. And I'm not against fixing the upstream.

baentsch:
> If I remember correctly, this was because, in addition to the inconsistency in the required flags, the optimized Dilithium code was not building cleanly on macOS. The macOS build issue might've been fixed now, but I haven't tried it in a while. And I'm not against fixing the upstream.

Ah, Yes - that was the one stated reason by the Dilithium team why we should pull straight from upstream, namely that OSX with AVX2 support was said to work OK there but not in PQClean. So wouldn't it be reasonable to remove this "hack" with this PQClean import PR and see whether this correction now also made it into PQClean? If it still fails (on OSX), I'd say we _have to_ pull from upstream for Dilithium/exclude Dilithium from a PQClean pull: Either option/outcome would make this hack become unnecessary and make the PR "more correct".

xvzcf:
> So wouldn't it be reasonable to remove this "hack" with this PQClean import PR and see whether this correction now also made it into PQClean? If it still fails (on OSX), I'd say we _have to_ pull from upstream for Dilithium/exclude Dilithium from a PQClean pull: Either option/outcome would make this hack become unnecessary and make the PR "more correct".

I'm not opposed to any of these options.

dstebila:
@baentsch I updated my PR here with the latest pull from PQClean.  Still running into some compilation problems with HQC code from PQClean having shadowed variables.  I thought that was fixed in https://github.com/PQClean/PQClean/pull/332 but seems not.  So that's the current blocker for this PR.

baentsch:
@dstebila Beyond HQC is there any other algorithm coming from PQClean that is improved and that we'd want (show/test as) upgraded? If so (at least NTRU if I read the above correctly), I' suggest we add a simple filter to "copy_from_pqclean.py" to omit specific algorithms from import so that we can (make) progress and not have an "all-or-nothing" import dependency. Please check #841 for a proposal how to easily do this.

dstebila:
Replaced by #846 for NTRU and other minor algorithm updates.  HQC and NTRU Prime tracked by issue #833

PR #841, opened 2020-10-25
selectively exclude NOT_READY algorithms from copy_from_pyclean

baentsch:
Allows excluding specific algorithm families from "copy_from_pyclean" by setting the environment variable "NOT_READY" to those names, e.g., `export NOT_READY=hqc`) when running that script.

This would allow piecemeal progress until all upstream errors are resolved (e.g., on PR #835 ).


dstebila:
Does NOT\_READY actually remove a scheme, or just not copy any updates for it?  For example, if HQC has changed but we don't want to pull that change in a copy-from-PQClean, will NOT\_READY=HQC ignore HQC changes from PQClean, or actually remove HQC from liboqs?

baentsch:
@dstebila  The latter. In your example, it will simply not copy over HQC code from PQClean. Existing (HQC) code (in liboqs) remains untouched.

dstebila (approved):

PR #843, opened 2020-10-25
Modifying naive random selection to Fisher-Yates shuffle in HQC 

pandasamanvaya:
HQC uses a naive random selection algorithm for creating a random vector of fixed weight which could be improved upon by Fisher-Yates shuffle algorithm. Navie random selection algorithm has a best-case time complexity of &Omega;(n<sup>2</sup>) whereas the Fisher–Yates shuffle has the worst-case complexity of O(n)(since random element is selected in O(1)). The downside is that while naive random selection uses O(&omega;<sub>r</sub>) &thickapprox; O(&Sqrt;n) memory, Fisher-Yates shuffle uses O(n) memory.

### DIFF
--- a/src/kem/hqc/pqclean_hqc-128-1-cca2_leaktime/vector.c
+++ b/src/kem/hqc/pqclean_hqc-128-1-cca2_leaktime/vector.c
@@ -139,54 +139,6 @@ void PQCLEAN_HQC1281CCA2_LEAKTIME_vect_set_random_fixed_weight(AES_XOF_struct *c
     }
 }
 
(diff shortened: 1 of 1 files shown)


baentsch:
@pandasamanvaya Thanks for your suggestion to improve this algorithm. However, our project only imports and integrates upstream HQC code: We thus regularly copy HQC from the PQClean project and code changed here would be overwritten at the next such import. So the most logical way to make your suggestion be suitably evaluated and integrated is by proposing it upstream [e.g., here](https://github.com/PQClean/PQClean/tree/master/crypto_kem/hqc-128).

pandasamanvaya:
Thanks for your feedback

PR #844, opened 2020-10-26
Remove keccak duplication from picnic

christianpaquin:
Removes keccak duplication from picnic, moving some missing code to oqs's sha3 folder.


dstebila (approved):
@christianpaquin I notice that some of the optimizations that were in the picnic/external folder aren't present in common/sha3, such as AVX2, ARM Neon, etc. Is performance still satisfactory?

christianpaquin:
> @christianpaquin I notice that some of the optimizations that were in the picnic/external folder aren't present in common/sha3, such as AVX2, ARM Neon, etc. Is performance still satisfactory?

I'll double check with the picnic team.

sebastinas:
The ARM Neon optimization wasn't used in liboqs. I was under the wrong impression that the liboqs SHAKE wrapper would use AVX2 under the hood if available. Without this optimization, the performance degrades noticeably even if the AVX2 x4 implementation is used where possible. I'm not yet sure what's easier, but the two options would be to either make the liboqs wrapper use the AVX2 implementation or also expose a Keccak code package style interface.

christianpaquin:
> The ARM Neon optimization wasn't used in liboqs. I was under the wrong impression that the liboqs SHAKE wrapper would use AVX2 under the hood if available. Without this optimization, the performance degrades noticeably even if the AVX2 x4 implementation is used where possible. I'm not yet sure what's easier, but the two options would be to either make the liboqs wrapper use the AVX2 implementation or also expose a Keccak code package style interface.

Thanks Sebastian. We'll take some time to review our internal common crypto code and see what's the best option for picnic. Again, we want to make sure we don't degrade the current performance.

PR #847, opened 2020-10-28
Documentation updates

dstebila:
Add note about Rainbow cryptanalysis.


baentsch (approved):
Fair summary of status

PR #851, opened 2020-11-13
Sync with PQClean

### LINKED ISSUE #833: Re-run copy from PQClean to pick up HQC revisions

dstebila:
Sync with PQClean.  Update HQC to Round 3.

Fixes #833 


dstebila:
Removed HQC round 2. Will need to re-add HQC round 3 before merging.

PR #853, opened 2020-11-27
disable pthreads testing for mem test

baentsch:
Implementation (preparation) for #439 (see discussion there) reusing existing test code.

@xvzcf , @dstebila : Please (re)confirm that `test_kem`/`test_sig` do not do more testing than desirable when setting the new `-mem` option.

* [ ] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [ ] Does this PR change the the list of algorithms available -- either adding or removing?  (If so, PRs in OQS-OpenSSL, OQS-BoringSSL, and OQS-OpenSSH will also be required by the time this is merged.)

PR #854, opened 2020-12-01
enable Valgrind memory use testing

### LINKED ISSUE #439: Recreate memory benchmark

baentsch:
Fixes #439, replaces #853 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,7 +63,7 @@ jobs:
           working_directory: build
       - run:
           name: Run tests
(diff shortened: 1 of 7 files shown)


dstebila (approved):
Should we run one execution of those for one algorithm in the CI to confirm they work?

baentsch:
> Should we run one execution of those for one algorithm in the CI to confirm they work?

Very reasonable. But one didn't cut it due to some algorithm-specific return values (the tmp-storage code didn't properly handle before). Will ask for re-review when tests finally pass on all platforms.

baentsch:
@dstebila Looks like I nailed the remaining Windows problem anyway. So we could leave mem_tests in. Please give the updated version a quick glance over before we merge (asking for re-review).

dstebila (approved):

Issue #859, opened 2020-12-04
Unintended cmake build logic resulting from misspellings of OQS_PORTABLE_BUILD


ljoy913:
In:
liboqs/src/common/CMakeLists.txt
liboqs/src/kem/bike/CMakeLists.txt

There are three total instances of **OQS_PORTABLE_BUILD** being spelled as **OQS_BUILD_PORTABLE**

**OQS_PORTABLE_BUILD** is defined and is ON, by default.
**OQS_BUILD_PORTABLE** is not defined anywhere and is, by default, OFF

The build logic that mistakenly uses **OQS_BUILD_PORTABLE** is building as a non-portable build in all cases.

Fixing these instances, results in the tests failing to link as the sha3/sha3x4.c sha-3 implementation would no longer be the default on portable builds.

dstebila:
Fixed by #864.

Issue #852, opened 2020-11-27
Current "main" branch: tests hang in test_free()


mouse07410:
MacOS Catalina 10.15.7, Xcode-12.2.

Config:
```
env RUN_TESTS_PARALLEL_UPPER=FALSE cmake -GNinja .. -DOQS_USE_OPENSSL=ON -DOQS_USE_CPU_EXTENSIONS=ON -DBUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=/opt/local -DOQS_KEM_DEFAULT="OQS_KEM_alg_ntru_hps4096821" -DOQS_SIG_DEFAULT="OQS_SIG_alg_falcon_1024" -DCMAKE_BUILD_TYPE=Release
```

Run tests:
```
$ ninja run_tests
[0/1] cd /Users/ur20980/src/liboqs && python3 -m pytest --verbose --numprocesses=auto
======================================== test session starts ========================================
platform darwin -- Python 3.8.6, pytest-6.1.1, py-1.9.0, pluggy-0.13.1 -- /opt/local/bin/python3
cachedir: .pytest_cache
rootdir: /Users/ur20980/src/liboqs
plugins: xdist-2.1.0, forked-1.3.0
gw0 I / gw1 I / gw2 I / gw3 I / gw4 I / gw5 I / gw6 I / gw7 I / gw8 I / gw9 I / gw10 I / gw11 I / gw1[gw0] darwin Python 3.8.6 cwd: /Users/ur20980/src/liboqs                                                                                                               
gw0 C / gw1 I / gw2 I / gw3 I / gw4 I / gw5 I / gw6 I / gw7 I / gw8 I / gw9 I / gw10 I / gw11 I / gw1[gw1] darwin Python 3.8.6 cwd: /Users/ur20980/src/liboqs
.  .  .  .  .
gw0 [380] / gw1 [380] / gw2 [380] / gw3 [380] / gw4 [380] / gw5 [380] / gw6 [380] / gw7 [380] / gw8 [380] / gw9 [380] / gw10 [380] / gw11 [380] / gw12 [380] / gw13 [380] / gw14 [380] / gw15 [380] / gw16 [380] / gw17 [380] / gw18 [380] / gw19 [380]
scheduling tests via LoadScheduling

tests/test_cmdline.py::test_examples[example_sig] 
tests/test_cmdline.py::test_kem[BIKE1-L3-CPA] 
[...]

PR #864, opened 2020-12-09
OQS_BUILD_PORTABLE -> OQS_PORTABLE_BUILD and other refactoring.

xvzcf:
Should address https://github.com/open-quantum-safe/liboqs/issues/859 and https://github.com/open-quantum-safe/liboqs/issues/852


baentsch (approved):
Logic is OK for me (particular for sha3); change in `tmp_store.c` will require an adaptation in the speed container, though. Will do when this is merged in.

dstebila (approved):

Issue #594, opened 2020-01-25
Add options for iOS and Android cross-compilation


nicholasfulton:
I'm attempting to build liboqs for both iOS and Android for a Xamarin.Forms project, and haven't been able to manipulate the included build and configuration files to make that work. If it would be possible, add configuration or documentation to help with cross-compiling the library for iOS and Android, ideally targeting multiple processor configurations. Thanks!

dstebila:
Hi Nicholas, a while ago someone created a cross-compilation script for Android (https://github.com/open-quantum-safe/liboqs/blob/master/configure-android) but it is not something we actively maintain since we don't have anyone in the project who uses it nor do we have any continuous integration testing set up for Android.  We don't have much expertise in cross-compilation.  If you are able to get it working, please let us know what you've done so we can incorporate it.

nicholasfulton:
I believe I've made good progress with developing a build script for Android. However, developing a more general solution runs into an issue with building using Clang. It seems that the build scripts specifically require GCC and disallow Clang. That's problematic because Clang is necessary to cross-compile for iOS and Android, as GCC doesn't include toolchains for those platforms. Is there a reason that liboqs requires GCC? Otherwise, I'll work on a more elegant way to allow building with Clang as well.

dstebila:
Hi @nicholasfulton, in general liboqs should compile just fine with clang -- our macOS continuous integration builds use the default compiler on macOS, which is clang.  

The now-unmaintained configure-android script does explicitly refer to gcc, and that was just because that was what the person putting that script together was able to get working, not intentionally made to restrict use to gcc-only.

zugaldia:
I've got basic Java bindings working, including Android, on https://github.com/zugaldia/javacpp-presets. It uses the JavaCPP infrastructure to build the bindings. A few notes for those willing to give it a try:
- The Linux bindings support the full API.
- The Android bindings only support a subset of the API and are much trickier to build. They should be considered a WIP:
  - [This patch](https://github.com/zugaldia/javacpp-presets/blob/master/liboqs/liboqs-cross.patch) enables cross-compilation.
  - NDK doesn't come with OpenSSL, I'm using the binaries provided by [Teska Labs](https://teskalabs.com/blog/openssl-binary-distribution-for-developers-static-library).
  - Unlike `configure-android` discussed above which uses `gcc`, I've only tested the latest NDK version with `clang`. 
  - `sig-picnic` needs to be disabled to avoid a compilation error.
[...]

PR #875, opened 2020-12-17
Added Android CircleCI config

nicholasfulton:
Adds Android CI/CD config for the CircleCI pipeline. Tests whether building library is successful.

Partially addresses Android cross-compilation (#594)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,6 +102,26 @@ jobs:
       - store_artifacts:
           path: build/test-results
 
+  android_emulated:
+  description: A template for running liboqs tests on emulated Android VMs.
+  parameters:
+    ABI:
+      description: "The desired Android ABI to target"
+      type: string
+    SDK:
+      description: "The desired SDK version to target"
+      type: string
+  docker:
+    - image: circleci/android:api-30-ndk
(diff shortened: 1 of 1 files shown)


baentsch:
At first glance I see nothing wrong with these configs. What is the reason they're not added to the active jobs/didn't get run in CCI?

nicholasfulton:
I had added them to the build workflow. I see now that that's explicitly not called on the main branch. What's the reason that the build workflow isn't called when committing to main? In light of this, should I move/copy the Android jobs to "commit-to-main"?

baentsch:
Our preference is to run all tests before new code is committed to main. For some reason I'm not seeing _any_ CCI runs as part of your PR... Are those runs maybe executed within your personal CCI account instead? If so, could you please share a link to those runs?

dstebila:
I've created a copy of this PR as #877 which is inside the OQS repositories so it should trigger the CI.

baentsch:
@nicholasfulton Thanks again for your PR and sorry for the delay. Would you be OK withdrawing this PR when #877 gets merged? Future PRs should be more smooth as we've relaxed the use of docker authentication for CCI.

PR #878, opened 2021-01-07
FrodoKEM portable build with CPU feature detection

### LINKED ISSUE #874: FrodoKEM optimizations not enabled for portable builds
Need to reorganize code so that AVX2 intrinsics can be used even in portable builds.

https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/frodokem/external/frodo_macrify_optimized.c

dstebila:
Fixes #874.


dstebila:
I've figured out the problem with this.  All of the code for a single Frodo variant is compiled into a single compilation unit.  In the portable build that also supports optimizations if available, the `-maes -mavx2` compiler options must be given to allow for compilation of the optimizations, but then this affects the entire compilation unit, making it no longer portable.  

@xvzcf Do you know if it's possible to have CMake compile the same file multiple times but with different headers, and outputting to different compilation units?

baentsch:
> All of the code for a single Frodo variant is compiled into a single compilation unit. In the portable build that also supports optimizations if available, the `-maes -mavx2` compiler options must be given to allow for compilation of the optimizations, but then this affects the entire compilation unit, making it no longer portable.

This precisely is the problem with the common code, too (say if we compile SHA/KE CPU-optimized): The moment we allow CPU-specifics, we lose portability. The only solution I see for this is to create separate compilation units -- with separate symbols as we need both units available in the executable for the portability property.

> All of the code for a single Frodo variant is compiled into a single compilation unit. 

Now if this is the case, a rather straightforward but unelegant solution would be to wrap the few external entry points such that they can be #define'd to be optimized or not and then compile the whole code twice (possibly copying it into different directories for cmake to work OK): Once with and once without optimization, setting the #define differently in each case. Generates twice as much code, but would serve the purpose. Definitely not elegant but basically what's happening with the PQClean code (coming in separate folders).

baentsch (approved):
For whatever it's worth: This is what I'd call "macro kung-fu" and it looks good to me (certainly the build-and-switch logic as far as I can judge it from reading through it). The other thing I'd do is run comparative performance tests with/without this patch (and obviously with/without AVX2+PORTABLE set) -- but then again, we've got automation for that so maybe we can leave that to after the merge assuming you already did a quick check -- and didn't see a significant regression for algorithms using SHA3 (and AVX2&&PORTABLE active)? Otherwise, please let me know so I'd trigger a profiling run against this branch to check this out before merge.

dstebila:
Thanks, @baentsch. I did see performance increases from this on portable builds in my testing.

PR #881, opened 2021-01-22
Use valgrind to check for secret-dependent branching

### LINKED ISSUE #812: Add constant time checks using Valgrind
See e.g. https://post-apocalyptic-crypto.org/timecop/index.html or https://github.com/microsoft/PQCrypto-LWEKE/compare/master...dstebila:ds-valgrind.

jschanck:
Here's one way to use Valgrind to check for secret-dependent branching (and resolve #812). It's based on  Moritz Neikes [TIMECOP](https://post-apocalyptic-crypto.org/timecop/index.html).

The idea behind TIMECOP is that if you tell Valgrind that secret data resides in uninitialized memory, then Valgrind will tell you when your program branches on secret data.

SUPERCOP's implementation of TIMECOP introduced "classify" and "declassify" terminology for the operation of telling Valgrind that memory is secret or non-secret. I've replicated that terminology by defining two macros in oqs.h:

```c
#define OQS_TIMECOP_CLASSIFY(addr, len)  VALGRIND_MAKE_MEM_UNDEFINED(addr, len)
#define OQS_TIMECOP_DECLASSIFY(addr, len)  VALGRIND_MAKE_MEM_DEFINED(addr, len)
```

Every secret can ultimately be traced back to a call to randombytes, so we can tell Valgrind that secrets reside in uninitialized memory by defining our own randombytes, e.g.

```c
void TEST_KEM_randombytes(uint8_t *random_array, size_t bytes_to_read) {
  OQS_randombytes_system(random_array, bytes_to_read);
  OQS_TIMECOP_CLASSIFY(random_array, bytes_to_read);
}
```
and making it the default
```c
  OQS_randombytes_custom_algorithm(&TEST_KEM_randombytes);
```

I've done this in test_kem.c and test_sig.c. If you run the new test_kem or test_sig outside of Valgrind there's no change. But in Valgrind you'll see something like this:

```sh
$ valgrind ./tests/test_kem sntrup761
[...]
== Thread 2:
== Conditional jump or move depends on uninitialised value(s)
==    at 0x2B4EBF: PQCLEAN_SNTRUP761_CLEAN_crypto_kem_keypair (kem.c:145)
==    by 0x10BCE8: kem_test_correctness (test_kem.c:94)
==    by 0x10C58B: test_wrapper (test_kem.c:177)
==    by 0x4CDAFA2: start_thread (pthread_create.c:486)
==    by 0x4DED4CE: clone (clone.S:95) 
[...]
```
(I've compiled with CMAKE_BUILD_TYPE=debug to get the line numbers were secret branching occurs.)

Of course, some non-secret data is derived from calls to randombytes, and examining pqclean_sntrup761_clean/kem.c we see that line 145 is part of a rejection sampling loop and the branch is not secret. We can remove the false positive by calling OQS_TIMECOP_DECLASSIFY on the offending memory.

There's a different example of declassifying memory in Kyber512 in the first commit linked to this PR.

--- 

None of changes to oqs.h, test_kem.c, or test_sig.c introduce a dependency on Valgrind. Users can still compile test_kem and test_sig and run them outside of Valgrind. All we need are some header files that the Valgrind project provides, for this purpose, under a BSD license. I've added those to src/common/valgrind.

**TODO:**
- [x] Configure CMake to include the local valgrind/memcheck.h and valgrind/valgrind.h.
- [x] Add a compilation flag to completely remove calls to VALGRIND_MAKE_MEM_DEFINED. Compiling the schemes with "-DNVALGRIND" will do the trick. Alternatively we can change the definitions of OQS_TIMECOP_(DE)CLASSIFY in oqs.h with an ifdef.
- [ ] Flag false positives with OQS_TIMECOP_DECLASSIFY.


baentsch:
Very nice approach to use valgrind to highlight possible random-dependent code paths (tagging random data as uninitialized memory). Three "nitpick"/curiousity questions if I may before you finalize this:
1) Is it _guaranteed_ that all secret data has been generated via the (centralized) randomness function? I have a fuzzy feeling it is (e.g., via a NIST competition demand to only use such function [Reference needed] but I'm not certain).
2) Shouldn't take constant runtime checks take all data (secret or not) into account? But possibly I read to much (of a goal) into #812.
3) Wouldn't it be better to have an option to _enable_ Valgrind-based testing, rather than disabling it? I'm personally no fan of `#ifndef(NO_FEATURE)`: Isn't `#ifdef FEATURE` more readable in code (e.g., for activating/deactivating `OQS_TIMECOP_CLASSIFY` as per your TODO)?

jschanck:
>     1. Is it _guaranteed_ that all secret data has been generated via the (centralized) randomness function? I have a fuzzy feeling it is (e.g., via a NIST competition demand to only use such function [Reference needed] but I'm not certain).

Any scheme that passes the liboqs KAT test generates all of its secret data using the centralized random function.

NIST's call for proposals did not guarantee KATs of exactly the form assumed by liboqs. But the form assumed by liboqs is a de facto standard---no one deviates from it.

>     2. Shouldn't take constant runtime checks take all data (secret or not) into account? But possibly I read to much (of a goal) into #812.

I'm currently "declassifying" public keys, ciphertexts, and (signature, message) pairs. There are contexts where it makes sense to treat each of these as secret, but I think it's good to keep a narrow scope.

>     3. Wouldn't it be better to have an option to _enable_ Valgrind-based testing, rather than disabling it? I'm personally no fan of `#ifndef(NO_FEATURE)`: Isn't `#ifdef FEATURE` more readable in code (e.g., for activating/deactivating `OQS_TIMECOP_CLASSIFY` as per your TODO)?
> Edit: Plus the question whether checking in valgrind header files is really necessary? Shouldn't installation of Valgrind suffice?

See latest commit on this PR. I don't have much of a preference, but I think there's some benefit to enabling timecop for all debug builds.

Debian users, for instance, might eventually get the option of installing "liboqs" and "liboqs-dbg". A Debian user who installs liboqs-dbg might want to run their own constant-time analysis, but the Debian packager should not have to have valgrind installed in order to create liboqs-dbg.

baentsch:
> See latest commit on this PR.

That (TIMECOP option) really looks good: Thanks!

> the Debian packager should not have to have valgrind installed in order to create liboqs-dbg

Why? It's just one more prereq/package -- Is debian packaging resource-constrained? For the actual package I guess all the dev/dbg variant would need is a(nother) dependency (the runtime package wouldn't need).

That said -- the underlying concern was "GPL license contagion" via valgrind code (that could make liboqs unattractive for some)... Upon reading the headers, it seems just these two files are not GPL'd. I didn't experience such "license splitting" of header files for a program yet, but this does all pretty nearly eliminate this concern.

dstebila:
Thanks John! I've sent an invite adding you to the OQS Core team on Github so that CI will run on your PRs.  

Having not looked at the code in detail, I would be inclined to have the Valgrind compilation be optional, not turned on for every debug build.  (Similar to how the clang sanitizers are done -- they are extras to be added on top of debug builds, but are not enabled automatically by turning on debug builds.)  And then we would add an extra CI job that does a TIMECOP build.

We'd also need to update documentation to describe the ability to run these tests, and to acknowledge the Valgrind/TIMECOP stuff we're relying on.

jschanck:
Thanks for the comments, Michael, Douglas. I've just pushed a slightly different approach which I think will be better in the long run.

Major changes:
- No instrumentation of code in /src. The new version uses valgrind "error suppression" files, stored in the /tests directory, which as of Valgrind 3.14.0 allow for suppression of errors by source file and line number.
- No valgrind headers in /common. There's a new build flag "OQS_ENABLE_TEST_CONSTANT_TIME" which introduces a compile time dependency on valgrind when building the tests.
- Documentation. It's currently at the top of test_constant_time.py, but it's more-or-less ready to copy to the wiki if/when this is merged. I've copied it below to explain, in more detail, how this all works.
- I've made suppression files for all of the secret dependant flow issues / false-positives that I run into on my laptop. My CPU is about 10 years old, so there will be more issues to document in the platform-specific implementations.

I still need to add a CI job. I could use some help with that.

# test_constant_time.py

The goal of this script is to ensure that every instance of secret-dependant
control flow in liboqs is documented. This script does not ensure that all of
the software in liboqs is constant time. Rather, it is intended to aid auditors
in their search for non-constant time behaviour.

WARNING: This script currently runs test_kem and test_sig on random seeds.
It is not coverage guided. It will miss instances of non-constant time
behaviour in code paths that are rarely executed.

This script requires Valgrind version >= 3.14.0, and it only gives meaningful
results if test_kem and test_sig have been compiled with CMAKE_BUILD_TYPE=Debug
and OQS_ENABLE_TEST_CONSTANT_TIME.


How this script works
---------------------
This script runs test_kem (and/or test_sig) through Valgrind's Memcheck tool.
Valgrind executes the test program and issues an error message if/whenever the
program's control flow depends on uninitialized data. As observed by Adam
Langley [1], if we tell Valgrind that secrets are uninitialized, then Valgrind
will tell us about secret-dependant control flow.

Assuming that each scheme in liboqs passes test_kat, our test_kem and test_sig
programs are structured such that all secret data can be traced back to a call
to OQS_randombytes. The tests intercept calls to OQS_randombytes and tell
Valgrind that every random byte is uninitialized. Hence, Valgrind will issue an
error if (but not only if!) our tests branch on secret data.

Since there may be false positives, we say that Valgrind identifies "suspected
non-constant time behaviour".

Again, the purpose of this script is to ensure that suspected non-constant time
behaviour is documented. This script ships with a collection of Valgrind
"suppression files". Each suppression file documents one or more instances of
suspected non-constant time behaviour in liboqs.

The suppression files are also used to silence errors from Valgrind. If this
script runs without error, then all of the suspected non-constant time behaviour
in liboqs has been documented. If this script fails, then a new suppression
file should be written.


How to write suppression files
------------------------------
Valgrind will output a suppression file template along with its error message.
It's your job to copy this template to the correct location, edit it, and tell
this script about the new file.

Suppression files for KEMs are stored in
    liboqs/tests/constant_time/kem/{passes,issues}/.
Suppression files for signature schemes are stored in
    liboqs/tests/constant_time/sig/{passes,issues}/.

This script does not differentiate between the passes and issues
subdirectories.  The label is for auditors. We "give a pass" to an error that
is known not to be a security threat, and we store the corresponding
suppression file in the "passes" subdirectory. We "raise an issue" about any
other error, and we store the corresponding suppression file in the "issues"
subdirectory.

If you are unsure where your suppression file belongs, then save it to the
"issues" subdirectory.

Once you've written a suppression file, give it a descriptive file name and
tell this script about it. There are json files called passes.json and
issues.json in
    liboqs/tests/constant_time/{kem,sig}/
These json files contain dictionaries of the form
    { "Scheme name" : ["list", "of", "suppression", "files"], ... }
Add the name of your suppression file to the appropriate list to suppress
the errors that you have documented.


How to write a good suppression file
------------------------------------
Here is an example of a suppression file:
    {
       Rejection sampling to produce public "A" matrix
       Memcheck:Cond
       fun:rej_uniform
       fun:PQCLEAN_KYBER*_CLEAN_gen_matrix
    }

The brackets wrap a single error that is to be suppressed. Within the brackets,
the first line is a comment. The remaining lines tell Valgrind to ignore
any "Memcheck:Cond" errors that occur when a function named rej_uniform is
called from a function whose name matches the glob pattern
PQCLEAN_KYBER*_CLEAN_gen_matrix.

Before this suppression file was written, a run of this script produced the
following output.

    ==594== Conditional jump or move depends on uninitialised value(s)
    ==594==    at 0x22550D: rej_uniform (indcpa.c:133)
    ==594==    by 0x225654: PQCLEAN_KYBER512_CLEAN_gen_matrix (indcpa.c:177)
    ==594==    by 0x2257D1: PQCLEAN_KYBER512_CLEAN_indcpa_keypair (indcpa.c:216)
    ==594==    by 0x1B6C1E: PQCLEAN_KYBER512_CLEAN_crypto_kem_keypair (kem.c:26)
    ==594==    by 0x1B6B9F: OQS_KEM_kyber_512_keypair (kem_kyber_512.c:56)
    ==594==    by 0x10D123: OQS_KEM_keypair (kem.c:818)
    ==594==    by 0x10AD07: kem_test_correctness (test_kem.c:103)
    ==594==    by 0x10B4E7: test_wrapper (test_kem.c:186)
    ==594==    by 0x4CDAFA2: start_thread (pthread_create.c:486)
    ==594==    by 0x4DED4CE: clone (clone.S:95)
    ==594==
    {
       <insert_a_suppression_name_here>
       Memcheck:Cond
       fun:rej_uniform
       fun:PQCLEAN_KYBER512_CLEAN_gen_matrix
       fun:PQCLEAN_KYBER512_CLEAN_indcpa_keypair
       fun:PQCLEAN_KYBER512_CLEAN_crypto_kem_keypair
       fun:OQS_KEM_kyber_512_keypair
       fun:OQS_KEM_keypair
       fun:kem_test_correctness
       fun:test_wrapper
       fun:start_thread
       fun:clone
    }

The lines beginning with "==" are a Valgrind error message. The bracketed text
is a suppression file template. To produce the final suppression file we
added a comment, replaced "512" with a wildcard (since an identical error occurs
in other Kyber parameter sets), and truncated the backtrace (since the extra lines
provide no interesting information to auditors).

The "fun:rej_uniform" line says to ignore _all_ Memcheck:Cond errors in
rej_uniform, but Valgrind told us that line 133 was the problem. Any
"fun:name" line in the backtrace can be replaced by an equivalent
"src:file:line", so we could have narrowed the scope of our suppression:
    {
       Rejection sampling to produce public "A" matrix
       Memcheck:Cond
       src:indcpa.c:133 # fun:rej_uniform
       fun:PQCLEAN_KYBER*_CLEAN_gen_matrix
    }
Here "# fun:rej_uniform" is a comment. An update to the Kyber source code might
break our suppression file by changing the line number, and leaving the function
name as a comment might help a future reviewer.

An ellipsis (...) can serve as a wildcard for a portion of the backtrace.
We could have written:
    {
       Rejection sampling to produce public "A" matrix
       Memcheck:Cond
       ...
       fun:PQCLEAN_KYBER*_CLEAN_gen_matrix
    }
But this is perhaps too concise. Remember that the goal here is to help auditors.

Further information can be found in Valgrind's manual. See
    https://www.valgrind.org/docs/manual/manual-core.html#manual-core.suppress
and
    https://www.valgrind.org/docs/manual/mc-manual.html#mc-manual.suppfiles

Credits
-------
The observation that Valgrind can be used to identify non-constant time
behaviour is due to Adam Langley [1, 2]. Mortiz Neikes' TIMECOP project applies
Langley's idea to the SUPERCOP benchmarking suite [3]. Versions of SUPERCOP
starting with 20200816 include TIMECOP and apply Langley's idea to randombytes
calls in particular [4]. We have borrowed the idea of instrumenting randombytes
calls from SUPERCOP.

[1] https://github.com/agl/ctgrind
[2] https://boringssl.googlesource.com/boringssl/+/a6a049a6fb51a052347611d41583a0622bc89d60
[2] https://post-apocalyptic-crypto.org/timecop/index.html
[3] http://bench.cr.yp.to/tips.html#timecop

baentsch:
> I still need to add a CI job. I could use some help with that.

Just looked into that: Just from reading, one first question regarding the skip logic: It seems this test is always run, even if  `OQS_ENABLE_TEST_CONSTANT_TIME` is not set. I guess that doesn't make sense/is not intended? Second issue: Requirement for valgrind > 3.14.0: This is not part of `apt` in ubuntu 18/bionic, so we'd need to build that into the CI image from source. Alternative: Upgrade CI to 20/focal (also as discussed in #884: When to upgrade ubuntu): 20.04 readily apt-packages valgrind 13.15.0. 

For now, proposed solutions: 
On 1) You may want to add this code to `test_constant_time.py`:
```
def is_constant_time_enabled():
    with open(os.path.join('build', 'include', 'oqs', 'oqsconfig.h')) as fh:
        for line in fh:
            if line.startswith("#define OQS_ENABLE_TEST_CONSTANT_TIME"):
                return True
    return False
```

and then guard test execution(s) with it, e.g., 
```
@pytest.mark.skipif(not is_constant_time_enabled(), reason="OQS_ENABLE_TEST_CONSTANT_TIME not enabled")
```

For 2), there's now https://github.com/open-quantum-safe/ci-containers/pull/47 adding valgrind 3.16.1 to our current CI image.

Once both things are done, this test can be easily activated for CI in `.circleci/config.yml` by adding this new job (also doing some more tests with a gcc-debug build while we're at it):
```
      - linux_x64:
          <<: *require_stylecheck
          name: constant-time
          context: openquantumsafe
          CONTAINER: openquantumsafe/ci-ubuntu-bionic-x86_64:latest
          CMAKE_ARGS: -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON
          PYTEST_ARGS: --ignore=tests/test_portability.py --numprocesses=auto
```

Final comment: stylecheck test fails. You may want to run `./tests/run_astyle.sh` to see where reformatting needs to be done. You may want to do a local test run as per [the wiki](https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide#continuous-integration-ci-testing), e.g. with jobname set to "constant-time" (CI job added above).

jschanck:
Thanks for the help, Michael!

I'm marking this as ready for review. There are some small outstanding issues, and some features that I would like to add, but this is useful as-is.

#### Issues
1. Some of the schemes (Classic McEliece, Rainbow, and SPHINCS, but maybe others) take a fairly long time to run in valgrind. I bumped the timeout from 8 minutes to 16 minutes in helpers.py, but this was not enough. My most recent commit bumps to 32 minutes, but hasn't run yet due to an authentication issue. I wonder whether this timeout is necessary at all. CircleCI probably enforces their own timeout.

2. We should run all available implementations through the constant time tests. The reference implementations work fine on the linux_x64 machine, but we only end up testing avx2 implementations because of the run-time CPU feature detection. We either have to run on a set of machines that covers the set of available implementations, or we need a mechanism for overriding the choice made based on CPU features.

baentsch:
> or we need a mechanism for overriding the choice made based on CPU features

If I understand the problem right, there is such mechanism: Just (un)set [OQS_PORTABLE_BUILD](https://github.com/open-quantum-safe/liboqs/wiki/Customizing-liboqs#oqs_portable_build) when building for that test. Also you may want to run with [OQS_USE_CPU_EXTENSIONS](https://github.com/open-quantum-safe/liboqs/wiki/Customizing-liboqs#oqs_use_cpu_extensions) OFF to test ref code.

> Some of the schemes (Classic McEliece, Rainbow, and SPHINCS, but maybe others) take a fairly long time to run in valgrind.

This seems like a bigger practical issue. What about moving this test out of CI and into our nightly [profiling](https://github.com/open-quantum-safe/profiling) test runs instead? a) we already have both -ref and -cpuext builds there and b) lots of time that CCI may not grant such "long-runners".

jschanck:
> > or we need a mechanism for overriding the choice made based on CPU features
> 
> If I understand the problem right, there is such mechanism: Just (un)set [OQS_PORTABLE_BUILD](https://github.com/open-quantum-safe/liboqs/wiki/Customizing-liboqs#oqs_portable_build) when building for that test. Also you may want to run with [OQS_USE_CPU_EXTENSIONS](https://github.com/open-quantum-safe/liboqs/wiki/Customizing-liboqs#oqs_use_cpu_extensions) OFF to test ref code.

Thanks, that will probably be sufficient for now.

> > Some of the schemes (Classic McEliece, Rainbow, and SPHINCS, but maybe others) take a fairly long time to run in valgrind.
> 
> This seems like a bigger practical issue. What about moving this test out of CI and into our nightly [profiling](https://github.com/open-quantum-safe/profiling) test runs instead? a) we already have both -ref and -cpuext builds there and b) lots of time that CCI may not grant such "long-runners".

I'm submitting another run with a longer timeout. It seems that python.subprocess uses *wallclock* time instead of cpu time. The jobs that are timing out do not take so long on my machine, so I think they're just getting sent to the background for too long on circleci. If this doesn't work then we can revisit.

baentsch:
I'm afraid we'd need to go over this again: The newly landed Dilithium+Kyber round3 code obviously requires new suppression files. @bhess: Might you have time & insight to help get this PR over the finish line?

bhess:
I will take a look at it.

jschanck:
There were also BIKE failures in the last constant-time-x64-extensions build. Turns out that Valgrind does not yet support AVX512. I've got a patch to disable AVX512 instructions when OQS_ENABLE_TEST_CONSTANT_TIME is on.
 
@bhess if you've already started the suppression files for round 3 kyber/dilithium, I'm happy to let you work on them. Otherwise I can wrap this up now.

bhess:
> @bhess if you've already started the suppression files for round 3 kyber/dilithium, I'm happy to let you work on them. Otherwise I can wrap this up now.

Unfortunately I got a bit stuck with another issue. Feel free to wrap it up. Thanks!

jschanck:
Hmm. ca5b2e4 doesn't seem to have worked. Any thoughts on a more reliable way to disable AVX512 when OQS_ENABLE_TEST_CONSTANT_TIME is on?

baentsch:
What's the plan with this PR now? The logic has already proven its value but it looks like it's clearly running for too long, e.g., to regularly allow CI to pass. What about the idea to remove this test from CI but mandate running it (locally) successfully before assigning a new release number? Adding a line item to do so to https://github.com/open-quantum-safe/liboqs/wiki/Release-process-(Under-construction) might be sufficient.

dstebila:
Do the Valgrind tests actually take that much longer?  If there are a few degenerate algorithms for which it takes exceedingly long, then we could skip those but run the CI on the rest of them.  I recall that we have in the past observed weird behaviour when the CI runs pytest with many threads and some of them seem to never resolve, and if I read the output correctly it did spawn many threads, so perhaps we should try `--numprocesses=1` or some other small constant.

jschanck:
@dstebila Testing without valgrind takes about an hour. Testing with valgrind takes about 3 hours.

@baentsch I don't think a policy of running the tests locally is sufficient. Ultimately these tests should run on several different platforms.

CircleCI supports filtering jobs based on branch name. So I would recommend that we make a special "audit" branch which runs more expensive tests like test_constant_time. And we make it a policy that changes to cryptographic code gets merged to "audit" first and then to "main" only once it has passed the more extensive tests.

baentsch:
> perhaps we should try --numprocesses=1 or some other small constant.

Good idea -- worthwhile trying to see if that gets it through CI.

> changes to cryptographic code gets merged to "audit" first and then to "main" only once it has passed the more extensive tests

Also a good thought -- begs the question, though, what constitutes "cryptographic code"? Any change in a subdirectory of `kem` and `sig`? Can one write a CI test policy in such way (only run this expensive test if code within specific directories changed)? Going this route for each more simple "plumbing"/adaptation/integration change feels "too heavyweight".

dstebila:
> @dstebila Testing without valgrind takes about an hour. Testing with valgrind takes about 3 hours.

Are there individual algorithms that take more than 30 minutes to run when under valgrind?  Because the specific failure in the CircleCI log is that it went 30 minutes without output, not that it exceeded the total runtime limit.

> CircleCI supports filtering jobs based on branch name. So I would recommend that we make a special "audit" branch which runs more expensive tests like test_constant_time. And we make it a policy that changes to cryptographic code gets merged to "audit" first and then to "main" only once it has passed the more extensive tests.

There's always been an intention to move some of the CI for PQClean and OQS to one of @cryptojedi's big servers, but it hasn't happened because we've always been able to stretch with what CircleCI provides.  If we are getting beyond that, now might be the time to revisit.

jschanck:
@dstebila The 30 minute timeout is a soft limit in CircleCI. I bumped it to an hour.

I also pushed a proof of concept for only running test_constant_time when the branch name is 'audit'. I think we can skirt the question of 'what is cryptographic code' by leaving it up to the person reviewing a PR whether they want to send it off to the audit branch or not.

baentsch:
> I also pushed a proof of concept for only running test_constant_time when the branch name is 'audit'. I think we can skirt the question of 'what is cryptographic code' by leaving it up to the person reviewing a PR whether they want to send it off to the audit branch or not.

What about extending this a bit and running it on any branch that has the term "audit" _in_ its name? This way, the creator of the PR already has it in his hands to trigger "the full monty" (and the reviewer already gets to see the results -- or request it if he things this moniker is necessary)?

jschanck:
> What about extending this a bit and running it on any branch that has the term "audit" _in_ its name? This way, the creator of the PR already has it in his hands to trigger "the full monty" (and the reviewer already gets to see the results -- or request it if he thinks this moniker is necessary)?

We can definitely make it run on any liboqs branch whose name matches the regex /^audit/, but if the PR is from a forked repo (like this one) then the branch name appears as something like 'pull/881' to circleci.

If someone wants to make an open-quantum-safe:audit branch, then I (or maybe any maintainer?) can retarget this PR and we can try out this workflow.

Issue #884, opened 2021-01-28
Contributing guide update


baentsch:
This is to foster a discussion on updates to [Contributing wiki](https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide)

> The code must build on Linux, macOS, and Windows.

This is not true for all algorithms, right? Suggested rephrase:

> The code must build at least on Linux (x86_64). Optimizations and other operating systems supported may be supported and must be suitably registered for each algorithm in https://github.com/open-quantum-safe/liboqs/blob/main/.CMake/alg_support.cmake

Also this statement isn't quite correct for all algorithms: 

> Source code must be formatted in accordance with our formatting requirements (make prettyprint).

Issues: a) We're not using `make` any more and b) this only applies to OQS core code, not the actual algorithms. So I suggest to introduce a new section "Common code" and move the above statement there changed as follows:

> Source code contributed to OQS core code must be formatted in accordance with our formatting requirements. Non-OQS core code is any code not originally created in this repository but only imported manually or automatically, such as specific quantum-safe algorithms. Adherence to our formatting requirements is successfully tested by executing `./tests/run_astyle.sh` in the liboqs project root folder without "Formatting" statements returned. Passing this test is a prerequisite for successful CI.

The `Automated inclusion` section needs to be re-written once #883 is landed.

Regarding [Testing](https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide#testing):

What about changing that heading to "Continuous Integration Testing (CI)" ?

Issues: We are not testing Centos 7 any more, so we need to drop the statement. 

Regarding Ubuntu: Would it be time to upgrade to a more recent LTS than 18.04? Canonical main stream is 20.04. -> I'd suggest to agree a time frame for this update.

Do we further want to posit

> Normal pull requests (PRs) must have all CI tests pass. Non-standard PRs must explain the reasons for breaking CI as well as a plan how to get CI to work again.

If so, we further face the issue that only core team members can actually trigger full CI. So should the further demand passing all CI tests locally be explicitly stated? Not really sure what to recommend here, so leaving a recommendation away at this time.

Next steps: Please comment below with changes, additions, etc. so we can agree in a next meeting how to close out this issue.

christianpaquin:
These changes look good to me.

dstebila:
All good suggestions, thanks for checking on this.  

[...]

PR #885, opened 2021-01-29
remove docker auth for now

baentsch:
Disables docker auth in CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,9 +19,10 @@ jobs:
     description: Validate formatting of code and documentation
     docker:
       - image: openquantumsafe/ci-ubuntu-bionic-x86_64:latest
-        auth:
-          username: $DOCKER_LOGIN
-          password: $DOCKER_PASSWORD
+# Re-enable iff docker enforces rate limitations without auth:
+#        auth:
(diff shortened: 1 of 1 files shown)


dstebila (approved):

.circleci/config.yml
```diff
@@ -19,9 +19,10 @@ jobs:
     description: Validate formatting of code and documentation
     docker:
       - image: openquantumsafe/ci-ubuntu-bionic-x86_64:latest
-        auth:
-          username: $DOCKER_LOGIN
-          password: $DOCKER_PASSWORD
+# Re-enable iff docker enforces rate limitations without auth:
```
thomwiggers, line 22: This will be happening soon though, right? https://www.docker.com/blog/docker-hub-image-retention-policy-delayed-and-subscription-updates/ says "mid 2021".

https://docs.docker.com/docker-hub/download-rate-limit/

baentsch: Not necessarily: See link at https://github.com/open-quantum-safe/liboqs/issues/884#issuecomment-769158300

thomwiggers: Ah I see :+1:

PR #883, opened 2021-01-27
Update copy_from_pqclean / copy_from_upstream

bhess:
Updates copy_from_pqclean / copy_from_upstream (#880)

* Renames copy_from_pqclean to copy_from_upstream.

* Adds 'upstreams' field to copy_from_upstream.yml, allowing to specify upstream repositories, branches and commits.

```
upstreams:
  -
    name: pqclean
    git_url: https://github.com/PQClean/PQClean.git
    git_branch: master
    git_commit: 3d7d2024fa892bd7f00dca3fff4122175f4a26dc
    kem_meta_path: 'crypto_kem/{pqclean_scheme}/META.yml'
    sig_meta_path: 'crypto_sign/{pqclean_scheme}/META.yml'
    kem_scheme_path: 'crypto_kem/{pqclean_scheme}'
    sig_scheme_path: 'crypto_sign/{pqclean_scheme}'
  -
    name: pqcrystals-kyber
    git_url: https://github.com/pq-crystals/kyber.git
    git_branch: master
    git_commit: 8c8d7da74ee92808eebb6128ca209fd5851f48dc
    kem_meta_path: '{pretty_name_full}_META.yml'
    kem_scheme_path: '.'
  -
    name: pqcrystals-dilithium
    git_url: https://github.com/pq-crystals/dilithium.git
    git_branch: master
    git_commit: bda9689704262b53f0942efdda6340b922acb58b
    sig_meta_path: '{pretty_name_full}_META.yml'
    sig_scheme_path: '.'
...
```

* Modifies "copy" command in copy_from_upstream: sources are pulled from specified git-repositories. Implementation folders in $LIBOQS_DIR/src will be prefixed with the upstream-name.

* Adds "verify" command in copy_from_upstream: Implementations in $LIBOQS_DIR/src are compared with expected upstream versions. Implementations that differ from upstream will be hightlighted.

* Prepares for copying pqclean, pqcrystals-kyber and pqcrystals-dilithium (Round 3 versions) from upstream.

How to use copy_from_upstream:

```
$ ./copy_from_upstream.py --help
usage: copy_from_upstream.py [-h] [-v VERBOSITY] [-k] {copy,verify}

positional arguments:
  {copy,verify}

optional arguments:
  -h, --help            show this help message and exit
  -v VERBOSITY, --verbosity VERBOSITY
  -k, --keep_data
```

Assumed that pqclean_saber_avx2 differs from upstream, the output will be as follows:

```
$ ./copy_from_upstream.py verify
-----
[...]

Issue #880, opened 2021-01-22
Update copy_from_pqclean


baentsch:
I'd support this proposal from Basil Hess and solicit further feedback to the below from @dstebila , @xvzcf : 

> Using kyber-upstream is now indeed a simple task thanks to what you already prepared.
> I really like the copy_from_pqclean mechanism. The META.yml definitions are obviously a pqclean-"invention", but they are a nice way to make an easy path for integrating also other implementations/libraries. Perhaps it could be renamed to copy_from_upstream.
> 
> I'm slightly updating the script with the following features. Please let me know if anything doesn't make sense here:
>
>- Adding support to checkout the upstream code from a git-repository: copy_from_pqclean.yml will contain a section "upstreams" with several git-repos (pqclean, pqcrystals-kyber, pqcrystals-dilithium, etc...), and the git-commit hash that will be used. This will replace the PQCLEAN_DIR env. variable. Instead, the git repo location will be used (this can be a github-url, a local path on the filesystem, etc.).
> - Adding a "verify" switch to the script: The goal is to be able to prove that the code used in liboqs origins from an upstream location with a specific commit. This will basically repeat the copy_from_pqclean process, hash of the code that is checked out, and compare it with the code in the liboqs src-directory. For someone using liboqs it should be useful to be able to reproduce and verify from where the crypto code origins.

dstebila:
That all sounds good to me!

xvzcf:
Same here.

jschanck:
I like the "verify" idea. Does liboqs maintain any patches on upstream code? If so then verify needs to apply those patches before the comparison.

If there are not any patches on upstream, and we intend to have a "verify" option that does not apply patches, then I'll have to take a different approach to silencing false positives in #881.

dstebila:
> I like the "verify" idea. Does liboqs maintain any patches on upstream code? If so then verify needs to apply those patches before the comparison.

For all the things imported via copy\_from\_pqclean, no, we do not add any patches.  (Things directly imported, like SIKE, Picnic, FrodoKEM, do make patches before being added.)

bhess:
I also thought about adding the possibility to apply patches before the comparison, if necessary. But if there is an alternative this might be cleaner.

For now the updated copy_from_pqclean script is on this branch https://github.com/bhess/liboqs/tree/feat.copy_from_upstream

The script is renamed to `copy_from_upstream` to reflect that other upstream sources are possible.
[...]

PR #890, opened 2021-02-02
Pull round 3 Dilithium from PQClean

jschanck:
This PR copies round 3 Dilithium from PQClean. An alternative is to wait for #883 to land and copy Dilithium directly from upstream. It's not clear to me how close #883 is to completion, hence this PR.

> Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)

Yes, there are significant changes between round 2 and round 3 Dilithium.

> Does this PR change the the list of algorithms available -- either adding or removing?  (If so, PRs in OQS-OpenSSL, OQS-BoringSSL, and OQS-OpenSSH will also be required by the time this is merged.)

Yes, it removes dilithium4, adds dilithium5, and adds the "_aes" variants of dilithium2, dilithium3, and dilithium5.


bhess:
Looks good to me, and a good alternative to accelerate having Dilithium round 3 in.

I believe #880 isn't far from completion though. There is also #891 to preview a version where #880 pulls Dilithium round 3 (and Kyber) from upstream.

baentsch:
> I believe #880 isn't far from completion though. 

As discussed, we're all agreed on this point. @jschanck Thus, we thank you very much for your work on this PR but prefer doing a direct "highest upstream" import in favour of a pull from PQClean/this PR. So, can we ask you to either close this PR or provide reasons why #890 should be prioritized over #880+#891?

jschanck:
My preference is also for pulling from the "highest upstream", but in most cases (including kyber and dilithium) we need to patch the upstream code before it will hook into our common code.

#891 imports redundant AES and fips202 code instead of doing this patching.

PQClean does the patching for reference implementations, but imports redundant AES and fips202 for platform specific implementations.

Neither is perfect.

I think we should extend copy_from_upstream with a patching mechanism, and deal with the common code issue. But I have no objection to merging #891 in the meantime, so I'll close this PR.

baentsch:
> I think we should extend copy_from_upstream with a patching mechanism, and deal with the common code issue.

Fully agree. Documented in #894

PR #902, opened 2021-02-09
Changing python platform check

### LINKED ISSUE #901: Portability tests on ubuntu-focal CI
Portability tests are skipped on the new ubuntu-focal CI, for example here:

https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/1218/workflows/50fa1837-6443-4ad6-aac7-fa8fe6d1716f/jobs/12325/parallel-runs/0/steps/0-104

The likely cause is

https://github.com/open-quantum-safe/liboqs/blob/1ead54fc9adb441b017d296ec5c01b1cb30dbbfe/tests/test_portability.py#L27

With python3 on Focal, the string "Ubuntu" is reported with `platform.version()` rather than `platform.platform()`

baentsch:
Fixes #901 

### DIFF
--- a/tests/test_portability.py
+++ b/tests/test_portability.py
@@ -9,9 +9,20 @@ MIN_CPUS = {}
 # set other CPU types for other architectures; Westmere supports cpuid but not avx2
 MIN_CPUS["x86_64"] = "Westmere"
 
+def check_ubuntu():
+   try:
+      with open("/etc/os-release") as f:
+         for line in f.readlines():
+            if "Ubuntu" in line:
+               return True
+   # if file /etc/os-release doesn't exist, we're definitely not on Ubuntu:
+   except:
+      return False
+   return False
+
 @helpers.filtered_test
 @pytest.mark.parametrize('kem_name', helpers.available_kems_by_name())
-@pytest.mark.skipif(not "Ubuntu" in platform.platform(), reason="Only supported on Ubuntu")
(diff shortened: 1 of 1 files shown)


tests/test_portability.py
```diff
@@ -9,22 +9,33 @@
 # set other CPU types for other architectures; Westmere supports cpuid but not avx2
 MIN_CPUS["x86_64"] = "Westmere"
 
+def check_ubuntu():
+   try:
+      with open("/etc/os-release") as f:
+         for line in f.readlines():
+            if "Ubuntu" in line:
+               return True
+   # if file /etc/os-release doesn't exist, we're not on Linux:
```
xvzcf, line 18: It seems like `os-release` is for Linux distributions that use `systemd` and can be (has been) adopted by other distributions. This would work to detect Ubuntu, but perhaps the comment is a bit inaccurate?

baentsch: ACK

xvzcf (approved):

dstebila (approved):

baentsch:
This is not really satisfactory: 
grafik

grafik

Even though this is just a simple comment-change, CI runs for 72mins and fails (3rd time). I also can't get myself to merge as CI doesn't pass. Yes, I know, #906 might be a work-around resolving the (not-really-existant) errors reported, but is it really reasonable that simple CI runs take more than 1 hour? Capturing also the discussions on this topic from our conversation I created #907 to resolve this for good.

Issue #888, opened 2021-02-01
Expert review needed for SIKE constant time warnings


jschanck:
Automated tests have identified a few instances of potentially non-constant time behaviour in SIKE. The tests produce a lot of false positives, so there's a good chance that these are benign, but I'm not expert enough to evaluate them. Help appreciated!

- Lines 590 and 993 of sidh_compressed.c invert an element of Fp^2 using a non-constant time binary gcd (fp2inv_mont_bingcd).
- Line 654 of ec_isogeny.c leaks the number of leading zeros in the "m" input to Ladder. This function is called from validate_ciphertext on line 1434 of sidh_compressed.c.

dstebila:
Tagging @bhess and @christianpaquin

christianpaquin:
I forwarded this to the SIKE library dev; waiting for review.

bhess:
Thank you for the report!

> * Lines 590 and 993 of sidh_compressed.c invert an element of Fp^2 using a non-constant time binary gcd (fp2inv_mont_bingcd).

This won't be an issue since the Fp^2 inversion is done on the public curve Montgomery coefficient.

> * Line 654 of ec_isogeny.c leaks the number of leading zeros in the "m" input to Ladder. This function is called from validate_ciphertext on line 1434 of sidh_compressed.c.

This is currently in review with the SIKE team.

jschanck:
Thanks for taking a look.

> > * Lines 590 and 993 of sidh_compressed.c invert an element of Fp^2 using a non-constant time binary gcd (fp2inv_mont_bingcd).
> 
> This won't be an issue since the Fp^2 inversion is done on the public curve Montgomery coefficient.

It's ever so slightly more subtle, no? I agree that it's safe to perform non-constant time operations on the public curve Montgomery coefficient "A". The issue here is that just before Line 590 we have A presented as (A24 : C24). Line 590 computes 1/C24 in variable time, and it's not obvious to me that it is safe to leak C24.

geovandro:
> Thanks for taking a look.
> 
> > > * Lines 590 and 993 of sidh_compressed.c invert an element of Fp^2 using a non-constant time binary gcd (fp2inv_mont_bingcd).
> > 
> > 
> > This won't be an issue since the Fp^2 inversion is done on the public curve Montgomery coefficient.
> 
[...]

PR #904, opened 2021-02-09
Use valgrind to check for secret-dependent branching on 'audit' branches

jschanck:
This PR should run the full set of constant-time checks on #881.


dstebila (approved):

dstebila:
Thanks John! To make sure I understand: The "issues" files log places that don't pass Valgrind's checks.  For each of these we've checked with the authors that they believe these are not problematic from a security perspective.  But an ideal version would have no issues files.

baentsch (approved):
LGTM. Should `OQS_ENABLE_TEST_CONSTANT_TIME` also be documented in https://github.com/open-quantum-safe/liboqs/wiki/Customizing-liboqs ? Leave it to you -- maybe too much documentation also confuses too many people :-)

jschanck:
> Thanks John! To make sure I understand: The "issues" files log places that don't pass Valgrind's checks. For each of these we've checked with the authors that they believe these are not problematic from a security perspective. But an ideal version would have no issues files.

Almost. The "issues" files log errors that are either awaiting review or are known to be actual problems. Once we have review of an "issue" it should be moved to "passes".

Here's the current status.

The open issues for SIDH/SIKE are:
* sidh_vartime_bingcd_final_C24 
* sike_vartime_bingcd_final_C24
* sike_compressed_verify_ladder
* sike_memcmp

The first two can be moved to "passes" based on reviews from #888. The second two will be fixed by #914. I'll resolve all four issues once #914 merges.

Bike and HQC have genuine non-constant time behaviour, documented by:
* bike_has_no_timing_protections
* hqc_fast_convolution_mult
* hqc_gf_mul
* hqc_vect_set_random_fixed_weight

And there's one falcon issue that can be fixed by pulling from pqclean.
* falcon_hash_to_point_vartime

jschanck:
BIKE failed on the last run---rather than try to whitelist the entire implementation I'm just disabling test_constant_time for BIKE.

dstebila:
Thanks John. I made #917 to track updating Falcon (and others) from PQClean.

baentsch:
> BIKE failed on the last run---rather than try to whitelist the entire implementation I'm just disabling test_constant_time for BIKE.

If we decide to merge in this state to move forward, may I suggest to create an issue to track such test-disablements (and/or document them in the algorithm-review-yml if we don't get this fixed before a release) to not forget this?

jschanck:
This is failing with a new type of timeout.

Options:
* Move this out of CI.
* Split into kem and sig jobs that run in separate containers.
* Skip some of the heavy schemes.

I'm happy to wait until after 0.5.0.

baentsch:
FWIW, I'd vote for 

> Skip some of the heavy schemes.

and merge now (plus create an issue to fix this long-term -- I'd suggest outside of CI, e.g., using a github action on release).

jschanck:
This is ready to go.

dstebila (approved):

PR #906, opened 2021-02-10
Potential work-around for ongoing SIGKILL issues

jschanck:
Some recent CI jobs have failed with
```
AssertionError: Got unexpected return code -9
```
and a backtrace pointing to:
```python
    try:
        out, _ = proc.communicate(input=input, timeout=480)
    except subprocess.TimeoutExpired:
        proc.kill()
        out, _ = proc.communicate()

    if not(ignore_returncode) and (proc.returncode != expected_returncode):
        print(out.decode('utf-8'))
        assert False, "Got unexpected return code {}".format(proc.returncode)
```

I see two possibilites:
1. CircleCI actually sent SIGKILL to our subprocess.
2. We hit our self-imposed timeout. The -9 return code came from our own call to proc.kill().

This PR adds logic to retry a subprocess if we're running in CircleCI and our subprocess receives a SIGKILL. It's harmless to retry, since CircleCI will kill our python process if it really wants us to shut down.

This PR also disambiguates the above possibilities by raising a more informative assertion error in the timeout case. 

```python
    try:
        out, _ = proc.communicate(input=input, timeout=480)
    except subprocess.TimeoutExpired as T:
        proc.kill()
        assert False, str(T)
```
Note: The python documentation suggests that the proc.kill() is unnecessary. I've left it in case it's there from experience with a race condition or something.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -17,14 +17,14 @@ def run_subprocess(command, working_dir='.', env=None, expected_returncode=0, in
     Helper function to run a shell command and report success/failure
     depending on the exit status of the shell command.
     """
+    env_ = os.environ.copy()
     if env is not None:
-        env_ = os.environ.copy()
         env_.update(env)
-        env = env_
+    env = env_
 
     # Detect if we're running in CircleCI.
     # https://circleci.com/docs/2.0/env-vars/
-    circleci = env is not None and 'CIRCLECI' in env
+    circleci = 'CIRCLECI' in env
 
     # CircleCI has its own timeout mechanism. Disable ours.
     timeout = None if circleci else 10*60
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -26,25 +26,32 @@ def run_subprocess(command, working_dir='.', env=None, expected_returncode=0, in
     # then print it, which pytest will then capture and
     # buffer appropriately
     print(working_dir + " > " + " ".join(command))
-    proc = subprocess.Popen(
-        command,
-        stdin=subprocess.PIPE,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.STDOUT,
-        cwd=working_dir,
-        env=env,
-    )
 
-    try:
-        out, _ = proc.communicate(input=input, timeout=480)
-    except subprocess.TimeoutExpired:
-        proc.kill()
-        out, _ = proc.communicate()
+    while 1: # We might need to re-run the command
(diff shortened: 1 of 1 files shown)


jschanck:
Seems that the recent address sanitizer failures were due to timeouts, not SIGKILL from CircleCI.

tests/helpers.py
```diff
@@ -17,34 +17,48 @@ def run_subprocess(command, working_dir='.', env=None, expected_returncode=0, in
...
+        try:
+            out, _ = proc.communicate(input=input, timeout=timeout)
+        except subprocess.TimeoutExpired as T:
+            proc.kill()
+            assert False, str(T)
+
+        # CircleCI is welcome to kill this python process, otherwise
+        # we'll keep trying the subprocess.
+        if circleci and proc.returncode == -9:
+            continue
```
baentsch, line 56: Given this is an endless loop, would it make sense to do an output if we hit this condition such that we're a bit more informed what's going on when reading the logs?

tests/helpers.py
```diff
@@ -17,34 +17,48 @@ def run_subprocess(command, working_dir='.', env=None, expected_returncode=0, in
...
         env_.update(env)
-        env = env_
+    env = env_
+
+    # Detect if we're running in CircleCI.
+    # https://circleci.com/docs/2.0/env-vars/
+    circleci = 'CIRCLECI' in env
+
+    # CircleCI has its own timeout mechanism. Disable ours.
+    timeout = None if circleci else 10*60
```
baentsch, line 30: Should we really have/permit sub-tests taking 10 minutes?

tests/helpers.py
jschanck: The infinite loop is not necessary at all, since the test runs of this PR revealed that CircleCI was not sending the SIGKILLs. I'll remove it.

tests/helpers.py
jschanck: Frankly we shouldn't be using the subprocess module's timeout mechanism at all. We have a clear problem with CircleCI sending jobs to the background for long enough that the timeout expires, which is why this PR disables the timeout in CircleCI. But other test environments are also likely to send jobs to the background.

@xvzcf what was the reason for switching from subprocess.run to Popen + communicate in dd739fc8

baentsch: This rings a bell, was it https://github.com/open-quantum-safe/liboqs/issues/852 ?

PR #908, opened 2021-02-11
Switch back to subprocess.run

jschanck:
This PR switches our helpers.run_subprocess function back to using Python's `subprocess.run`. It also includes a pure python implementation of test_free which serves as an alternative fix for #852. The previous fix for #852 in commit 346305fc added a timeout to our subprocess calls which we've been hitting in CI.

The timeout was a less-than-ideal solution for killing long running jobs since it is based on wall-clock time.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -17,34 +17,29 @@ def run_subprocess(command, working_dir='.', env=None, expected_returncode=0, in
     Helper function to run a shell command and report success/failure
     depending on the exit status of the shell command.
     """
+    env_ = os.environ.copy()
     if env is not None:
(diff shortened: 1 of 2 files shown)


baentsch (approved):
Replacing the process-heavy & deadlock-prone `test_free` implementation with a pure Python one is a very good (and more efficient) solution for the original problem in #852. I'm then also in favour of reverting `run_subprocess` -- keeping in mind (maybe as a comment?) that it shouldn't be used for heavily parallel tasks as per the [subprocess](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) documentation.

PR #913, opened 2021-02-12
First attempt at a CODEOWNERS file [skip ci]

dstebila:
Activates Github's [code owners feature](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) to help direct code reviews.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,7 @@
 # https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
 
 * @dstebila @xvzcf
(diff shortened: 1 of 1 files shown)


baentsch (commented):
If order matters(?) it should be for `copy_from_upstream` @bhess @baentsch @xvzcf ; I'd also suggest adding `.circleci` tagging @jschanck @baentsch @xvzcf . Rest LGTM.

xvzcf:
Agree with @baentsch; you could also add me to `src/CMakeLists.txt` and `src/oqsconfig.h.cmake`, as well as specify a wildcard to add me to all `CMakeLists.txt` files.

PR #915, opened 2021-02-15
Support OpenSSL on Windows

thomwiggers:
Attempts to add support for OpenSSL on Windows. We probably should add this to the CI matrix as well.

See also https://github.com/open-quantum-safe/liboqs-rust/issues/17


thomwiggers:
It might be better to not include `rand_nist` in `common` anyway, if it's only necessary for tests.

thomwiggers:
AFAICT it works with openssl with this change.

xvzcf:
Nice. We should probably also remove/rework the Windows guard [here](https://github.com/open-quantum-safe/liboqs/blob/main/.CMake/alg_support.cmake#L12).

dstebila (approved):

dstebila:
This is my first time dealing with testapproval.  Should we run the tests being held?  Who would trigger that?  Or can we decide it is not necessary for this PR (since the remaining tests are arm tests, unaffected by this PR) and just merge as is?

baentsch:
> Who would trigger that? Or can we decide it is not necessary for this PR (since the remaining tests are arm tests, unaffected by this PR) and just merge as is?

I'd suggest we run these tests once to (have the system) ascertain that indeed this doesn't impact the other platforms. As all else passed I just triggered those runs. We then have a clean decision base to approve/merge. I personally would make the actual merge decision dependent on whether @xvzcf's comment 

> Nice. We should probably also remove/rework the Windows guard [here](https://github.com/open-quantum-safe/liboqs/blob/952c628051f2b88aee02952abc46be0ac8f09ab7/.CMake/alg_support.cmake#L12-L23).

is addressed in this PR or via creation of a new issue.

thomwiggers:
WRT the Windows guards, I'm not exactly sure what's going on there.

thomwiggers:
From the meeting minutes

> Note that this should also allow us to build with BIKE support on Windows.

So probably this PR should also go into that.

thomwiggers:
Why was BIKE even disabled on Win32 if that's just because of OpenSSL, because it works fine without OpenSSL on Linux and Mac...

xvzcf:
> Why was BIKE even disabled on Win32 if that's just because of OpenSSL, because it works fine without OpenSSL on Linux and Mac...

Yeah, I went back to my original CMake commit to jog my memory, and it turns out I misspoke in the meeting. I disabled BIKE on Windows because it was not listed as being supported [here](https://github.com/open-quantum-safe/liboqs/blob/fb458165698c5850dcf08d7268a2b6c90e47e378/VisualStudio/winconfig.h#L4), not because of OpenSSL; and from quickly skimming through the PR's most recent AppVeyor logs, it seems a bit more work might be required to get BIKE working on Windows.

thomwiggers:
The Windows guard was changes so that OpenSSL is disabled by default on Windows and enabled by default otherwise.

BIKE is unavailable on Windows now, and enabled by default otherwise using `cmake_dependent_option`:
```cmake
 cmake_dependent_option(OQS_ENABLE_KEM_BIKE "" ON "NOT WIN32" OFF)
 ```

xvzcf (approved):

Issue #791, opened 2020-07-15
Undefined behaviour in KeccakP-1600-times4-SIMD256


dstebila:
When enabling clang's UndefinedBehaviorSanitizer using the build sequence
```
mkdir build && cd build
cmake -DCMAKE_C_COMPILER=clang-9 -DCMAKE_BUILD_TYPE=Debug -DUSE_SANITIZER=Undefined -GNinja ..
ninja
```
there are several undefined behaviours in many algorithms, related to the KeccakP-1600-times4-SIMD256 operation.  I can't tell at this point whether the undefined behaviour is present in the Keccak code package or in how code in liboqs/implementations are calling it.  Here's an example:

```
tests/test_kem Kyber512
Configuration info
==================
Target platform:  x86_64-Linux-4.19.76-linuxkit
Compiler:         clang (9.0.0 (tags/RELEASE_900/final))
OQS version:      0.3.1-dev
Git commit:       53937db0c2c5e40afc50c4a8be2362379ca407de
OpenSSL enabled:  Yes (OpenSSL 1.1.1  11 Sep 2018)
AES:              OpenSSL
SHA-2:            OpenSSL
SHA-3:            OpenSSL

================================================================================
Sample computation for KEM Kyber512
================================================================================
/usr/lib/llvm-9/lib/clang/9.0.0/include/avxintrin.h:3145:10: runtime error: load of misaligned address 0x000001220830 for type 'const __m256i' (vector of 4 'long long' values), which requires 32 byte alignment
0x000001220830: note: pointer points here
 00 00 00 00  07 00 01 02 03 04 05 06  0f 08 09 0a 0b 0c 0d 0e  17 10 11 12 13 14 15 16  1f 18 19 1a
              ^ 
UndefinedBehaviorSanitizer:DEADLYSIGNAL
==5033==ERROR: UndefinedBehaviorSanitizer: SEGV on unknown address 0x000000000000 (pc 0x000000bc2081 bp 0x7fa205e65fb0 sp 0x7fa205e1d640 T5034)
==5033==The signal is caused by a READ memory access.
==5033==Hint: address points to the zero page.
    #0 0xbc2080 in KeccakP1600times4_PermuteAll_24rounds /root/liboqs/build/../src/common/sha3/keccak4x/KeccakP-1600-times4-SIMD256.c
[...]

Issue #502, opened 2019-06-27
Use OpenSSL SHA3 for incremental API


dstebila:

dstebila:
May also want to switch SHA3 implementation to PQClean's, since OQS's is missing several things, including SHA3-384.

dstebila:
#517 started to use OpenSSL for some SHA-3 functions, but only the non-incremental API.  This issue is now about using OpenSSL for the SHA-3 incremental API.  Note that this may need a change in the PQClean SHA-3 API to use an opaque context object.

dstebila:
Closing due to #922.

PR #891, opened 2021-02-03
Sync upstream with copy_from_upstram / pqcrystals pull

bhess:
This PR is to show the changes after copy_from_upstream with #880.

It will also be an alternative to #890.

With this configuration, it is pulling the latest Kyber round 3 and Dilithium round 3 from pqcrystals-upstream.

The updates can be reproduced like this:

```
$ ./copy_from_upstream copy
```

To verify that the synced version matches with upstream versions:

```
$ ./copy_from_upstream.py verify
-----
Total schemes: 151 - 151 match with upstream, 0 differ
-----
```

Note that this PR might have to be rebased after #880 is merged.

* [x] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)

Changes to Dilithium Round 3 KAT. Adds Dilithium Round 3 algorithms (2, 3, 5, 2-AES, 3-AES, 5-AES)

* [x] Does this PR change the the list of algorithms available -- either adding or removing?  (If so, PRs in OQS-OpenSSL, OQS-BoringSSL, and OQS-OpenSSH will also be required by the time this is merged.)

PR #922, opened 2021-02-22
Refactor SHA3 implementations and improve API

jschanck:
This is an attempt to restructure our sha3 api to meet the needs of the remaining schemes and avoid code duplication.

Changes:
* Added a complete four-way parallel shake interface made public through sha3x4.h.  This interface does not require AVX2 instructions. It accepts any implementation of "Keccak1600P-times4" from the XKCP low-level interface, and it quietly falls back to a serial implementation using whatever (possibly platform optimized) implementation of Keccak1600P is available.
* Scripted the importing/patching of implementations of the low-level XKCP interface. (scripts will be submitted in a followup commit or pr). The implementations are made available in src/common/sha3/xkcp_low and a suitable implementation is chosen at compile time by src/common/sha3/xkcp_low/CMakeLists.txt.
* Replaced fips202.c, sha3_c.c, and sha3_avx.c with xkcp_sha3.c, which implements sha3.h using the low-level XKCP interface.
* Replaced sha3_x4.c, and sha3x4_avx2.c with xkcp_sha3x4.c, which implements sha3x4.h using the low-level XKCP interface.
* Added ossl_sha3.c and ossl_sha3x4.c which implement sha3.h and sha3x4, including the incremental APIs (#502), using openssl.
* Removed (unused) cshake functions.
* Removed non-incremental shake functions like OQS_SHA3_shake128_absorb, but keep simple shake functions like OQS_SHA3_shake128.
* Added "sha3_*_ctx_reset" which re-initializes an already allocated context.
* Changed OQS_SHA3_sha3_*_inc_finalize so that it does not free the context. This allows you to re-use a context with "reset".
* Removed Picnic's duplicate sha3 code

Issues:
* OpenSSL does not implement a proper "squeeze" function. There's work in progress to fix this (https://github.com/openssl/openssl/pull/7921), but we should not expect it to be available anytime soon. The implementation of the incremental API that I've written is quadratic in the total output length if O(1) bytes are squeezed at a time, e.g. one byte per squeeze, one block per squeeze, etc. Some care should be taken to minimize the number of sequential squeeze calls in schemes.

TODO:
* Import optimized Keccak routines for other architectures and select appropriately at compile time. There are several ARM implementations available. There's also AVX512.
* Shims for pqclean and remove unnecessary fips202x4 code, etc, from Kyber, Dilithium, and SPHINCS.


christianpaquin:
I asked the Picnic team to review this PR.

dstebila:
I'm a little worried about the patching mechanism and the size of the patches.  Suppose there are changes to the upstream implementations that invalidate the patches.  How would we regenerate patches in the future if need be?  

And perhaps to ask a question before that: why were these patches needed in the first place?  It looks like the main reason is to change something to do with the hash context object allocation/releasing.  Is there something about OQS's architecture that is so different from PQClean's that we want to manage hash context objects different?

baentsch:
>  Suppose there are changes to the upstream implementations that invalidate the patches. How would we regenerate patches in the future if need be?

I'd echo this concern. Wouldn't it be better to provide the SHA3 API (and prove its value for one implementation) and then let the upstreams use/cater to it (and then just pick that code up) - saving us the hassle/responsibility (to keep patching the implementations)? Should the upstreams not pick up the API wouldn't we then not know it's not quite right?

jschanck:
> How would we regenerate patches in the future if need be?

I wrote documentation for the patching mechanism, but didn't include it with the PR. Here's a copy.
https://gist.github.com/jschanck/c67d37551da2a5f6e06c640816d2c99f

> And perhaps to ask a question before that: why were these patches needed in the first place? It looks like the main reason is to change something to do with the hash context object allocation/releasing. 

Upstream Kyber and Dilithium allocate their shake contexts on the stack. We use the heap. So we have to insert some additional calls to free contexts.

> Is there something about OQS's architecture that is so different from PQClean's that we want to manage hash context objects different?

When we were pulling from PQClean we could make it all work with shims. But we don't pull from PQClean anymore. Even if we did, it only worked because PQClean already applies these patches (https://github.com/jschanck/package-pqclean/tree/main/kyber/patches). If we're going to pull directly from upstream we need to apply them ourselves.

baentsch:
> Upstream Kyber and Dilithium allocate their shake contexts on the stack. We use the heap. So we have to insert some additional calls to free contexts.

Wouldn't there be a chance that the pqcrystals team would be willing to add a "free_context(&state)" call to their uses of the SHA3 API? For stack-based implementations it could simply be a NOP. It would mean their code could easily benefit from downstream shared, possibly better vetted, SHA3 code without being subject to lots of patching (which always carries risks if upstream code changes). Such downstream uptake should be attractive to the upstream team(s), no?

In addition, shouldn't any upstream be interested in not having to maintain their own --optimized-- SHA3 code as it conceptually has nothing to do with the core of the PQC competition? What are all other implementations doing that need fast SHA3? Do they all "roll their own"? To make the core PQC code comparable, shared SHA3 code (also for the optimized case) would be generally beneficial, no? 

> If we're going to pull directly from upstream we need to apply them ourselves.

Should we then maybe reconsider this decision if it otherwise seems to cause us to take responsibility for "correct patching"? 

If we don't want to revert this decision we should make an informed decision whether we want to a) simply copy code properly maintained upstream or b) patching such code to integrate our own library into such code. Personally, option b) causes me headaches (of increased "responsibility" for proper operation of algorithms we didn't author and that can and will change without our influence).

Thinking all options through, I've got to ask what is our key motivation to create and integrate this optimized SHA3 code into upstream code bases? Do we only want to save on object code size? Do we want to enable performance comparisons of the core PQC logic (not of things like SHA3)? Do we want to have a more robust code base for shared code?

jschanck:
This discussion seems like it might spiral out of control....

So I'm splitting this PR into the SHA3 refactoring (this PR #922), the copy from upstream patch mechanism (#944), and the patches themselves (which I'll hold locally for now).

@baentsch I found it hard to follow your comment. I understand that you have concerns about the patching mechanism. Could you please separate those concerns from the core question of whether we want to provide a SHA3 interface, like what this PR provides, that eliminates the need for (say) Picnic to bring their own implementations?

mouse07410:
> @baentsch I found it hard to follow your comment. I understand that you have concerns about the patching mechanism. Could you please separate those concerns from the core question of whether we want to provide a SHA3 interface, like what this PR provides, that eliminates the need for (say) Picnic to bring their own implementations?

Let me offer my (possibly naïve) view. 
1. I think it would be beneficial for this library to offer its own optimized and portable versions of SHA3.
2. I think it makes sense to approach the upstream maintainers and offer them `liboqs` API to SHA3 (this API probably won't differ much from what they have anyway).
3. I doubt `liboqs` maintainers want to take upon themselves any more patching of the upstream code than absolutely necessary. That means, we probably don't want to patch SHA3 for the upstream repos - potentially large pain for likely small gain.

So, to answer @jschanck questions (@baentsch can correct where our views diverge):
1. Yes, it would be good for `liboqs` to provide its own SHA3 interface.
2. It does not look attractive (aka, manageable) to patch SHA3 implementations of the upstream vendors.

baentsch:
> @baentsch I found it hard to follow your comment. I understand that you have concerns about the patching mechanism. Could you please separate those concerns from the core question of whether we want to provide a SHA3 interface, like what this PR provides, that eliminates the need for (say) Picnic to bring their own implementations?

I (hopefully) [did reformulate my concerns in a more understandable way in the discussion for #944](https://github.com/open-quantum-safe/liboqs/pull/944#issuecomment-799993138)

christianpaquin:
Testing this. Is there a way to run the tests without qemu (@xvzcf? unrelated to this PR, I get qemu failures because it's not installed, and we don't document how to install it in the readme).

jschanck:
@christianpaquin If you're running tests for a performance comparison you should compile with OQS_PORTABLE_BUILD=OFF, which has the side effect of disabling the test that uses qemu.

dstebila (changes requested):
I noticed that xkcp_low does not have SPDX license identification headers.  Is this so that those files can be copied from xkcp without changes?  If so, that's fine, but then the license section of README.md should be updated to indicate clearly the license(s) that apply to this directory.

dstebila:
As for merging, should I merge to an audit branch first to get the full test suite?

jschanck:
> I noticed that xkcp_low does not have SPDX license identification headers. Is this so that those files can be copied from xkcp without changes? If so, that's fine, but then the license section of README.md should be updated to indicate clearly the license(s) that apply to this directory.

I updated the README. We don't copy directly from xkcp because we need to patch a bunch of undefined behaviour. I've added the scripts that I use to do this patching.

> As for merging, should I merge to an audit branch first to get the full test suite?
Not worth it. Right now we're only running address sanitizer on audit branches. That's not going to turn up anything new. I could re-enable UBSan, but that's disabled because of #791. This PR no longer includes the patches for Kyber so it no longer fixes #791, so we already know UBSan will fail.

baentsch:
> Not worth it. Right now we're only running address sanitizer on audit branches. That's not going to turn up anything new. I could re-enable UBSan, but that's disabled because of #791. This PR no longer includes the patches for Kyber so it no longer fixes #791, so we already know UBSan will fail.

Wasn't (the by now merged) #891 meant to provide a fix for #791 ? Did I [misunderstand this statement](https://github.com/open-quantum-safe/liboqs/issues/791#issuecomment-772431186)?

jschanck:
> Wasn't (the by now merged) #891 meant to provide a fix for #791 ? Did I [misunderstand this statement](https://github.com/open-quantum-safe/liboqs/issues/791#issuecomment-772431186)?

Ah right, #891 "fixed" #791 because some global variables got properly aligned by chance. This "fix" could break with a compiler update, but #791 is no longer reproducible.

So we could re-enable UBSan and it might pass.

dstebila (approved):

jschanck:
I pushed to audit.sha3 and circleci started a run. Results will be here: https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/1341/workflows/7ff9763a-030a-4521-8159-30f81fc92e6e

jschanck:
Address sanitizer passed on audit branch. UBSan failed because of unrelated (and recently fixed https://github.com/microsoft/PQCrypto-SIDH/commit/67981447d36dfdcfc849d8d88963f727a31ba772) UB in SIKE.

PR #927, opened 2021-03-01
Fix OQS_PORTABLE_BUILD logic for Frodo

jschanck:
The Frodo code currently ignores OQS_PORTABLE_BUILD if it is compiled on a machine without AVX2 instructions.

https://github.com/open-quantum-safe/liboqs/blob/5b2ed97f08ba94c59ea2769b5e52760accc51f44/src/kem/frodokem/external/frodo_macrify_optimized.c#L27-L40

My understanding of OQS_PORTABLE_BUILD is that a less powerful machine should be able to compile a library that takes advantage of the features of a more powerful machine (without fixing the target machine at compile time). This PR brings the Frodo code in line with my expectations.

There are other places where the OQS_PORTABLE_BUILD logic could be improved. E.g we ignore OQS_PORTABLE_BUILD in alg_support.cmake while setting the value of the "OQS_ENABLE_X" variables. I'll open separate issues or PRs for those.


src/common/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ if(OQS_USE_AES_OPENSSL)
     set(AES_IMPL aes/aes_ossl.c)
 else()
    set(AES_IMPL aes/aes.c aes/aes_c.c)
-    if (OQS_USE_AES_INSTRUCTIONS)
+   if (OQS_PORTABLE_X86_64_BUILD OR OQS_USE_AES_INSTRUCTIONS)
```
baentsch, line 17: Wouldn't it be sufficient (even more correct) to add `ARCH_X86_64 AND  OQS_USE_AES_INSTRUCTIONS`? Presence/absence of this code should be independent of OQS_PORTABLE_BUILD.

src/kem/frodokem/CMakeLists.txt
```diff
@@ -22,19 +24,20 @@ if(OQS_ENABLE_KEM_frodokem_1344_shake)
...
-            set(SRCS_AVX2 external/frodo640aes_avx2.c)
-        endif()
-        if(OQS_ENABLE_KEM_frodokem_976_aes)
-            set(SRCS_AVX2 ${SRCS_AVX2} external/frodo976aes_avx2.c)
-        endif()
-        if(OQS_ENABLE_KEM_frodokem_1344_aes)
-            set(SRCS_AVX2 ${SRCS_AVX2} external/frodo1344aes_avx2.c)
-        endif()
+set(SRCS_AVX2 "")
+if(OQS_PORTABLE_X86_64_BUILD OR OQS_USE_AVX2_INSTRUCTIONS)
```
baentsch, line 30: Same thing here: Why the dependency on OQS_PORTABLE_BUILD? Dependency on architecture is understood, but code presence IMO should otherwise (only) depend on OQS_USE_CPU_EXTENSIONS. @xvzcf : What's your take?

jschanck: No, OQS_USE_CPU_EXTENSIONS is not the right flag here.

The first thing we do with OQS_USE_CPU_EXTENSIONS is compile a program to test for the CPU features of the machine that we're compiling on (in .CMake/gcc_clang_intrinsics.cmake). We then set OQS_USE_AVX2_INSTRUCTIONS based on the output of that program.

The OQS_PORTABLE_BUILD flag enables runtime detection of CPU features and call dispatching. When it is enabled, we have to include this code whether or not the _compiling_ machine has AVX2 instructions. When it's not enabled, we include the code only if OQS_USE_AVX2_INSTRUCTIONS is set.

baentsch: I do not suggest adding OQS_USE_CPU_EXTENSIONS here -- that must have been set for OQS_USE_AVX2_INSTRUCTIONS even to be(come) active, right? I do propose checking for ARCH_X86_64 _and_ OQS_USE_AVX2_INSTRUCTIONS (independent of OQS_PORTABLE_BUILD, thus independent of OQS_PORTABLE_X86_64_BUILD).

jschanck: OK. I see your point w.r.t. OQS_USE_AES_INSTRUCTIONS, but not for OQS_USE_AVX2_INSTRUCTIONS.

The name "AVX2" refers to an X86 extension, so checking ARCH_X86_64 is redundant---we will never set OQS_USE_AVX2_INSTRUCTIONS if the architecture is not X86_64.

For AES I think it would be better to rename the flag "OQS_USE_AES_NI_INSTRUCTIONS", as "AES-NI" unambiguously refers to the X86 extension. If we ever implement a compile-time check for ARM CPU features we should use a different name like "OQS_USE_ARMV8_AES_INSTRUCTIONS".

For this PR, I think it's fine for "OQS_USE_AES_INSTRUCTIONS" to refer to AES-NI for now.

dstebila (approved):

PR #928, opened 2021-03-02
Increasing message counter during sign-benchmark

bhess:
The purpose of this PR is to better capture message-dependencies in the sign-benchmarks. Currently, the message is selected at random once and is then fixed during the benchmark run. If the runtime is message-dependent, such as in deterministic Dilithium or Rainbow, the runtimes vary across benchmark runs.

To address this issue, this code change varies the message in the benchmark-loop. This results in stable total runtimes, but captures the variance in the "std. dev." column of the benchmark. To avoid the overhead of generating a new random message after each sign-iteration, a message counter is increased inside the message. This is sufficient for signatures that follow a hash-then-sign paradigm.

### DIFF
--- a/tests/speed_sig.c
+++ b/tests/speed_sig.c
@@ -21,6 +21,7 @@ static OQS_STATUS sig_speed_wrapper(const char *method_name, uint64_t duration, 
 	uint8_t *secret_key = NULL;
 	uint8_t *message = NULL;
 	uint8_t *signature = NULL;
+	size_t *message_ctr = NULL;
 	size_t message_len = 50;
 	size_t signature_len = 0;
 	OQS_STATUS ret = OQS_ERROR;
@@ -40,11 +41,13 @@ static OQS_STATUS sig_speed_wrapper(const char *method_name, uint64_t duration, 
 		goto err;
 	}
 
+	message_ctr = (size_t *)message;
+
 	OQS_randombytes(message, message_len);
 
 	printf("%-30s | %10s | %14s | %15s | %10s | %25s | %10s\n", sig->method_name, "", "", "", "", "", "");
 	TIME_OPERATION_SECONDS(OQS_SIG_keypair(sig, public_key, secret_key), "keypair", duration)
-	TIME_OPERATION_SECONDS(OQS_SIG_sign(sig, signature, &signature_len, message, message_len, secret_key), "sign", duration)
+	TIME_OPERATION_SECONDS(OQS_SIG_sign(sig, signature, &signature_len, message, message_len, secret_key); (*message_ctr)++, "sign", duration)
 	TIME_OPERATION_SECONDS(OQS_SIG_verify(sig, message, message_len, signature, signature_len, public_key), "verify", duration)
 
 	if (printInfo) {


baentsch:
Allow me to repeat one more time publicly what I kept saying in personal communication now for more than a year: I think it is not a good idea to change a test to allow single algorithms vulnerable to side channel attacks (in this case: That have a runtime strongly dependent on the message) to hide/let this property go unnoticed when being tested. 

Actually, this test was the reason that unearthed (at least for me and what I still perceive as) a practical cryptographic weakness of the original Dilithium code base. This has been resolved via the now available randomized signing variant which clearly does not need this change as visible here (who can spot the day of merging Dilithium 3.1 without deterministic signing :-)?
grafik

I thus won't approve this PR. To the contrary, I'd even propose creating more tests checking for data-dependencies of these algorithms: "Classic" crypto has gone through many pains of fixing such issues, sometimes only after attackers exploited such implementations. Wouldn't it be a reasonable goal for our project to try to expose/help fix such problems before exploitation?

bhess:
> Allow me to repeat one more time publicly what I kept saying in personal communication now for more than a year: I think it is not a good idea to change a test to allow single algorithms vulnerable to side channel attacks (in this case: That have a runtime strongly dependent on the message) to hide/let this property go unnoticed when being tested.

Capturing the variance is exactly what this PR aims to achieve. Consider a benchmark with (deterministic) Dilithium before this PR:

```
Speed test
==========
Operation                      | Iterations | Total time (s) | Time (us): mean | pop. stdev | CPU cycles: mean          | pop. stdev
------------------------------ | ----------:| --------------:| ---------------:| ----------:| -------------------------:| ----------:
Dilithium3                     |            |                |                 |            |                           |           
sign                           |      24926 |          3.000 |         120.360 |      9.955 |                    311817 |      25680
```

And after the PR:
```
Speed test
==========
Operation                      | Iterations | Total time (s) | Time (us): mean | pop. stdev | CPU cycles: mean          | pop. stdev
------------------------------ | ----------:| --------------:| ---------------:| ----------:| -------------------------:| ----------:
Dilithium3                     |            |                |                 |            |                           |           
sign                           |      17937 |          3.000 |         167.257 |    106.252 |                    433349 |     275380
```

Notice the difference in "pop. stdev", which can be interpreted without the need for a time series.

> I'd even propose creating more tests checking for data-dependencies of these algorithms: "Classic" crypto has gone through many pains of fixing such issues, sometimes only after attackers exploited such implementations. Wouldn't it be a reasonable goal for our project to try to expose/help fix such problems before exploitation?

Creating an extra test is a good idea. Message-dependency is something to be aware of (and the Dilithium specification properly documents this for its deterministic version).
But in a benchmark I'd expect the timing mean plus the std. dev. to get a picture how it behaves in practice.

dstebila:
I accept that we shouldn't change test suites to avoid exposing data-dependent behaviour.  But that specific concern for Dilithium has been put to bed by #923.  

So for me the question is now whether a signature timing benchmark (for messages of a fixed length) makes more sense to be run on a fixed non-random message of that length (which we're not doing), a random-once-and-then-fixed-after-that message of that length (which we are currently doing), a distinct-each-time message of that length (which is what this PR would do), or another option.

I would say it's a separate issue of whether we think we should have tests that check for message-dependent signing behaviour. Doing something via Vagrind/TIMECOP à la #881 is not immediately straightforward, as I understand it, since we can reasonably expect all signature schemes to accept variable length messages and thus at least depend on the length of the message being signed.

baentsch:
> I accept that we shouldn't change test suites to avoid exposing data-dependent behaviour. 

Thanks! In addition, changing this code here now would make all performance data points collected over the past months incomparable with new ones as indicated by

> And after the PR:
> 
> Speed test
> ==========
> Operation                      | Iterations | Total time (s) | Time (us): mean | pop. stdev | CPU cycles: mean          | pop. stdev
> ------------------------------ | ----------:| --------------:| ---------------:| ----------:| -------------------------:| ----------:
> Dilithium3                     |            |                |                 |            |                           |           
> sign                           |      17937 |          3.000 |         167.257 |    106.252 |                    433349 |     275380


I also agree with 

> we can reasonably expect all signature schemes to accept variable length messages and thus at least depend on the length of the message being signed

However, isn't this a direct consequence of the NIST APIs? If so, pretty hard for us to change given we build on those APIs.

Using different types of contents of (same-length) messages (fixed, fully random, somewhat changing as per this PR) may be interesting. But then again, the IBM team asked a reasonable question we may want to answer first:

> it would be helpful for the  OQS team to clarify what the purpose of the tests are:
> 
> - Comparing algorithms vs Comparing algorithm evolution
> - Testing algorithm runtime variation for a given message vs testing algorithms runtime over a distribution of messages

-> Is it possible this whole discussion really is one about what tests should visualize rather than what this particular test does? So maybe better to create an issue in https://github.com/open-quantum-safe/profiling and work on a (re)solution together there rather than changing this test and then forgetting about it? Reason:

> Capturing the variance is exactly what this PR aims to achieve. 

seems to indicate interest to contribute generating new numbers (and provide answers to the "goal questions") -- both topics of interest in the profiling subproject: What about we continue this conversation there? See https://github.com/open-quantum-safe/profiling/issues/40

bhess:
> So for me the question is now whether a signature timing benchmark (for messages of a fixed length) makes more sense to be run on a fixed non-random message of that length (which we're not doing), a random-once-and-then-fixed-after-that message of that length (which we are currently doing), a distinct-each-time message of that length (which is what this PR would do), or another option.

I think this is really the question that it boils down to. Shouldn’t a benchmark give a “realistic” picture of what one can expect from the performance of an algorithm (with metrics such as mean, standard deviation, perhaps its distribution)? 
I’d expect that in practice, one signs distinct messages. Message lengths will also be variable. But if all algorithms first internally hash the messages, the lengths could as well be kept fixed in the benchmark.
What about having one "default" benchmark (that the user runs via `speed_sig`)? And then more tests or profiling runs that address practical use cases like variable-length messages, varying CPU load, etc. (topics for open-quantum-safe/profiling#40).

> I accept that we shouldn't change test suites to avoid exposing data-dependent behaviour.

I don’t dispute this. But it seems to imply that this PR does this, which I don’t see and it was never the intention. In my previous reply I tried to show that if anything, the contrary is the case.

> What about we continue this conversation there? See [open-quantum-safe/profiling#40](https://github.com/open-quantum-safe/profiling/issues/40)

Agreed!

PR #938, opened 2021-03-12
Fix small error in comments of kem.h and sig.h

thomwiggers:
They made a reference to example files that had been moved at some point.

### DIFF
--- a/src/kem/kem.h
+++ b/src/kem/kem.h
@@ -2,7 +2,7 @@
  * \file kem.h
  * \brief Key encapsulation mechanisms
  *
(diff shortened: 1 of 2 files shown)


baentsch (approved):
LGTM. Suggest adding "[skip ci]" next time only comments are changed.

thomwiggers:
Yeah I'd of course realised that as soon as the commit had been pushed — I wasn't sure if fixing it would just leave those builds running anyway...

baentsch:
> I wasn't sure if fixing it would just leave those builds running anyway...

It would have left them running (you can tell I fell into that myself before :). Even a `commit --amend` didn't help.

PR #939, opened 2021-03-12
guard Frodo AVX2 code if CPU_EXTENSIONS not requested

baentsch:
#927 breaks builds when [OQS_USE_CPU_EXTENSIONS=OFF](https://github.com/open-quantum-safe/liboqs/wiki/Customizing-liboqs#oqs_use_cpu_extensions): See e.g. [here](https://app.circleci.com/pipelines/github/open-quantum-safe/oqs-demos/338/workflows/84da16f9-8f90-446e-b371-25d7e6d88258/jobs/413). This PR is one way to fix it. Another alternative is to not set `OQS_PORTABLE_X86_64_BUILD` if `OQS_USE_CPU_EXTENSIONS=OFF`: Any preference, @jschanck ?

### DIFF
--- a/src/kem/frodokem/external/frodo_macrify_optimized.c
+++ b/src/kem/frodokem/external/frodo_macrify_optimized.c
@@ -27,7 +27,7 @@ int frodo_mul_add_as_plus_e(uint16_t *out, const uint16_t *s, const uint16_t *e,
 { // Generate-and-multiply: generate matrix A (N x N) row-wise, multiply by s on the right.
   // Inputs: s, e (N x N_BAR)
   // Output: out = A*s + e (N x N_BAR)
-#if defined(OQS_PORTABLE_X86_64_BUILD)
+#if defined(OQS_USE_AVX2_INSTRUCTIONS) && defined(OQS_PORTABLE_X86_64_BUILD)
     OQS_CPU_EXTENSIONS available_cpu_extensions = OQS_get_available_CPU_extensions();
     if (available_cpu_extensions.AVX2_ENABLED) {
       return frodo_mul_add_as_plus_e_avx2(out, s, e, seed_A);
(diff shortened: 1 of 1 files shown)


dstebila (approved):

jschanck:
The issue is
https://github.com/open-quantum-safe/liboqs/blob/e116748c50ef3adbbf08f4bc1ad0c18ab8f7b7df/src/common/common.c#L13

Since OQS_PORTABLE_BUILDs are the only builds that use runtime CPU feature detection, this should be

```c
#if defined(OQS_PORTABLE_BUILD)
```

The only use I see for OQS_USE_CPU_EXTENSIONS is the combination `OQS_PORTABLE_BUILD=OFF` + `OQS_USE_CPU_EXTENSIONS=OFF`, which allows one to build for the local machine while disabling the use of the local machine's CPU extensions.

xvzcf:
> The issue is
> 
> https://github.com/open-quantum-safe/liboqs/blob/e116748c50ef3adbbf08f4bc1ad0c18ab8f7b7df/src/common/common.c#L13
> 
> Since OQS_PORTABLE_BUILDs are the only builds that use runtime CPU feature detection, this should be
> 
> ```c
> #if defined(OQS_PORTABLE_BUILD)
> ```
> 
> The only use I see for OQS_USE_CPU_EXTENSIONS is the combination `OQS_PORTABLE_BUILD=OFF` + `OQS_USE_CPU_EXTENSIONS=OFF`, which allows one to build for the local machine while disabling the use of the local machine's CPU extensions.

Agreed, and the `OQS_PORTABLE_BUILD=OFF + OQS_USE_CPU_EXTENSIONS=OFF` seems like a marginal use case to me.

jschanck:
I will submit a separate PR with a potential fix.

baentsch:
> Agreed, and the OQS_PORTABLE_BUILD=OFF + OQS_USE_CPU_EXTENSIONS=OFF seems like a marginal use case to me.

Agreed. But that combination wasn't the case here. What I've been looking for in this case was a portable image without the need for/cost of CPU feature checks. And that --I thought-- `OQS_USE_CPU_EXTENSIONS=OFF` should yield.

Whatever, I'll be looking forward to

> I will submit a separate PR with a potential fix.

Issue #940, opened 2021-03-12
Current master - McEliece AVX2 is broken


mouse07410:
MacOS 10.15.7, Xcode-12.4, current master.
```
-- The C compiler identification is AppleClang 12.0.0.12000032
-- The ASM compiler identification is Clang
-- Found assembler: /opt/local/bin/clang
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /opt/local/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Found OpenSSL: /opt/local/lib/libcrypto.dylib (found suitable version "1.1.1j", minimum required is "1.1.1")  
-- Looking for OPENSSL_NO_EC2M
-- Looking for OPENSSL_NO_EC2M - not found
-- Found Doxygen: /opt/local/bin/doxygen (found version "1.9.1") found components: doxygen dot 
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/ur20980/src/liboqs/build
[1272/2236] Building ASM object src/kem/classic_mceliece/CMakeFile...ic_mceliece_8192128_avx.dir/pqclean_mceliece8192128_avx/consts.S.
FAILED: src/kem/classic_mceliece/CMakeFiles/classic_mceliece_8192128_avx.dir/pqclean_mceliece8192128_avx/consts.S.o 
[...]

PR #941, opened 2021-03-12
Fix OQS_PORTABLE_BUILD logic for upstream imports

jschanck:
The first commit fixes the OQS_PORTABLE_BUILD logic for the schemes that are imported using copy_from_upstream. Previously you would only get functional runtime CPU feature detection if you compiled on a machine with every CPU feature.

The second commit fixes our interpretation of "required_flag: pclmulqdq" in META.yml. The presence of this flag should cause us to compile with "-mpclmul" not "-mpclmulqdq". Upstream will not change META.yml, see https://github.com/PQClean/PQClean/pull/381#issuecomment-799186507.

The third commit is a mitigation for #946.


jschanck:
Side effect of these changes is that we observed #940 in CI. The fix in 3637b73 appears to work.

jschanck:
macOS is failing because the OQS_PORTABLE_BUILD logic for sha3 is broken on main. Not going to fix this, since main's sha3 is obsolete. This PR is on pause until #922 is merged.

dstebila (approved):

PR #942, opened 2021-03-12
Make OQS_USE_CPU_EXTENSIONS imply OQS_PORTABLE_BUILD=OFF; alternative…

jschanck:
#941 fixes the broken `OQS_PORTABLE_BUILD=ON` + `OQS_USE_CPU_EXTENSIONS=OFF` combination (#939) as part of a larger effort to fix OQS_PORTABLE_BUILD. Unfortunately that's now waiting on new sha3 to merge.

@baentsch, in #939, writes:
> What I've been looking for in this case was a portable image without the need for/cost of CPU feature checks. And that --I thought-- `OQS_USE_CPU_EXTENSIONS=OFF` should yield.

Here it sounds like you actually do want  `OQS_PORTABLE_BUILD=OFF` + `OQS_USE_CPU_EXTENSIONS=OFF`.

That doesn't change the fact that we should have a short-term mitigation for the broken configuration. I'm fine with forcing OQS_PORTABLE_BUILD=OFF when OQS_USE_CPU_EXTENSIONS=OFF. This PR does that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,11 @@ else()
     message(FATAL_ERROR "Unknown or unsupported processor: " ${CMAKE_SYSTEM_PROCESSOR})
 endif()
 
+if(NOT OQS_USE_CPU_EXTENSIONS AND OQS_PORTABLE_BUILD)
+    message(WARNING "OQS_PORTABLE_BUILD requires OQS_USE_CPU_EXTENSIONS")
+    set(OQS_PORTABLE_BUILD OFF)
+endif()
+
 if(OQS_PORTABLE_BUILD AND ARCH_X86_64)
     set(OQS_PORTABLE_X86_64_BUILD ON)
 endif()


dstebila (approved):

dstebila:
I'm a bit confused as to the sequence in which #922, #939, #941, and this one should be merged.

jschanck:
> I'm a bit confused as to the sequence in which #922, #939, #941, and this one should be merged.

#939 should be closed---it re-introduces a bug in the portable build logic for Frodo.

I promised a proper fix in the comments to #939. That proper fix is part of #941 which is waiting on #922 (and #941 probably needs additional work even after #922 is merged).

This PR fixes the issue from #939 by simply disabling the broken combination of build flags.

baentsch:
> Here it sounds like you actually do want OQS_PORTABLE_BUILD=OFF + OQS_USE_CPU_EXTENSIONS=OFF.
> That doesn't change the fact that we should have a short-term mitigation for the broken configuration. I'm fine with forcing OQS_PORTABLE_BUILD=OFF when OQS_USE_CPU_EXTENSIONS=OFF. This PR does that.

I was looking for a maximally portable binary. Getting this by setting `OQS_PORTABLE_BUILD=OFF` is somewhat counter-intuitive. But I may simply be confused by the naming:

So, allow me to state which configurations I'd been using & thought sensible:
1) "Reference": A configuration permitting benchmarking of all algorithms' baseline, reference code. Build anywhere, run anywhere. Slow. No feature-test overhead.
2) "Performance": A configuration permitting benchmarking of all algorithms' best optimized code. Build on CPUs with all required features, executing OK on CPUs with those features, crashing on CPUs without these features. No feature-test overhead.
3) "Practical"/Release: A configuration running at high-speed where possible, slower where not (e.g., CPU features not present); but surely never crashing; carries CPU feature detection overhead and thus not good for benchmarking. Builds on CPUs with all CPU features available but executes also on CPUs without them (with feature-test overhead).

Could we agree these configurations make sense? If not, let's discuss.

If we'd agree they make sense, these 3 configurations were delivered by (in my mind -- but possibly that's the problem):
1) OQS_USE_CPU_EXTENSIONS=OFF (OQS_PORTABLE_BUILD not relevant)
2) OQS_USE_CPU_EXTENSIONS=ON&&OQS_PORTABLE_BUILD=OFF
3) OQS_USE_CPU_EXTENSIONS=ON&&OQS_PORTABLE_BUILD=ON

If I understand the PR right, OQS_PORTABLE_BUILD must be OFF when OQS_USE_CPU_EXTENSIONS=OFF. Basically fine with me, but I'd then probably suggest renaming OQS_PORTABLE_BUILD to something like OQS_CPU_RT_EXTENSION_CHECK to avoid future confusion. This would also make it obvious that the combination OQS_USE_CPU_EXTENSIONS=OFF&&OQS_CPU_RT_EXTENSION_CHECK=ON is pointless.

Last point, should we add CMAKE_BUILD_TYPES named as per the 3 configs above, also documenting what we "mean" by the config flags (whatever their name)?

jschanck:
Renaming OQS_PORTABLE_BUILD sounds good to me. It should wait until after #941 is merged and should not be tacked onto this PR. @baentsch would you mind opening a discussion thread instead?

dstebila:
Okay, I'll merge this one.  Then we're waiting on the Picnic team feedback on #922, then #941, then some new work on renaming OQS\_PORTABLE\_BUILD. Thanks guys.

PR #944, opened 2021-03-15
Patch mechanism for copy from upstream

jschanck:
Opening this PR for discussion around whether we want a patch mechanism for copy from upstream.

This PR contains a functioning patch mechanism. The documentation is here:
https://gist.github.com/jschanck/c67d37551da2a5f6e06c640816d2c99f

### DIFF
--- a/scripts/copy_from_upstream/copy_from_upstream.py
+++ b/scripts/copy_from_upstream/copy_from_upstream.py
@@ -45,6 +45,11 @@ def file_put_contents(filename, s, encoding=None):
     with open(filename, mode='w', encoding=encoding) as fh:
         fh.write(s)
 
+def shell(command, expect=0):
+    subprocess_stdout = None if DEBUG > 0 else subprocess.DEVNULL
+    ret = subprocess.run(command, stdout=subprocess_stdout, stderr=subprocess_stdout)
+    if ret.returncode != expect:
+        raise Exception("'{}' failed with error {}. Expected {}.".format(" ".join(command), ret, expect))
 
 def generator(destination_file_path, template_filename, family, scheme_desired):
     template = file_get_contents(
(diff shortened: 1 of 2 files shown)


bhess:
The changes look good to me. It's a useful extension for copy_from_upstream when integrating upstream implementations.

Perhaps we could briefly describe in https://gist.github.com/jschanck/c67d37551da2a5f6e06c640816d2c99f the alternatives to (big) patches, such as
- Adding shim API (similar to `src/common/pqclean_shims`) to avoid having to patch namespaces, function names and incompatible API. 
- Offer the patch to upstream, if it is not only for OQS-internal reasons.

bhess (commented):
he changes look good to me. It's a useful extension for copy_from_upstream when integrating upstream implementations.

Perhaps we could briefly describe in https://gist.github.com/jschanck/c67d37551da2a5f6e06c640816d2c99f the alternatives to (big) patches, such as

Adding shim API (similar to `src/common/pqclean_shims`) to avoid having to patch namespaces, function names and incompatible API.
Offer the patch to upstream, if it is not only for OQS-internal reasons.

baentsch:
Conceptually, a patch mechanism for upstream code is a good thing, particularly for quick fixes we fully understand. 

My key concern doing this on a regular and long-term basis is this: `liboqs` changes from being an independent maintainer of a collection of unchanged upstream code bases into being the active maintainer (taking responsibility of the internal operations) of certain (such patched) algorithms.

I thus fully agree with @bhess comments
> - Adding shim API (similar to src/common/pqclean_shims) to avoid having to patch namespaces, function names and incompatible API.
> - Offer the patch to upstream, if it is not only for OQS-internal reasons.

and would amend them as follows
- all patches need to be designed such that they can be taken up by upstream
- if the upstream maintainers don't take up the patch by the next run of `copy_from_upstream` and provide their own implementation for the function in question, the patch shall be removed again from liboqs.

This way we can introduce quick improvements without moving into the position of becoming a long-term maintainer of upstream patches. 

If we agree and document all of this, I'm all for this (PR).

baentsch (approved):

baentsch:
Suggest to update https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide#automated-inclusion after merge, capturing the jist of the conversation above: Would you want to do this, @jschanck or shall I give it a try after you merged?

jschanck:
> Suggest to update https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide#automated-inclusion after merge, capturing the jist of the conversation above: Would you want to do this, @jschanck or shall I give it a try after you merged?

I added https://github.com/open-quantum-safe/liboqs/wiki/Patching-upstream-sources and a brief note to https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide#automated-inclusion

PR #948, opened 2021-03-23
Patch Kyber, Dilithium, and SPHINCS to use our SHA3

jschanck:
This PR patches Kyber, Dilithium, and SPHINCS to use our SHA3 .

The Kyber and Dilithium patches are bit more minimalistic than the ones that were originally part of #922. I've added shims for the "absorb_once" and "squeezebytes" functions.

Dilithium's fips202.c differs from Kyber's and PQClean's, so the patches here make some name changes like "shake128_absorb" to "shake128_inc_absorb". This prevents us from needing a parallel set of shims just for Dilithium.

The SPHINCS patches just remove unused files.


baentsch (commented):
Patches for Kyber and Dilithium appear small/understandable (based on which I'd approve; @bhess might still want to take a look, maybe also to comment regarding a possible upstream uptake); The SPHINCS patch in turn looks like a more drastic change: For that part I can't give feedback, thus.

bhess (commented):
The pqcrystals patches looks good to me.
Is there a performance change after switching to the internal implementation (e.g. due to allocating the context on the heap)?
I've forwarded the patches to the pqcrystals team to check if they can be adopted in the upstream.

jschanck:
@baentsch 
> The SPHINCS patch in turn looks like a more drastic change: For that part I can't give feedback, thus.

The SPHINCS code has 12 copies of fips202x4.c and fips202x4.h that are no longer necessary after inserting shims. The patch just removes those. It's a quirk of `git apply` instead of `patch` that the 12 files appear in full in the .patches.

@bhess 
> Is there a performance change after switching to the internal implementation

Yes, but it should be minor. I've tried to minimize the overhead by using `shake{128,256}_inc_ctx_reset` to re-use contexts when possible.

>I've forwarded the patches to the pqcrystals team to check if they can be adopted in the upstream.

Thanks!

dstebila (approved):

Issue #775, opened 2020-06-26
When CPU_EXTENSIONS becomes supported in cross-compilation, update cross-compilation instructions to use it


kevinmkane:
CPU_EXTENSIONS currently causes a CMake error when configuring for cross-compilation. If and when it becomes capable of being used in cross-compilation, two sets of instructions need to be updated to remove the `-DOQS_USE_CPU_EXTENSIONS=OFF` directive currently present:

1. In liboqs's README.md, undoing where it was added in #772.
2. In OQS-OpenSSL's [cross-compilation wiki](https://github.com/open-quantum-safe/openssl/wiki/Cross-compiling-for-Windows-on-Linux).

dstebila:
@jschanck Do you think #942 brings us any closer to this one?

jschanck:
@dstebila Not quite. But Michael and I discussed some changes (https://github.com/open-quantum-safe/liboqs/discussions/943, specifically OQS_TARGET_ARCH) that would allow one to target a particular microarchitecture during cross-compilation. I'll do a mock-up of that idea soon.

PR #951, opened 2021-03-26
Replace OQS_PORTABLE_BUILD and OQS_USE_CPU_EXTENSIONS

jschanck:
Following the discussion in https://github.com/open-quantum-safe/liboqs/discussions/943, this replaces OQS_PORTABLE_BUILD and OQS_USE_CPU_EXTENSIONS with more flexible (and less confusing) alternatives.

* OQS_DIST_BUILD (default: OFF): "Build distributable library with optimized code for several CPU microarchitectures. Enables run-time CPU feature detection."
* OQS_OPT_TARGET (default: "auto"): "The target microarchitecture for optimization."

OQS_DIST_BUILD is a drop in replacement for the previous OQS_PORTABLE_BUILD. A library built with OQS_DIST_BUILD will include implementations for several microarchitectures, and it will dispatch calls to the correct implementation based the CPU features that are available at run time. The OQS_DIST_BUILD option is refined to one of OQS_DIST_X86_64_BUILD, OQS_DIST_ARM64v8_BUILD, or OQS_DIST_ARM32v7_BUILD at compile time depending on CMAKE_SYSTEM_PROCESSOR. This is useful for cross compiling and it also simplifies some logic throughout the build.

OQS_OPT_TARGET can take a value of "auto", "generic", or the name of a CPU like "haswell". When OQS_DIST_BUILD=ON we force OQS_OPT_TARGET=generic.

On x86_64, "generic" means we'll compile with "-march=x86-64". On ARM64v8 "generic" means "-mcpu=cortex-a53". On ARM32v7 "generic" means "-mcpu=cortex-a5". This is like OQS_USE_CPU_EXTENSIONS=OFF.

On any architecture, "auto" will try to determine the current CPU using "march=native" (if it's supported). This is like the previous OQS_USE_CPU_EXTENSIONS=ON

Downstream builds that set OQS_PORTABLE_BUILD or OQS_USE_CPU_FEATURES will fail. So we should prepare for that if we decide to make these changes.

*edit*: Removing OQS_USE_CPU_EXTENSIONS had the side effect that our ARM jobs tried to execute detect_gcc_clang_intrinsics.c. I took the opportunity to add both compile-time detection of ARM crypto features and neon. It is now possible to cross compile for ARM and optimize for the target CPU (relevant to #775). I also took the opportunity to add run-time detection of ARM features, and I replaced OQS_get_available_CPU_extensions (01efe50) for efficiency reasons (previous implementation was copying ~40 bytes on return).


baentsch:
Thanks for working on this! Conceptually, it looks good. I'd like to be sure I understood correctly how this now delivers the three main configs for different uses:
1) "Release" (fastest, with all possible CPU feature support, but never-crashing/RT feature detection): In docker images
2) "Reference" (without any CPU feature support): Benchmarking ref code
3) "Performance" (possibly crashing if not all required features available at run-time): (Aiming to get) Best benchmarking numbers

Am I understanding the above correct such as to set up those 3 configurations up (for x86_64) with the following switches:
1) OQS_OPT_TARGET="skylake" & OQS_DIST_BUILD=ON
2) OQS_OPT_TARGET="generic"; OQS_DIST_BUILD=don't care (or _must_ it be OFF?)
3) OQS_OPT_TARGET="skylake" (or "auto" if building and running on the same host) & OQS_DIST_BUILD=OFF

For ARM64, I guess building with "auto" and only setting OQS_DIST_BUILD=ON when creating docker images is best.

Did I get this all right? In any case, what about the suggestion to create something like CMAKE_BUILD_TYPE(s) designating sensible configurations -- incl. one that doesn't trip up the unwary ("Release"? Or rather something like "Development": OQS_OPT_TARGET="auto" & OQS_DIST_BUILD=OFF)?

jschanck:
Thanks for taking a look.

> 1. "Release" (fastest, with all possible CPU feature support, but never-crashing/RT feature detection): In docker images
> [...] OQS_OPT_TARGET="skylake" & OQS_DIST_BUILD=ON

You don't have to set OQS_OPT_TARGET here (in fact it's ignored).

> 2. "Reference" (without any CPU feature support): Benchmarking ref code
> [...] OQS_OPT_TARGET="generic"; OQS_DIST_BUILD=don't care (or _must_ it be OFF?)

OQS_DIST_BUILD must be OFF in this case (it is off by default so the user only has to specify OQS_OPT_TARGET=generic).

> 3. "Performance" (possibly crashing if not all required features available at run-time): (Aiming to get) Best benchmarking numbers 
> [...]  OQS_OPT_TARGET="skylake" (or "auto" if building and running on the same host) & OQS_DIST_BUILD=OFF

Right.

> For ARM64, I guess building with "auto" and only setting OQS_DIST_BUILD=ON when creating docker images is best.

Agreed.

> Did I get this all right? In any case, what about the suggestion to create something like CMAKE_BUILD_TYPE(s) designating sensible configurations -- incl. one that doesn't trip up the unwary ("Release"? Or rather something like "Development": OQS_OPT_TARGET="auto" & OQS_DIST_BUILD=OFF)?

We have four CMAKE_BUILD_TYPEs right now.  "" (empty string), "Release", "Debug", and "RelWithDebInfo". But we barely use these. "Debug" and "RelWithDebInfo" are the same. Release and "" are the same. This PR makes OQS_OPT_TARGET=auto and OQS_DIST_BUILD=OFF the default for all four configurations.

So the question is whether we want some build type that enables OQS_DIST_BUILD by default. Would we want this to be the "Release"/"RelWithDebInfo" types? Or would we add "Distribution"/"DistWithDebInfo"?

baentsch:
Looks like I'm slowly appearing to be on the same page now: Thanks for the explanations. Just two more comments:

> You don't have to set OQS_OPT_TARGET here (in fact it's ignored).

Hm, so setting `OQS_DIST_BUILD` turns off (evaluation of) `OQS_OPT_TARGET`? Wasn't the latter meant to provide also a way to trigger cross-compilation? So one has to be on "the right"/target platform building when setting `OQS_DIST_BUILD`? If so, what if that platform does not have all possible CPU features? In other words, setting `OQS_DIST_BUILD` on an old machine (e.g., without AVX2) will not generate AVX2 extension code?

> So the question is whether we want some build type that enables OQS_DIST_BUILD by default. Would we want this to be the "Release"/"RelWithDebInfo" types? Or would we add "Distribution"/"DistWithDebInfo"?

I'd vote for "Release". A Release-with-Debuginfo doesn't sound needed, is it?

jschanck:
> Hm, so setting `OQS_DIST_BUILD` turns off (evaluation of) `OQS_OPT_TARGET`? Wasn't the latter meant to provide also a way to trigger cross-compilation? So one has to be on "the right"/target platform building when setting `OQS_DIST_BUILD`? If so, what if that platform does not have all possible CPU features? In other words, setting `OQS_DIST_BUILD` on an old machine (e.g., without AVX2) will not generate AVX2 extension code?

No. If OQS_DIST_BUILD=ON and you're compiling on any x86_64 machine (or compiling for x86_64 from arm64, etc, etc) then you will get a library with AVX2 implementations.

We briefly discussed using OQS_OPT_TARGET in combination with OQS_DIST_BUILD to trigger some types of cross-compilation (x86_64 compiling for arm64, etc), but that doesn't actually work. The right way to do that sort of cross-compilation is with a toolchain file. I've provided some examples with this PR.

> > So the question is whether we want some build type that enables OQS_DIST_BUILD by default. Would we want this to be the "Release"/"RelWithDebInfo" types? Or would we add "Distribution"/"DistWithDebInfo"?
> 
> I'd vote for "Release". A Release-with-Debuginfo doesn't sound needed, is it?

Let's discuss this on the Wednesday call.

.circleci/config.yml
```diff
@@ -289,7 +289,7 @@ workflows:
           name: ubuntu-focal-shared-noopenssl
           context: openquantumsafe
           CONTAINER: openquantumsafe/ci-ubuntu-focal-x86_64:latest
-          CMAKE_ARGS: -DCMAKE_C_COMPILER=gcc-7 -DCMAKE_BUILD_TYPE=Release -DOQS_USE_OPENSSL=OFF -DBUILD_SHARED_LIBS=ON
+          CMAKE_ARGS: -DCMAKE_C_COMPILER=gcc-7 -DCMAKE_BUILD_TYPE=Release -DOQS_DIST_BUILD=ON -DOQS_USE_OPENSSL=OFF -DBUILD_SHARED_LIBS=ON
```
baentsch, line 292: After today's discussion, isn't "-DCMAKE_BUILD_TYPE=Release -DOQS_DIST_BUILD=ON" a contradiction?

jschanck: It's not so much that Release means "OQS_DIST_BUILD=OFF && OQS_OPT_TARGET=auto" but that Release has no effect and those are the defaults. This is just overriding the default by setting OQS_DIST_BUILD=ON.

There's no reason to ever specify OQS_BUILD_TYPE=Release, so maybe the best thing to do here is remove references to it from the .circleci/config.yml file.

jschanck:
15e3684: removes some unnecessary references to CMAKE_BUILD_TYPE=Release. This should make it clear that CMAKE_BUILD_TYPE is only to be used for turning on debug info (with CMAKE_BUILD_TYPE=Debug). If you want a distributable build you compile with OQS_DIST_BUILD=ON. This is compatible with all settings of CMAKE_BUILD_TYPE.


f9025e8: MSVC has very limited options for platform specific compilation, setting OQS_DIST_BUILD=ON for our appveyor builds means we will at least test compilation (if not execution) of most code on Windows.

dstebila (approved):

dstebila:
Can you update the [wiki page](https://github.com/open-quantum-safe/liboqs/wiki/Customizing-liboqs) as well?

jschanck:
> Can you update the [wiki page](https://github.com/open-quantum-safe/liboqs/wiki/Customizing-liboqs) as well?

Done

thomwiggers:
`OQS_DIST_BUILD (default: OFF)` — this is a change from the previously-default-on `OQS_PORTABLE_BUILD` right?

jschanck:
@thomwiggers that's right.

thomwiggers:
Okay; maybe something to add to the release notes https://github.com/open-quantum-safe/liboqs/blob/main/RELEASE.md#release-notes

PR #958, opened 2021-04-02
Manually remove fips202 files left by copy_from_upstream

jschanck:
Manually remove some files that should have been removed by copy_from_upstream.

The issue is that copy from upstream does:
`for x in upstream/scheme: copy x to liboqs/scheme`.
But this doesn't handle upstream deletions. It should completely remove liboqs/scheme first.


dstebila:
I see the files that were deleted, but I don't see a change in copy\_from\_upstream to actually do the proposed deletion of the directory before the copy process, did you intend to add that?

jschanck:
The changes were made in #948 by patching the upstream META.yml files: https://github.com/jschanck/liboqs/blob/b228c41156424c5e8ef5b29ec43269567184b65e/scripts/copy_from_upstream/patches/pqcrystals-kyber-yml.patch. These patches instruct copy_from_upstream to skip the fips202 files, but there were copies already in our repo.

dstebila (approved):

PR #959, opened 2021-04-02
Update windows cross-compiling toolchain and documentation [skip ci]

### LINKED ISSUE #775: When CPU_EXTENSIONS becomes supported in cross-compilation, update cross-compilation instructions to use it

jschanck:
Resolves #775


dstebila:
Is wine now a dependency for cross-compiling for Windows?  That is, did something change making that a new dependency, or it was already a dependency and just wasn't explicit?  Or something else?

jschanck:
It was already a dependency if you wanted to build with OQS_USE_CPU_EXTENSIONS=ON flag. It is only a dependency now if you want to build with OQS_DIST_BUILD=OFF.

The issue is that for non-dist builds we compile and run "detect_gcc_clang_intrinsics.c". If you're cross compiling then you need to run the result in an emulator.

We could potentially work around this, since we can get the same information from the compiler w/o running anything using  `cc -E -dM detect_gcc_clang_intrinsics.c`. But that would require some knowledge of CMake's behaviour across platforms.

dstebila (approved):

dstebila:
Thanks for the explanation. Merging as is.

PR #962, opened 2021-04-06
Update RELEASE notes with new build flags [skip ci]

baentsch:
As per discussion in #951

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -40,3 +40,8 @@ What's New
 This release includes the NIST PQC round 3 parameters for SPHINCS.
 
 ### Other changes
+
+#### Build options changed
+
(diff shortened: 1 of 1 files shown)


jschanck:
Alternatively, I'd be open to changing the default value of OQS_DIST_BUILD.

baentsch:
> Alternatively, I'd be open to changing the default value of OQS_DIST_BUILD.

I don't mind either way: Doesn't highlighting the new flags in the release notes make sense regardless? Personally I think the way it is now is "the norm": Build on and for a given platform (only). Portability/Distributability is a good feature to have but does it have to be default? It also significantly slows down some algorithms (that fall back to reference code w/o CPU feature detection code).

PR #964, opened 2021-04-06
Speed up ARM jobs in CI

jschanck:
Our ARM jobs currently compile liboqs in an emulated environment. With this PR they will cross compile from the x86_64 host and only use emulation to run the tests. This should speed up the ARM jobs significantly. It might take some trial and error to get it working well.


baentsch:
@jschanck Are you aware that you can run specific CI jobs locally by executing in your project's main directory e.g., `circleci local execute --job arm64`? Improved iteration cycles substantially for me (installation info at https://circleci.com/docs/2.0/local-cli/)

jschanck:
OK, not quite as fast as I hoped, but for arm64 we're down to 20min, from 1hr+.

@baentsch the solution here involves building inside of our debian-buster x86_64 container. I currently need to download some packages for cross compiling. It would be nice if the container came with
```
qemu-user-static
gcc-aarch64-linux-gnu
gcc-arm-linux-gnueabi
gcc-arm-linux-gnueabihf
```
The qemu-user-static package is only needed for running compile-time CPU feature checks in non-dist builds, so we could avoid that dependency by compiling dist builds.


I'm currently testing without OpenSSL support. To cross-compile with OpenSSL support for ARM64v8 I would need the following commands to be run while building the container:
```
dpkg --add-architecture arm64
apt-get update
apt-get install -y libssl-dev:arm64
```
(Similar commands if we want to test OpenSSL support on armel or armhf.)

For the record: I'm not sure the armel test tells us much. I'm in favor of dropping it unless someone has a particular CPU that they'd like us to support.

baentsch:
> I'm not sure the armel test tells us much. I'm in favor of dropping it unless someone has a particular CPU that they'd like us to support.

I personally also wouldn't mind dropping it. @xvzcf @dstebila : Do you know whether anyone needs `armel`?

dstebila:
> > I'm not sure the armel test tells us much. I'm in favor of dropping it unless someone has a particular CPU that they'd like us to support.
> 
> I personally also wouldn't mind dropping it. @xvzcf @dstebila : Do you know whether anyone needs `armel`?

I don't know much about various ARM architectures and how widely they're used. Looking at Debian's [supported hardware page](https://www.debian.org/releases/stretch/armel/ch02s01.html.en) I don't see anything in their armel category that is relevant to any uses of liboqs I know about, so I'm fine with dropping it.

jschanck:
Latest run skips armel completely and also skips test_mem for armhf and arm64. Full CircleCI workflow completes in 18 minutes. The two ARM jobs take about 16 minutes each. I'm done fiddling with it now.

dstebila (approved):

Issue #961, opened 2021-04-06
SIGILL in SIDH code


thomwiggers:
I'm having trouble reproducing this locally, but on Github Actions for `liboqs-rust` I'm getting SIGILL in SIDH consistently.

https://github.com/open-quantum-safe/liboqs-rust/runs/2276909119#step:12:62

```
test kem::SidhP434::test_encaps_decaps ... error: test failed, to rerun pass '-p oqs --lib'

Caused by:
  process didn't exit successfully: `/home/runner/work/liboqs-rust/liboqs-rust/target/debug/deps/oqs-c655d61dbe5addf1 --test-threads 1` (signal: 4, SIGILL: illegal instruction)
```

For debugging purposes, I already pulled out the support CPU flags on Github Actions:

```
 flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse 
sse2 ss ht syscall nx pdpe1gb rdtscp lm constant_tsc rep_good nopl xtopology cpuid pni pclmulqdq ssse3 fma cx16 
pcid sse4_1 sse4_2 movbe popcnt aes xsave avx f16c rdrand hypervisor lahf_lm abm invpcid_single pti fsgsbase bmi1 
avx2 smep bmi2 erms invpcid xsaveopt md_clear
```

The difference with my local set of `flags`:

* My computer doesn't have `hypervisor`
* GHA don't have `{'ida', 'flush_l1d', 'tsc_adjust', 'hwp_epp', 'vpid', 'ibpb', 'bts', 'tm', 'xsaves', 'ssbd', 'acpi', 'dtes64', 'dts', 'vmx', 'hwp_notify', 'pts', 'stibp', 'x2apic', 'intel_pt', 'xsavec', 'tpr_shadow', 'xgetbv1', 'ept', 'hwp_act_window', 'rdseed', 'hwp', 'adx', 'tsc_deadline_timer', 'sdbg', 'smap', 'tm2', 'est', 'ibrs', 'arat', 'aperfmperf', 'pebs', 'dtherm', 'ept_ad', 'clflushopt', 'mpx', 'epb', 'pdcm', 'nonstop_tsc', 'monitor', 'cpuid_fault', 'xtpr', '3dnowprefetch', 'pln', 'art', 'arch_perfmon', 'vnmi', 'pbe', 'ds_cpl', 'flexpriority'}`

jschanck:
[...]

PR #965, opened 2021-04-12
SIKE/CMake: recognize both x64+BMI2 and x64+BMI2+ADX

jschanck:
While looking into https://github.com/open-quantum-safe/profiling/issues/47 I noticed that the bugfix in #963 was slightly too restrictive.

SIKE provides assembly routines for both x64+BMI2 and x64+BMI2+ADX systems, but it requires different #defines in the two cases. Prior to #963 we were setting the wrong #defines for x64+BMI2, which caused #961. I assumed that SIKE only supported x64+BMI2+ADX in #963, and I disabling support for x64+BMI2.

This PR re-enables support for x64+BMI2 and sets the right #defines for both configurations.


src/kem/sike/CMakeLists.txt
```diff
@@ -82,8 +83,11 @@ if(OQS_ENABLE_KEM_sike_p434 OR
         target_compile_definitions(sike PRIVATE _GENERIC_ _X86_)
     elseif(ARCH_X86_64)
         target_compile_definitions(sike PRIVATE _AMD64_)
-        if(OQS_USE_ADX_INSTRUCTIONS AND NOT OQS_DIST_BUILD)
-            target_compile_definitions(sike PRIVATE _MULX_ _ADX_)
+        if(OQS_USE_BMI2_INSTRUCTIONS)
```
baentsch, line 86: Woudn't it be more logical/readable to have the two options be independent? Or would the code break if only ADX is present but not BMI2?

jschanck: The two options are not independent. Every machine with ADX instructions has BMI2 instructions.

src/kem/sike/CMakeLists.txt
```diff
@@ -92,7 +96,8 @@ if(OQS_ENABLE_KEM_sike_p434 OR
             if(OQS_ENABLE_KEM_sike_p751 OR OQS_ENABLE_KEM_sidh_p751 OR OQS_ENABLE_KEM_sike_p751_compressed OR OQS_ENABLE_KEM_sidh_p751_compressed)
                 target_compile_definitions(sike PRIVATE USE_SIKEP751_ASM)
             endif()
-            if(OQS_USE_ADX_INSTRUCTIONS AND NOT OQS_DIST_BUILD)
+            if(OQS_USE_BMI2_INSTRUCTIONS)
```
baentsch, line 99: The switch from checking for ADX to BMI2 I understand (if the latter also includes/implies the former or the code builds regardless of ADX presence) but leaving away `OQS_DIST_BUILD` here seems to imply that the resultant code will work regardless of CPU features found, no? Asked another way, what code difference results for SIKE with `OQS_DIST_BUILD` ON or OFF?

jschanck: The OQS_USE_XXX_INSTRUCTIONS are only set when OQS_DIST_BUILD=OFF.

https://github.com/open-quantum-safe/liboqs/blob/efd7b1926ac485bd2c284986cb94e51f4d50450c/.CMake/compiler_opts.cmake#L51-L53

baentsch: OK, I was confused how then run-time feature detection works when OQS_DIST_BUILD=ON given this (example Kyber, other places equivalent) code: https://github.com/open-quantum-safe/liboqs/blob/efd7b1926ac485bd2c284986cb94e51f4d50450c/src/kem/kyber/kem_kyber_512.c#L44-L56 . But this https://github.com/open-quantum-safe/liboqs/blob/efd7b1926ac485bd2c284986cb94e51f4d50450c/.CMake/alg_support.cmake#L178-L180 in combination with this https://github.com/open-quantum-safe/liboqs/blob/efd7b1926ac485bd2c284986cb94e51f4d50450c/CMakeLists.txt#L33-L35 makes it clear again. So case closed. Thanks.

baentsch (approved):
After the explanations, LGTM.

Issue #949, opened 2021-03-25
Add constant-time plain C AES implementation when building without AES-NI


dstebila:
There's such an implementation available in PQClean that we could use.

dstebila:
I looked at the AES C implementation in PQClean and it looks like it would mostly fit in with what we need, except that our API includes ECB decryption operations, whereas the implementation in PQClean doesn't provide that.  However, as far as I can tell, the only place we call decryption is in the AES test harness (to see if decryption successfully decrypts encryption).  We have other known-answer tests for AES encryption in our test harness, so I think it would be okay to drop the decryption operations from our API.  Any thoughts?

baentsch:
FWIW: Sounds reasonable to me: If no algorithm uses that function by now I cannot really imagine that that will happen in the future. I just validated "practically" the former by checking that `ninja run_tests` passes after quickly removing `OQS_AES128_ECB_dec[_sch]`, making `OQS_AES128_ECB_load_schedule` fail with param `for_encryption=0` and also ifdef'ing away the `*_dec` functionality and performance tests in `tests/aes.c`. If you want to keep aes_dec support as an option (e.g., in testing), I put this into the branch [no-aesdec](https://github.com/open-quantum-safe/liboqs/tree/no-aesdec) where you can toggle it (presently disabled): https://github.com/open-quantum-safe/liboqs/blob/4e3e5419a08b1f81eaf1672e7fc05af908f58c01/src/common/aes/aes.h#L12

jschanck:
+1. Removing decryption makes it more clear that AES is only provided for use as a pseudorandom function (/generator).

I'd actually prefer if we removed "encrypt" too and provided an API like the one used by Kyber and Dilithium.

baentsch:
> I'd actually prefer if we removed "encrypt" too and provided an API like the one used by Kyber and Dilithium.

Wouldn't that necessitate changes to all algorithms/many "upstream API integration discussions"?

jschanck:
> Wouldn't that necessitate changes to all algorithms/many "upstream API integration discussions"?

I wouldn't want it to hold up the work of getting a constant time plain C AES. But it's not so bad.

Frodo and Bike use our AES API directly. Those would need (small) patches.

Everything else that uses our API uses shims. Shims are an easy change.

Kyber and Dilithium have redundant AES implementations. We want to remove those. Replicating their API would make it easy to do w/ shims.

PR #970, opened 2021-04-14
Remove AES decryption functions

dstebila:
This is the first step towards accomplishing #949: removing the existing AES decryption code.  Thanks @baentsch for getting this started.  Once this lands, I'll make a second PR for the rest of #949.


baentsch (approved):
OK if you don't really want to keep the option for re-enabling this functionality -- which as per [this discussion](https://github.com/open-quantum-safe/liboqs/issues/949#issuecomment-816756393) is unlikely, so LGTM.

PR #974, opened 2021-04-19
Enable liboqs build in UEFI/EDKII.

jyao1:
UEFI/EDKII used openssl as crypto lib with OPENSSL_SYS_UEFI.

This patch adds __UEFI__ macro in liboqs to indicate special build in UEFI/EDKII.
With this patch, we can start building liboqs in UEFI/EDKII.

For algorithm specific patches, we will submit one by one later.

Signed-off-by: Jiewen Yao <jiewen.yao@intel.com>

### DIFF
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -75,7 +75,7 @@ extern "C" {
 #define OQS_API __attribute__((visibility("default")))
 #endif
 
-#if defined(__UEFI__)
+#if defined(OQS_SYS_UEFI)
 #undef OQS_API
 #define OQS_API
 #endif
@@ -205,7 +205,7 @@ void OQS_MEM_aligned_free(void *ptr);
 #define UNUSED
 #endif
 
-#if defined(__UEFI__)
+#if defined(OQS_SYS_UEFI)
 #undef UNUSED
 #define UNUSED
 #endif
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -75,6 +75,11 @@ extern "C" {
 #define OQS_API __attribute__((visibility("default")))
 #endif
 
+#if defined(__UEFI__)
+#undef OQS_API
+#define OQS_API
+#endif
+
 /**
  * Represents return values from functions.
  *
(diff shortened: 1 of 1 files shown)


dstebila:
Hi Jiewen, I'm not sure I understand the context here. Can you explain a bit more?

jyao1:
Sure.
UEFI specification (https://uefi.org/) describes the interface between system firmware (also known as BIOS) and operating system. EDKII (https://www.tianocore.org/, https://github.com/tianocore/edk2) is an implementation for the UEFI specification and is widely deployed in Personal Computer and Server.

EDKII is using openssl (https://github.com/tianocore/edk2/tree/master/CryptoPkg/Library/OpensslLib) as crypto engine for security features, such as UEFI secure boot image verification (https://github.com/tianocore/edk2/tree/master/SecurityPkg/Library/DxeImageVerificationLib), and HTTPS boot (https://github.com/tianocore/edk2/tree/master/NetworkPkg/HttpDxe).

Now, we are evaluating post-quantum cryptography (PQC) support in UEFI/EDKII, and have UEFI proof-of-concept (https://github.com/jyao1/CryptoEx/tree/master/QuantumSafePkg). We are using liboqs (https://github.com/jyao1/liboqs/tree/UefiSupport).

UEFI/EDKII can be built with Microsoft visual studio, GCC, or llvm. But UEFI/EDKII cannot link with any std lib. In order to use liboqs, we need change the code to make it buildable in the UEFI/EDKII environment.

We take same approach as what we have done for openssl (https://github.com/openssl/openssl/blob/master/include/openssl/e_os2.h#L48).

baentsch:
Thanks for the explanations, @jyao1 -- Could you explain why you are using the pretty generic `__UEFI__` here to toggle this behaviour as opposed to the more "controlled" `OPENSSL_SYS_UEFI` in OpenSSL? Couldn't it be so that some package inadvertently sets this #define? Or asked another way around, what about using a define such as `OQS_SYS_UEFI` instead?

jyao1:
No special reason.
I am happy to use OQS_SYS_UEFI. I will change it.

jyao1:
I will close this one and create a new one.

PR #975, opened 2021-04-20
Enable liboqs build in UEFI/EDKII.

jyao1:
UEFI/EDKII used openssl as crypto lib with OPENSSL_SYS_UEFI.

This patch adds OQS_SYS_UEFI macro in liboqs to indicate special build in UEFI/EDKII.
With this patch, we can start building liboqs in UEFI/EDKII.

For algorithm specific patches, we will submit one by one later.

Signed-off-by: Jiewen Yao <jiewen.yao@intel.com>

### DIFF
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -75,6 +75,11 @@ extern "C" {
 #define OQS_API __attribute__((visibility("default")))
 #endif
 
+#if defined(OQS_SYS_UEFI)
+#undef OQS_API
+#define OQS_API
+#endif
+
 /**
  * Represents return values from functions.
  *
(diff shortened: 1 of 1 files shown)


baentsch:
Thanks for the update. LGTM. [Background link](https://github.com/open-quantum-safe/liboqs/pull/974#issuecomment-822878040).

dstebila (approved):

dstebila:
Thanks for the explanation. I've merged this. If you can think about whether it's possible to add something to our continuous integration checks to ensure that this functionality is tested, that would be best, otherwise we'll not know if any subsequent changes regress this.

PR #985, opened 2021-05-03
Fix potential NULL dereference in fix_test_kem

thomwiggers:
The `malloc` result was used before it was checked.

PR #986, opened 2021-05-03
Test without alignment

thomwiggers:
This changes the test_kem.c functions like those in PQClean:
  * also prefix magic bytes for checking for out-of-bounds access
  * use odd numbers for breaking alignment

Includes #985

### DIFF
--- a/tests/test_kem.c
+++ b/tests/test_kem.c
@@ -72,17 +72,17 @@ static OQS_STATUS kem_test_correctness(const char *method_name) {
 	shared_secret_e = malloc(kem->length_shared_secret + sizeof(magic_t));
 	shared_secret_d = malloc(kem->length_shared_secret + sizeof(magic_t));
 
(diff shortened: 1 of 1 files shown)


baentsch (approved):
Good & clearly supersedes #985. Please close that first.

thomwiggers:
I'd thought this change might be slightly more controversial than #985 — either way, if #985 gets merged, this PR will become more compact, or if this one gets merged, #985 gets closed automatically.

baentsch:
I don't see why this should be controversial: You basically put & test a fence around the ops memory + break alignment just for functional testing. In the `speed_kem.c` file that might have raised some eyebrows but here it's a good extension in my view. Besides [speed_kem.c](https://github.com/open-quantum-safe/liboqs/blob/main/tests/speed_kem.c) basically tests the original (whatever the compiler considers good :) alignment.

dstebila (approved):

Issue #978, opened 2021-04-28
No documentation available to easily cross compile a library for Windows


zerjioang:
I wanted to compile a liboqs library for windows using linux and I was unable to find a way to do it just following the documentation. We should have a section in readme specifying how to properly cross compile. For example

## Cross compiling LIBOQS from Ubuntu to Windows
To build a `.dll` file

```bash
sudo apt install cmake gcc gcc-mingw-w64 ninja-install
git clone https://github.com/open-quantum-safe/liboqs
cd liboqs
mkdir build && cd build
cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=../.CMake/toolchain_windows-amd64.cmake -DOQS_DIST_BUILD=ON -DBUILD_SHARED_LIBS=ON ..
ninja
```

The result of ninja process, will generate a `bin/liboqs.dll` file on project root directory

baentsch:
Isn't [this](https://github.com/open-quantum-safe/liboqs#cross-compilation) what you're suggesting?

zerjioang:
@baentsch  yes, to add it to the readme as part of offical guidelines for newcomers

baentsch:
@zerjioang Thanks for the suggestion. When I run your suggestion to build a DLL above, I encounter some issues:
- `ninja-install` doesn't seem to be an ubuntu-installable APT package; only `ninja-build` works for me
- After running `ninja`, I only seem to be getting a file `liboqs.dll.a` which subsequently cannot be executed on Windows.

--> Could you please update your description as to what needs to be done to arrive at a Windows-executable DLL? Would you want to do that as part of a PR to update [README.md](https://github.com/open-quantum-safe/liboqs/blob/main/README.md)? Are you using this DLL to handle https://github.com/open-quantum-safe/liboqs-python/issues/36 (i.e., are cross-building the Windows-DLL for the python wrapper and not build it on Windows natively)? May this explain the naming issue there?

zerjioang:
@baentsch  Okey, I'll do a PR with changes for review & approval

FYI: I choose to try cross compilation first, because I already had setup the environment for compilation in my Ubuntu. That's the only reason, to try this way first. Because I can always rely on Windows, download msys2 and do the compilation process again to generate the DLL

baentsch:
Thanks for the explanation (and in advance for the PR). Does the issue https://github.com/open-quantum-safe/liboqs-python/issues/36 also surface if you build "natively" on Windows?

[...]

PR #989, opened 2021-05-04
[#978] - readme updated as clarification of how to cross compile and …

zerjioang:
Regarding #978, a documentation update for clarification

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,19 +143,36 @@ Binaries can be generated using Visual Studio 2019 with the [CMake Tools](https:
 
 You can cross compile liboqs for various platform by supplying CMake with an appropriate [toolchain](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html) file.
 
+#### Cross compiling LIBOQS from Ubuntu to Raspberry Pi
+
 For example, to cross compile for a Raspberry Pi from Ubuntu Bionic:
 
 	apt install gcc-8-arm-linux-gnueabihf
 	mkdir build && cd build
 	cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=../.CMake/toolchain_rasppi.cmake -DOQS_USE_OPENSSL=OFF ..
 	ninja
-	
-Or to compile for Windows AMD64 from Ubuntu Bionic:
+
+#### Cross compiling LIBOQS from Ubuntu to Raspberry Pi
+
+To compile for Windows AMD64 from Ubuntu Bionic:
 
 	apt install gcc-mingw-w64
 	mkdir build && cd build
 	cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=../.CMake/toolchain_windows-amd64.cmake -DOQS_DIST_BUILD=ON ..
 	ninja
+
+#### Cross compiling LIBOQS from Ubuntu to Windows as library
+
+To compile for Windows AMD64 from Ubuntu Bionic and create a `liboqs.dll` file
+
+	sudo apt install cmake gcc gcc-mingw-w64 ninja-build
+	git clone https://github.com/open-quantum-safe/liboqs
+	cd liboqs
+	mkdir build && cd build
+	cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=../.CMake/toolchain_windows-amd64.cmake -DOQS_DIST_BUILD=ON -DBUILD_SHARED_LIBS=ON ..
+	ninja
+
+The result of ninja process, will generate a `bin/liboqs.dll` file on project root directory that you will need to copy to Windows and configure it to be available via PATH environment variable
 
 ## Documentation
 


README.md
```diff
@@ -143,20 +143,37 @@ Binaries can be generated using Visual Studio 2019 with the [CMake Tools](https:
...
 	apt install gcc-mingw-w64
 	mkdir build && cd build
 	cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=../.CMake/toolchain_windows-amd64.cmake -DOQS_DIST_BUILD=ON ..
 	ninja
 
+#### Cross compiling LIBOQS from Ubuntu to Windows as library
+
+To compile for Windows AMD64 from Ubuntu Bionic and create a `liboqs.dll` file
+
+	sudo apt install cmake gcc gcc-mingw-w64 ninja-build
```
baentsch, line 168: please remove "sudo" here to be in line with the other documentation

README.md
```diff
@@ -143,20 +143,37 @@ Binaries can be generated using Visual Studio 2019 with the [CMake Tools](https:
...
+	git clone https://github.com/open-quantum-safe/liboqs
+	cd liboqs
+	mkdir build && cd build
+	cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=../.CMake/toolchain_windows-amd64.cmake -DOQS_DIST_BUILD=ON -DBUILD_SHARED_LIBS=ON ..
+	ninja
+
+The result of ninja process, will generate a `bin/liboqs.dll` file on project root directory that you will need to copy to Windows and configure it to be available via PATH environment variable
```
baentsch, line 175: I'm not a native speaker but shouldn't there be a "the" between `via` and `PATH`? Also, the DLL is created in the `build/bin` directory, not the project root directory.

baentsch (changes requested):
Please address the 2 small comments made: Thanks in advance!

README.md
```diff
@@ -143,20 +143,37 @@ Binaries can be generated using Visual Studio 2019 with the [CMake Tools](https:
 
 You can cross compile liboqs for various platform by supplying CMake with an appropriate [toolchain](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html) file.
 
+#### Cross compiling LIBOQS from Ubuntu to Raspberry Pi
```
thomwiggers, line 146:
```suggestion
#### Cross compiling liboqs from Ubuntu to Raspberry Pi
```

README.md
```diff
@@ -143,20 +143,37 @@ Binaries can be generated using Visual Studio 2019 with the [CMake Tools](https:
...
 For example, to cross compile for a Raspberry Pi from Ubuntu Bionic:
 
 	apt install gcc-8-arm-linux-gnueabihf
 	mkdir build && cd build
 	cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=../.CMake/toolchain_rasppi.cmake -DOQS_USE_OPENSSL=OFF ..
 	ninja
-	
-Or to compile for Windows AMD64 from Ubuntu Bionic:
+
+#### Cross compiling LIBOQS from Ubuntu to Windows
```
thomwiggers, line 155:
```suggestion
#### Cross compiling liboqs from Ubuntu to Windows
```

README.md
```diff
@@ -143,20 +143,37 @@ Binaries can be generated using Visual Studio 2019 with the [CMake Tools](https:
...
+
+To compile for Windows AMD64 from Ubuntu Bionic:
 
 	apt install gcc-mingw-w64
 	mkdir build && cd build
 	cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=../.CMake/toolchain_windows-amd64.cmake -DOQS_DIST_BUILD=ON ..
 	ninja
 
+#### Cross compiling LIBOQS from Ubuntu to Windows as library
```
thomwiggers, line 164:
```suggestion
#### Cross compiling liboqs from Ubuntu to Windows as library
```

README.md
thomwiggers:
```suggestion
	apt install cmake gcc gcc-mingw-w64 ninja-build
```

README.md
thomwiggers:
```suggestion
This will generate a `bin/liboqs.dll` file in the `build` directory that you will need to copy to Windows. You should configure it to be available via the PATH environment variable.
```

thomwiggers (changes requested):
Some minor changes necessary for the text.

However, some of these instructions are maybe slightly too specific scenarios for the README, and it might be better to put them in the wiki? Otherwise we might start seeing all possible permutations of (linux os, platform)...

baentsch:
> Some minor changes necessary for the text.
> 
> However, some of these instructions are maybe slightly too specific scenarios for the README, and it might be better to put them in the wiki? Otherwise we might start seeing all possible permutations of (linux os, platform)...

Agree -- but is there a way to bring in Wiki changes via the PR mechanism?

thomwiggers:
> Agree -- but is there a way to bring in Wiki changes via the PR mechanism?

AFAIK no, unless you extract the wiki into a repository (it _is_ accessible as a git repo) and set up CI to push that version to the wiki.

baentsch:
> AFAIK no, unless you extract the wiki into a repository (it _is_ accessible as a git repo) and set up CI to push that version to the wiki.

Yikes. This feels to be "a bit over the top" for just adding a single clarification. Unless we see a need for more of these (?), I'm tempted to suggest directly editing the Wiki as per your latest suggestions and then closing this PR. Any objections, @zerjioang @thomwiggers ?

zerjioang:
@baentsch @thomwiggers No objections from my side. But, you are right in putting everything in the readme might  be 'too much'. I think the best way to have a documentation under git control is to create a some sort of mkdocs or similar project, in where we can just write markdown files, and they are rendered as fancy HTML website. Everything could be handled via issues, pull-request and even automate it with Github actions. 

Take a look to an example documentation look like: https://squidfunk.github.io/mkdocs-material/getting-started/

baentsch:
@zerjioang Thanks for the PR changes and the good suggestion/pointer! Tagging @dstebila @xvzcf for thoughts/feedback. I personally feel this would be nice, but not utterly urgent. If someone volunteers changing over all documentation, perfect :-) Without feedback, I'll manually amend the Wiki as per this PR.

zerjioang:
@baentsch  okey then, so I close this PR in favor to wiki changes

baentsch:
Thanks again, @zerjioang: Update now [in the Wiki](https://github.com/open-quantum-safe/liboqs/wiki/Platform-specific-notes-for-building-liboqs#cross-building-a-windows-dll).

Issue #1000, opened 2021-05-13
Collecting per-build environment algorithm failures


baentsch:
In support of #892 this is to document build failures by algorithm and build environment.

Using our CI tests, downstream integrations and manual tests supporting [the wiki build instructions](https://github.com/open-quantum-safe/liboqs/wiki/Platform-specific-notes-for-building-liboqs), all algorithms not listed below should be assumed to build and work OK on
- Ubuntu, Debian, Alpine x86_64 using gcc 7, gcc 8 and clang 9
- MacOS using XCode 11.3
- Windows using MSVC19, cygwin (gcc 9 and 10)
- ARM: 64, hf

Failures:
BIKE: fails to build completely under MSVC, thousands of errors, e.g.,
![grafik](https://user-images.githubusercontent.com/57787676/118084839-a8623300-b3c1-11eb-8b14-c636b7ece5d3.png)

BIKE: fails also under cygwin (less dramatic), e.g.
```
[22/1510] Building ASM object src/kem/.../bike1_l1_fo.dir/additional/gf_mul.S.o
FAILED: src/kem/bike/CMakeFiles/bike1_l1_fo.dir/additional/gf_mul.S.o
/usr/bin/cc -DAVX2 -DFUNC_PREFIX=OQS_KEM_bike1_l1_fo -DLEVEL=1 -Iinclude -I/usr/include -fPIC -march=native -Werror -Wall -Wextra -Wpedantic -Wstrict-prototypes -Wshadow -Wformat=2 -Wfloat-equal -Wwrite-strings -O3 -fomit-frame-pointer -fdata-sections -ffunction-sections -Wl,--gc-sections -Wno-maybe-uninitialized -mavx2 -include /home/Administrator/git/liboqs/src/kem/bike/functions_renaming.h -MD -MT src/kem/bike/CMakeFiles/bike1_l1_fo.dir/additional/gf_mul.S.o -MF src/kem/bike/CMakeFiles/bike1_l1_fo.dir/additional/gf_mul.S.o.d -o src/kem/bike/CMakeFiles/bike1_l1_fo.dir/additional/gf_mul.S.o -c ../src/kem/bike/additional/gf_mul.S
../src/kem/bike/additional/gf_mul.S: Assembler messages:
../src/kem/bike/additional/gf_mul.S:70: Error: unknown pseudo-op: `.hidden'
[...]

PR #1001, opened 2021-05-14
BIKE and Picnic enablement for cygwin

baentsch:
Resolves some issues in #1000 as per input from @jschanck . Will eliminate caveats in https://github.com/open-quantum-safe/liboqs/wiki/Platform-specific-notes-for-building-liboqs#using-cygwin.


christianpaquin:
I don't have cygwin to test, but changes look good.

baentsch:
> I don't have cygwin to test, but changes look good.

I did check on a Windows VM with cygwin installed. Do you know whether one can also run cygwin in AppVeyor?

baentsch:
Please feel free to suggest a shorter matrix of configs to run: [cygwin building seems to be much slower than MSVC](https://ci.appveyor.com/project/dstebila/liboqs/builds/39183926).

Final question: Rainbow and Sphincs are completely exempt from CI: Is that really intentional/still OK? Why not at least test them in one configuration under (the faster) MSVC or msys2 environments?

dstebila (approved):

PR #1005, opened 2021-05-21
Consistent type in KEM decaps API (#1004)

dstebila:

### DIFF
--- a/src/kem/kem.h
+++ b/src/kem/kem.h
@@ -313,7 +313,7 @@ OQS_API OQS_STATUS OQS_KEM_encaps(const OQS_KEM *kem, uint8_t *ciphertext, uint8
  * @param[in] secret_key The secret key represented as a byte string.
  * @return OQS_SUCCESS or OQS_ERROR
  */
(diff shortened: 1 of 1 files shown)


jschanck:
Copy from upstream needs a few changes:
https://github.com/open-quantum-safe/liboqs/blob/3a56677a935e52b6b76edaa8f9312f0ba46a0398/scripts/copy_from_upstream/src/kem/family/kem_family.h#L17
https://github.com/open-quantum-safe/liboqs/blob/3a56677a935e52b6b76edaa8f9312f0ba46a0398/scripts/copy_from_upstream/src/kem/family/kem_scheme.c#L131

Also all of the lines starting with extern int here:
https://github.com/open-quantum-safe/liboqs/blob/3a56677a935e52b6b76edaa8f9312f0ba46a0398/scripts/copy_from_upstream/src/kem/family/kem_scheme.c#L34-L80

I will note that NIST and SUPERCOP use unsigned char instead of uint8_t. PQClean arguably leaves it up to the implementation. I can see an argument for unsigned char based on alignment rules. On the other hand, I think we're safe assuming that uint8_t is typedef'd to unsigned char for the systems that we support.

dstebila:
@jschanck The most recent commit should address your points.

jschanck (approved):

baentsch (approved):
Downstream changes should be minimal. Good to merge as far as I can see.

PR #1010, opened 2021-05-31
ARM64 via CCI machine

### LINKED ISSUE #1007: Change CircleCI to use native arm64 runners
https://circleci.com/docs/2.0/arm-resources/

baentsch:
Fixes #1007 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -384,8 +384,8 @@ workflows:
       - arm_machine:
           <<: *require_buildcheck
           name: arm64
-          PYTEST_ARGS: --ignore=tests/test_portability.py --numprocesses=auto --maxprocesses=10
-          CMAKE_ARGS: -DOQS_ENABLE_SIG_SPHINCS=OFF -DOQS_DIST_BUILD=ON
+          PYTEST_ARGS: --numprocesses=auto --maxprocesses=10
+          CMAKE_ARGS: -DOQS_DIST_BUILD=ON
       - arm_emulated:
           <<: *require_buildcheck
           name: armhf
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -346,10 +375,16 @@ workflows:
...
+      #- arm_emulated:
+      #    <<: *require_buildcheck
+      #    name: arm64
+      #    ARCH: arm64
+      #    CMAKE_ARGS: -DOQS_ENABLE_SIG_SPHINCS=OFF -DOQS_DIST_BUILD=ON
+      - arm_machine:
           <<: *require_buildcheck
           name: arm64
-          ARCH: arm64
+          PYTEST_ARGS: --ignore=tests/test_portability.py --numprocesses=auto --maxprocesses=10
```
jschanck, line 387: We should run test_portability. We don't have any code dispatching on ARM yet, but we will soon.

.circleci/config.yml
```diff
@@ -346,10 +375,16 @@ workflows:
...
           CMAKE_ARGS: -DOQS_ENABLE_SIG_SPHINCS=OFF -DOQS_DIST_BUILD=ON
```
jschanck, line 388: The test is completing in <4 minutes, so let's run SPHINCS.

jschanck (approved):
Looks good. Two minor suggestions above. Both are optional.

.circleci/config.yml
baentsch: Done with update

.circleci/config.yml
baentsch: Done with update

PR #1021, opened 2021-06-16
Add OQS_PERMIT_UNSUPPORTED_ARCHITECTURE flag

jschanck:
This PR adds a build flag (`OQS_PERMIT_UNSUPPORTED_ARCHITECTURE`) which bypasses the unsupported architecture error. This should only be used for testing on new architectures.


baentsch (approved):
LGTM. Only thought: What about adding a `cmake` message along the lines "Consider implementing suitable CPU optimization options function in `src/common/common.c` and contributing back the code" if `OQS_PERMIT_UNSUPPORTED_ARCHITECTURE` is set? That would not be as drastic as outputting such warning as a `printf` in the empty `set_available_cpu_extensions` function.

dstebila:
Adding some extra text in the compilation warning as Michael suggests seems like a good idea. Also we talked yesterday about a warning around risks of other architectures possibly having non-constant-time instructions. Would that go in there too, or somewhere else?

PR #1023, opened 2021-06-17
Remove Picnic's compatibility layer

### LINKED ISSUE #1022: OQS-OpenSSH v8 fails to build on macOS due to explicit_bzero
On macOS, building [our fork of OpenSSH v8](https://github.com/open-quantum-safe/openssh) against the latest liboqs commit fails with the following error:

```
ld: 1 duplicate symbol for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [ssh-agent] Error 1
duplicate symbol '_explicit_bzero' in:
    openbsd-compat//libopenbsd-compat.a(explicit_bzero.o)
    /Users/wxyz/Projects/openssh/src/oqs/lib/liboqs.a(compat.c.o)
```

The fix should include removing `explicit_bzero` [here](https://github.com/open-quantum-safe/liboqs/blob/main/tests/test_namespace.py#L35).

jschanck:
We've had some problems with duplicate symbols being introduced by picnic's OS/libc compatibility layer, which needs:
* aligned_alloc
* aligned_free
* explicit_bzero
* timingsafe_bcmp

We already provide everything except for timingsafe_bcmp in our common code. This PR adds an implementation of "timingsafe_bcmp" (`OQS_MEM_secure_bcmp`) to our common code, and then removes Picnic's compatibility layer.

Resolves #1022


christianpaquin (approved):

src/common/common.c
```diff
@@ -125,6 +125,18 @@ OQS_API void OQS_init(void) {
...
+OQS_API int OQS_MEM_secure_bcmp(const void *a, const void *b, size_t len) {
+	/* Assume CHAR_BIT = 8 */
+	uint8_t r = 0;
+
+	for (size_t i = 0; i < len; i++) {
+		r |= ((const uint8_t *)a)[i] ^ ((const uint8_t *)b)[i];
+	}
+
+	// We have 0 <= r < 256, and unsigned int is at least 16 bits.
+	return 1 & ((-(unsigned int)r) >> 8);
```
dstebila, line 137: I'm not very good with bit-fiddling code.  Just wanted to check you're confident this is correct.  I recall that a [recent FrodoKEM commit](https://github.com/microsoft/PQCrypto-LWEKE/commit/669522db63850fa64d1a24a47e138e80a59349db#diff-40aa21688f1ac58ddf8ef5f0f4e472e6075411f5c4f75b9c1d512b7e03a2cbabR120) to this type of code got it wrong so it's worth double-checking.

jschanck: The FrodoKEM code tries to work entirely with 16 bit integers, which makes it more complicated than it needs to be.

The trick is to cast to a larger width integer before doing the negation.

As the comment says, we have 0 <= r < 256, and unsigned int is guaranteed to be at least 16 bits. So we have
```
 (unsigned int) r = ... 0000 0000 xxxx xxxx
```
where by "..." I mean there might be more than 8 leading 0s.
There are two cases to consider, r=0 and r!=0.

If r=0 then
```
- (unsigned int) r = ... 0000 0000 0000 0000
```
Shifting right 8 places and extracting the low bit gives 0.

If 1 <= r < 255 then, -(unsigned int) r is between UINT_MAX-255 and UINT_MAX (inclusive).
So in binary it looks like:
```
- (unsigned int) r = ... 1111 1111 xxxx xxxx
```
Shifting right 8 places and extracting the low bit gives 1.

dstebila: Thanks for the explanation!

PR #1030, opened 2021-06-23
New datasheets

xvzcf:
Opening this as a draft PR to get feedback. The main thing to look at is `update_pqclean_alg_docs.py` in the latest commit.

- hqc.yml is formatted differently as it was generated by PyYAML (the other non-pqclean datasheets can also be similarly re-generated without much work)
- `update_pqclean_alg_docs.py` can be incorporated into `copy_from_upstream.py` once we're set.


baentsch (commented):
As requested a quick review: 1) the "import" logic for SIGs seems to be missing. 2) The code simply seems to be overriding fields possibly already existing in the liboqs .yml with PQClean META.yml input, right? Wouldn't it be reasonable to output something in case this happens (such as to clean up our data)?

xvzcf:
> As requested a quick review: 1) the "import" logic for SIGs seems to be missing. 

Yeah, will just add this in late.

> The code simply seems to be overriding fields possibly already existing in the liboqs .yml with PQClean META.yml input, right? Wouldn't it be reasonable to output something in case this happens (such as to clean up our data)?

I'm not sure I follow, do you mean having a "verbose" mode that logs the replacements? If so, we could also have a "dry-run" mode that just lists the changes without actually making them.

baentsch:
>  If so, we could also have a "dry-run" mode that just lists the changes without actually making them.

Yes - that would do.

bhess:
As discussed yesterday, I updated sike.yml with its different variants/implementations to the new yml structure. Also adapted the yml to pass a `ymllint` check (indentions, document start, line lengths).

baentsch:
>  Also adapted the yml to pass a ymllint check (indentions, document start, line lengths).

Sounds like a good idea! What about adding it as a script for all YMLs (and execute in CI)?

bhess:
> What about adding it as a script for all YMLs (and execute in CI)?

In 3886b87 I added a yamllint run on docs/algorithms as part of test_code_conventions. In the openquantumsafe/ci-ubuntu-focal-x86_64 image, yamllint would need to be installed (available with apt).

baentsch:
Tagging @dkostic : Would you want to check `docs/algorithms/kem/bike.yml` in this PR for correctness (as it's the basis for our documentation on BIKE going forward)? Please feel free to push necessary changes directly into this branch. @xvzcf @dstebila :  Even `yamllint` passes on this file, so  I'm considering done what I promised unless I hear otherwise.

dkostic:
> Tagging @dkostic : Would you want to check `docs/algorithms/kem/bike.yml` in this PR for correctness (as it's the basis for our documentation on BIKE going forward)?

The only thing I'm not sure about are lines 39 and 63, `implementations-switch-on-runtime-cpu-features`. I think that should be set to `true` because our implementation has a runtime cpu features switch? If you agree with that I can push the change.

baentsch:
>  If you agree with that I can push the change.

Can you please point to that in the code? I probably was blind in that regard...

dkostic:
> Can you please point to that in the code? I probably was blind in that regard...

Here it is:
https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/bike/additional_r3/decode_internal.h#L61-L82
https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/bike/additional_r3/gf2x_internal.h#L131-L175
https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/bike/additional_r3/sampling_internal.h#L46-L61

baentsch:
Thank you very much, @dkostic : Just updated `implementations-switch-on-runtime-cpu-features` suitably.

dstebila:
@xvzcf I've updated FrodoKEM-640-AES based on what I saw in some of the other YML files.  Can you take a look and let me know if that seems right? In terms of implementations, there's a single upstream implementation (which I've just called 'master') but if AVX2 is present then it will do some optimizations. Please check if I represented that appropriately in the YAML. Once I get the go-ahead I'll do the remaining 5 variants.

xvzcf:
> @xvzcf I've updated FrodoKEM-640-AES based on what I saw in some of the other YML files. Can you take a look and let me know if that seems right? In terms of implementations, there's a single upstream implementation (which I've just called 'master') but if AVX2 is present then it will do some optimizations. Please check if I represented that appropriately in the YAML. Once I get the go-ahead I'll do the remaining 5 variants.

Yes, the YAML looks good, you can go ahead.

xvzcf:
@dstebila this is ready to go ([here's](https://github.com/open-quantum-safe/liboqs/blob/new-datasheets/docs/algorithms/kem/ntru.md) an example of a generated markdown file). There are some TODOs but none of them are blockers. To generate new markdown files:

1) (Optional) Make sure the PQClean YAML files are current by running the following in the liboqs root directory:

```
./scripts/copy_from_upstream/update_pqclean_alg_docs.py --liboqs-root=. -v 1 -w
```
2) (Optional) Ensure the YAML files are uniformly formatted.

```
python3 scripts/format_docs_yaml.py
```

3) Generate the KEM markdown files:

```
python3 scripts/gen_docs_md_from_yaml.py
```

This process can also be improved as we add the signature YAML files (and can be added to the wiki).

dstebila (approved):

dstebila:
Thanks @xvzcf and everyone else who contributed!

PR #1033, opened 2021-06-24
Make BIKE compile for older CMake versions when AVX512 is used

### LINKED ISSUE #1032: BIKE does not compile on some X86_64 architectures with AVX512
See the thread https://github.com/open-quantum-safe/boringssl/issues/64 for the description of the issue and a suggestion for a fix.

geovandro:
BIKE compilation fails when AVX512 and older CMake versions (like 3.10.*) are to be used.

Fixes #1032, resolves https://github.com/open-quantum-safe/boringssl/issues/64.


baentsch (approved):
LGTM. Thanks again!

dstebila (approved):

PR #1039, opened 2021-06-29
PQClean / NTRU Prime update

### LINKED ISSUE #1026: Pull updated NTRU Prime from PQClean

baentsch:
- Fixes #1026 
- Updates Kyber768 reference
- Removes McEliece patch


src/common/pqclean_shims/compat.h
```diff
@@ -0,0 +1,49 @@
...
+UNALIGNED_VECTOR_POLYFILL_GCC
+#  endif
+
+#elif defined(MSVC)
+// MSVC simply doesn't have these types
+#define __m256_u    __m256
+#define __m256d_u   __m256d
+#define __m256i_u   __m256i
+
+#else
```
jschanck, line 43: ICC doesn't have the __m256x_u types either. Honestly I think it was a mistake for gcc to introduce them, and I can't see a situation where they're necessary. I would change the `#elif defined(MSVC)` to `#else` and remove the error case.

jschanck (approved):

src/common/pqclean_shims/compat.h
baentsch: So done.

Issue #1038, opened 2021-06-29
SPHINCS+ memcpy source and destination overlap - undefined behavior


bhess:
The SPHINCS+ implementations have cases of memcpy use where the source and destination overlap. According to the C standard and posix, memcpy behavior is undefined if memory regions overlap.

Detected using valgrind on ppc64le/Ubuntu focal. Memcpy implementations vary, so it seems to be not detected with valgrind on x86_64.

The cause in `gen_chain`:

https://github.com/open-quantum-safe/liboqs/blob/d9fb4e0e81ccd0a0972b38743404d982fe0a90c1/src/sig/sphincs/pqclean_sphincs-haraka-128f-robust_clean/wots.c#L37-L44

Used for example by `wots_gen_pk`, where src and dst are the same. The replicated code of all variants is affected:

https://github.com/open-quantum-safe/liboqs/blob/d9fb4e0e81ccd0a0972b38743404d982fe0a90c1/src/sig/sphincs/pqclean_sphincs-haraka-128f-robust_clean/wots.c#L122-L123

Using memmove would be the safe alternative, or avoid memcpy if src and dst are the same.

Below is the valgrind log. It's part of a constant-time check, but the issues detected are because of overlapping memory.
[ppc64le.txt](https://github.com/open-quantum-safe/liboqs/files/6731356/ppc64le.txt)

baentsch:
Yikes -- good catch! Hard to believe the upstreams didn't encounter this before, at least PQClean: Shouldn't this issue then rather be opened there?

bhess:
> PQClean: Shouldn't this issue then rather be opened there?

Yes, thanks. Created an issue in PQClean that links to the one here.

baentsch:
Is this still an issue with the new Sphincs+ code or can this be closed by now?

bhess:
I checked and it appears to not be an issue anymore with the current version. Closing.

PR #1041, opened 2021-06-29
Suppression file update after ppc64le constant-time check

bhess:
Minor suppression file updates after a constant-time check on ppc64le (see https://github.com/open-quantum-safe/liboqs/pull/1037):

- Dilithium: (i) two line numbers changed, (ii) using explicit function name (fun:rej_uniform) for generic Dilithium
- ~~Adds the new BIKE variants (adding to issues.json like the previous variants, expert review might be needed here)~~

The CI log with BIKE and Dilithium reports is available here: [ppc64le.txt](https://github.com/open-quantum-safe/liboqs/files/6733709/ppc64le.txt)
The SPHINCS+ reports are documented separately in #1038.

The reported issues (except #1038) should also be reproducible using the `constant-time-x64-extensions` / `constant-time-x64` circleci jobs. After this fix, all constant-time checks pass on x86_64 / ppc64le. BIKE will have to be added separately after this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: c
+before_script:
+  - sudo apt -y install astyle cmake gcc ninja-build libssl-dev python3-pytest python3-pytest-xdist unzip xsltproc doxygen graphviz valgrind
+jobs:
+  include:
+    - arch: ppc64le         # The IBM Power LXD container based build for OSS only
+      os: linux             # required for arch different than amd64
+      dist: focal           # or bionic | xenial with xenial as default
+      compiler: gcc
+      script:
+        - mkdir build && cd build && cmake -GNinja .. && cmake -LA .. && ninja
+        - cd build & ninja run_tests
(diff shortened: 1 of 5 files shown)


jschanck:
The new BIKE code has some non-constant time subroutines for some platforms, but there's definitely an attempt to be constant time on recent Intel CPUs. It will take quite a bit of work to fully document this.

Before we start, it also looks like we're not always using the best BIKE code available. For instance we're using `gf2x_mul_base_port` instead of `gf2x_mul_base_pclmul` when compiling specifically for Haswell, which has PCLMUL. This is causing a lot of constant time warnings. The `cpu_features.h` file needs to be rewritten to be aware of the dist build setting and it needs to fall back to the oqs_config.h defines for non-dist builds.

dstebila:
> The new BIKE code has some non-constant time subroutines for some platforms, but there's definitely an attempt to be constant time on recent Intel CPUs. It will take quite a bit of work to fully document this.
> 
> Before we start, it also looks like we're not always using the best BIKE code available. For instance we're using `gf2x_mul_base_port` instead of `gf2x_mul_base_pclmul` when compiling specifically for Haswell, which has PCLMUL. This is causing a lot of constant time warnings. The `cpu_features.h` file needs to be rewritten to be aware of the dist build setting and it needs to fall back to the oqs_config.h defines for non-dist builds.

Should we create a separate issue to track this, allowing us to merge this PR without waiting on that?

bhess:
> Should we create a separate issue to track this, allowing us to merge this PR without waiting on that?

I can remove the BIKE part from this PR, so this can be tracked and handled in a separate issue.

bhess:
Removed the BIKE part of this PR for the time being.

dstebila:
Is this ready to merge?

bhess:
yes, it's ready from my side.

bhess:
Replaced by #1048 (with branch on open-quantum-safe project to be able to test TravisCI)

Issue #1043, opened 2021-06-29
Remove OQS_KEM_DEFAULT and OQS_SIG_DEFAULT


dstebila:
The original motivation for these (allowing easy swapping of which algorithm is used in the OQS-OpenSSL integration) is no longer needed since we have all algorithms now available in OQS-OpenSSL.

Some people use this to create a minimal build with just one algorithm (combined with the MINIMAL\_BUILD flag).  While it's possible to do that with the existing CMake flags, it woudl require manually setting many of them to off.  So we should create a better set of flags for this.

baentsch:
> So we should create a better set of flags for this.

Please check #1044 for a proposal for this (basically, passing a semicolon-delimited list of algorithms to _enable_ -- all others will automatically be disabled).

baentsch:
Closed with #1044

PR #1044, opened 2021-06-30
Removing default algs

baentsch:
Begins to address #1043, retaining `OQS_MINIMAL_BUILD´ option: The latter now can be fed with a string of algorithms to be enabled, disabling all others.

Current state meant as intermediate state to check with @jschanck whether this approach is OK. If so, actual removal of DEFAULT algs will be added to this PR.

### DIFF
--- a/.CMake/alg_support.cmake
+++ b/.CMake/alg_support.cmake
@@ -2,11 +2,11 @@
 
 include(CMakeDependentOption)
 
-if(NOT DEFINED OQS_KEM_DEFAULT)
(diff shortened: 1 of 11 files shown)


dstebila:
Thanks Michael, looks like a great start! Syntax and approach seems reasonable to me, so if it satisfies @jschanck I think it's good to go.

jschanck:
OK by me. Thanks!

dstebila (approved):

baentsch:
Merging this will most likely cause a conflict in #1045. So please decide which PR to merge first, @dstebila @jschanck .

PR #1045, opened 2021-06-30
Remove old versions of BIKE

jschanck:
- Does this PR change the the list of algorithms available -- either adding, removing, or renaming?  (If so, PRs in OQS-OpenSSL, OQS-BoringSSL, and OQS-OpenSSH will also be required by the time this is merged.)

This PR removes `bike1_l1_cpa`, `bike1_l3_cpa`, `bike1_l1_fo`, and `bike1_l3_fo`. Resolves #1024.

This PR also improves the compilation of round 3 BIKE. In particular the `cpu_features.h` file is now aware of both dist and non-dist builds and it selects the best available code in both cases.

PR #1047, opened 2021-07-01
removing BIKE R2 from README

baentsch:
Correcting reference removal in #1045.
This intentionally does not fix the equally incorrect https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/kem/bike.md assuming that will be fixed soon by #1030.


drucker-nir (approved):
For future BIKE related issues and PRs, I recommend making Dusan Kostic (@dkostic) a reviewer.

PR #1052, opened 2021-07-20
Adding armv8 crypto extensions to SHA256/224

Martyrshot:
This PR adds the armv8 crypto extensions accelerated implementation of sha256 hashblocks from supercop. This was tested on a ROCK64, a RASPI 3B+, and on an Intel mac (to ensure I didn't break x86 builds by mistake).

* [ ] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.) - **NO**
* [ ] Does this PR change the the list of algorithms available -- either adding, removing, or renaming?  (If so, PRs in OQS-OpenSSL, OQS-BoringSSL, and OQS-OpenSSH will also be required by the time this is merged.) - **NO**


dstebila:
Stylecheck failed because you were using an old version of the Doxygen configuration file.  I've done a rebase on main and force pushed.

Martyrshot:
Ah shoot, I thought I forgot something. Sorry about that!

src/common/CMakeLists.txt
```diff
@@ -30,7 +30,11 @@ endif()
 if(${OQS_USE_SHA2_OPENSSL})
     set(SHA2_IMPL sha2/sha2_ossl.c)
 else()
-    set(SHA2_IMPL sha2/sha2_c.c)
+    set(SHA2_IMPL sha2/sha2.c sha2/sha2_c.c)
+    if (OQS_DIST_ARM64v8_BUILD OR OQS_USE_ARM_SHA2_INSTRUCTIONS)
+      set(SHA2_IMPL ${SHA2_IMPL} sha2/sha2_ni.c)
+      # set_source_files_properties(sha2/sha2_ni.c PROPERTIES COMPILE_FLAGS -march=armv8-a+crypto)
```
dstebila, line 36: Is this commented-out line required?

src/common/sha2/sha2.c
```diff
@@ -0,0 +1,132 @@
+// SPDX-License-Identifier: Public domain
```
dstebila, line 1: Can you change this to MIT?

src/common/sha2/sha2.c
```diff
@@ -0,0 +1,132 @@
...
+    } while(0)
+#elif defined(OQS_USE_ARM_SHA2_INSTRUCTIONS)
+#define C_OR_NI(stmt_c, stmt_ni) \
+    stmt_ni
+#else
+#define C_OR_NI(stmt_c, stmt_ni) \
+    stmt_c
+#endif
+void OQS_SHA2_sha224_inc_init(OQS_SHA2_sha224_ctx *state) {
+	oqs_sha2_sha224_inc_init_c((sha224ctx *) state);
```
dstebila, line 26: Why are these casts necessary?

src/common/sha2/sha2_c.c
```diff
@@ -2,51 +2,53 @@
...
 #include <oqs/oqs.h>
 
-#include "sha2.h"
+#include "sha2_local.h"
+#include <stdio.h>
 
 #define PQC_SHA256CTX_BYTES 40
 #define PQC_SHA512CTX_BYTES 72
-
+/*
```
dstebila, line 10: Why is this commented out code still here?

dstebila (changes requested):

src/common/sha2/sha2_c.c
Martyrshot: Good question, I'll remove it!

src/common/sha2/sha2.c
Martyrshot: Will do!

src/common/sha2/sha2.c
Martyrshot: I added them because the generic OQS_SHA2_sha*_ctx structs contain a void pointer, and I was getting typing errors for using the more specific sha*ctx structs which contain a uint8_t pointer. This seemed like the easiest way to keep the OQS_SHA2_sha*_ctx structs general while not having to cast the void pointer to a uint_8t every access. Do you have any recommendations on how to do this in a simpler/cleaner way?

src/common/CMakeLists.txt
Martyrshot: Looks like in some cases it is, so I've updated this to be more flexible.

src/common/sha2/sha2.c
dstebila: No I guess that makes sense.

Martyrshot:
Do we want to consider changing the sha2 context structures to contain a uint8_t pointer rather than a void pointer? Not exactly related to adding armv8 crypto extensions, but since the code is getting updated anyway it could be a simple change to do while we are here?

src/common/CMakeLists.txt
```diff
@@ -30,7 +30,14 @@ endif()
 if(${OQS_USE_SHA2_OPENSSL})
     set(SHA2_IMPL sha2/sha2_ossl.c)
 else()
-    set(SHA2_IMPL sha2/sha2_c.c)
+    set(SHA2_IMPL sha2/sha2.c sha2/sha2_c.c)
+    if (OQS_DIST_ARM64v8_BUILD)
+       set(SHA2_IMPL ${SHA2_IMPL} sha2/sha2_ni.c)
+       set_source_files_properties(sha2/sha2_ni.c PROPERTIES COMPILE_FLAGS -mcpu=cortex-a53+crypto)
+    elseif (OQS_USE_ARM_SHA2_INSTRUCTIONS)
+       # Assume if we are here we are comopiling native
```
baentsch, line 38: Typo in comment. But I'm wondering more what/who sets `OQS_USE_ARM_SHA2_INSTRUCTIONS`? Or should this go into the build instructions Wiki?

baentsch:
> Do we want to consider changing the sha2 context structures to contain a uint8_t pointer rather than a void pointer? 

Conceptually and personally I'd be all for accuracy wrt data structure types -- however, the `void *` in my eyes is an indication that it's _not_ always a `uint8_t`: In the OSSL case, it's effectively a `EVP_MD_CTX*`. Thus, I'd keep it as is, but defer to @dstebila who put it like it is right now.

src/common/CMakeLists.txt
dstebila: `OQS_USE_ARM_SHA2_INSTRUCTIONS` is set by CMake in `.CMake/detect_gcc_clang_intrinsics.c`.

dstebila:
> > Do we want to consider changing the sha2 context structures to contain a uint8_t pointer rather than a void pointer?
> 
> Conceptually and personally I'd be all for accuracy wrt data structure types -- however, the `void *` in my eyes is an indication that it's _not_ always a `uint8_t`: In the OSSL case, it's effectively a `EVP_MD_CTX*`. Thus, I'd keep it as is, but defer to @dstebila who put it like it is right now.

Indeed, the point of containing a `void *` pointer was to permit multiple instantiations which use different types of context structures.

src/common/CMakeLists.txt
Martyrshot: Thank you for catching that! I just updated it.

Martyrshot:
Makes sense to me!

src/common/CMakeLists.txt
```diff
@@ -30,7 +30,14 @@ endif()
 if(${OQS_USE_SHA2_OPENSSL})
     set(SHA2_IMPL sha2/sha2_ossl.c)
 else()
-    set(SHA2_IMPL sha2/sha2_c.c)
+    set(SHA2_IMPL sha2/sha2.c sha2/sha2_c.c)
+    if (OQS_DIST_ARM64v8_BUILD)
+       set(SHA2_IMPL ${SHA2_IMPL} sha2/sha2_ni.c)
+       set_source_files_properties(sha2/sha2_ni.c PROPERTIES COMPILE_FLAGS -mcpu=cortex-a53+crypto)
```
dstebila, line 36: Actually just realizing that this line might be problematic.  It's setting `-mcpu=cortex-a53+crypto` but is guarded only by `if (OQS_DIST_ARM64v8_BUILD)`; but there will be certainly many other ARM64v8 CPU's other than Cortex-A53, and even some which don't have crypto extensions (e.g., our Raspberry Pi 3B's which don't versus our ROCK64's which do).  Have you tried compiling and running this on other ARM64v8 processors?  We have the rasp3b's, the rock64's, as well as an Apple Silicon.

Martyrshot: I have tried it on the roc, raspis and aws instance. I'd be happy to test on apple silicon, but I would need access to a machine since I only have an intel mac. The reason I picked that flag, in particular, is because I have to specify a core in order to "turn on" crypto extensions when compiling on a system that doesn't have them (the `+crypto` portion of `-mcpu`), and `a53` is the `generic` core used by liboqs. Without that compile flag the compiler will error saying that the extension isn't supported, but since we do a runtime check when `OQS_DIST_ARM64v8_BUILD` is specified, so we need to have those intrinsics compiled into the library.

Martyrshot: Looks like there is an unrelated compiling issue when using `OQS_DIST_BUILD` I get the following error:

```
/Users/jgoertzen/stock-oqs/liboqs/src/common/common.c:74:10: fatal error: 'sys/auxv.h' file not found
#include <sys/auxv.h>
         ^~~~~~~~~~~~
1 error generated.
```

If I don't use `OQS_DIST_BUILD` my branch builds and all tests pass. If this is intentional then the above concern isn't an issue, otherwise, I can take a look into finding a solution.

dstebila: I think we do want to be able to compile in that configuration, so can you look into it?  I think what might be going on there is that the lines 74-92 of common.c are protected by `OQS_DIST_ARM64v8_BUILD` which in principle would include both Linux ARM (e.g., Raspberry Pi's) and Apple ARM, but the code in that section is solely Linux-focused; for example I think neither `<sys/auxv.h>` nor `getauxval` is available on macOS.  [This site](https://github.com/google/cpu_features/pull/150) shows one example of reading the CPU features on macOS. Although for Apple chips it seems like the features we want are always available (at least for now), so maybe we can simplify and just set them automatically without testing?

Martyrshot: Sounds good, I'll see what I can figure out. Would you like me to create a new PR or issue specifically for detecting apple M1 cpu features?

dstebila: I think I've done it. It runs for me our M1, but I haven't tested to make sure it doesn't break things on our other ARMs. Can you give a try?

Martyrshot: Oh, for sure! Thanks for doing that! I'll try it out right away.

src/common/common.c
```diff
@@ -71,6 +71,27 @@ static void set_available_cpu_extensions(void) {
...
+	} else {
+		return (p != 0) ? 1 : 0;
+	}
+}
+static void set_available_cpu_extensions(void) {
+	/* mark that this function has been called */
+	cpu_ext_data[OQS_CPU_EXT_ARM_AES] = 1;
+	cpu_ext_data[OQS_CPU_EXT_ARM_SHA2] = 1;
+	cpu_ext_data[OQS_CPU_EXT_ARM_SHA3] = macos_feature_detection("hw.optional.armv8_2_sha3");
+	cpu_ext_data[OQS_CPU_EXT_ARM_NEON] = macos_feature_detection("hw.optional.neon");
```
Martyrshot, line 91: Should `macos_feature_detection` return an unisnged int? Currently, this causes warnings for implicit conversion

Martyrshot: Looking at the function, I think returning an unsigned int makes the most sense. I've pushed an updated version and am testing it on all the arm platforms I have access to.

src/common/CMakeLists.txt
Martyrshot: I made some small changes and tested on the roc, raspi, was, and apple silicon machine and all tests passed. I also verified that on the apple silicon machine that the sha256 instructions were being used rather than the c implementation. So it should be good to go!

PR #1058, opened 2021-07-30
Add CI job for x86

dstebila:
Add a continuous integration job for building liboqs on a 32-bit Intel platform.

Fixes #1049.

Depends on https://github.com/open-quantum-safe/ci-containers/pull/55 and https://github.com/PQClean/PQClean/pull/409.

PR #1061, opened 2021-07-30
Sync with PQClean

dstebila:
Sync with PQClean https://github.com/PQClean/PQClean/commit/6c1ea921ee4a06a6b50c742ca540bb9b5e51aee2. 

Related to #1058.


baentsch (changes requested):
The contents look good, but the documentation didn't get updated (git commit hash unchanged in docs/algorithms/sig/falcon.XXX): It seems to be necessary to manually run `update_pqclean_alg_docs.py` -> would it be an idea to trigger this when running `copy_from_upstream.py`?

dstebila:
> The contents look good, but the documentation didn't get updated (git commit hash unchanged in docs/algorithms/sig/falcon.XXX): It seems to be necessary to manually run `update_pqclean_alg_docs.py` -> would it be an idea to trigger this when running `copy_from_upstream.py`?

Good catch, Michael. Fixed now.

baentsch (approved):

xvzcf:
It would be nice to trigger the documentation update in `copy_from_upstream.py`. 

My thought was to first extend `update_pqclean_alg_docs.py` to also support non-PQClean algorithms (such as using the META.yml at https://github.com/pq-crystals/dilithium), and subsequently fold this script into `copy_from_upstream.py`.

baentsch:
Makes sense. Now tracked in #1064.

PR #1063, opened 2021-08-02
use doc YMLs to count non-copy-from-upstream algs [skip ci]

### LINKED ISSUE #1040: Automate non-upstream algorithm inclusion
As per [this discussion](https://github.com/open-quantum-safe/liboqs/issues/1026#issuecomment-869970147) this issue shall document the need to correct the `copy_from_upstream` logic once #1030 has landed.

baentsch:
Fixes #1040 

### DIFF
--- a/scripts/copy_from_upstream/copy_from_upstream.py
+++ b/scripts/copy_from_upstream/copy_from_upstream.py
@@ -17,6 +17,8 @@ import json
 # kats of all algs
 kats = {}
 
+non_pqclean_lengths = {}
+
 parser = argparse.ArgumentParser()
 parser.add_argument("-v", "--verbosity", type=int)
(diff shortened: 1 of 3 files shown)


bhess (approved):
Looks good to me, thanks.
Just a nit-pick, I think technically it's not "non-PQClean algs" but "non-copy_from_upstream" algs. (i.e. the CRYSTALS algs are not from pqclean but still automatically added to the count)

dstebila (approved):

dstebila:
I'll assume that you ran the updated copy\_from\_upstream script and it yielded the same output as the previous version.

baentsch:
> I'll assume that you ran the updated copy_from_upstream script and it yielded the same output as the previous version.

Yes.

PR #1065, opened 2021-08-03
Remove UNUSED macro from OQS API

dstebila:
The UNUSED macro in common.h is not actually being used in the liboqs codebase anymore, so it can be removed. This also reduces risk of namespace conflict between macros defined via oqs.h and other application.

As suggested by @Martyrshot.

### DIFF
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -205,23 +205,6 @@ void *OQS_MEM_aligned_alloc(size_t alignment, size_t size);
  */
 void OQS_MEM_aligned_free(void *ptr);
 
-/**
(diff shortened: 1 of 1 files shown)


baentsch:
The change conceptually looks OK and I'm not aware of downstream code under our control that depends on it. But before approving it I'd like to also make @jyao1 aware of this change as it may have an impact on the UEFI integration: Could you please check/give feedback on this proposed change, @jyao1 ?

jyao1:
Please remove it. I have no concern from UEFI side.

baentsch (approved):

PR #1067, opened 2021-08-04
Compile-time guard for some BIKE AVX2 and AVX512 code

dstebila:
I encountered a BIKE build failure on a non-AVX512 platform. BIKE's run-time dispatch protects against calls to a AVX512-dependent function, but the reference to that AVX512-dependent function is present despite that function not being compiled, which was leading to a linker error.

@dkostic Can you review?


dstebila:
Also needed around some similar AVX2 code.

dkostic (approved):

PR #1068, opened 2021-08-05
Update Valgrind suppression files for recent implementation changes

dstebila:
Updating the Valgrind suppression files to deal with the following recent changes:
- NTRU Prime code changed slightly so line offset has to change.
- BIKE variants changed names.

### DIFF
--- a/tests/constant_time/kem/issues.json
+++ b/tests/constant_time/kem/issues.json
@@ -1,8 +1,6 @@
 {
-  "BIKE1-L1-CPA": ["bike_has_no_timing_protections"],
-  "BIKE1-L1-FO": ["bike_has_no_timing_protections"],
-  "BIKE1-L3-CPA": ["bike_has_no_timing_protections"],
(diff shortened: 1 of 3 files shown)


dstebila:
I've run the `constant-time-x64` CircleCI job locally which does constant-time testing on the generic implementation and it passes.  Currently locally running (probably will take all night) the `constant-time-x64-extensions` job which does constant-time testing when configured with `OQS_OPT_TARGET=auto`. Will report on results once it's done. 

Note that the `constant-time-x64-extensions` job won't compile unless #1067 is applied, but we can merge these two PRs independently since that CI job is actually commented out and only ever run manually (at least for now).

dstebila:
> I've run the `constant-time-x64` CircleCI job locally which does constant-time testing on the generic implementation and it passes. Currently locally running (probably will take all night) the `constant-time-x64-extensions` job which does constant-time testing when configured with `OQS_OPT_TARGET=auto`. Will report on results once it's done.

The ``constant-time-x64-extensions` job passed on my machine.

Issue #1025, opened 2021-06-18
Deal with CI backlog


dstebila:
@xvzcf You had some concerns about backlog of incomplete coverage in CI, e.g. some algorithms being disabled on AppVeyor.  Can you document those in this issue?

xvzcf:
Yes, was going to do so later today.

xvzcf:
We might want to go over these before adding new CI jobs:

- [appveyor_build.bat](https://github.com/open-quantum-safe/liboqs/blob/main/appveyor_build.bat#L10): SPHINCS and Rainbow are disabled
- [.circleci/config.yml](https://github.com/open-quantum-safe/liboqs/blob/main/.circleci/config.yml#L344): There are several disabled jobs
- A commit to `main` triggers various downstream sub-project jobs, and our CircleCI plan has a limit of 1 concurrent job
- This triggering mechanism currently only applies to CircleCI jobs, and not to Appveyor or GitHub Actions ones
- OpenSSL and OpenSSH (possibly others) CI jobs don't test the full suite of algorithms because only a subset is enabled by default (this generally hasn't caused us too many problems)
- Our plan allows "1 concurrent job with access to Linux and Windows builds on Small and Medium sizes", and it would be nice to consolidate CI systems (subject to all the considerations stemming from the previous points)

I also noticed [here](https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/1554/workflows/21c8fdf0-1572-47dd-bd8c-d88dbcb95998/jobs/14491) that the triggering mechanism doesn't fail if the branch wasn't found. I introduced this bug, will fix it.

baentsch:
If I get it right, every single item above was triggered by the need to save (CI) time, either because the algorithms are too slow (i.e., we effectively did pick (on :) "losers") or the tests are too time-consuming/not adding serious value (to warrant their regular execution).

So I see some alternatives for resolution:
1) We buy more CCI resources (for parallel execution).
2) We create a more convoluted "tier system" of tests (PR run, audit-run, release-run, etc.) with different levels of "thoroughness" (i.e., time/resource demands)
3) We "mix" testing options, i.e., not doing full matrix tests, but do testing of specific features/algorithms only on specific setups (say, Rainbow only on shared x86 builds and Sphincs only on static arm builds). Absolutely non-ideal but may allow us to test more within the given CI resource budget.
[...]

PR #1069, opened 2021-08-05
Use Github Actions for some CI

dstebila:
In hopes of partially alleviating pressure on our CircleCI quota as per #1025, this PR switches some jobs to use Github Actions rather than CircleCI.  In particular, I've started off with more "standalone" type jobs, which include:
- alpine (2 jobs - with OpenSSL and without)
- address sanitizer (1 job) (currently disabled until #1067 lands)
- ARM emulation jobs (1 job - armhf)

Nothing stops us from moving other jobs over as well, such as the main Ubuntu x86_64 and i386 jobs, or the macOS jobs, or even the Windows jobs.  But I wanted to get initial feedback and discussion of strategy.  This currently doesn't affect triggering of downstream jobs (when a merge to main happens) or Travis/Appveyor (except for filtering branches of the form `ghactionsonly-*` from being run on those other CI platforms, allowing us to test just this one CI system).

Although the build shows as red, if the only failure is CircleCI, then that's due to there being no jobs that are meant to be run on CircleCI due to the `ghactionsonly-*` filtering currently going on.  That problem should not be present on other branches.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,56 +118,6 @@ jobs:
       - store_artifacts:
           path: build/test-results
 
-  arm_emulated:
-    description: A template for running liboqs tests on emulated ARM Docker VMs
-    parameters:
-      ARCH:
-        description: "The desired ARM architecture to be emulated."
-        type: string
-      CMAKE_ARGS:
-        description: "Arguments to pass to CMake."
-        type: string
-      PYTEST_ARGS:
-        description: "Arguments to pass to pytest."
-        type: string
-    machine:
-      image: ubuntu-1604:201903-01
-    steps:
-      - checkout
-      - run:
-          name: Install the emulation handlers
-          command: docker run --rm --privileged multiarch/qemu-user-static:register --reset
-      - run:
-          name: Build in an x86_64 container
-          command: |2
-            docker run --rm \
-                       -v `pwd`:`pwd` \
-                       -w `pwd` \
-                       openquantumsafe/ci-debian-buster-amd64:latest /bin/bash \
-                       -c "mkdir build && \
-                           (cd build && \
-                            cmake .. -GNinja << parameters.CMAKE_ARGS >> \
-                                     -DCMAKE_TOOLCHAIN_FILE=../.CMake/toolchain_<< parameters.ARCH >>.cmake && \
-                            cmake -LA .. && \
-                            ninja)"
-      - run:
-          name: Run the tests in an << parameters.ARCH >> container
-          command: |2
-            docker run --rm -e SKIP_TESTS=style,mem_kem,mem_sig \
-                            -v `pwd`:`pwd` \
-                            -w `pwd` \
-                            openquantumsafe/ci-debian-buster-<< parameters.ARCH >>:latest /bin/bash \
-                            -c "mkdir -p tmp && \
-                                python3 -m pytest --verbose \
-                                                  --numprocesses=auto \
-                                                  << parameters.PYTEST_ARGS >> \
-                                                  --ignore=tests/test_code_conventions.py \
-                                                  --junitxml=build/test-results/pytest/test-results.xml"
-      - store_test_results:
-          path: build/test-results
-      - store_artifacts:
-          path: build/test-results
-
   arm_machine:
     description: A template for running liboqs tests on ARM(presently only 64) machines
     parameters:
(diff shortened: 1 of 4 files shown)


baentsch:
Thanks @dstebila for setting this up/giving this a try! This seems to be a reasonable way out of our (C)CI timing problems -- assuming GHActions doesn't impose similar limitations once more tests are running there at the same time: Does anyone know about that -- short of trying it by moving over all jobs? 

Judging from the code and run logs, the approach looks good to me: Same baseline containers, very similar test code, better runtime.

On the negative side, this adds to our CI "zoo". Or is the idea to move _everything_ over to GHActions? If so (my personal preference to not deal with different CI system's quirks), I'd think we'd need to add `ssh(d)` support to our CI containers and do something similar to [this](https://github.com/marketplace?category=testing&type=actions&query=ssh+) to enable SSH access to debug problems with CI.

.circleci/config.yml
```diff
@@ -318,8 +272,11 @@ workflows:
   version: 2.1
   build:
     when:
-      not:
-        equal: [ main, << pipeline.git.branch >> ]
+      and:
+        - not:
+            equal: [ main, << pipeline.git.branch >> ]
+        - not:
+            matches: { pattern: "^ghactionsonly-.*", value: << pipeline.git.branch >> }
```
baentsch, line 279: Why not simply leave this away to ensure the remaining tests run OK -- until we move completely away from CCI?

baentsch (commented):
After digging some more, I'm now a bit concerned reading [this](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers) considering moving over OSX and Windows jobs: This way, we could easily exceed [these quotas/limits](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes) bringing us to where we stand right now: Up against a "pay wall". I've got the nagging feeling now that we have to decide whether we want to pay CircleCI (Amazon?) or Github (Microsoft?), to keep our level of testing if we want to avoid the "CI zoo". If not, we pay with our time (juggling/staying within the quotas of different CI systems).

dstebila:
> Thanks @dstebila for setting this up/giving this a try! This seems to be a reasonable way out of our (C)CI timing problems -- assuming GHActions doesn't impose similar limitations once more tests are running there at the same time: Does anyone know about that -- short of trying it by moving over all jobs?

PQClean does much of its testing on Github Actions, with many more jobs than we have, and does not seem to encounter usage limitations.

> Judging from the code and run logs, the approach looks good to me: Same baseline containers, very similar test code, better runtime.
> On the negative side, this adds to our CI "zoo". 

Given that the jobs are very similar, I feel like this isn't too much of a burden. Getting started with a new CI provider is a cost, but now that I've done that hopefully it's not too much of a burden to keep it going.  

> Or is the idea to move _everything_ over to GHActions? If so (my personal preference to not deal with different CI system's quirks), I'd think we'd need to add `ssh(d)` support to our CI containers and do something similar to [this](https://github.com/marketplace?category=testing&type=actions&query=ssh+) to enable SSH access to debug problems with CI.

I do not intend at this point to move everything over to Github Actions.  I just wanted to reduce some of the CI backlog we'd been encountering due to over-reliance on the free part of CircleCI.  And Github Actions has a few drawbacks compared to CircleCI -- including no native support for SSH access and no fancy test results reporting.

dstebila:
> After digging some more, I'm now a bit concerned reading [this](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers) considering moving over OSX and Windows jobs: This way, we could easily exceed [these quotas/limits](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes) bringing us to where we stand right now: Up against a "pay wall". I've got the nagging feeling now that we have to decide whether we want to pay CircleCI (Amazon?) or Github (Microsoft?), to keep our level of testing if we want to avoid the "CI zoo". If not, we pay with our time (juggling/staying within the quotas of different CI systems).

Since we're a public repository, we don't have a quota limit and pay for anything, and in principle the 10x / 2x cost for macOS and Windows jobs doesn't affect us.

.circleci/config.yml
dstebila: Lines 276-277 were already there (though the PR touches them because 275 and 278-279 were added).  I added 278-279 so that I could iterate on the Github Actions tests without triggering tests on the other systems.  I guess in principle I could remove that now for merging, but it also seemed like it could be useful to have in the future. I'm fine either way.

baentsch:
> Since we're a public repository, we don't have a quota limit and pay for anything

That sounds too good to be true: When reading "[GitHub Actions usage is free for both public repositories and self-hosted runners.](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions)" I thought that would mean we're in the "Github Free" category -- which does have a monthly quota. But if all jobs indeed are free, so much the better.

dstebila:
> > Since we're a public repository, we don't have a quota limit and pay for anything
> 
> That sounds too good to be true: When reading "[GitHub Actions usage is free for both public repositories and self-hosted runners.](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions)" I thought that would mean we're in the "Github Free" category -- which does have a monthly quota. But if all jobs indeed are free, so much the better.

So far so good.  I just checked our usage and it's still listed as 0 minutes.

dstebila:
There is a tool called [act](https://github.com/nektos/act) that can be used to run Github Action jobs locally, similar to `circleci local execute`.

PR #1070, opened 2021-08-05
Add test suite comparing parameters in C with parameters in YML

dstebila:
Based on discussion in https://github.com/open-quantum-safe/openssl/pull/316#issuecomment-893521915.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,6 +127,9 @@ jobs:
       CMAKE_ARGS:
         description: "Arguments to pass to CMake."
         type: string
+      PYTEST_ARGS:
+        description: "Arguments to pass to pytest."
+        type: string
     machine:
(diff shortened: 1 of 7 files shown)


baentsch (approved):
Together with https://github.com/open-quantum-safe/openssl/pull/316 this does what we wanted for resolving https://github.com/open-quantum-safe/openssl/issues/315. Thanks! The Unicode issue in CI (probably regarding some authors names) probably needs addressing  before a merge, though.

dstebila:
> The Unicode issue in CI (probably regarding some authors names) probably needs addressing before a merge, though.

Fixing the Unicode issue now. As for the other platforms where it fails due to lack of the Python YAML package being installed, I'll just skip this test on those platforms; it isn't critical this test is executed on all platforms.

PR #1073, opened 2021-08-07
Fix a bug where intrinsics were ignored when compiling for apple silicon

Martyrshot:
This pr fixes an issue for apple silicon compile-time feature detection.

* [ ] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.) - *NO*
* [ ] Does this PR change the the list of algorithms available -- either adding, removing, or renaming?  (If so, PRs in OQS-OpenSSL, OQS-BoringSSL, and OQS-OpenSSH will also be required by the time this is merged.) - *NO*


dstebila:
Replacing with #1074 for clean commit history and so CI runs.

PR #1074, opened 2021-08-08
Fix a bug where intrinsics were ignored when compiling for apple silicon

dstebila:
This pr fixes an issue for apple silicon compile-time feature detection.

Replaces #1073.

### DIFF
--- a/.CMake/detect_gcc_clang_intrinsics.c
+++ b/.CMake/detect_gcc_clang_intrinsics.c
@@ -54,7 +54,7 @@ int main(void) {
 #if defined(__ARM_FEATURE_AES)
 	printf("ARM_AES;");
 #endif
-#if defined(__ARM_FEATURE_SHA2)
+#if (defined(__APPLE__) && defined(__aarch64__)) || defined(__ARM_FEATURE_SHA2)
 	printf("ARM_SHA2;");
 #endif
 #if defined(__ARM_FEATURE_SHA3)
--- a/tests/system_info.c
+++ b/tests/system_info.c
@@ -186,6 +186,8 @@ static void print_oqs_configuration(void) {
 #endif
 #if defined(OQS_USE_SHA2_OPENSSL)
 	printf("SHA-2:            OpenSSL\n");
+#elif defined(OQS_USE_ARM_SHA2_INSTRUCTIONS)
+	printf("SHA-2:            C and ARM CRYPTO extensions\n");
 #else
 	printf("SHA-2:            C\n");
 #endif


dstebila:
I tried this on our Apple Silicon machine and it looks like the code is being used when compiled with `-DOQS_USE_OPENSSL=OFF`

Martyrshot:
Just to clarify, do you mean the c code, or intrinsics code? When testing on the Apple Silicon machine with this change it used intrinsics for me. I did have to specify not to use openssl with `-DOQS_USE_OPENSSL=OFF` in order for the intrinsics to be used.

dstebila:
> Just to clarify, do you mean the c code, or intrinsics code? When testing on the Apple Silicon machine with this change it used intrinsics for me. I did have to specify not to use openssl with `-DOQS_USE_OPENSSL=OFF` in order for the intrinsics to be used.

I compiled with `cmake -GNinja -DOQS_USE_OPENSSL=OFF ..` and the intrinsics code was used.

I would have that compiling with just `cmake -GNinja ..` would cause OpenSSL to be used for SHA2 (and indeed that seems to be what's happening in a build I just tried).  Are you experiencing something different?

baentsch:
> I would have that compiling with just cmake -GNinja .. would cause OpenSSL to be used for SHA2

This is exactly what should happen. Anything else would be incorrect/unintended behaviour of the system.

I'd approve this PR if someone could point to documentation stating that all Apple ARM64 platforms have optimized SHA2 instructions.

Martyrshot:
@dstebila 
> Are you experiencing something different?
No, that's what I'm experiencing too, but I wanted to clarify that's how I tested that intrinsics were being used since you mentioned it here:
> I tried this on our Apple Silicon machine and it looks like the code is being used when compiled with `-DOQS_USE_OPENSSL=OFF`

@baentsch although this isn't apple technical documentation, this PR on google's CPU features detection tool says that all apple chips since the A7 have the crypto extensions: [https://github.com/google/cpu_features/pull/150#issuecomment-755998695](https://github.com/google/cpu_features/pull/150#issuecomment-755998695)

baentsch:
> this isn't apple technical documentation, this PR on google's CPU features detection tool says that all apple chips since the A7 have the crypto extensions: google/cpu_features#150 (comment)

Given Apple purportedly doesn't give technical documentation to it's own engineers without NDA that's probably as good as it gets :-)

baentsch (approved):
LGTM

dstebila:
@baentch Any objections to me merging this into 0.7.0 without a new release candidate?

baentsch:
No, i.e., fine with me: This should not cause issues (....famous last words... :-), we have CI and I'll anyway run the same tests over the final release.

PR #1075, opened 2021-08-18
Added seeds for Falcon 1024 and 512

jsrobson10:
I wanted to be able to generate Falcon 512 and Falcon 1024 keypairs from seeds. This gives extra functionality to shrink the "key pair" to a much smaller size. 

* [ ] These changes don't change the behaviour of Falcon 512 and Falcon 1024 during "normal" use, but they do add the ability to change the seed used from a randomly generated one to one passed as a parameter. 
* [ ] These changes don't add any new algorithms to OQS.

### DIFF
--- a/src/sig/falcon/pqclean_falcon-1024_avx2/api.h
+++ b/src/sig/falcon/pqclean_falcon-1024_avx2/api.h
@@ -19,7 +19,7 @@
  * Return value: 0 on success, -1 on error.
  */
 int PQCLEAN_FALCON1024_AVX2_crypto_sign_keypair(
-    uint8_t *pk, uint8_t *sk);
+    uint8_t *pk, uint8_t *sk, uint8_t *seed, size_t seedlen);
 
 /*
  * Compute a signature on a provided message (m, mlen), with a given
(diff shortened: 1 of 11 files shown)


dstebila:
Hi Joshua. Because there's a risk of undermining security of key generation by taking an external seed that may not actually be uniform, I'm hesitant to incorporate this. I think the right approach would be to first check with the Falcon team to see if they endorse this mode of operation for Falcon.  If they do, then because of how we integrate algorithms into liboqs, it would first have to go into PQClean, and then we'd get the updated code when we pull from PQClean.

dstebila:
One other thought is that there is no guarantee that two different implementations of Falcon would construct the same (pk, sk) pair given the same seed. The only guarantee from the Falcon specification is that if two implementations use the same (pk, sk) pair, they should be able to interoperate -- verify each others' signatures correctly.

baentsch:
@jsrobson10 : Would you proceed along the lines suggested by @dstebila, please? If OK'd & incorporated by Falcon and PQClean this change would flow automatically into `liboqs`. Thus, this PR would be a duplication of effort. Accordingly closing here.

PR #1077, opened 2021-08-20
Add instructions on how to build shared/dynamic library to the README

Wandering-Consciousness:
I was trying to build the .NET (C#) wrapper project (liboqs-dotnet) and its README referenced this repo/README to build the oqs shared library but I noticed the README only mentioned that a static library was built. The Golang wrapper did have instructions on how to do so, so for consistency and benefit to future readers, I updated this repo's one too.

No

* [ no] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [ no] Does this PR change the the list of algorithms available -- either adding, removing, or renaming?  (If so, PRs in OQS-OpenSSL, OQS-BoringSSL, and OQS-OpenSSH will also be required by the time this is merged.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ Various `cmake` build options to customize the resultant artifacts are available
 
 The following instructions assume we are in `build`.
 
-3. The main build result is `lib/liboqs.a`, a static library. If you want to build a shared/dynamic library, append `-DBUILD_SHARED_LIBS=ON` to the `cmake -GNinja ..` command above and the result will be `lib/liboqs.so|dylib|dll`. The public headers are located in the `include` directory. There are also a variety of programs built under the `tests` directory:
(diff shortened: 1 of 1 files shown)


baentsch (changes requested):
Thanks for this suggestion: Makes good sense, as not everyone reads/finds the Wiki. Two suggestions for improvement, though: Could you begin section 3 with "By default [the main build result...]" and continue with your addition as written? Second, it may be a good idea to put a hyperlink for BUILD_SHARED_LIBS (to https://github.com/open-quantum-safe/liboqs/wiki/Customizing-liboqs#build_shared_libs). Otherwise LGTM.

Wandering-Consciousness:
Thank you. I've made the changes you suggested.

baentsch (approved):
Thanks again. LGTM.

dstebila (approved):

PR #1079, opened 2021-08-26
Updated SIKE to latest upstream commit

### LINKED ISSUE #1056: inline keyword in SIKE
SIKE uses `__inline` to mark some inline code, but this appears to be an MSVC-specific keyword.  The more portable way seems to be either `inline` or `__inline__`, although there are some subtle differences between these two that I'm not entirely sure of.  @christianpaquin do you know if `__inline` was chosen for a specific reason?

christianpaquin:
Updated SIKE to upstream commit effa60, addressing issue 1056: replaces `__inline` with `inline`; upstream also catches up to some changes already in OQS (for Apple M1 support).

Fixes #1056.


christianpaquin:
> Somewhat related and after checking the upstream reference (incl. the M1 support), may I ask what speaks against adding the "FAST" build option too when importing the code to `liboqs`? It would be great to have it for more representative profiling results.

Optimizations enabled by the FAST upstream makefile option have been mapped to our build system's optimizations.

PR #1085, opened 2021-09-05
increase constant time test timeout allowing slow runners to complete [skip ci]

baentsch:
This should allow constant-time testing to pass despite some algorithms being very slow (we won't name them :).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -353,12 +353,16 @@ workflows:
       - linux_oqs:
           name: constant-time-x64
           context: openquantumsafe
+          # some algorithms _are_ slow:
+          no_output_timeout: 45m
           CONTAINER: openquantumsafe/ci-ubuntu-bionic-x86_64:latest
           CMAKE_ARGS: -DOQS_OPT_TARGET=generic -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON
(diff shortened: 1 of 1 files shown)


dstebila (approved):

baentsch:
@dstebila  - this change was premature, sorry: The `no_output_timeout` only works on `run`, not `jobs`, so please revert this on main. We also already have a 1h timeout, so I now really don't understand [last week's test error](https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/1766/workflows/bacfb0fa-fa72-49d7-b40a-30fc6aec05da/jobs/15531). Ideas welcome and sorry for writing this correction twice (but it's not here any more....).

dstebila:
> @dstebila - this change was premature, sorry: The `no_output_timeout` only works on `run`, not `jobs`, so please revert this on main. We also already have a 1h timeout, so I now really don't understand [last week's test error](https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/1766/workflows/bacfb0fa-fa72-49d7-b40a-30fc6aec05da/jobs/15531). Ideas welcome and sorry for writing this correction twice (but it's not here any more....).

Okay, I've pushed a revert directly in commit https://github.com/open-quantum-safe/liboqs/commit/65f090d55e4bb050a57fb9c7dea69e704125671a. If this failure persists, we could disable the constant time tests for McEliece on the job and see if that solves it.

Issue #1110, opened 2021-10-17
Fails to build on FreeBSD on powerpc64le


pkubaj:
Error:
```
/wrkdirs/usr/ports/security/liboqs/work/liboqs-0.7.0/tests/ds_benchmark.h:142:45: error: implicit conversion from 'long' to 'double' may lose precision [-Werror,-Wimplicit-int-float-conversion]
        return (uint64_t)(time.tv_sec * 1e9 + time.tv_nsec);
```

This is FreeBSD 13.0-RELEASE on powerpc64le with LLVM 11.0.1.

baentsch:
@pkubaj Thanks for the error report. However, this is not reproducible as it seems to pertain to new code (from you?): The code line in the error message above is CPP-guarded `__s390x__` (which arguably is _not_ PPC :) and reads differently in the main branch:
https://github.com/open-quantum-safe/liboqs/blob/fff05ccf92e3abcb54e4313fc0057d96772212ea/tests/ds_benchmark.h#L142

-> Please update your report as otherwise we cannot do anything about it.

baentsch:
@pkubaj Could you please re-test? #1225 just merged could/should have resolved this

baentsch:
Closing due to inactivity (assuming resolved).

PR #1112, opened 2021-10-18
Add powerpc64le name as alternative to ppc64le

dstebila:
FreeBSD uses powerpc64le name for 64-bit POWER little endian.

Replaces #1109.


baentsch:
This will not fix #1110.

Issue #1101, opened 2021-10-05
Endianness issues with FrodoKEM-SHAKE and BIKE


bhess:
When building liboqs on s390x (#1100), everything builds and tests pass, except the KAT for FrodoKEM-SHAKE and BIKE. The most plausible cause to me are endianness issues (s390x is a big-endian architecture).
I open the issue here because Frodo and BIKE are direct contributions to liboqs.

See the build log:
https://app.travis-ci.com/github/bhess/liboqs/jobs/541322083

Frodo seems to already have macros defined to support big-endian and the -AES versions pass the tests.
Looking at the code, I think the way `a_cols` is accessed here could be problematic:
https://github.com/open-quantum-safe/liboqs/blob/36d8a49c886d6b2e58f8b3b0e06d0920eeea5b6d/src/kem/frodokem/external/frodo_macrify_shake_portable.c#L42

BIKE seems to be explicit about it that the code only supports little-endian:
https://github.com/awslabs/bike-kem/blob/master/cmake/arch.cmake
It could therefore be best to just disable it in big-endian builds for now.

baentsch:
> It could therefore be best to just disable it in big-endian builds for now.

Explicitly tagging @dkostic for an opinion on this.

PR #1115, opened 2021-10-21
Disable BIKE build on s390x (big endian)

bhess:
Disables BIKE build on s390x, see #1101.


dstebila:
Thanks Basil!

PR #1117, opened 2021-11-01
Adding support for multiple upstream to copy_from_upstream, as well as pqclean's aarch64 kyber implementation

Martyrshot:
There are two parts to this PR:
1) Adding support to copy_from_upstream.py to specify specific upstreams for arch-specific implementations
2) Pulling in pqclean's aarch64 optimized implementation of Kyber.


scripts/copy_from_upstream/patches/pqclean-kyber-armneon-yml.patch
```diff
@@ -0,0 +1,60 @@
...
+diff --git a/crypto_kem/kyber512/META.yml b/crypto_kem/kyber512/META.yml
+index 3c4d1ee..d1ac307 100644
+--- a/crypto_kem/kyber512/META.yml
++++ b/crypto_kem/kyber512/META.yml
+@@ -37,9 +37,7 @@ implementations:
+     - name: aarch64
+       version: https://github.com/neon-ntt/neon-ntt/tree/b011eeff3515fb168aa4dbaa671d760009d98dbb
+       supported_platforms:
+-        - architecture: arm_8
++        - architecture: ARM64_V8
```
baentsch, line 36: What is the reason for changing an architecture's name? Also: Has this been PR'd to upstream (PQClean) so that eventually OQS won't need these patches any more?

baentsch (changes requested):
Do I understand this PR right in that it enables architecture-specific pulls from different upstream repositories (and doing this for ARM64 Kyber)? It then seems documentation generation needs to be improved as all KEM docs (.yml and .md) are touched _except_ those for Kyber -- which seems weird: Can you please check/amend that suitably? 
Only as a question: As this seems to introduce a somewhat different code base for Kyber, would it be worth while/sensible/possible to add constant time tests for this code/platform -- or is this over the top (maybe to be added as a new issue: "constant time checks for aarch64")?
Finally, for curiosity: Is this code markedly faster than the ARM code presently generated? Worthwhile updating [profiling](https://github.com/open-quantum-safe/profiling) with this?

scripts/copy_from_upstream/patches/pqclean-kyber-armneon-yml.patch
Martyrshot: > Do I understand this PR right in that it enables architecture-specific pulls from different upstream repositories (and doing this for ARM64 Kyber)? 
Yes, exactly.
>It then seems documentation generation needs to be improved as all KEM docs (.yml and .md) are touched _except_ those for Kyber -- which seems weird: Can you please check/amend that suitably?
Will do.
>Only as a question: As this seems to introduce a somewhat different code base for Kyber, would it be worth while/sensible/possible to add constant time tests for this code/platform -- or is this over the top (maybe to be added as a new issue: "constant time checks for aarch64")?
I think as our codebase keeps adding optimized aarch64 implementations this would be worthwhile, but perhaps we should create another issue to discuss this. I'll make a new issue shortly.
>Finally, for curiosity: Is this code markedly faster than the ARM code presently generated? Worthwhile updating [profiling](https://github.com/open-quantum-safe/profiling) with this?
I hope so, but I haven't performed any benchmarking yet. This was a case of @dstebila asking me to pull in pqclean's kyber code and blindly doing it. :-) I've only skimmed the associated [paper](https://eprint.iacr.org/2021/986.pdf), but it looks like there is a decent speed improvement over past kyber implementations.

Martyrshot: > What is the reason for changing an architecture's name? Also: Has this been PR'd to upstream (PQClean) so that eventually OQS won't need these patches any more?

I patched this so that it matches out naming convention of ARM64_V8. I did this so that the fragment files would match our style. I don't think upstream should be changed to match our naming convention, unless they like our naming better.

src/kem/kyber/kem_kyber_1024.c
```diff
@@ -51,6 +57,16 @@ OQS_API OQS_STATUS OQS_KEM_kyber_1024_keypair(uint8_t *public_key, uint8_t *secr
 		return (OQS_STATUS) pqcrystals_kyber1024_ref_keypair(public_key, secret_key);
 	}
 #endif /* OQS_DIST_BUILD */
+#elif defined(OQS_ENABLE_KEM_kyber_1024_aarch64)
+#if defined(OQS_DIST_BUILD)
+	if () {
```
dstebila, line 62: Why is this `if ()` empty?

src/kem/saber/CMakeLists.txt
```diff
@@ -20,6 +20,14 @@ if(OQS_ENABLE_KEM_saber_lightsaber_avx2)
     set(_SABER_OBJS ${_SABER_OBJS} $<TARGET_OBJECTS:saber_lightsaber_avx2>)
 endif()
 
+if(OQS_ENABLE_KEM_saber_lightsaber_aarch64)
```
dstebila, line 23: Why are we getting Saber changes in this PR?

src/kem/saber/kem_saber_firesaber.c
```diff
@@ -51,6 +57,16 @@ OQS_API OQS_STATUS OQS_KEM_saber_firesaber_keypair(uint8_t *public_key, uint8_t
 		return (OQS_STATUS) PQCLEAN_FIRESABER_CLEAN_crypto_kem_keypair(public_key, secret_key);
 	}
 #endif /* OQS_DIST_BUILD */
+#elif defined(OQS_ENABLE_KEM_saber_firesaber_aarch64)
+#if defined(OQS_DIST_BUILD)
+	if (OQS_CPU_has_extension(OQS_CPU_EXT_ASIMD)) {
```
dstebila, line 62: I thought you eliminated ASIMD from PQClean META.yml?

dstebila (changes requested):
Shouldn't there be an update in docs/algorithms/kem/kyber.yml?  Also shouldn't all the docs/algorithms/kem/*.md files be regenerated?

src/kem/saber/CMakeLists.txt
Martyrshot: This likely happened when I ran copy_from_upstream. I'll revert this.

src/kem/saber/kem_saber_firesaber.c
Martyrshot: I did for the kyber meta.ymls but not sabre's. I'll end up reverting the sabre code so this will disappear.

src/kem/kyber/kem_kyber_1024.c
Martyrshot: A missed case in the template fragment. Thanks for catching it!

Martyrshot:
> Shouldn't there be an update in docs/algorithms/kem/kyber.yml? Also shouldn't all the docs/algorithms/kem/*.md files be regenerated?

Updating docs is the next thing on my list. Sorry I forgot to mention that I haven't updated the doc generation scripts yet. I wanted to get a PR for the copy_from_upstream script started early incase I had to change anything major.

src/kem/saber/CMakeLists.txt
baentsch: > Why are we getting Saber changes in this PR?
> This likely happened when I ran copy_from_upstream. I'll revert this.

Would it be a problem also updating Saber (or anything else that got upgraded in PQClean) at the same time? Shouldn't our CI (or doc generation) tell us if such "automatic" upgrade would be bad?

Martyrshot:
One thing I didn't think of until I pulled in sabre by mistake:

Should I add support for specifying specific commits and/or branches for arch specific upstreams?

Martyrshot:
Before I go updating the doc generation script to support multiple upstreams, I would like some feedback on the changes I'd like to make. :)
Here is a snippet of what kyber would look like (trimmed so only 512's implementation is included, but this wouldn't be the case for the real thing.

https://gist.github.com/Martyrshot/9b6acc809353f428d4bfd409fe4c108b

The big update is listing "optimized implementation" sources in addition to the "primary" implementation source, and then denoting which implementation characteristic belongs to which source. (Currently denoted as primary and the roman numeral associated with the listed optimized source)

Any suggestions or feedback would be greatly appreciated!

baentsch:
LGTM. Maybe to make it even easier to navigate, would it be very difficult to make the "Upstream" (leading entries in the second table) hyperlinks to the text line actually pointing to/documenting the effective upstream (in the text above)? But on second glance, the concept became clear (after spotting the "i" in the table I also found the "i" in the  "Optimized Implementation sources" and then knew where the "aarch64" code comes from.

Martyrshot:
> LGTM. Maybe to make it even easier to navigate, would it be very difficult to make the "Upstream" (leading entries in the second table) hyperlinks to the text line actually pointing to/documenting the effective upstream (in the text above)?

I've updated the documentation and made the (formally called) "upstream" entries links to the appropriate info.

I have the modified update_docs python script in another branch for now. If everything looks good I can add it here. It will currently only support multiple upstreams for kems, but wanted to keep the scope small so we can get kyber the optimized kyber merged asap.

I'm not sure what entries the "No branching-on-secrets claimed?" and "No branching-on-secrets checked by valgrind?" fields should have. They are currently set the same as rows that use the primary source.

Martyrshot:
> One thing I didn't think of until I pulled in sabre by mistake:
> 
> Should I add support for specifying specific commits and/or branches for arch specific upstreams?

As per weekly meeting discussion last week to keep things simple, we aren't going ahead with this idea.

baentsch:
> I've updated the documentation and made the (formally called) "upstream" entries links to the appropriate info.

Thanks very much -- IMO much easier to navigate/understand.

> I'm not sure what entries the "No branching-on-secrets claimed?" and "No branching-on-secrets checked by valgrind?" fields should have. They are currently set the same as rows that use the primary source.

My take is that the first value/statement/assertion should come from the authors of the code and the second one should be set to `true` only iff the first value is `true` _and_ we have a constant_time test for this (in this case for `aarch64`). Thus --in the absence of dissenting views (@dstebila @jschanck ?)-- my take would be that these values should be `true` and `false` respectively.

baentsch (approved):

docs/algorithms/kem/kyber.yml
```diff
@@ -51,6 +57,19 @@ parameter-sets:
...
+    supported-platforms:
+    - architecture: ARM64_V8
+      operating_systems:
+      - Linux
+      - Darwin
+      required_flags:
+    common-crypto:
+    - SHA3: liboqs
+    no-secret-dependent-branching-claimed: true
+    no-secret-dependent-branching-checked-by-valgrind: true
```
dstebila, line 71: Are actually checking the aarch64 implementation using Valgrind? I thought that's a TODO based on https://github.com/open-quantum-safe/liboqs/issues/1119.  As such this should (for now) be false.

```suggestion
    no-secret-dependent-branching-checked-by-valgrind: false
```

docs/algorithms/kem/kyber.yml
```diff
@@ -185,6 +225,19 @@ parameter-sets:
...
```
dstebila, line 239:
```suggestion
    no-secret-dependent-branching-checked-by-valgrind: false
```

docs/algorithms/kem/kyber.yml
```diff
@@ -118,6 +141,19 @@ parameter-sets:
...
```
dstebila, line 155:
```suggestion
    no-secret-dependent-branching-checked-by-valgrind: false
```

dstebila (changes requested):

dstebila:
> > I've updated the documentation and made the (formally called) "upstream" entries links to the appropriate info.
> 
> Thanks very much -- IMO much easier to navigate/understand.
> 
> > I'm not sure what entries the "No branching-on-secrets claimed?" and "No branching-on-secrets checked by valgrind?" fields should have. They are currently set the same as rows that use the primary source.
> 
> My take is that the first value/statement/assertion should come from the authors of the code and the second one should be set to `true` only iff the first value is `true` _and_ we have a constant_time test for this (in this case for `aarch64`). Thus --in the absence of dissenting views (@dstebila @jschanck ?)-- my take would be that these values should be `true` and `false` respectively.

Agreed, if my understanding is correct. I've marked that in my new review.

baentsch (approved):
Thx.

Martyrshot:
I've updated the kyber yml and md files to reflect the requested changes.

dstebila (approved):

baentsch:
@Martyrshot Can I ask whether the command `python3 copy_from_upstream.py copy` has been successfully executed on your local machine before this PR was completed? When running it locally for me it fails (now on `main`) as follows:

```
$ python3 copy_from_upstream.py copy
Traceback (most recent call last):
  File "copy_from_upstream.py", line 656, in <module>
    copy_from_upstream()
  File "copy_from_upstream.py", line 585, in copy_from_upstream
    update_upstream_alg_docs.do_it(os.environ['LIBOQS_DIR'])
  File "/home/mib/git/oqs/liboqs/scripts/copy_from_upstream/update_upstream_alg_docs.py", line 270, in do_it
    update_upstream_kem_alg_docs(liboqs_root, instructions['kems'], instructions['upstreams'], write_changes)
  File "/home/mib/git/oqs/liboqs/scripts/copy_from_upstream/update_upstream_alg_docs.py", line 110, in update_upstream_kem_alg_docs
    oqs_yaml['upstream'] = rhs_if_not_equal(oqs_yaml['upstream'], ("{}/commit/{}"+patches_done).format(upstream_base_url, ui['git_commit']), "upstream")
KeyError: 'upstream'

```

Thanks in advance for checking!

Martyrshot:
Ah, this is happening because I left out the updated generation script because it would crash for old documentation ymls. I can make a pr with the updated script (still needs updates for signatures, but it shouldn't crash) and the updated ymls. Sorry about that.

Martyrshot:
Oh, actually it's not what I thought it was. I'll let you know when I figure it out. (It was working on my local machine when I originally made this pr, so I must have left something out of the PR by mistake...)

PR #1089, opened 2021-09-13
skip algorithms for constant_time tests

baentsch:
Fixes #1087. [Release documentation](https://github.com/open-quantum-safe/liboqs/wiki/Release-process) also updated.

* [ ] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [ ] Does this PR change the the list of algorithms available -- either adding, removing, or renaming?  (If so, PRs in OQS-OpenSSL, OQS-BoringSSL, and OQS-OpenSSH will also be required by the time this is merged.)

Issue #1087, opened 2021-09-12
Investigate constant time test errors


baentsch:
Constant time tests keep failing when CPU extensions are activated: [Sep 5](https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/1766/workflows/bacfb0fa-fa72-49d7-b40a-30fc6aec05da/jobs/15531) , [Sep 12](https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/1772/workflows/22a6d567-df7c-486d-bb39-6ff9c15aa3ae/jobs/15550).

A quick analysis:
```
> python3 testcompare.py tests-20210905.log 
85 tests started, 84 passed: 
  SPHINCS+-SHA256-256s-robust not passed
> python3 testcompare.py tests-20210912.log 
85 tests started, 80 passed: 
  SPHINCS+-SHA256-128s-robust not passed
  SPHINCS+-SHA256-192s-robust not passed
  SPHINCS+-SHA256-256s-robust not passed
  Rainbow-V-Compressed not passed
  SPHINCS+-SHA256-192s-simple not passed
```

Suggestions for improvement solicited: Extending `no_output_timeout` beyond 1h? Dropping some algs from testing? Re-running and debugging locally?...

dstebila:
I think we should drop these algorithms from the test, and make a note on the release page for the wiki that these tests should be run locally before release.

PR #1120, opened 2021-11-04
Skip non-enabled algs in constant time tests

christianpaquin:
Skip non-enabled algs in constant time tests.

The current script depended on a `SKIP_ALGS` variable, that isn't documented and isn't set by cmake when using the `OQS_MINIMAL_BUILD ` build macro. I'm not sure if that code can safely replaced by the common helper check (as this PR does), or if it has a deeper purpose I don't understand. (@jschanck?)

### DIFF
--- a/tests/test_constant_time.py
+++ b/tests/test_constant_time.py
@@ -230,10 +230,6 @@ def get_ct_issues(t, name):
 @pytest.mark.parametrize('kem_name', helpers.available_kems_by_name())
 def test_constant_time_kem(kem_name):
     if not(helpers.is_kem_enabled_by_name(kem_name)): pytest.skip('Not enabled')
-    if ('SKIP_ALGS' in os.environ) and len(os.environ['SKIP_ALGS'])>0:
-        for algexp in os.environ['SKIP_ALGS'].split(','):
-            if len(re.findall(algexp, kem_name))>0:
-               pytest.skip("Test disabled by alg filter")
     passes = get_ct_passes('kem', kem_name)
     issues = get_ct_issues('kem', kem_name)
     output = helpers.run_subprocess(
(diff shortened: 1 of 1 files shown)


jschanck:
Looks good to me. I prefer this solution to `SKIP_ALGS`. The change makes the `helpers.available_{sigs,kems}_by_name` decorator redundant. You could remove it, or add a new `helpers.enabled_{sigs,kems}_by_name` decorator instead.

jschanck:
Although it looks like @baentsch wrote `SKIP_ALGS`, not me. So maybe there's more to it.

christianpaquin:
> Although it looks like @baentsch wrote `SKIP_ALGS`, not me. So maybe there's more to it.

I've asked @baentsch to review :)

baentsch (changes requested):
This approach surely is fine with me, but it needs to be complemented with suitable changes to CI: `SKIP_ALGS` parameter is [documented here](https://github.com/open-quantum-safe/liboqs/blob/468533abea77e7a7900ee9fdd82856ed25d06c2b/.circleci/config.yml#L97) and configured [here](https://github.com/open-quantum-safe/liboqs/blob/468533abea77e7a7900ee9fdd82856ed25d06c2b/.circleci/config.yml#L374). Reason were "irregular failures" of these algorithms in weekly tests. 
FYI, original issue: https://github.com/open-quantum-safe/liboqs/issues/1087  and original PR: https://github.com/open-quantum-safe/liboqs/pull/1089

christianpaquin:
Ah I see (my grep missed the `.circleci` folder, oops), you need skipping some algs even if they are compiled in the library. So these checks are not redundant, but complementary. I'll add back the SKIP_ALGS check.

PR #1127, opened 2021-11-19
Fix gcc 11 error for Kyber on arm64

### LINKED ISSUE #1125: GCC 11 throws `mismatched bound` errors on arm64
I am interested in packaging liboqs for Debian. Debian build of ba5b61a7 on arm64 fails due to `mismatched bound` errors:
```
/<<PKGBUILDDIR>>/src/kem/kyber/pqclean_kyber512_aarch64/kem.c:24:57: error: argument 1 of type ‘uint8_t[800]’ {aka ‘unsigned char[800]’} with mismatched bound [-Werror=array-parameter=]
   24 | int PQCLEAN_KYBER512_AARCH64_crypto_kem_keypair(uint8_t pk[KYBER_PUBLICKEYBYTES],
      |                                                 ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
In file included from /<<PKGBUILDDIR>>/src/kem/kyber/pqclean_kyber512_aarch64/kem.c:2:
/<<PKGBUILDDIR>>/src/kem/kyber/pqclean_kyber512_aarch64/kem.h:14:58: note: previously declared as ‘uint8_t *’ {aka ‘unsigned char *’}
   14 | int PQCLEAN_KYBER512_AARCH64_crypto_kem_keypair(uint8_t *pk, uint8_t *sk);
      |                                                 ~~~~~~~~~^~
```
([full build log](https://buildd.debian.org/status/fetch.php?pkg=liboqs&arch=arm64&ver=0.7.0.27.gba5b61a7%2Bdfsg-1&stamp=1637141594&raw=0); note Debian turns all GCC warning to errors)

Similar issue was reported as #1035 and seemingly has been fixed since then. Interestingly, most of the other architectures build just fine.

dstebila:
Fixes #1125.


scripts/copy_from_upstream/patches/pqclean-kyber-armneon-yml.patch
```diff
@@ -16,7 +16,7 @@ index 28e2014..56a4b5f 100644
 +++ b/crypto_kem/kyber1024/META.yml
 @@ -37,7 +37,7 @@ implementations:
      - name: aarch64
-       version: https://github.com/neon-ntt/neon-ntt/tree/b011eeff3515fb168aa4dbaa671d760009d98dbb
+       version: https://github.com/neon-ntt/neon-ntt/tree/c6ffbcd1e5590e4dddbc604fc80d2ad756485b4d
        supported_platforms:
 -        - architecture: arm_8
 +        - architecture: ARM64_V8
```
baentsch, line 22: This "architecture" difference makes me uneasy. Can you explain this, @Martyrshot or @bhess ?

baentsch (approved):
Except the one comment on "architecture" LGTM (but again, I'd defer to @bhess and @Martyrshot on that). Also, maybe @merkys would like to check whether this indeed fixes #1125.

scripts/copy_from_upstream/patches/pqclean-kyber-armneon-yml.patch
dstebila: I think it's just a matter of us using this string directly in places that expect ARM64_V8 whereas PQClean happens to populate it with arm_8.  @Martyrshot?

bhess: I wonder if this part of `update_upstream_alg_docs.py` will still work as expected, as it seems to expect `arm_8` to take the branch:

https://github.com/open-quantum-safe/liboqs/blob/2c98b97fa468cba89e11cf469613ed255c86bce1/scripts/copy_from_upstream/update_upstream_alg_docs.py#L209

merkys:
I am running the build right now, looks promising!

scripts/copy_from_upstream/patches/pqclean-kyber-armneon-yml.patch
Martyrshot: > I think it's just a matter of us using this string directly in places that expect ARM64_V8 whereas PQClean happens to populate it with arm_8. @Martyrshot?

Yes, exactly. It's just so we can use that string directly in our build system. Otherwise we'd have to add a check and replace everywhere that string is used.

> I wonder if this part of update_upstream_alg_docs.py will still work as expected, as it seems to expect arm_8 to take the branch:

It should be ok, as that line is reading straight from the upstream meta which doesn't get updated until line 210.

bhess: > It should be ok, as that line is reading straight from the upstream meta which doesn't get updated until line 210.

But the branch would also remove the "asimd" flag. Now that the upstream-yml is patched, shouldn't the patch remove this flag from the yml?

merkys:
Tested building on aarch64 using Debian build settings, all builds fine, thus #1125 no longer surfaces.

scripts/copy_from_upstream/patches/pqclean-kyber-armneon-yml.patch
Martyrshot: > But the branch would also remove the "asimd" flag. Now that the upstream-yml is patched, shouldn't the patch remove this flag from the yml?

Oh good catch. It looks like the patch isn't being applied when the repo is first cloned. I have a fix the completely removes the yml patch. I can push it to this pr if you'd like @dstebila.

dstebila: Sure, go for it.

Martyrshot: This should be resolved now, and the yml patch has been removed.

Issue #1122, opened 2021-11-10
Add SABER aarch64 implementation


dstebila:
Depends on #1117.

Issue #1119, opened 2021-11-02
Constant time checks for aarch64


Martyrshot:
As per a comment by @baentsch on this [PR](https://github.com/open-quantum-safe/liboqs/pull/1117#pullrequestreview-793886320), do we want to add constant time checks for aarch64?

dstebila:
That does seem like a good idea. We'll need to run those test suites on an ARM machine though, since it's a runtime test.

baentsch:
> That does seem like a good idea. We'll need to run those test suites on an ARM machine though, since it's a runtime test.

Yes. FWIW/reminder: We have an `aarch64` VM at AWS that's regularly running profiling on that architecture and which might be used for this, too.

Issue #1114, opened 2021-10-21
Investigate correct values for "no-secret-dependent-branching-checked-by-valgrind"


baentsch:
Set to `true` for all algorithms. Is this really correct? Maybe need to document what this really means. Possibly drop it if it simply means we've run `test_constant_time` (which we do weekly). Maybe change to `false` for algorithms which we had to exempt from weekly testing.

PR #1131, opened 2021-11-21
Adding arm64_v8 optimized versions of saber.

Martyrshot:
This PR adds arm optimized versions of lightsaber, firesaber, and saber from PQClean.

This PR resolves: #1122, but looks like there is an issue with upstream where the frame pointer gets corrupted. (PQCLEAN: issue [#431](https://github.com/PQClean/PQClean/issues/431#issue-1059280939))


jschanck:
It doesn't look like saber manages the frame pointer on aarch64. liboqs adds `fno-omit-frame-pointer` in clang debug builds. (I don't think there's any reason to do this, but maybe someone knows better.) Try removing this flag, or explicitly passing `fomit-frame-pointer`.

Martyrshot:
Thanks for catching that John!

It's looking like the actual issue is the stackpointer is getting corrupted after `PQCLEAN_LIGHTSABER_AARCH64_asm_asymmetric_mul()` is called. I'll update the pqclean pr.

Martyrshot:
This is pending a fix to upstream. Not sure on the time frame yet.

Martyrshot:
Assuming all these tests pass, this should be ready to merge.

scripts/copy_from_upstream/.CMake/alg_support.cmake/add_enable_by_alg.fragment
```diff
@@ -16,11 +16,11 @@ endif()
 {% if platform['operating_systems'] %}if(CMAKE_SYSTEM_NAME MATCHES "{{ platform['operating_systems']|join('|') }}")
 {% endif -%}
 if((OQS_DIST_ARM64_V8_BUILD OR (OQS_USE_ARM_NEON_INSTRUCTIONS{% for flag in platform['required_flags'] -%} {%- if not loop.last or loop.first %} AND {% endif -%}OQS_USE_ARM_{{ flag|upper }}_INSTRUCTIONS {%- endfor -%})))
-{% if family['name'] == "kyber" and impl['upstream']['name'] == 'pqclean' -%}
-if(((CMAKE_C_COMPILER_ID STREQUAL "GNU") AND (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "9.4.0")) OR ((CMAKE_CXX_COMPILER_ID STREQUAL "GNU") AND (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "9.4.0")))
+{% if impl['upstream']['name'] == 'pqclean' -%}
```
baentsch, line 19: Is this change to enable (or disable) _all_ PQClean algs as per rather complex build environment rules intentional? This not only impacts Saber.

baentsch (commented):
LGTM -- if the change regarding algorithm enablement is confirmed as intentional and doesn't limit the activation of algorithms unduly.

docs/algorithms/kem/saber.yml
```diff
@@ -44,6 +44,18 @@ parameter-sets:
...
+  - upstream-id: aarch64
+    supported-platforms:
+    - architecture: ARM64_V8
+      operating_systems:
+      - Linux
+      - Darwin
+    common-crypto:
+    - SHA3: liboqs
+    no-secret-dependent-branching-claimed: false
+    no-secret-dependent-branching-checked-by-valgrind: true
```
dstebila, line 56: How can it be that there's it is not claiming no-secret-dependent-branching but we're checking it anyway?

baentsch: Oops; indeed: We don't have tests: #1119 . This must be corrected. Reverted my "LGTM" above :)

scripts/copy_from_upstream/.CMake/alg_support.cmake/add_enable_by_alg.fragment
Martyrshot: Yes it is. Since we don't advertise a minimum compiler version, Douglas suggested adding a compiler version check for the minimum version of gcc required to get access to the intrinsics we need. The modification in this PR allows (apple)clang to using the aarch64 optimized algorithms.

The reason the if statement doesn't check for "kyber" anymore is because these intrinsics appear in saber as well, so I made it more general.

docs/algorithms/kem/saber.yml
Martyrshot: I've updated the aarch64 no-secret-dependent-branching-checked-by-valgrind to false, however I noticed it on the avx2 implementations as well. Is that intentional?

scripts/copy_from_upstream/.CMake/alg_support.cmake/add_enable_by_alg.fragment
baentsch: Understood. The question was whether this (same check) applies _to all_ algorithms imported from PQClean (not just kyber and saber)? But then again, we only have aarch64 optimized code for those algorithms, so it's fine. Thanks for the explanation.

docs/algorithms/kem/saber.yml
baentsch: > Is that intentional?

I can't answer that: Just checked all "no-secret-dependent-branching-checked-by-valgrind" in main branch and they're _all_ `true`, so any change to this value must be a result of the latest run of `update_docs_from_yaml.py`... Also see https://github.com/open-quantum-safe/liboqs/issues/1114

dstebila: We do have a [CI job](https://github.com/open-quantum-safe/liboqs/blob/f0678827043112e4c2f3854f40a1258180a0665b/.circleci/config.yml#L367) that runs the constant-time analysis on Haswell which includes AVX2.

README.md
```diff
@@ -40,23 +40,37 @@ Details on each supported algorithm can be found in the [docs/algorithms](https:
 
 <!--- OQS_TEMPLATE_FRAGMENT_LIST_KEXS_START -->
 - **BIKE**: BIKE-L1, BIKE-L3
+- **BIKE**: BIKE-L1, BIKE-L3
+- **Classic McEliece**: Classic-McEliece-348864†, Classic-McEliece-348864f†, Classic-McEliece-460896†, Classic-McEliece-460896f†, Classic-McEliece-6688128†, Classic-McEliece-6688128f†, Classic-McEliece-6960119†, Classic-McEliece-6960119f†, Classic-McEliece-8192128†, Classic-McEliece-8192128f†
```
dstebila, line 44: Duplicate lines?

Martyrshot: Hm, I didn't the portion of the update_docs script that updates the README.md. For now I'm pushing a manually fixed version of the README, but I'll have to look into why duplicate lines are appearing now.

Martyrshot: Oh, I see the issue. It's on my system. It looks like a bunch of 'family 2.yml's were made for some reason. Either way, the README should be ok now.

baentsch (approved):

PR #1135, opened 2021-11-23
more code style checks

### LINKED ISSUE #1129: CR+LF line endings in some of files
I have noticed that `appveyor.yml` and `src/kem/sike/external/P751/AMD64/fp_x64_asm.S` have CR+LF (`\r\n`) line endings. While this is a very minute issue, some of tools employed in Debian packaging have minor problems with it (I should probably find a way to deal with them too). But if there is no difference, maybe the same newline convention could be used throughout the project? If this could not be done for `appveyor.yml` due to usage in Windows, maybe then line endings could be unified throughout all `*asm.S` files?

baentsch:
Fixes #1129 (now also guarding against regressions).

Also, problematic files are highlighted better and advice is output what needs to be done on error.

### DIFF
--- a/tests/run_astyle.sh
+++ b/tests/run_astyle.sh
@@ -1,17 +1,21 @@
 #!/bin/bash
 # SPDX-License-Identifier: MIT
 
+rv=0
+
 # check style of non-external code:
 find src tests -name '*.[ch]' | grep -v '/external/' | grep -v 'kem/.*/.*/.*' | grep -v 'sig/.*/.*/.*' | xargs astyle --dry-run --options=.astylerc | grep Format
 if [ $? -ne 1 ]; then
-   echo "Error: Some files need reformatting. Check output above. Exiting."
-   exit -1
+   echo "Error: Some files need reformatting. Check output above."
+   rv=-1
 fi
 
 # check _all_ source files for CRLF line endings:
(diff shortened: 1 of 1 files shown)


tests/run_astyle.sh
```diff
@@ -1,4 +1,17 @@
 #!/bin/bash
 # SPDX-License-Identifier: MIT
 
-find src tests -name '*.[ch]' | grep -v '/external/' | grep -v 'kem/.*/.*/.*' | grep -v 'sig/.*/.*/.*' | xargs astyle --dry-run --options=.astylerc
+# check style of non-external code:
+find src tests -name '*.[ch]' | grep -v '/external/' | grep -v 'kem/.*/.*/.*' | grep -v 'sig/.*/.*/.*' | xargs astyle --dry-run --options=.astylerc | grep Format
+if [ $? -ne 1 ]; then
+   echo "Error: Some files need reformatting. Check output above. Exiting."
+   exit -1
```
dstebila, line 8: Suggest not exiting here but saving return value and outputting it at the end, so that the user also gets to see the results of running the next check on CRLF.

dstebila (approved):

tests/run_astyle.sh
baentsch: Makes sense. See new commit.

PR #1136, opened 2021-11-24
Added supported-platforms compile guards in the SIKE CMakeLists

### LINKED ISSUE #1128: SIKE not properly disabled on unsupported architectures
With Debian now trying to [package liboqs](https://tracker.debian.org/pkg/liboqs), they build on a lot more platforms than we support, many of which [the build fails on](https://buildd.debian.org/status/package.php?p=liboqs).  I'm not suggesting we try to support those at this point in time.  However, one set of build failures (see e.g. [this build log](https://buildd.debian.org/status/fetch.php?pkg=liboqs&arch=mips64el&ver=0.7.0.27.gba5b61a7%2Bdfsg-1&stamp=1637141544&raw=0)) seems to be related to SIKE's config.h header file thinking it's on an unsupported architecture, but apparently CMake still tried to compile SIKE.  In other words, there seems to be a mismatch between architectures where CMake thinks it should try compiling SIKE and architectures that SIKE's config.h header file is willing to support.

christianpaquin:
Added compile guards to SIKE's CMakeLists file to avoid downstream preprocessor failures in config.h when building on unsupported platforms.

Fixes #1128.


dstebila (approved):

baentsch:
When looking the code change over I wonder whether it properly solves the issue: I _think_ that the SIKE logic this way would not become part of liboqs (and not be compiled, so probably solving the original issue) in the case of mipsel, BUT I'm afraid that the algorithm #defines would be out of step with the library then: Wouldn't it be more prudent to (also) disable all `OQS_ENABLE_KEM_sike*` #defines when an unsupported platform is encountered? Back to your question of having a real platform for testing, @christianpaquin .... The best I could offer to check for good is to try to set up a qemu-based mipsel Debian. Or alternatively, we could ask for a test-run of this code in the Debian build harness before closing this for good.

dstebila:
> When looking the code change over I wonder whether it properly solves the issue: I _think_ that the SIKE logic this way would not become part of liboqs (and not be compiled, so probably solving the original issue) in the case of mipsel, BUT I'm afraid that the algorithm #defines would be out of step with the library then: Wouldn't it be more prudent to (also) disable all `OQS_ENABLE_KEM_sike*` #defines when an unsupported platform is encountered? 

Oh, that seems like a more robust solution, good point Michael.

dstebila (changes requested):
I think Michael's suggestion about moving the guard elsewhere makes sense.

christianpaquin:
@baentsch, @dstebila, yeah, good idea. I added some code in the main CMakeLists. I left the guards in the SIKE CMakeLists; I'm not sure if this is redundant.

I'm about to disconnect for the long Thanksgiving weekend, so in the interest of creating the RC soon, feel free to take over. I'm not a liboqs build master :) so feel free to adjust this PR accordingly.

baentsch:
> I added some code in the main CMakeLists. I left the guards in the SIKE CMakeLists; I'm not sure if this is redundant.

I'd say the latter. IMO the latter commit alone (and more correctly) resolves #1128. 

I now set up a mips `qemu` image to check this "for real". It runs for ages emulating every compiler instruction... _lots_ of them...... But on the positive side: It just confirmed the correctness of https://github.com/open-quantum-safe/liboqs/pull/1136/commits/9fa530ba299dbfba4d472ad27924f9ef5712a1c3

baentsch (approved):
The second commit alone would suffice.

dstebila (approved):

PR #1137, opened 2021-11-29
add downlevel compiler warning

baentsch:
This informs users if the installed compiler will not create optimized ARM code due to a compiler compatibility issue otherwise going unnoticed (as per https://github.com/open-quantum-safe/profiling/issues/68).


.CMake/alg_support.cmake
```diff
@@ -182,6 +182,8 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux|Darwin")
 if((OQS_DIST_ARM64_V8_BUILD OR (OQS_USE_ARM_NEON_INSTRUCTIONS)))
 if(((CMAKE_C_COMPILER_ID STREQUAL "GNU") AND (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "9.4.0")) OR ((CMAKE_CXX_COMPILER_ID STREQUAL "GNU") AND (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "9.4.0")) OR ((NOT (CMAKE_C_COMPILER_ID STREQUAL "GNU")) AND (NOT (CMAKE_CXX_COMPILER_ID STREQUAL "GNU"))))
     cmake_dependent_option(OQS_ENABLE_KEM_kyber_512_aarch64 "" ON "OQS_ENABLE_KEM_kyber_512" OFF)
+else()
+    message(WARNING "  ARM optimizations not active due to backlevel compiler")
```
dstebila, line 186: "backlevel" is a weird word to me -- not a real English word as far as I know and I'm not really sure what precisely you're trying to say. "out-of-date"? "unsupported"?

baentsch: Strange indeed. I meant to write what I wrote in the commit message: downlevel. Will update.

Martyrshot (changes requested):
Should this be added to scripts/copy_from_upstream/.CMake/alg_support.cmake/add_enable_by_alg.fragment? If we run copy_from_upstream.py now it's likely that these changes get overwritten.

Martyrshot (approved):
Looks good to me!

.CMake/alg_support.cmake
baentsch: More words: "out-of-date" doesn't "feel" quite right: Older compiler versions can still have value today (as evidenced by some of them still receiving support - also making the word "unsupported" quite unsuitable). 

Also, my intention was merely to make people aware that their compiler environment is not fully suitable for a specific feature of `liboqs` -- without making them change their build environment "just" for this feature.

Hence the word "downlevel". But it seems to be a marketing term (https://en.wiktionary.org/wiki/downlevel), so we could also use something else. Suggestions welcome by any native speaker (without marketing background :-)

baentsch: So done

dstebila (approved):

PR #1141, opened 2021-12-02
Enabling use of ARM PMU for benchmarking when compiler flag is enabled

tedeaton:
This pull request enables the use of ARMv8's performance monitoring unit (PMU) in the benchmarking script to allow for a more precise cycle count on that architecture when possible. It is enabled by the OQS_SPEED_USE_ARM_PMU compiler flag. Note that loading a kernel module (such as https://github.com/mupq/pqax#enable-access-to-performance-counters ) is required to access the PMU from user mode.

### DIFF
--- a/tests/ds_benchmark.h
+++ b/tests/ds_benchmark.h
@@ -181,6 +181,15 @@ static void _bench_init_perfcounters(int32_t do_reset, int32_t enable_divider) {
 	__asm__ volatile("mcr p15, 0, %0, c9, c12, 3\t\n" ::"r"(0x8000000f));
 }
 #elif defined(SPEED_USE_ARM_PMU)
+
+/* Enabling access to ARMv8's Performance Monitoring Unit
+ * cannot be done from user mode. A kernel module to
+ * enable access must be loaded. This generally will
+ * require superuser permissions. A module that has
+ * been found to work on some platforms can be found at
+ * https://github.com/mupq/pqax#enable-access-to-performance-counters
+ */
+
 static void _bench_init_perfcounters(void) {
 	__asm__ volatile("MSR PMCR_EL0, %0" ::"r"(1));
 	__asm__ volatile("MSR PMCNTENSET_EL0, %0" ::"r"(0x80000000));
--- a/tests/ds_benchmark.h
+++ b/tests/ds_benchmark.h
@@ -135,6 +135,12 @@ static uint64_t _bench_rdtsc(void) {
 	__asm__ volatile("mrc p15, 0, %0, c9, c13, 0\t\n"
 	                 : "=r"(value));
 	return value;
+#elif defined(SPEED_USE_ARM_PMU)
+	/* Use the Performance Monitoring Unit */
+	uint64_t value;
+	/* Read the PMU register */
+	__asm__ volatile("mrs %0, PMCCNTR_EL0" : "=r" (value));
+	return value;
 #elif defined(__s390x__)
 #define USING_TIME_RATHER_THAN_CYCLES
 	uint64_t tod;
@@ -174,6 +180,11 @@ static void _bench_init_perfcounters(int32_t do_reset, int32_t enable_divider) {
 	/* Clear overflows */
 	__asm__ volatile("mcr p15, 0, %0, c9, c12, 3\t\n" ::"r"(0x8000000f));
 }
+#elif defined(SPEED_USE_ARM_PMU)
+static void _bench_init_perfcounters(void) {
+	__asm__ volatile("MSR PMCR_EL0, %0" ::"r"(1));
+	__asm__ volatile("MSR PMCNTENSET_EL0, %0" ::"r"(0x80000000));
+}
 #endif
 
 #define DEFINE_TIMER_VARIABLES                                                                              \
@@ -193,6 +204,15 @@ static void _bench_init_perfcounters(int32_t do_reset, int32_t enable_divider) {
     _bench_cycles_M2 = 0.0;         \
     _bench_time_cumulative = 0;     \
     _bench_time_mean = 0.0;         \
+    _bench_time_M2 = 0.0;
+#elif defined(SPEED_USE_ARM_PMU)
+#define INITIALIZE_TIMER        \
+    _bench_init_perfcounters(); \
+    _bench_iterations = 0;      \
+    _bench_cycles_mean = 0.0;   \
+    _bench_cycles_M2 = 0.0;     \
+    _bench_time_cumulative = 0; \
+    _bench_time_mean = 0.0;     \
     _bench_time_M2 = 0.0;
 #else
 #define INITIALIZE_TIMER        \
--- a/tests/speed_kem.c
+++ b/tests/speed_kem.c
@@ -11,6 +11,9 @@
 #if defined(USE_RASPBERRY_PI)
 #define _RASPBERRY_PI
 #endif
+#if defined(OQS_SPEED_USE_ARM_PMU)
+#define SPEED_USE_ARM_PMU
+#endif
 #include "ds_benchmark.h"
 #include "system_info.c"
 
--- a/tests/speed_sig.c
+++ b/tests/speed_sig.c
@@ -11,6 +11,9 @@
 #if defined(USE_RASPBERRY_PI)
 #define _RASPBERRY_PI
 #endif
+#if defined(OQS_SPEED_USE_ARM_PMU)
+#define SPEED_USE_ARM_PMU
+#endif
 #include "ds_benchmark.h"
 #include "system_info.c"
 
--- a/tests/test_aes.c
+++ b/tests/test_aes.c
@@ -10,6 +10,9 @@
 #if defined(USE_RASPBERRY_PI)
 #define _RASPBERRY_PI
 #endif
+#if defined(OQS_SPEED_USE_ARM_PMU)
+#define SPEED_USE_ARM_PMU
+#endif
 #include "ds_benchmark.h"
 #include "system_info.c"
 


dstebila (approved):

dstebila:
Code looks good, thanks Ted!  Can you add some documentation in the code with instructions about loading the kernel module, as well as add some notes about the compiler option on these two wiki pages [1](https://github.com/open-quantum-safe/liboqs/wiki/Platform-specific-notes-for-building-liboqs), [2](https://github.com/open-quantum-safe/liboqs/wiki/Customizing-liboqs)?

Which platforms have you been able to run it successfully on?

- [ ] AWS ARM
- [ ] ROCK64
- [ ] Raspberry Pi
- [ ] Apple M1 Mac Mini

tedeaton:
Tested and ran successfully on:

- [x] AWS ARM
- [x] ROCK64
- [x] Raspberry Pi
 
It's not clear to me if the M1 has the PMU, or how to access it from user mode if it does, since MacOS handles kernel modules differently than linux.

dstebila (approved):

baentsch:
> Tested and ran successfully on:
>
> AWS ARM

@tedeaton Can I ask in which OS environment you did this? When I tried activating this for a standard AWS Linux ARM VM, all kinds of problems occurred (from having to find kernel headers for RHEL then to missing KConfig kernel file documentation...). Will try Ubuntu20 on t4.micro tomorrow for a change....

baentsch:
> @tedeaton Can I ask in which OS environment you did this?

Answering my own question: aarch64-Ubuntu20 on AWS installs `enable_ccr` just fine. 

When running the resultant code then, some questions come up: Before asking them below, can I ask you to check whether these figures are as you'd expect them? The first run is with `OQS_SPEED_USE_ARM_PMU=ON`, the second without.

```
ubuntu@ip-172-31-17-142:~/liboqs/build$ ./tests/speed_kem Kyber512
Configuration info
==================
Target platform:  aarch64-Linux-5.11.0-1022-aws
Compiler:         gcc (10.3.0)
Compile options:  [-mcpu=native;-Werror;-Wall;-Wextra;-Wpedantic;-Wstrict-prototypes;-Wshadow;-Wformat=2;-Wfloat-equal;-Wwrite-strings;-O3;-fomit-frame-pointer;-fdata-sections;-ffunction-sections;-Wl,--gc-sections;-Wbad-function-cast]
OQS version:      0.7.0
Git commit:       b5cbb0d71a6f87f70198c986d85114beec1cb5ab
OpenSSL enabled:  Yes (OpenSSL 1.1.1f  31 Mar 2020)
AES:              OpenSSL
SHA-2:            OpenSSL
SHA-3:            C
CPU exts compile-time:  AES SHA2 NEON

Speed test
==========
Started at 2021-12-06 06:50:26
Operation                      | Iterations | Total time (s) | Time (us): mean | pop. stdev | CPU cycles: mean          | pop. stdev
------------------------------ | ----------:| --------------:| ---------------:| ----------:| -------------------------:| ----------:
Kyber512                       |            |                |                 |            |                           |           
keygen                         |     133108 |          3.000 |          22.538 |      5.628 |                     32581 |      24735
encaps                         |     115278 |          3.000 |          26.024 |      0.854 |                     32648 |      29978
decaps                         |     129473 |          3.000 |          23.171 |      0.645 |                     32752 |      26171
Ended at 2021-12-06 06:50:35
ubuntu@ip-172-31-17-142:~/liboqs/build$ ../build-nopmu-gcc10/tests/speed_kem Kyber512
Configuration info
==================
Target platform:  aarch64-Linux-5.11.0-1022-aws
Compiler:         gcc (10.3.0)
Compile options:  [-mcpu=native;-Werror;-Wall;-Wextra;-Wpedantic;-Wstrict-prototypes;-Wshadow;-Wformat=2;-Wfloat-equal;-Wwrite-strings;-O3;-fomit-frame-pointer;-fdata-sections;-ffunction-sections;-Wl,--gc-sections;-Wbad-function-cast]
OQS version:      0.7.0
Git commit:       b5cbb0d71a6f87f70198c986d85114beec1cb5ab
OpenSSL enabled:  Yes (OpenSSL 1.1.1f  31 Mar 2020)
AES:              OpenSSL
SHA-2:            OpenSSL
SHA-3:            C
CPU exts compile-time:  AES SHA2 NEON

Speed test
==========
Started at 2021-12-06 06:50:54
Operation                      | Iterations | Total time (s) | Time (us): mean | pop. stdev | High-prec time (ns): mean | pop. stdev
------------------------------ | ----------:| --------------:| ---------------:| ----------:| -------------------------:| ----------:
Kyber512                       |            |                |                 |            |                           |           
keygen                         |     144331 |          3.000 |          20.786 |      0.781 |                     20719 |        644
encaps                         |     123267 |          3.000 |          24.338 |      0.773 |                     24272 |        662
decaps                         |     139472 |          3.000 |          21.510 |      0.655 |                     21446 |        445
Ended at 2021-12-06 06:51:03
```

Questions: 

1) Why does the runtime drop (nearly 10%) with `OQS_SPEED_USE_ARM_PMU=ON`? Does PMU read-out have such a large overhead?
2) Would it be sensible to change our simple "speed" (iterations/runtime) display in https://openquantumsafe.org/benchmarking/visualization/speed_kem.html to showing the cycle-counts instead (now that we finally have it for both x64 and aarch64)? This possibly depending on the answer to the question above...

baentsch:
And another very weird run result (Ubuntu-ARM64 on AWS) probably warranting the answer "No" to my second question above:

```
ubuntu@ip-172-31-17-142:~/liboqs/build$ ./tests/speed_kem Kyber512
Configuration info
==================
Target platform:  aarch64-Linux-5.11.0-1022-aws
Compiler:         gcc (10.3.0)
Compile options:  [-mcpu=native;-Werror;-Wall;-Wextra;-Wpedantic;-Wstrict-prototypes;-Wshadow;-Wformat=2;-Wfloat-equal;-Wwrite-strings;-O3;-fomit-frame-pointer;-fdata-sections;-ffunction-sections;-Wl,--gc-sections;-Wbad-function-cast]
OQS version:      0.7.0
Git commit:       b5cbb0d71a6f87f70198c986d85114beec1cb5ab
OpenSSL enabled:  Yes (OpenSSL 1.1.1f  31 Mar 2020)
AES:              OpenSSL
SHA-2:            OpenSSL
SHA-3:            C
CPU exts compile-time:  AES SHA2 NEON

Speed test
==========
Started at 2021-12-06 13:37:30
Operation                      | Iterations | Total time (s) | Time (us): mean | pop. stdev | CPU cycles: mean          | pop. stdev
------------------------------ | ----------:| --------------:| ---------------:| ----------:| -------------------------:| ----------:
Kyber512                       |            |                |                 |            |                           |           
keygen                         |     132182 |          3.000 |          22.696 |     13.299 |                     27137 |      25555
encaps                         |     114048 |          3.000 |          26.305 |      1.103 |                     18426 |      27465
decaps                         |     125402 |          3.000 |          23.923 |     18.013 |           147100876138328 | 52091367556182272
Ended at 2021-12-06 13:37:39
```

tedeaton:
> Why does the runtime drop (nearly 10%) with OQS_SPEED_USE_ARM_PMU=ON? Does PMU read-out have such a large overhead?

I'm not entirely sure. This was the reason we originally started investigating benchmarking on ARM, because we were noticing strange overhead on AES. It's possible(?) that the use of `clock_gettime` has more overhead than using the PMU, causing the difference.

> And another very weird run result (Ubuntu-ARM64 on AWS)

Interesting. I notice that `log_2( 147100876138328 * 125402)` is approximately equal to 64. I suspect that in one of the executions the PMU overflowed and reset, causing the cycle count to be 2^64. for that one execution. I should have checked for overflow, that's a mistake on my part. But hopefully it should be an easy fix, I'll check it out today.

tedeaton:
Ah, my apologies, I misread your previous outputs. Yes, that is very strange, and I have no theories why the time is increasing. Since the only difference is reading off the PMU vs. using `clock_gettime`  I suppose the only reasonable explanation is that there's some overhead reading the PMU.

The PMU should still give a more accurate comparison though (as long as I can fix that bug). For example, when benchmarking AES we were seeing a less than 2x improvement in clock time for C vs. intrinsics, but over 20x improvement on cycle counts (which was what was expected).

baentsch:
> when benchmarking AES we were seeing a less than 2x improvement in clock time for C vs. intrinsics, but over 20x improvement on cycle counts (which was what was expected).

Yikes. That's _a lot_. Which configurations did you compare to each other to get these differentials? If  cycle count improves 20x how can clock time (regardless of measurement mechanism -- even a watch should do if done 1M times :) only improve 2x? Or is my assumption flawed that there's always the same & fixed number of cycles in a given clock period? Is the optimized code (and architecture) 10x parallelized? But even then, one should see 20x more iterations in the same test execution period if the cycle count drops 20x, no?

dstebila:
This sounds like a low-resolution clock versus a high-resolution cycle counter.  The way the [timing code](https://github.com/open-quantum-safe/liboqs/blob/main/tests/ds_benchmark.h) works is that it starts and stops the timer before each iteration.  With a low-resolution clock and many iterations, it would probably be better to start and stop the clock once at the beginning/end, which should minimize overhead and over many iterations the time observed with the low-resolution clock should approach that of the high-resolution cycle counter.

baentsch:
Thanks for the explanation, @dstebila. I should have read the code earlier, sorry. In my mind, a test timed to run for 3 secs would be started to run in the tightest possible loop, never again looking at counters (or timers) until a (timer/interrupt) occurs in order to not unduly influence the results with counter/timer-checking code. This indeed isn't the case with ds_benchmark. 

Would it be worth while considering changing the code to my understanding (at least in the case when there's no high-precision cycle counter available)? Or is there a reason speaking against using a (CPU-)timer interrupt? Something like the following "pseudo"code:

```
static int done = 0;

static void handler(int sig, siginfo_t *si, void *uc)
{
   done = 1;
}

// main test code
    timer_t timerid;
    struct sigevent sev;
    struct itimerspec its;
    long long freq_nanosecs = 3000000000;
    sigset_t mask;
    struct sigaction sa;

    sa.sa_flags = SA_SIGINFO;
    sa.sa_sigaction = handler;
    sigemptyset(&sa.sa_mask);
    if (sigaction(SIGRTMIN, &sa, NULL) == -1)
        errExit("sigaction");

    sev.sigev_notify = SIGEV_SIGNAL;
    sev.sigev_signo = SIGRTMIN;
    sev.sigev_value.sival_ptr = &timerid;
    if (timer_create(CLOCK_REALTIME, &sev, &timerid) == -1)
        errExit("timer_create");
    
    its.it_value.tv_sec = freq_nanosecs / 1000000000;
    its.it_value.tv_nsec = freq_nanosecs % 1000000000;
    its.it_interval.tv_sec = its.it_value.tv_sec;
    its.it_interval.tv_nsec = its.it_value.tv_nsec;

    if (timer_settime(timerid, 0, &its, NULL) == -1)
         errExit("timer_settime");

    while(!done) tstcnt++; //do the real op-under-test here

    printf("Cnt = %d\n", tstcnt);
```

Admittedly more convoluted than ds_benchmark -- and less portable. But arguably more accurate, no? 

From a portability perspective, is it better to rely on timer interrupts or specific CPU features (that possibly even need to be activated) for cycle counting to work correctly? Maybe something to discuss in the context of https://github.com/open-quantum-safe/profiling/discussions/67 ?

dstebila:
I've never used timers / interrupts so I don't have any insight into whether they're a preferred approach.  We can certainly give it a try...

bhess:
Is it intentional that this PR was not merged to main?

The main branch seems to only contain 32-bit cycle counters for ARM which would lead to overflows.

Regarding clock, multiplying tv_sec by 1e6 or 1e9 could lead to portability issues since time_t is 32-bit long on some platforms, leading to overflows:

https://github.com/open-quantum-safe/liboqs/blob/8e22871ff6cfdc69dba6748bbd902ed0fb316fc3/tests/ds_benchmark.h#L228

https://github.com/open-quantum-safe/liboqs/blob/8e22871ff6cfdc69dba6748bbd902ed0fb316fc3/tests/ds_benchmark.h#L147

dstebila:
> Is it intentional that this PR was not merged to main?

Oops, I didn't realize that this PR merged to a non-main branch.  Via #1147, I have now merged all of this down to main.

> The main branch seems to only contain 32-bit cycle counters for ARM which would lead to overflows.
> 
> Regarding clock, multiplying tv_sec by 1e6 or 1e9 could lead to portability issues since time_t is 32-bit long on some platforms, leading to overflows:
> 
> https://github.com/open-quantum-safe/liboqs/blob/8e22871ff6cfdc69dba6748bbd902ed0fb316fc3/tests/ds_benchmark.h#L228
> 
> https://github.com/open-quantum-safe/liboqs/blob/8e22871ff6cfdc69dba6748bbd902ed0fb316fc3/tests/ds_benchmark.h#L147

Ah, interesting.  So should we cast tv_sec to a uint64_t before multiplying?

bhess:
> Ah, interesting. So should we cast tv_sec to a uint64_t before multiplying?

Yes I think this would be safer.

Issue #1143, opened 2021-12-03
Add common crypto performance tests


baentsch:
As suggested by @bhess (https://github.com/open-quantum-safe/profiling/issues/68#issuecomment-984507492) this issue is to track adding performance tests for "common code" similar to `speed_kem` and `speed_sig` to better understand the performance impact of "core crypto" operations -- either provided by `liboqs` common code or by third-party routines (like OpenSSL):
- [ ] AES (anything needed beyond [this](https://github.com/open-quantum-safe/liboqs/blob/188e6a38b2bbc1795a105e21bc6925ac266df90b/tests/test_aes.c#L152-L158) ?)
- [ ] SHA2
- [ ] SHA3

 (others?)

baentsch:
Fixed by #1144

PR #1144, opened 2021-12-03
Adding a speed_common script to benchmark OQS AES and hash algorithms

tedeaton:
This adds a speed_common script, similar to speed_kem and speed_sig, but for benchmarking the encryption and hashing algorithms that `liboqs` uses. Specifically, it benchmarks AES128, AES256, SHA256, SHA384, SHA512, SHA3-256, SHA3-384, SHA3-512, SHA3-shake128, and SHA3-shake256. Currently it only benchmarks the main API of each of the hash algorithms, not the incremental version (that is something that could be added later if it's needed).

The script has a few options to change how it is run:
`--msglen n` : Each algorithm is benchmarked with a message length of `n` bytes (must be a multiple of 16 to benchmark AES)
`--outlen n` : For algorithms with variable-length output (e.g., shake128), specifies the output length.
`<alg>` : Only benchmark a specific algorithm.

Fulfills request in #1143 

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -45,6 +45,9 @@ if(NOT WIN32)
     add_executable(test_sha3 test_sha3.c ${COMMON_OBJS})
     target_link_libraries(test_sha3 PRIVATE ${INTERNAL_TEST_DEPS})
 
+    add_executable(speed_common speed_common.c ${COMMON_OBJS})
+    target_link_libraries(speed_common PRIVATE ${INTERNAL_TEST_DEPS})
+
     set(UNIX_TESTS test_aes test_hash test_sha3)
 
     set(PYTHON3_EXEC python3)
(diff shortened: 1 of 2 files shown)


baentsch (changes requested):
Thanks for taking on this issue! The logic in this PR departs from the pre-existing `speed_kem|sig` (and `test_aes -b`) apps though, in that it _requires_ an algorithm parameter to run. Instead, not passing an algorithm name should lead to all algs being tested. Further, it would be prudent to have a default set of message lengths be run if that parameter is not given (analogous to `openssl speed`).

tedeaton:
That should be how the script works already, unless I messed up the input parsing. If no algorithm is specified, then `single_alg` never gets set, and each algorithm should be run. Similarly, `message_len` is set to 64 bytes by default.

baentsch (approved):

baentsch:
> That should be how the script works already

Sincere apologies: Indeed I overlooked that in the code and didn't actually run the code before. After doing this now, this is to confirm this PR indeed completely implements #1143. Thanks!

PR #1156, opened 2021-12-14
docs: set license MIT to Falcon Signature

ax1:
https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/sig/falcon.md states **license CC0-1.0**

But the license in the header file at https://falcon-sign.info/impl/falcon.h.html is  ** license MIT** , the same at https://github.com/PQClean/PQClean/blob/master/crypto_sign/falcon-1024/clean/LICENSE

No


dstebila (approved):

dstebila:
Good catch, thanks.

PR #1158, opened 2021-12-15
Update Picnic to 3.0.7

sebastinas:
This PR updates the Picnic copy to version 3.0.7. Additionally, it slightly reduces the diff between our code and the copy in liboqs by providing a code path for calling `OQS_randombytes` in `randomness.c` and by defining `PICNIC_EXPORT` instead of `PICNIC_STATIC`. `PICNIC_STATIC` is intended for the static build for tests and benchmarking and is intended as an internal flag. With these changes future imports of the Picnic code base should be a bit smoother and should only require a check that `oqs_picnic_macros.h` is still included in the correct places.

* [x] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.): It doesn't.
* [x] Does this PR change the the list of algorithms available -- either adding, removing, or renaming?  (If so, PRs in OQS-OpenSSL, OQS-BoringSSL, and OQS-OpenSSH will also be required by the time this is merged.): It doesn't.


christianpaquin:
Not sure why it fails on macOS; can you repro @sebastinas?

sebastinas:
I forgot to change back one of the `explicit_bzero`s. That's now fixed.

dstebila:
Thanks Sebastian.  We're just about to release liboqs 0.7.1, but if I understand correctly this Picnic update does not involve any change to KATs or security issues, so my plan is to go ahead with the 0.7.1 release and then merge this after that.

sebastinas:
I'll take the time to fix the AVX2/SSE2/NEON handling to better match what's done for other implementations.

sebastinas:
Following the release of 0.7.1, I've rebased my changes and updated to Picnic 3.0.8. The Picnic now follows the same pattern to enable SSE2/AVX2/NEON support. Compiling with `-mavx2` or `-msse2` is not necessary for Picnic.

christianpaquin (approved):
Looks good. Thanks Sebastian!

baentsch:
Hmm -- this merge seems to have been premature: AppVeyor fails since the merge. I'm puzzled, though why this failure didn't materialize/show up in the PR CI (?). @sebastinas Would you want to take a look or shall I do (as I added the cygwin support -- where picnic now fails)?

sebastinas:
@baentsch I cannot reproduce the issue in a local cygwin installation. Neither the KATs for the upstream Picnic code, nor the liboqs tests fail for me. In any case, if you have fixes that are required for Picnic on cygwin, please send them upstream: https://github.com/IAIK/Picnic

baentsch:
OK, will need to investigate as the error message is rather severe (stack trace dumped). See #1163 for further tracking.

PR #1160, opened 2021-12-20
Add support for powerpc64

pkubaj:
This PR adds support for 64-bit POWER on big-endian.
There is already support for one 64-bit big-endian architecture (s390x).
There is also support for ppc64le (64-bit POWER little-endian).
So this PR shouldn't add much overhead.


dstebila:
Thanks for the PR. Do you know of any resources for running continuous integration on this platform?

pkubaj:
I don't think there are any that provide out-of-the-box. But e.g. OpenSSL crosscompiles for plenty of platforms: https://github.com/openssl/openssl/blob/master/.github/workflows/cross-compiles.yml

That said, I think it may be ok if liboqs builds on both ppc64le and s390x.

dstebila (approved):

baentsch:
> I don't think there are any that provide out-of-the-box. But e.g. OpenSSL crosscompiles for plenty of platforms: https://github.com/openssl/openssl/blob/master/.github/workflows/cross-compiles.yml
> That said, I think it may be ok if liboqs builds on both ppc64le and s390x.

Indeed, the OpenSSL matrix also only tests s390x and ppc64le -- as does `liboqs` via Travis. If I'm not mistaken, the only BE PPC are IBM i/AIX machines... So, maybe [this link promising i/AIX "in the cloud"](https://help.skytap.com/skytap-on-ibm-cloud-overview.html) might be an option? Tagging @bhess for an opinion as to whether there's a way (& interest) to add IBM i/AIX to the `liboqs` CI.

pkubaj:
Actually, all the POWER CPUs up to Power10 support BE. It's just that the current trend is to move to LE, but it's not like you can't run BE. In fact, I have a dual boot on my POWER9 workstation with both LE and BE.

CMakeLists.txt
```diff
@@ -59,6 +59,12 @@ elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64le|powerpc64le")
     if(${OQS_DIST_BUILD})
         set(OQS_DIST_PPC64LE_BUILD ON)
     endif()
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "(ppc64|powerpc64)")
+    set(ARCH "ppc64")
```
baentsch, line 63: What about adding a warning message here about "suboptimal" CI for this platform?

baentsch:
> Actually, all the POWER CPUs up to Power10 support BE. It's just that the current trend is to move to LE, but it's not like you can't run BE. In fact, I have a dual boot on my POWER9 workstation with both LE and BE.

Good to know. But if we can't do CI for the "generic" ppc64be platform I'd urge caution to adding it as an apparently fully supported platform to `liboqs`. See for example [this discussion](https://github.com/open-mpi/ompi/issues/4349) for what might ensue (say if some "big endian users" expected `liboqs` to work out of the box). 

So what about the suggestion to add a `cmake` warning message stating "this configuration not tested via continuous integration." (see comment above)?

baentsch (changes requested):
Please add warning message about lack of CI for generic/big-endian PPC in CMakeLists.txt.

bhess:
> Indeed, the OpenSSL matrix also only tests s390x and ppc64le -- as does `liboqs` via Travis. If I'm not mistaken, the only BE PPC are IBM i/AIX machines... So, maybe [this link promising i/AIX "in the cloud"](https://help.skytap.com/skytap-on-ibm-cloud-overview.html) might be an option? Tagging @bhess for an opinion as to whether there's a way (& interest) to add IBM i/AIX to the `liboqs` CI.

I believe most Linux distributions moved to a ppc64le target since little endian (switchable) is supported in POWER8 and later. Using qemu might be an option on Linux. AIX is a good point, but I don't think our CIs support an integration.

> So what about the suggestion to add a `cmake` warning message stating "this configuration not tested via continuous integration." (see comment above)?

Adding a warning is a good idea.

pkubaj:
> I believe most Linux distributions moved to a ppc64le target since little endian (switchable) is supported in POWER8 and later. Using qemu might be an option on Linux. AIX is a good point, but I don't think our CIs support an integration.
Currently, Void Linux, Adelie and Gentoo support ppc64. Void and Adelie are ELFv2, but Gentoo is still ELFv1 on ppc64.
There are also FreeBSD and OpenBSD (both ELFv2).
This PR was done on FreeBSD.

CMakeLists.txt
pkubaj: Done.

pkubaj (commented):
Comment has been addressed.

baentsch (approved):

PR #1161, opened 2021-12-20
Mark stack non-executable when compiling with clang or gcc

### LINKED ISSUE #1159: liboqs.so: found executable STACK entry: GNU_STACK RWE
I decided to enable strict ELF checker in ALT Linux when building liboqs package, and it reported that `liboqs.so` have executable stack:

`verify-elf: ERROR: ./usr/lib64/liboqs.so.0.7.1: found executable STACK entry:   GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RWE 0x10`

This is output of `readelf --wide --segments -- ./usr/lib64/liboqs.so.0.7.1` and executable stack is marked with letter `E' in flags column.

Executable stack is rare occurrence these days so you might want to fix it somehow.

dstebila:
Fixes #1159


vt-alt:
Thanks. We using gcc to build liboqs (and solution with `-Wa,...` is tested to work). In that case you just adding `-Wl,-z,noexecstack`. I tested this commit and it still produces executable stack. (Maybe some linking is still missing this flag?) Why not to always add `-Wa,--noexecstack` to `CMAKE_ASM_FLAGS` for both compilers?

```
$ grep '\.S' log2 | grep -v noexec -c
4
$ grep '\.S' log2 | grep -v noexec
[2350/2366] : && /usr/bin/cc -pipe -frecord-gcc-switches -Wall -fdiagnostics-color=always -fdiagnostics-color=always -fdiagnostics-color=always -g -O2 -flto=auto -O2 -g -DNDEBUG  src/common/sha3/xkcp_low/CMakeFiles/xkcp_low_keccakp_1600_plain64.dir/KeccakP-1600/plain-64bits/KeccakP-1600-opt64.c.o src/common/sha3/xkcp_low/CMakeFiles/xkcp_low_keccakp_1600times4_serial.dir/KeccakP-1600times4/serial/KeccakP-1600-times4-on1.c.o src/common/sha3/xkcp_low/CMakeFiles/xkcp_low_keccakp_1600_avx2.dir/KeccakP-1600/avx2/KeccakP-1600-AVX2.S.o src/common/sha3/xkcp_low/CMakeFiles/xkcp_low_keccakp_1600times4_avx2.dir/KeccakP-1600times4/avx2/KeccakP-1600-times4-SIMD256.c.o src/common/CMakeFiles/common.dir/aes/aes_ossl.c.o src/common/CMakeFiles/common.dir/sha2/sha2_ossl.c.o src/common/CMakeFiles/common.dir/sha3/xkcp_sha3.c.o src/common/CMakeFiles/common.dir/sha3/xkcp_sha3x4.c.o src/common/CMakeFiles/common.dir/common.c.o src/common/CMakeFiles/common.dir/pqclean_shims/nistseedexpander.c.o src/common/CMakeFiles/common.dir/pqclean_shims/fips202.c.o src/common/CMakeFiles/common.dir/pqclean_shims/fips202x4.c.o src/common/CMakeFiles/common.dir/rand/rand.c.o src/common/CMakeFiles/common.dir/rand/rand_nist.c.o tests/CMakeFiles/test_hash.dir/test_hash.c.o -o tests/test_hash  /usr/lib64/libcrypto.so  -lm && :
...
```
So something is linked without `noexecstack`.

baentsch (changes requested):
`main` branch build of shared lib: output of `$ readelf --wide --segments -- lib/liboqs.so`:

```
Elf-Datei-Typ ist DYN (geteilte Objektadatei)
Entry point 0xc2f0
There are 8 program headers, starting at offset 64

Programm-Header:
  Typ            Offset   VirtAdr            PhysAdr            DateiGr  SpeiGr   Flg Ausr.
  LOAD           0x000000 0x0000000000000000 0x0000000000000000 0x7dc508 0x7dc508 R E 0x200000
  LOAD           0x7dc7b0 0x00000000009dc7b0 0x00000000009dc7b0 0x002538 0x0025e8 RW  0x200000
  DYNAMIC        0x7dd2c0 0x00000000009dd2c0 0x00000000009dd2c0 0x0001f0 0x0001f0 RW  0x8
  NOTE           0x000200 0x0000000000000200 0x0000000000000200 0x000024 0x000024 R   0x4
  TLS            0x7dc7b0 0x00000000009dc7b0 0x00000000009dc7b0 0x000000 0x000008 R   0x4
  GNU_EH_FRAME   0x77f7f8 0x000000000077f7f8 0x000000000077f7f8 0x00c42c 0x00c42c R   0x4
  GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RWE 0x10
  GNU_RELRO      0x7dc7b0 0x00000000009dc7b0 0x00000000009dc7b0 0x001850 0x001850 R   0x1

 Abbildung von Sektion nach Segment:
  Segmentsektionen...
   00     .note.gnu.build-id .gnu.hash .dynsym .dynstr .gnu.version .gnu.version_r .rela.dyn .rela.plt .init .plt .plt.got .text .fini .rodata .eh_frame_hdr .eh_frame 
   01     .init_array .fini_array .data.rel.ro .dynamic .got .got.plt .data .bss 
   02     .dynamic 
   03     .note.gnu.build-id 
   04     .tbss 
   05     .eh_frame_hdr 
   06     
   07     .init_array .fini_array .data.rel.ro .dynamic .got
```

Same command run on `ds-noexecstack`:

```
Elf-Datei-Typ ist DYN (geteilte Objektadatei)
Entry point 0xc2f0
There are 8 program headers, starting at offset 64

Programm-Header:
  Typ            Offset   VirtAdr            PhysAdr            DateiGr  SpeiGr   Flg Ausr.
  LOAD           0x000000 0x0000000000000000 0x0000000000000000 0x7dc508 0x7dc508 R E 0x200000
  LOAD           0x7dc7b0 0x00000000009dc7b0 0x00000000009dc7b0 0x002538 0x0025e8 RW  0x200000
  DYNAMIC        0x7dd2c0 0x00000000009dd2c0 0x00000000009dd2c0 0x0001f0 0x0001f0 RW  0x8
  NOTE           0x000200 0x0000000000000200 0x0000000000000200 0x000024 0x000024 R   0x4
  TLS            0x7dc7b0 0x00000000009dc7b0 0x00000000009dc7b0 0x000000 0x000008 R   0x4
  GNU_EH_FRAME   0x77f7f8 0x000000000077f7f8 0x000000000077f7f8 0x00c42c 0x00c42c R   0x4
  GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RWE 0x10
  GNU_RELRO      0x7dc7b0 0x00000000009dc7b0 0x00000000009dc7b0 0x001850 0x001850 R   0x1

 Abbildung von Sektion nach Segment:
  Segmentsektionen...
   00     .note.gnu.build-id .gnu.hash .dynsym .dynstr .gnu.version .gnu.version_r .rela.dyn .rela.plt .init .plt .plt.got .text .fini .rodata .eh_frame_hdr .eh_frame 
   01     .init_array .fini_array .data.rel.ro .dynamic .got .got.plt .data .bss 
   02     .dynamic 
   03     .note.gnu.build-id 
   04     .tbss 
   05     .eh_frame_hdr 
   06     
   07     .init_array .fini_array .data.rel.ro .dynamic .got 
```

--> Not an awful lot changed: The "E" markings are still there (which shouldn't be the case if I read https://github.com/open-quantum-safe/liboqs/issues/1159#issue-1084234445 correctly). And indeed, this command lets the "E" vanish from the GNU_STACK flags: `cmake -GNinja -DCMAKE_ASM_FLAGS=-Wa,--noexecstack -DBUILD_SHARED_LIBS=ON ..`. So I'm afraid the PR doesn't cut it yet.

baentsch:
@dstebila : Changing the new `add_compile_options` to `add_link_options` seems to do the trick.

dstebila:
> @dstebila : Changing the new `add_compile_options` to `add_link_options` seems to do the trick.

Yeah but then I get build failures with gcc-11 on macOS that I haven't been able to resolve.

dstebila:
> > @dstebila : Changing the new `add_compile_options` to `add_link_options` seems to do the trick.
> 
> Yeah but then I get build failures with gcc-11 on macOS that I haven't been able to resolve.

Wait, maybe not... checking again...

dstebila:
> > > @dstebila : Changing the new `add_compile_options` to `add_link_options` seems to do the trick.
> > 
> > 
> > Yeah but then I get build failures with gcc-11 on macOS that I haven't been able to resolve.
> 
> Wait, maybe not... checking again...

That seems to have resolved my build failure.  I've just pushed with that change, please check again @vt-alt.

dstebila:
Build is failing on our Ubuntu bionic i386 image.  Seems to be related to an old version of CMake that doesn't support add\_link\_options.  Can we upgrade CMake in that image?  (The option was added in CMake 3.13, apparently; which came out in November 2018.)

vt-alt:
> > > > @dstebila : Changing the new `add_compile_options` to `add_link_options` seems to do the trick.
> > > 
> > > 
> > > Yeah but then I get build failures with gcc-11 on macOS that I haven't been able to resolve.
> > 
> > 
> > Wait, maybe not... checking again...
> 
> That seems to have resolved my build failure. I've just pushed with that change, please check again @vt-alt.

This still reports executable stack. Apparently a lot of binaries still not affected by the flags:
```
altair:~/src/liboqs (sisyphus)$ grep -i '\.S' log2 | grep -v noexecstack -c
229
altair:~/src/liboqs (sisyphus)$ grep -c noexecstack log2
16
```
I think `-Wa,--noexecstack` should be added when building `.S` (as this is `as` flag affecting compilation of .S) or `-Wl,-z,noexecstack` should be added when linking [already compiled] `.S.o`.

This patch is tested to work (on GCC11):
```patch
diff --git a/.CMake/compiler_opts.cmake b/.CMake/compiler_opts.cmake
index 9dd5b323..a7ef0756 100644
--- a/.CMake/compiler_opts.cmake
+++ b/.CMake/compiler_opts.cmake
@@ -67,6 +67,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang")
     add_compile_options(-Wextra)
     add_compile_options(-Wpedantic)
     add_compile_options(-Wno-unused-command-line-argument)
+    set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,--noexecstack")

     if(NOT ${OQS_BUILD_ONLY_LIB})
         set(THREADS_PREFER_PTHREAD_FLAG ON)
@@ -117,6 +118,7 @@ elseif(CMAKE_C_COMPILER_ID STREQUAL "GNU")
     add_compile_options(-Wformat=2)
     add_compile_options(-Wfloat-equal)
     add_compile_options(-Wwrite-strings)
+    set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,--noexecstack")

     if(NOT ${OQS_BUILD_ONLY_LIB})
         set(THREADS_PREFER_PTHREAD_FLAG ON)
```
This is just tested to work too too:
```
diff --git a/.CMake/compiler_opts.cmake b/.CMake/compiler_opts.cmake
index 9dd5b323..a66ce44b 100644
--- a/.CMake/compiler_opts.cmake
+++ b/.CMake/compiler_opts.cmake
@@ -67,6 +67,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang")
     add_compile_options(-Wextra)
     add_compile_options(-Wpedantic)
     add_compile_options(-Wno-unused-command-line-argument)
+    set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,--noexecstack")

     if(NOT ${OQS_BUILD_ONLY_LIB})
         set(THREADS_PREFER_PTHREAD_FLAG ON)
@@ -117,6 +118,7 @@ elseif(CMAKE_C_COMPILER_ID STREQUAL "GNU")
     add_compile_options(-Wformat=2)
     add_compile_options(-Wfloat-equal)
     add_compile_options(-Wwrite-strings)
+    add_link_options(-Wl,-z,noexecstack)

     if(NOT ${OQS_BUILD_ONLY_LIB})
         set(THREADS_PREFER_PTHREAD_FLAG ON)
```
I am suggested (by @ldv-alt) that you use `-Wa,--noexecstack` in ASM FLAGS as it's better known solution.

dstebila:
The first option caused build failures when compiling with gcc-11 on my Mac, which I wasn't able to solve.  I can't recall what happened with the second option but will try in the morning.

vt-alt:
> The first option caused build failures when compiling with gcc-11 on my Mac, which I wasn't able to solve.

I am curios what problem it was? Also, is it creates executable stack there too?

dstebila:
> > The first option caused build failures when compiling with gcc-11 on my Mac, which I wasn't able to solve.
> 
> I am curios what problem it was? Also, is it creates executable stack there too?

Here's the error I get.    I ran `cmake -GNinja -DCMAKE_C_COMPILER=gcc-11 ..` and then `ninja`.  What's very confusing to me is that the compilation command says `gcc-11` but then the error message says `clang: error`.

```
[7/2354] Building ASM object src/kem/classic_mceliece/CMakeFiles/cl...eliece_348864f_avx.dir/pqclean_mceliece348864f_avx/syndrome_asm.S.o
FAILED: src/kem/classic_mceliece/CMakeFiles/classic_mceliece_348864f_avx.dir/pqclean_mceliece348864f_avx/syndrome_asm.S.o 
/usr/local/bin/gcc-11  -I/Users/dstebila/Dev/OQS/liboqs/build-gcc/include -I/Users/dstebila/Dev/OQS/liboqs/src/kem/classic_mceliece/pqclean_mceliece348864f_avx -I/Users/dstebila/Dev/OQS/liboqs/src/common/pqclean_shims -Wa,--noexecstack -fPIC -march=native -Werror -Wall -Wextra -Wpedantic -Wstrict-prototypes -Wshadow -Wformat=2 -Wfloat-equal -Wwrite-strings -O3 -fomit-frame-pointer -fdata-sections -ffunction-sections -Wl,-dead_strip -mavx2 -mpopcnt -mbmi -Wno-language-extension-token -MD -MT src/kem/classic_mceliece/CMakeFiles/classic_mceliece_348864f_avx.dir/pqclean_mceliece348864f_avx/syndrome_asm.S.o -MF src/kem/classic_mceliece/CMakeFiles/classic_mceliece_348864f_avx.dir/pqclean_mceliece348864f_avx/syndrome_asm.S.o.d -o src/kem/classic_mceliece/CMakeFiles/classic_mceliece_348864f_avx.dir/pqclean_mceliece348864f_avx/syndrome_asm.S.o -c /Users/dstebila/Dev/OQS/liboqs/src/kem/classic_mceliece/pqclean_mceliece348864f_avx/syndrome_asm.S
clang: error: unsupported option '--noexecstack'
```

dstebila:
The second option (`add_link_options(-Wl,-z,noexecstack)`) compiles fine for me on macOS with gcc-11.  (Well, the build fails, but that's due to #1166 not due to this, as far as I can tell.)

vt-alt:
> > > The first option caused build failures when compiling with gcc-11 on my Mac, which I wasn't able to solve.
> > 
> > I am curios what problem it was? Also, is it creates executable stack there too?
> 
> Here's the error I get. I ran `cmake -GNinja -DCMAKE_C_COMPILER=gcc-11 ..` and then `ninja`. What's very confusing to me is that the compilation command says `gcc-11` but then the error message says `clang: error`.
> 
> ```
> [7/2354] Building ASM object src/kem/classic_mceliece/CMakeFiles/cl...eliece_348864f_avx.dir/pqclean_mceliece348864f_avx/syndrome_asm.S.o
> FAILED: src/kem/classic_mceliece/CMakeFiles/classic_mceliece_348864f_avx.dir/pqclean_mceliece348864f_avx/syndrome_asm.S.o 
> /usr/local/bin/gcc-11  -I/Users/dstebila/Dev/OQS/liboqs/build-gcc/include -I/Users/dstebila/Dev/OQS/liboqs/src/kem/classic_mceliece/pqclean_mceliece348864f_avx -I/Users/dstebila/Dev/OQS/liboqs/src/common/pqclean_shims -Wa,--noexecstack -fPIC -march=native -Werror -Wall -Wextra -Wpedantic -Wstrict-prototypes -Wshadow -Wformat=2 -Wfloat-equal -Wwrite-strings -O3 -fomit-frame-pointer -fdata-sections -ffunction-sections -Wl,-dead_strip -mavx2 -mpopcnt -mbmi -Wno-language-extension-token -MD -MT src/kem/classic_mceliece/CMakeFiles/classic_mceliece_348864f_avx.dir/pqclean_mceliece348864f_avx/syndrome_asm.S.o -MF src/kem/classic_mceliece/CMakeFiles/classic_mceliece_348864f_avx.dir/pqclean_mceliece348864f_avx/syndrome_asm.S.o.d -o src/kem/classic_mceliece/CMakeFiles/classic_mceliece_348864f_avx.dir/pqclean_mceliece348864f_avx/syndrome_asm.S.o -c /Users/dstebila/Dev/OQS/liboqs/src/kem/classic_mceliece/pqclean_mceliece348864f_avx/syndrome_asm.S
> clang: error: unsupported option '--noexecstack'
> ```

Thanks, Perhaps, manual run of this command with `-v` would show why gcc runs clang. It maybe possible that there is non GNU `as` is used.

dstebila:
> Thanks, Perhaps, manual run of this command with `-v` would show why gcc runs clang. It maybe possible that there is non GNU `as` is used.

As discussed in https://github.com/open-quantum-safe/liboqs/issues/1166#issuecomment-1002843644, it is the case that a non-GNU `as` is being used on macOS, even when compiling with gcc-11.  So if it doesn't cause problems elsewhere, then I would go with the second option (`add_link_options(-Wl,-z,noexecstack)`).  The other macOS gcc-11 build problem observed in #1166 is soon to be fixed by #1173; once that lands, my plan is to rebase this PR and use the second option here.

dstebila:
#1173 has landed so I've rebased.  

I tried using the second option (`add_link_options(-Wl,-z,noexecstack)`) but that still causes problems, this time because the linker that's being used is Apple's linker, which doesn't recognize a `-z noexecstack` option.  However, as far as I can tell from [this Apple documentation](https://developer.apple.com/library/archive/documentation/Security/Conceptual/SecureCodingGuide/Articles/BufferOverflows.html) and [a random StackOverflow comment](https://stackoverflow.com/questions/31017817/trying-to-implement-enable-execute-stack-mac-os-x), macOS's linker marks the stack non-executable by default, and you have to explicitly specify `-allow_stack_execute` if you want it to be executable.  if that's to be believed, then I've side-stepped the problem by wrapping the `add_link_options(-Wl,-z,noexecstack)` step in a check that it's not on Darwin.

However I don't know how to check whether the stack is executable or not, other than asking @vt-alt to check.  Ideally we would add such a check to our test suite to automate this.

vt-alt:
I think such test should work:
```
if readelf --wide --segments -- liboqs.so | grep -P 'GNU_STACK\s+(\S+ ){5}\S*E\S*\s'; then
        echo 'Stack is executable'
else
        echo 'Stack is non-executable'
fi
```

dstebila:
> I think such test should work:
> 
> ```
> if readelf --wide --segments -- liboqs.so | grep -P 'GNU_STACK\s+(\S+ ){5}\S*E\S*\s'; then
>         echo 'Stack is executable'
> else
>         echo 'Stack is non-executable'
> fi
> ```

@xvzcf where do you think we should add that in our unit test suite?

xvzcf:
> @xvzcf where do you think we should add that in our unit test suite?

What about adding it in [test_namespace.py](https://github.com/open-quantum-safe/liboqs/blob/main/tests/test_namespace.py) (with the grepping done by some python construction) and renaming the file to something like `test_binary.py` or `test_solib.py`?

dstebila:
> > @xvzcf where do you think we should add that in our unit test suite?
> 
> What about adding it in [test_namespace.py](https://github.com/open-quantum-safe/liboqs/blob/main/tests/test_namespace.py) (with the grepping done by some python construction) and renaming the file to something like `test_binary.py` or `test_solib.py`?

That seems like a good place, perhaps renaming to `test_library.py`.  

@vt-alt or anyone else, do you know a command for checking this on macOS or any platform other than Linux?

vt-alt:
@dstebila No I'm not familiar with macos. But, googling says there possible to install `greadelf` tool, also if you have recent llvm it possible have `llvm-readelf`.

I just experimented on freebsd 13 & openbsd 7 & netbsd 9: `readelf --wide --segments` works the same, but there is no `grep -P` (grep with perl regex, so `\S`/`\s` should be replaced).

vt-alt:
AH I just noticed there is no `GNU_STACK` on openbsd 7 & netbsd 9, but it present on freebsd 13.

vt-alt:
I just experimented on `MacOSX 12.1`. `objdump`, in theory, could show executable stack too (as `STACK` segment in Program Header view with `-x` or `-p` options), and there on macos `objdump` is aliased to `llvm-objdump`. But I don't see `STACK` section there (for any `dylib` I tried looking to). Perhaps, this trick does not work on macos.

vt-alt:
And finally I tried Solaris 10 and 11, there is `readelf.py` which is incapable of showing stack, and there was no `objdump`. There is `elfdump` tool which can show `SUNW_NXSTACK` or `SUNW_SX_NXSTACK` tags (with `-d` option) for executables, unfortunately it does not show this for libraries. Perhaps, it does not even support this for libraries:, quoting https://en.wikipedia.org/wiki/Executable_space_protection#Solaris

> Solaris has supported globally disabling stack execution on SPARC processors since Solaris 2.6 (1997); in Solaris 9 (2002), support for disabling stack execution on a per-executable basis was added.

So it's only global and per-binary thing there.

dstebila:
Thanks @vt-alt for trying to figure it out on lots of different platforms.  

I've added a unit test in the newly renamed `test_binary.py` which only runs on Linux and only if a `liboqs.so` file exists, in which case it checks for the right flags in the `GNU_STACK` portion of the `readelf` output as suggested by https://github.com/open-quantum-safe/liboqs/pull/1161#issuecomment-1005146830.  Will wait to see how it does on all our various test platforms, I imagine there might be some quirks that have to be worked in.

.CMake/compiler_opts.cmake
```diff
@@ -117,6 +118,9 @@ elseif(CMAKE_C_COMPILER_ID STREQUAL "GNU")
     add_compile_options(-Wformat=2)
     add_compile_options(-Wfloat-equal)
     add_compile_options(-Wwrite-strings)
+    if (NOT CMAKE_SYSTEM_NAME STREQUAL "Darwin")
```
baentsch, line 121: Is this guard necessary? Why then is it not active for Clang above?

baentsch (approved):
Test ran OK on linux shared lib:

> [gw18] [ 15%] PASSED tests/test_binary.py::test_non_executable_stack

(in https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/1968/workflows/7da4b1f6-dfb6-42e6-93f5-544c8115b55d/jobs/16665)

.CMake/compiler_opts.cmake
dstebila: Yes it's necessary (I get a build failure if it's not there). But it is not necessary for clang above -- and I can't explain why. All I know at this point is that this configuration seems to work.

vt-alt:
Btw, cmake have this built-in variable https://cmake.org/cmake/help/v3.0/variable/APPLE.html

PR #1170, opened 2021-12-27
adding constant time test as weekly github action

baentsch:
Tries to implement #1169. Shall replace [CCI weekly run](https://github.com/open-quantum-safe/liboqs/blob/5ea49c2a446945d6ddece9c0a11d931b7c3f040f/.circleci/config.yml#L368-L402) if this works OK for a few weeks.

AppVeyor still failing because of #1163

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -32,5 +32,5 @@ jobs:
         working-directory: build
       - name: Run tests
         timeout-minutes: 60
-        run: mkdir -p tmp && python3 -m pytest --verbose tests/test_code_conventions.py ${{ matrix.PYTEST_ARGS }}
+        run: mkdir -p tmp && python3 -m pytest --verbose ${{ matrix.PYTEST_ARGS }}
 
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -1,0 +1,36 @@
+name: Weekly constant time tests
+
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/weekly.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Configure
+        run: mkdir build && cd build && cmake -GNinja ${{ matrix.CMAKE_ARGS }} .. && cmake -LA ..
+      - name: Build
+        run: ninja
+        working-directory: build
+      - name: Run tests
+        timeout-minutes: 60
+        run: mkdir -p tmp && python3 -m pytest --verbose tests/test_code_conventions.py ${{ matrix.PYTEST_ARGS }}
```
dstebila, line 35: Why are you running test_code_conventions.py?  Looking at the corresponding CircleCI job, the run command seems to be: `python3 -m pytest --verbose --ignore=tests/test_code_conventions.py --junitxml=build/test-results/pytest/test-results.xml << parameters.PYTEST_ARGS >>`

dstebila (changes requested):

.github/workflows/weekly.yml
baentsch: Indeed superfluous. Thanks for the catch.

dstebila (approved):

PR #1172, opened 2022-01-01
extend weekly run timeout [skip ci]

baentsch:
Extending maximum weekly constant runtime test time to 2h.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -31,6 +31,6 @@ jobs:
         run: ninja
         working-directory: build
       - name: Run tests
-        timeout-minutes: 60
+        timeout-minutes: 120
         run: mkdir -p tmp && python3 -m pytest --verbose ${{ matrix.PYTEST_ARGS }}
 


dstebila (approved):

baentsch:
This change allowed the constant time tests to proceed from [71%](https://github.com/open-quantum-safe/liboqs/runs/4676886314?check_suite_focus=true) (1h timeout) to [76%](https://github.com/open-quantum-safe/liboqs/runs/4745481918?check_suite_focus=true) (2h timeout) -> How many hours do we want to allow? Any algorithms to prune instead?

dstebila:
> This change allowed the constant time tests to proceed from [71%](https://github.com/open-quantum-safe/liboqs/runs/4676886314?check_suite_focus=true) (1h timeout) to [76%](https://github.com/open-quantum-safe/liboqs/runs/4745481918?check_suite_focus=true) (2h timeout) -> How many hours do we want to allow? Any algorithms to prune instead?

We could probably prune SPHINCS and Rainbow?

baentsch:
> We could probably prune SPHINCS and Rainbow?

Completely ditching some families sounds not right. The more I think about it, we should keep all (until NIST prunes :) and extend the permitted runtime to 10 hours. Alternatively we should clearly state (in the docs, e.g.) that we do (not) test some algorithms at the same level as others (because they're prohibitively expensive -- which may possibly be of interest to NIST).

dstebila:
> > We could probably prune SPHINCS and Rainbow?
> 
> Completely ditching some families sounds not right. The more I think about it, we should keep all (until NIST prunes :) and extend the permitted runtime to 10 hours. Alternatively we should clearly state (in the docs, e.g.) that we do (not) test some algorithms at the same level as others (because they're prohibitively expensive -- which may possibly be of interest to NIST).

Are we allowed 10 hour jobs?  If so then yes let's do that.  At least we do a full constant time test run prior to release as per our [release process](https://github.com/open-quantum-safe/liboqs/wiki/Release-process).

baentsch:
> Are we allowed 10 hour jobs? If so then yes let's do that. 

I don't know, but why not: #1175 . If that fails, we'll prune SPHINCS and Rainbow as per your suggestion, @dstebila

Issue #1166, opened 2021-12-21
Keccak build failures on macOS Intel with gcc-11


dstebila:
Building liboqs with gcc-11 on an Apple Intel Mac yields build failures related to the Keccak xkcp_low code.

Reproduce by checking out main, and then running `cmake -GNinja -DCMAKE_C_COMPILER=gcc-11 ..`.

```
[1354/2353] Building ASM object src/common/sha3/xkcp_low/CMakeFiles/xkcp_low_keccakp_1600_avx2.dir/KeccakP-1600/avx2/KeccakP-1600-AVX2.S.
FAILED: src/common/sha3/xkcp_low/CMakeFiles/xkcp_low_keccakp_1600_avx2.dir/KeccakP-1600/avx2/KeccakP-1600-AVX2.S.o 
/usr/local/bin/gcc-11  -I/Users/dstebila/Dev/OQS/liboqs/build/include -fPIC -march=native -Werror -Wall -Wextra -Wpedantic -Wstrict-prototypes -Wshadow -Wformat=2 -Wfloat-equal -Wwrite-strings -O3 -fomit-frame-pointer -fdata-sections -ffunction-sections -Wl,-dead_strip -Wbad-function-cast -Wa,-defsym,macOS=1 -MD -MT src/common/sha3/xkcp_low/CMakeFiles/xkcp_low_keccakp_1600_avx2.dir/KeccakP-1600/avx2/KeccakP-1600-AVX2.S.o -MF src/common/sha3/xkcp_low/CMakeFiles/xkcp_low_keccakp_1600_avx2.dir/KeccakP-1600/avx2/KeccakP-1600-AVX2.S.o.d -o src/common/sha3/xkcp_low/CMakeFiles/xkcp_low_keccakp_1600_avx2.dir/KeccakP-1600/avx2/KeccakP-1600-AVX2.S.o -c /Users/dstebila/Dev/OQS/liboqs/src/common/sha3/xkcp_low/KeccakP-1600/avx2/KeccakP-1600-AVX2.S
clang: error: no such file or directory: 'macOS=1'
[1367/2353] Building C object src/common/sha3/xkcp_low/CMakeFiles/x...s4_avx2.dir/KeccakP-1600times4/avx2/KeccakP-1600-times4-SIMD256.c.o
ninja: build stopped: subcommand failed.
```

baentsch:
[...]

PR #1173, opened 2022-01-01
Update XKCP and avoid assembler .ifdef directives

jschanck:
This is an attempt to fix Issue #1166. I don't have a machine to reproduce the issue on, so I'll need someone else to test the fix. The first pair of commits updates XKCP to the latest version. This alone won't fix the issue. The second pair of commits switches some `.ifdef` and `.ifndef` assembler directives to C preprocessor `#ifdef` and `#ifndef`, so we don't have to pass the `defsym` arguments which seem to be the problem in #1166.


baentsch:
Thanks for this fix. I just tried to push an update to add gcc-11 to OSX-CI to your repo but got a `permission denied`, so here's the diff if you'd want to give it a try:

```
git diff issue.1166
diff --git a/.circleci/config.yml b/.circleci/config.yml
index c0715647..6cb911a4 100644
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,7 +180,7 @@ jobs:
       - checkout # change this from "checkout" to "*localCheckout" when running CircleCI locally
       - run:
           name: Install dependencies
-          command: env HOMEBREW_NO_AUTO_UPDATE=1 brew install cmake ninja && pip3 install pytest pytest-xdist pyyaml
+          command: env HOMEBREW_NO_AUTO_UPDATE=1 brew install cmake ninja gcc@11 && pip3 install pytest pytest-xdist pyyaml
       - run:
           name: Get system information
           command: sysctl -a | grep machdep.cpu
@@ -349,6 +349,10 @@ workflows:
           name: arm64
           PYTEST_ARGS: --numprocesses=auto --maxprocesses=10
           CMAKE_ARGS: -DOQS_DIST_BUILD=ON
+      - macOS:
+          <<: *require_buildcheck
+          name: macOS-gcc11
+          CMAKE_ARGS: -DCMAKE_C_COMPILER=gcc-11
       - macOS:
           <<: *require_buildcheck
           name: macOS-noopenssl
```

dstebila (approved):

dstebila:
It works for me, thanks John!  I've pushed a commit with the new gcc-11 macOS CI job that Michael wanted to add, and assuming that passes then we'll merge it in.

baentsch (approved):

Issue #1149, opened 2021-12-09
Add M1 to CI


baentsch:
Maybe #1069 could be expanded to also run CI on M1 (adding OSX-aarch64 testing).

Unfortunately, [github runners are not available for M1](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#supported-architectures-and-operating-systems-for-self-hosted-runners). On the bright side, [CCI OSX-M1 self-hosted runners are supported](https://circleci.com/docs/2.0/runner-overview/#supported).

baentsch:
@dstebila : Would you be OK if I'd install a CCI self-hosted runner onto your M1 box? Would you want to create a separate, possibly more limited user for that?

PR #1175, opened 2022-01-10
extend timeout to 10h

baentsch:
Further extension of weekly runs as per #1172

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -31,6 +31,6 @@ jobs:
         run: ninja
         working-directory: build
       - name: Run tests
-        timeout-minutes: 120
+        timeout-minutes: 600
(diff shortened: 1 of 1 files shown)


dstebila (approved):
I note that this adds .dsci.yml, if you're intentionally adding dsci support in this PR that's fine.

baentsch:
> I note that this adds .dsci.yml, if you're intentionally adding dsci support in this PR that's fine.

Otherwise there's a (the M1) test failing (b/o lack of .dsci.yml file).... As it's activated on the repo, we may as well add the file....

baentsch:
Fixes #1149. OK to move this into use then, @dstebila ?

baentsch:
Now we know [the maximum runtime: 6h](https://github.com/open-quantum-safe/liboqs/runs/4823965410?check_suite_focus=true). It moved us to 92% (up from 76% within 2h). I'll now prune most Rainbow and SPHINCS tests...

Issue #1164, opened 2021-12-21
Build failures with gcc on Apple M1


dstebila:
Building liboqs with gcc-11 on an Apple M1 Mac yields build failures related to Kyber and Saber NTT code.

Reproduce by checking out 0.7.1, and then running `cmake -GNinja -DCMAKE_C_COMPILER=gcc-11 ..`.  (Note that there are other build failures on HEAD when building with gcc-11 on macOS which are independent of this which I'll create a separate issue for.)

```
[667/1230] Building ASM object src/kem/kyber/CMakeFiles/kyber_1024_aarch64.dir/pqclean_kyber1024_aarch64/__asm_base_mul.S.o
FAILED: src/kem/kyber/CMakeFiles/kyber_1024_aarch64.dir/pqclean_kyber1024_aarch64/__asm_base_mul.S.o 
/opt/homebrew/bin/gcc-11  -I/Users/dstebila/Dev/OQS/liboqs/build/include -I/Users/dstebila/Dev/OQS/liboqs/src/kem/kyber/pqclean_kyber1024_aarch64 -I/Users/dstebila/Dev/OQS/liboqs/src/common/pqclean_shims -arch arm64 -fPIC -mcpu=native -Werror -Wall -Wextra -Wpedantic -Wstrict-prototypes -Wshadow -Wformat=2 -Wfloat-equal -Wwrite-strings -O3 -fomit-frame-pointer -fdata-sections -ffunction-sections -Wl,-dead_strip -MD -MT src/kem/kyber/CMakeFiles/kyber_1024_aarch64.dir/pqclean_kyber1024_aarch64/__asm_base_mul.S.o -MF src/kem/kyber/CMakeFiles/kyber_1024_aarch64.dir/pqclean_kyber1024_aarch64/__asm_base_mul.S.o.d -o src/kem/kyber/CMakeFiles/kyber_1024_aarch64.dir/pqclean_kyber1024_aarch64/__asm_base_mul.S.o -c /Users/dstebila/Dev/OQS/liboqs/src/kem/kyber/pqclean_kyber1024_aarch64/__asm_base_mul.S
/Users/dstebila/Dev/OQS/liboqs/src/kem/kyber/pqclean_kyber1024_aarch64/__asm_base_mul.S:10:3: error: unknown directive
  .type PQCLEAN_KYBER1024_AARCH64_asm_point_mul_extended, %function
  ^
[...]

PR #1181, opened 2022-01-25
Update to picnic 3.0.11 (fixes #1178)

### LINKED ISSUE #1178: Execution failures with picnic on M1 when using gcc-11
Continuation of #1164 (just for picnic)

sebastinas:
This update fixes Picnic built with GCC on M1 (#1178).


baentsch (approved):
FWIW, this builds (and runs) OK on M1 static and shared. SHA2 common code still has a problem (#1164) but that shouldn't stop a merge of this. Thanks, @sebastinas !

dstebila (approved):

PR #1184, opened 2022-02-01
Resolves an issue when building sha2 using arm crypto extensions with gcc11

Martyrshot:
This PR resolves a build issue when using gcc11 on Darwin.

The issue is resolved by specifying the architecture version and explicitly enabling crypto extensions. After running `speed_common`, it appeared that there were no significant performance differences between armv8-a and armv8.5-a, so armv8-a was used to maximize compatibility.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -40,6 +40,9 @@ else()
     elseif (OQS_USE_ARM_SHA2_INSTRUCTIONS)
        # Assume we are compiling native
        set(SHA2_IMPL ${SHA2_IMPL} sha2/sha2_ni.c)
+       if ((CMAKE_SYSTEM_NAME MATCHES "Darwin") AND (${CMAKE_C_COMPILER_ID} STREQUAL "GNU"))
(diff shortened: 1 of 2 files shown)


Martyrshot:
I guess my thought was that we would be able to support older A series (iphone ipad) chips, but as those are mobile processors and it's likely that there won't be people running Darwin on them it probably doesn't make sense to support anything older than what the M1 is built on?

iOS and iPadOS are both based on Darwin, but I'm not sure how practical it is for an individual to build a c library and ship it with their app.

baentsch:
> older A series (iphone ipad) chips,

I stand corrected. I never thought about those platforms. Then I'm fine leaving this PR as-is -- just in case someone wants `liboqs` on such platform. The suggestion to point out how to (slightly) speed up things on M1 in the build-Wiki remains valid, though.

Martyrshot:
Since there are so many version of the arm instruction set, it might be worth building in the ability to specify which version to build for rather than the compiler picking by itself. It would kind of be a pseudo-cross-compilation I guess? Just instead of cross compiling for architecture, it would be instruction set versions. I'm not sure if it's worth it or not.

baentsch:
> I'm not sure if it's worth it or not.

Me neither. Maybe do an issue for this and see whether someone takes it on?

PR #1189, opened 2022-02-16
Suppress constant time check for public matrix generation on Kyber AVX2

### LINKED ISSUE #1187: Constant time checks for Kyber on AVX2 failing
This was picked up by the weekly builds: https://github.com/open-quantum-safe/liboqs/runs/5080898885?check_suite_focus=true#step:6:339

Not sure what changed -- no changes to the Kyber Intel code in a few months, as far as I can tell.

dstebila:
Fixes #1187.


dstebila:
I think this fixes it, but I couldn't reproduce locally, so I'm not certain. I've temporarily added a Github actions job that will run the test immediately rather than waiting for the weekly job; that should be removed from `.github/workflows/linux.yml` before merging.

baentsch (commented):
FWIW, LGTM. Amazing to see this test passes in 8sec when just testing Kyber* (as opposed to 2h when testing most algs...). Thanks also for updating to the "focal" image.

bhess:
Thanks @dstebila. I'll double check it locally.

bhess (changes requested):
I ran it a few times and the new rule works fine. In the suggested change I added one more rule that captures the related Memcheck:Value8 exception, which was also reported in the failed weekly run.

bhess (approved):

PR #1194, opened 2022-03-04
ARMv8 optimized Dilithium

Martyrshot:
This PR adds multi-upstream support for signatures to copy_from_upstream.

Arm optimized version of Dilithium2, 3, 5 have also been pulled in from pqclean. I could of things to point out:
1) pqclean's implementation of Dilithium does not support randomized signing. After a discussion with @dstebila, I have patched the arm implementation so that it only supports randomized signing. This was done instead of pushing to upstream since we are only pulling one implementation of dilithium from pqclean. Once upstream gets patched to support randomized signing, then this patch will need to be revised.

2) The `pqclean_name`s in the `copy_from_upstream.yml` file have been updated to match the formatting in pqclean's repo. This caused several of the dilithium folders to be renamed. I think that might be part of why the ci is failing currently, but I'm not sure. All tests pass on my M1 mac.


docs/algorithms/sig/dilithium.md
```diff
@@ -22,57 +28,60 @@
...
+|:-----------------------------------:|:-------------------------|:----------------------------|:--------------------------------|:------------------------|:-----------------------------------|:-----------------------------------------------|:----------------------|
+|  [Primary Source](#primary-source)  | ref                      | All                         | All                             | None                    | True                               | True                                           | False                 |
+|  [Primary Source](#primary-source)  | avx2                     | x86\_64                     | Darwin,Linux                    | AVX2,POPCNT             | True                               | True                                           | False                 |
+| [pqclean-aarch64](#pqclean-aarch64) | aarch64                  | ARM64\_V8                   | Linux,Darwin                    | None                    | True                               | False                                          | False                 |
```
bhess, line 35: Does the aarch64 version use any flags that can be documented, e.g. neon?

bhess, line 35: Does the aarch64 version use any CPU flags that can be documented, e.g. neon?

src/kem/kyber/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ endif()
 if(OQS_ENABLE_KEM_kyber_512_90s_avx2 OR OQS_ENABLE_KEM_kyber_768_90s_avx2 OR OQS_ENABLE_KEM_kyber_1024_90s_avx2)
     add_library(kyber_common_aes OBJECT pqcrystals-kyber_common_aes/aes256ctr.c)
     target_include_directories(kyber_common_aes PRIVATE ${CMAKE_CURRENT_LIST_DIR}/pqcrystals-kyber_common_aes)
-    target_compile_options(kyber_common_aes PRIVATE -maes -msse2 -mssse3)
+    target_compile_options(kyber_common_aes PRIVATE-maes-msse2-mssse3)
```
bhess, line 17: I don't think this will compile without the whitespaces.

src/sig/dilithium/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ endif()
 if(OQS_ENABLE_SIG_dilithium_2_aes_avx2 OR OQS_ENABLE_SIG_dilithium_3_aes_avx2 OR OQS_ENABLE_SIG_dilithium_5_aes_avx2)
     add_library(dilithium_common_aes OBJECT pqcrystals-dilithium_common_aes/aes256ctr.c)
     target_include_directories(dilithium_common_aes PRIVATE ${CMAKE_CURRENT_LIST_DIR}/pqcrystals-dilithium_common_aes)
-    target_compile_options(dilithium_common_aes PRIVATE -maes -msse2 -mssse3)
+    target_compile_options(dilithium_common_aes PRIVATE-maes-msse2-mssse3)
```
bhess, line 17: whitespaces, also below

docs/algorithms/sig/picnic.md
```diff
@@ -25,128 +27,128 @@
 |   picnic3\_L3    | EUF-CMA          |                    3 |                        49 |                        73 |                    35028 |
 |   picnic3\_L5    | EUF-CMA          |                    5 |                        65 |                        97 |                    61028 |
 
-## picnic\_L1\_FS implementation characteristics
+## picnic_L1_FS implementation characteristics
```
bhess, line 30: The backslash before underscore may be needed in markdown.

bhess (commented):
The dilithium patch looks ok to me.

src/sig/dilithium/CMakeLists.txt
Martyrshot: I believe this didn't cause issues on my machine is because the only builds that were effected were avx2 optimized implementations. I *believe* it's resolved now, and it looks like the CI is passing.

docs/algorithms/sig/dilithium.md
baentsch: Agree (with the removal), if "aarch64" == "armv8" == "neon always present". Is there some documentation stating this?

src/sig/dilithium/CMakeLists.txt
baentsch: Confirm it's resolved on my machine (which basically is the same as CI). Thanks!

docs/algorithms/sig/dilithium.md
Martyrshot: According to [this](https://developer.arm.com/documentation/den0024/a/AArch64-Floating-point-and-NEON):
```
Both floating-point and NEON are required in all standard ARMv8 implementations. However, implementations targeting specialized markets may support the following combinations:

No NEON or floating-point.
Full floating-point and SIMD support with exception trapping.
Full floating-point and SIMD support without exception trapping.
```

It looks like there is a chance that NEON might not be on a chip, but only in specialized settings.

baentsch: That really seems like nothing we ought to worry about then. Thanks for the pointer!

baentsch (approved):
Everything now looks OK for me; also  a test-run of `copy_from_upstream.py` was all clean (except the expected diff from the recently manually added warning regarding Rainbow).

dstebila (approved):

Martyrshot:
I have just finished testing this on our rock64 and raspi, and all tests pass. This should be ready to merge.

PR #1200, opened 2022-04-07
Edits to key scheduling for AES on ARM to be constant time

tedeaton:
This changes how key scheduling is handled for AES when building for platforms with ARM intrinsics. The AES key scheduler we were using was not constant time, so this PR removes it and uses the C implementation already in aes_c.c. Since the ARM AES implementation doesn't use bit slicing, a version of key scheduling without bit slicing is added to aes_c.c.


dstebila (approved):

dstebila:
Can you confirm this passes Valgrind's constant time tests?

tedeaton:
I can! Valgrind's constant time tests no longer flag key scheduling anywhere.

PR #1219, opened 2022-06-03
Use `SecRandomCopyBytes` for system randomness on iOS (to allow building on iOS)

### LINKED ISSUE #1217: Build fails on iOS due to missing `sys/random.h`
I'm having trouble building this project for iOS, running into missing header issues with `sys/random.h`.

It looks like `src/common/rand/rand.c` includes this header for any platform that defines `__APPLE__`, which I believe iOS does (apologies, I'm not actually an iOS developer).

Unfortunately, the iOS SDK does not provide `sys/random.h` (I believe access to either `getentropy` or even directly to `urandom` is not permitted in the iOS sandbox...)

Using `OQS_USE_OPENSSL` doesn't help, because the `#include <sys/random.h>` isn't guarded by that flag and so the header is included regardless.

Potentially, a relatively-easy fix would be to change the `ifdef` guards to not attempt to `#include <sys/random.h>` if `OQS_USE_OPENSSL` is set. This would, I believe, allow building for iOS by just relying on OpenSSL's PRNG.

Additionally, it might be nice to support iOS's PRNG API `SecRandomCopyBytes`.

I apologize if I've missed something either in this project or in iOS development!

zanebeckwith:
Building on iOS is currently broken (cf. #1217), due to the `rand` functionality attempting to include the `sys/random.h` header (which is not available on iOS).

This series allows building for iOS by:
- Using the `SecRandomCopyBytes` secure system randomness function when building for iOS, rather than using `getentropy` or reading from `urandom`
  - This means we don't include the `sys/random.h` header when building for iOS
  - Note that the CMake check for `getentropy` will already find that `getentropy` is not available on iOS, so otherwise (without this patch) on iOS we would be attempting to read from `urandom` (which is also not allowed on iOS)
- Links against the iOS `Security` "framework", when building for iOS

Fixes #1217

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -80,6 +80,10 @@ endif()
 if(${OQS_USE_OPENSSL})
     target_link_libraries(oqs PRIVATE ${OPENSSL_CRYPTO_LIBRARY})
 endif()
+if(IOS)
+    find_library(IOS_SECURITY Security)
+    target_link_libraries(oqs PUBLIC ${IOS_SECURITY})
+endif()
 
 target_include_directories(oqs
                            PUBLIC
--- a/src/common/rand/rand.c
+++ b/src/common/rand/rand.c
@@ -9,7 +9,12 @@
 #include <unistd.h>
 #include <strings.h>
 #if defined(__APPLE__)
+#include <TargetConditionals.h>
+#if TARGET_OS_IPHONE || TARGET_IPHONE_SIMULATOR
+#include <Security/SecRandom.h>
+#else
 #include <sys/random.h>
+#endif
 #else
 #include <unistd.h>
 #endif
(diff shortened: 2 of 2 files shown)


baentsch (approved):
Thanks for your contribution! The changes seem logical and [CI passes within the OQS namespace -- see here](https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/2039/workflows/a97f52a1-fc08-41da-b125-f6b752324be0)

zanebeckwith:
> Thanks for your contribution! The changes seem logical and [CI passes within the OQS namespace -- see here](https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/2039/workflows/a97f52a1-fc08-41da-b125-f6b752324be0)

Ah, thank you! I was wondering how to get CI working!

Thank you for the review!

dstebila:
Thanks @zanebeckwith for the code, I've merged it. 

And thanks @baentsch for running it, I couldn't figure out how to convince CircleCI to run it, what did you do?

baentsch:
> what did you do?

Move the PR's files into a separate (open-quantum-safe/)`liboqs` branch just for testing.

Issue #1218, opened 2022-06-03
Windows-msys2 build failing on picnic


baentsch:
There's a recurring [CI error](https://ci.appveyor.com/project/dstebila/liboqs/builds/43744511/job/4f372pki566y29s9) on main branch that's apparently now also holding #1210. Who knows about/looks after that environment? @christianpaquin ? Or shall we drop this? Anyone using this? But then again, the consistent error message about an array index error in picnic may be worth while checking out...

christianpaquin:
I'm not interested in the platform, personally. I'm ok to drop picnic per #1220.

sebastinas:
I'd appreciate to be pinged on issues with Picnic.

baentsch:
@sebastinas thanks for your interest in getting involved with Picnic issues we find. What's your preference on this issue then: Do you want to look into this/provide a solution or shall we close/drop/disregard it?

sebastinas:
My understanding is that msys2 support is being dropped in #1222. So I'll eventually look into fixing it upstream, but it's not a priority.

PR #1210, opened 2022-05-13
Adding aarch64 CPU feature detection for FreeBSD

Martyrshot:
This PR adds aarch64 CPU feature detection support when compiling for FreeBSD.

* [ ] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [ ] Does this PR change the the list of algorithms available -- either adding, removing, or renaming?  (If so, PRs in OQS-OpenSSL, OQS-BoringSSL, and OQS-OpenSSH will also be required by the time this is merged.)

PR #1220, opened 2022-06-06
workaround for picnic under msys2

baentsch:
Disable testing picnic under msys2 & update documentation.

Temporary workaround for #1218 unblocking #1210: @Martyrshot : If this lands, please rebase and re-run #1210.


docs/algorithms/sig/picnic.md
```diff
@@ -9,6 +9,9 @@
   - **Source**: https://github.com/pq-crystals/kyber/commit/faf5c3fe33e0b61c7c8a7888dd862bf5def17ad2 with copy_from_upstream patches
   - **Implementation license (SPDX-Identifier)**: CC0-1.0
 
+## Test limitation
```
dstebila, line 12: I guess this note will get wiped out when the Markdown documentation is regenerated, but hopefully the main problem will be solved by then.

dstebila (approved):

Issue #1222, opened 2022-06-08
Drop msys2 build in Windows CI


dstebila:
#1218 brought to light that we're currently doing CI builds on msys2 on Windows, but there are failures.  After discussion, none of the core team is making use of msys2. So the proposal is to drop msys2 from the CI builds. Please speak up here if you object, otherwise we'll do so around June 22.

mcoret:
msys2 is the only toolchain (in my experience), which allows testing SIKE primitives as a shared library under Windows. Mingw64/MSVC builds only expose SIDH interface in DLL exports (along all other non-isogenious primitives). There's no SIKE. Is it a bug?

baentsch:
I just tried on a Windows VM and SIKE appears (and operates) like any other algorithm when compiled into a DLL (`cmake` parameter `BUILD_SHARED_LIB` ) using MSVC. So can you please elaborate a bit more what you mean by the above?

mcoret:
Sure, just compiled on Visual Studio Community 2022, using
cmake .. -G Ninja -DBUILD_SHARED_LIBS=ON
ninja
There are OQS_KEM_sidh_* exports in the resulting DLL, no sike.

baentsch:
(Now) understood (and confirmed: `dumpbin /exports oqs.dll` reports no SIKE, but indeed SIDH symbols). As I have no real experience with the Windows toolchain (and so far thought both algorithm families are handled identically in `liboqs`), this is to ask @christianpaquin for help.

As stated above, both algorithms _are_ enabled:
```
>.\tests\Debug\test_kem.exe
Usage: test_kem algname
[...]

PR #1224, opened 2022-06-11
Fix SIKE exports in DLL

mcoret:
Following the discussion in #1222, this PR addresses the SIKE functions export issue, by introducing a new directive `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)` in the CMakeLists.txt file.


baentsch (approved):
Thanks for your contribution! LGTM. Before merging, @christianpaquin may want to comment, too: DLL size increased from 9.09MB to 9.36MB and instead of 358 symbols, a grand total of 4565 symbols are now exported by the DLL (including all SIKE-subroutine symbols, obviously). So instead of merging we could also add the one-liner to https://github.com/open-quantum-safe/liboqs/wiki/Platform-specific-notes-for-building-liboqs#using-msvc to not grow the DLL by default....

christianpaquin:
I'm not sure what are the implications (and side effects) of using this directive (we are not using it elsewhere in the library). The impact on the DLL feels like a big change, so that might be better for the next OQS version update. Seems like the SIKE code might need to export the required symbols explicitly; documenting the issue might be a good solution for now.

dstebila:
The ideal solution would be to just fix the visibility of these particular SIKE symbols.

We use the macros `OQS_API` to indicate which symbols should be exported; it's defined in https://github.com/open-quantum-safe/liboqs/blob/main/src/common/common.h#L73.  Is there something in the SIKE code where `OQS_API` is not being properly attached to the missing SIKE functions?

christianpaquin:
> The ideal solution would be to just fix the visibility of these particular SIKE symbols.

Correct. We'd need to look into what is missing.

baentsch:
> Is there something in the SIKE code where OQS_API is not being properly attached to the missing SIKE functions?

At first glance, no: All SIKE APIs are "adorned" in the same way as the properly exported SIDH functions. A bit of a mystery...

> The impact on the DLL feels like a big change, 

Completely agree. So then I would suggest closing this PR without merge, add a comment pointing to this `cmake`variable in https://github.com/open-quantum-safe/liboqs/wiki/Platform-specific-notes-for-building-liboqs#building-on-windows as a "workaround" and leave #1222 unresolved until the SIKE symbols get properly exported. OK for everyone?

baentsch:
Thanks again for the contribution. The underlying issue will be resolved for real in #1227 and the workaround is documented for now in https://github.com/open-quantum-safe/liboqs/wiki/Platform-specific-notes-for-building-liboqs#using-msvc

PR #1241, opened 2022-06-28
Bump Picnic to 3.0.15

sebastinas:
This one should also fix the remaining scan-build warning.


dstebila (approved):

baentsch (commented):
> This one should also fix the remaining scan-build warning.

Confirmed, thanks. The [versioning information](https://github.com/sebastinas/liboqs/blob/908d7e3421ad5c8631aed46ef840b4533899de18/src/sig/picnic/sig_picnic.c#L128) seems out of step now, though.

baentsch (approved):

PR #1250, opened 2022-07-07
Deal with the issue identified by valgrind

splasky:
* Fix memory leak detected by valgrind

This commit fixed the memory leak problem inside the tests directory.

No.
No.


baentsch (approved):
Thanks for this catch and your contribution!

dstebila (approved):

Issue #1192, opened 2022-03-03
Rainbow update


baentsch:
Following the [discussions here](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/KFgw5_qCXiI?pli=1) this issue is to track required adaptations to Rainbow: Already discussed:

- [x] Adding warnings to the liboqs documentation
- [ ] Removing Rainbow level 1
- [ ] Downgrading higher Rainbow levels
- [ ] Rebuilding all OQS downstream software packages (incl. new oqs-openssl hybrids) 

Final resolution of this issue is expected when NIST announces the next PQC evaluation decisions.

baentsch:
Suggest to close as being covered by #1245

dstebila:
As much as it might cause some downstream pain, I am now thinking we probably should remove Rainbow level 1 before the 0.7.2 release, since it is known-broken.

baentsch:
> we probably should remove Rainbow level 1 before the 0.7.2 release, since it is known-broken

No objections. To the contrary, I now feel vindicated for doing [this](https://github.com/openssl/openssl/blob/2837b19fcba4bf4ff2ecdc8435c650bf18c27552/test/recipes/95-test_external_oqsprovider_data/oqsprovider.sh#L57)...

baentsch:
> As much as it might cause some downstream pain, 

Thinking about that part of the statement again, what pain would the _complete removal_ of Rainbow cause? Not doing a release simply means running all templating scripts and CI on all project "main" branches. Should be smooth. Doing a release (for all components) just for that would be tedious. A re-set of strengths (and retaining) some Rainbow variants would have been tedious. Outright pruning should be straightforward. We just need to make sure no IDs assigned to Rainbow variants are (re)used, e.g., in OpenSSL. If there are no objections, will do a PR right away...

baentsch:
Confirmed (by doing it): Complete removal of Rainbow works well through the whole stack.

[...]

PR #1263, opened 2022-07-27
Remove Rainbow level 1

### LINKED ISSUE #1260: Remove Rainbow level 1

dstebila:
Fixes #1260

* [YES] Does this PR change the the list of algorithms available -- either adding, removing, or renaming?  (If so, PRs in OQS-OpenSSL, OQS-BoringSSL, and OQS-OpenSSH will also be required by the time this is merged.)


baentsch (approved):

dstebila:
@baentsch Should I have also changed RainbowIII and RainbowV to have security level 1 and 3?  Within liboqs that should have no effect other than to change some documentation and an information member variable. But will that cause downstream consumers to change the things they pair Rainbow with in hybrid modes? Since the Rainbow team downgraded their claimed security of Rainbow, it would be preferable to have that reflected here, but I don't want to have to go through the process of reassigning everything downstream only for us to remove it a few weeks later. I think we discussed this earlier but I can't remember the decision.

baentsch:
> Should I have also changed RainbowIII and RainbowV to have security level 1 and 3? [...] I think we discussed this earlier but I can't remember the decision.

That was part of the discussion [here](https://github.com/open-quantum-safe/liboqs/issues/1192) and there you stated 

> We'll remove Rainbow level 1 for 0.7.2 via https://github.com/open-quantum-safe/liboqs/pull/1263 and then remove the remaining Rainbow after that.

As you closed the issue there that contained all those steps with all the knock-on effects discussed there (new OIDs, etcpp), my take is "No": Lots of work for an algorithm that's broken and was not part of the NIST competition at those reduced levels. It was dropped (at the original levels). It never was nor ever will be used by anyone at those reduced levels. Thus, changing the Rainbow sec levels doesn't make any sense. Dropping the whole family would have been the most logical solution. Dropping RainbowI is an understandable "bare minimum" and that's what I understood this PR is about (assuming you'd reached agreement on this "bare minimum solution" in last week's call that I couldn't attend).

baentsch:
@dstebila In the future, please avoid merging such a change to `liboqs` that we know breaks oqs-provider without also merging the matching [oqs-provider PR](https://github.com/open-quantum-safe/oqs-provider/pull/61) at the same time: See the flurry of (in hindsight unnecessary) activity triggered by this: OpenSSL depends on the two components to work OK together; https://github.com/openssl/openssl/pull/18899 (@bhess, @christianpaquin @xvzcf also FYI). Thanks in advance!

dstebila:
> @dstebila In the future, please avoid merging such a change to `liboqs` that we know breaks oqs-provider without also merging the matching [oqs-provider PR](https://github.com/open-quantum-safe/oqs-provider/pull/61) at the same time: See the flurry of (in hindsight unnecessary) activity triggered by this: OpenSSL depends on the two components to work OK together; [openssl/openssl#18899](https://github.com/openssl/openssl/pull/18899) (@bhess, @christianpaquin @xvzcf also FYI). Thanks in advance!

Sorry, will try to remember for next time.

PR #1273, opened 2022-08-03
adding library version retrieval function

baentsch:
Provide a mechanism to retrieve at runtime, e.g., in a shared library, the current library version string. For use e.g. in language wrappers to display the true library version wrapped.

A common API is added, though. Downstream project's CI should be observed.

### DIFF
--- a/src/common/common.c
+++ b/src/common/common.c
@@ -191,6 +191,10 @@ OQS_API void OQS_init(void) {
 	return;
 }
 
+OQS_API const char *OQS_version(void) {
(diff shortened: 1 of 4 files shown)


baentsch:
If CI passes and @vsoftco and @dstebila concur, I'd suggest merging this before doing the 0.7.2 RC.

vsoftco (approved):
All good from my side, thanks Michael

dstebila (approved):

PR #1275, opened 2022-08-03
Addition of stateful hash-based signatures

kj3moraes:
## Addition of XMSS to the codebase

I have added code for XMSS into the `src/sig_stateful` directory. The algorithm variants supported are :- 
* all the XMSS variants (SHA256, SHA512, SHAKE256, SHAKE512)
* all the RFC recommended XMSS^MT variants. 

The continuous integration tests pass and I have made the respective test files for testing the SIG_STFL API.


baentsch:
I hope this question from many thousand miles away is permitted: Why is this PR in "Ready for Review" state? IMO it's passing way too few tests.... Some failures appear very simple to correct (`include/oqs/sig_stfl.h:224:32: error: no newline at end of file [-Werror,-Wnewline-eof]`)...

kj3moraes:
I've added the newlines at EOF for the files that failed in the CircleCI checks. However, many of the tests are failing because they timed out. XMSS-keygen is incredibly slow, and it takes about 40 minutes total for all the variants to be tested.

baentsch:
> I've added the newlines at EOF

Thanks.

> XMSS-keygen is incredibly slow, and it takes about 40 minutes total for all the variants to be tested.

All right -- is it then possible to create a "lightweight" (CI-only) test, maybe not running all variants but "key" ones (whatever that means)? We similarly [skip some ultra-slow tests in "standard CI"](https://github.com/open-quantum-safe/liboqs/blob/3cf9849cc43545ad7fdeb3092a90f94ff1881680/tests/helpers.py#L115) but can run them locally (or at release).

kj3moraes:
> All right -- is it then possible to create a "lightweight" (CI-only) test, maybe not running all variants but "key" ones (whatever that means)? We similarly [skip some ultra-slow tests in "standard CI"](https://github.com/open-quantum-safe/liboqs/blob/3cf9849cc43545ad7fdeb3092a90f94ff1881680/tests/helpers.py#L115) but can run them locally (or at release).

Yes, maybe we could do select variants and exclude the slow ones. However, I don't know how to exclude certain variants from being tested. The ones I would exclude are : 
1. all the SHAKE variants
2. XMSS-SHA2_16_512 and XMSS-SHA2_20_512
3. XMSSMT-SHA2_60/*_256 variants

baentsch:
> I don't know how to exclude certain variants from being tested.

Good point: My pointer only would have permitted exclusion of whole tests. Thankfully, @dstebila just added the general mechanism to exclude single algorithms in #1282: Maybe you want to rebase to that and set "SKIP_ALGS" suitably as per your suggestion.

baentsch (commented):
I didn't really check the contents of your PR yet: It would be good to see it pass CI first. Adding the slowest new algorithms to `.circleci/config.yml`, e.g., [here](https://github.com/open-quantum-safe/liboqs/blob/d87e800740babba159e5aa3cca22a5221759117a/.circleci/config.yml#L361-L366) by setting "SKIP_ALGS" env var suitably (to exclude those slow algorithms) should do this. See example for that [here](https://github.com/open-quantum-safe/liboqs/blob/d3c10161e3b13bf9352ad4e53dddfde955b04c71/.github/workflows/linux.yml#L78). You can test this locally by running `circleci` [CLI](https://circleci.com/docs/local-cli) with the single jobs until they pass (get below 1h overall runtime).

src/sig_stateful/sig_stfl.h
```diff
@@ -0,0 +1,224 @@
...
+ * @brief SECRET KEY object for stateful signature schemes *
+ */
+typedef struct OQS_SECRET_KEY OQS_SECRET_KEY;
+
+typedef struct OQS_SECRET_KEY {
+
+	/** The (maximum) length, in bytes, of secret keys for this signature scheme. */
+	unsigned long long length_secret_key;
+
+	/** The physical secret key stored in memory as an array of bytes*/
```
dstebila, line 111:
```suggestion
	/** The secret key stored in memory as an array of bytes. */
```

src/sig_stateful/sig_stfl.h
```diff
@@ -0,0 +1,224 @@
...
+	uint8_t *secret_key;
+
+	unsigned long long (*sigs_total)(const OQS_SECRET_KEY *secret_key);
+
+	unsigned long long (*sigs_left)(const OQS_SECRET_KEY *secret_key);
+
+	OQS_STATUS (*lock_key)(OQS_SECRET_KEY *sk);
+	OQS_STATUS (*oqs_save_updated_sk_key)(const OQS_SECRET_KEY *sk);
+	OQS_STATUS (*release_key)(OQS_SECRET_KEY *sk);
```
dstebila, line 120: Missing documentation of these methods.

src/sig_stateful/sig_stfl.h
```diff
@@ -0,0 +1,224 @@
...
+
+} OQS_SECRET_KEY;
+
+OQS_API OQS_SECRET_KEY *OQS_SECRET_KEY_new(const char *method_name);
+
+OQS_API void OQS_SECRET_KEY_free(OQS_SECRET_KEY *sk);
```
dstebila, line 126: Missing documentation

src/sig_stateful/sig_stfl.h
```diff
@@ -0,0 +1,224 @@
...
+	/**
+	 * Printable string representing the version of the cryptographic algorithm.
+	 *
+	 * Implementations with the same method_name and same alg_version will be interoperable.
+	 * See README.md for information about algorithm compatibility.
+	 */
+	const char *alg_version;
+
+	/** The NIST security level (1, 2, 3, 4, 5) claimed in this algorithm's original NIST submission. */
+	uint8_t claimed_nist_level;
```
dstebila, line 145: How are we setting these for XMSS?  Since it's not a NIST submission perhaps this field doesn't make sense.

src/sig_stateful/sig_stfl.h
```diff
@@ -0,0 +1,224 @@
...
+OQS_API OQS_STATUS OQS_SIG_STFL_keypair(const OQS_SIG_STFL *sig, uint8_t *pk, OQS_SECRET_KEY *sk);
+
+OQS_API OQS_STATUS OQS_SIG_STFL_sign(const OQS_SIG_STFL *sig, uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, OQS_SECRET_KEY *secret_key);
+
+OQS_API OQS_STATUS OQS_SIG_STFL_verify(const OQS_SIG_STFL *sig, const uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *public_key);
+
+OQS_API OQS_SECRET_KEY *OQS_SIG_STFL_derive_subkey(const OQS_SIG_STFL *sig, OQS_SECRET_KEY *master_key, const unsigned long long number_of_signatures);
+
+OQS_API void OQS_SIG_STFL_free(OQS_SIG_STFL *sig);
+
```
dstebila, line 212: Missing documentation

src/sig_stateful/xmss/external/xmss.c
```diff
@@ -0,0 +1,205 @@
...
+    int ret = xmss_core_keypair(&params, pk + XMSS_OID_LEN, sk->secret_key + XMSS_OID_LEN);
+    if (ret != 0) return ret;
+
+    unsigned long long max = pow(2, params.full_height) - (unsigned long) 1;
+    ull_to_bytes(sk->secret_key + sk->length_secret_key - params.bytes_for_max, params.bytes_for_max, max);
+    
+    return ret;
+}
+
+int xmss_derive_subkey(OQS_SECRET_KEY *master, OQS_SECRET_KEY *subkey, unsigned long long number_of_sigs) {
```
dstebila, line 39: Make sure to lock/save/release the master key in the appropriate places.

src/sig_stateful/xmss/sig_stfl_xmss_sha256.c
```diff
@@ -0,0 +1,185 @@
...
+OQS_SIG_STFL *OQS_SIG_STFL_alg_xmss_sha256_h10_new(void) {
+
+	OQS_SIG_STFL *sig = (OQS_SIG_STFL *)malloc(sizeof(OQS_SIG_STFL));
+	if (sig == NULL) {
+		return NULL;
+	}
+	memset(sig, 0, sizeof(OQS_SIG_STFL));
+
+	sig->method_name = "XMSS-SHA2_10_256";
+	sig->alg_version = "...";
```
dstebila, line 35: Need to set the version field

src/sig_stateful/xmss/sig_stfl_xmss_sha256.c
```diff
@@ -0,0 +1,185 @@
...
+
+	// Check how true this is
+	sig->claimed_nist_level = 2;
```
dstebila, line 38: Need to decide whether we're keeping the claimed_nist_level or not.

src/sig_stateful/xmss/sig_stfl_xmss_xmssmt_functions.c
```diff
@@ -0,0 +1,140 @@
...
+	uint32_t oid = 0;
+	for (i = 0; i < XMSS_OID_LEN; i++) {
+		oid |= secret_key->secret_key[XMSS_OID_LEN - i - 1] << (i * 8);
+	}
+	xmss_parse_oid(&params, oid);
+
+	unsigned long long max = OQS_SIG_STFL_alg_xmss_xmssmt_sigs_total(secret_key);
+	unsigned long long idx = 0;
+	for (i = 0; i < params.index_bytes; i++) {
+		idx |= ((unsigned long long)secret_key->secret_key[i]) << 8 * (params.index_bytes - 1 - i);
```
dstebila, line 70:
```suggestion
		idx |= ((unsigned long long)secret_key->secret_key[i]) << (8 * (params.index_bytes - 1 - i));
```

src/sig_stateful/xmss/sig_stfl_xmss_xmssmt_functions.c
```diff
@@ -0,0 +1,140 @@
...
+	xmss_parse_oid(&params, oid);
+
+	unsigned long long max = 0;
+	for (unsigned int j = params.bytes_for_max; j > 0; j--) {
+		max |= ((unsigned long long)secret_key->secret_key[params.sk_bytes - XMSS_OID_LEN - j] << 8 * (j - 1));
+	}
+	return max;
+}
+
+void perform_key_allocation(OQS_SECRET_KEY *sk) {
```
dstebila, line 95: Missing namespacing or should be `static`

src/sig_stateful/xmss/sig_stfl_xmss_xmssmt_functions.c
```diff
@@ -0,0 +1,140 @@
...
+	sk->sigs_total = OQS_SIG_STFL_alg_xmss_xmssmt_sigs_total;
+
+	// Initialize the key with length_secret_key amount of bytes.
+	sk->secret_key = (uint8_t *)malloc(sk->length_secret_key * sizeof(uint8_t));
+	memset(sk->secret_key, 0, sk->length_secret_key);
+}
+
+OQS_SECRET_KEY *OQS_SIG_STFL_alg_xmss_derive_subkey(OQS_SECRET_KEY *master_key, const unsigned long long number_of_sigs) {
+
+	OQS_SECRET_KEY *subkey = (OQS_SECRET_KEY *)malloc(sizeof(OQS_SECRET_KEY));
```
dstebila, line 108: Need to check for malloc failure

src/sig_stateful/xmss/external/xmss.h
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Verifies a given message signature pair using a given public key.
+ *
+ * Note: m and mlen are pure outputs which carry the message in case
+ * verification succeeds. The (input) message is assumed to be contained in sm
+ * which has the form [signature || message].
+ */
+int xmss_sign_open(const uint8_t *m, unsigned long long *mlen,
+                   const uint8_t *sm, unsigned long long smlen,
+                   const uint8_t *pk);
```
dstebila, line 63: Is this sign_open or verify?

src/sig_stateful/xmss/external/xmss.h
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Verifies a given message signature pair using a given public key.
+ *
+ * Note: m and mlen are pure outputs which carry the message in case
+ * verification succeeds. The (input) message is assumed to be contained in sm
+ * which has the form [signature || message].
+ */
+int xmssmt_sign_open(const uint8_t *m, unsigned long long *mlen,
+                     const uint8_t *sm, unsigned long long smlen,
+                     const uint8_t *pk);
```
dstebila, line 91: Is this sign_open or verify?

tests/CMakeLists.txt
```diff
@@ -97,6 +97,23 @@ endif()
...
+else ()
+    target_link_libraries(test_sig_stfl PRIVATE ${API_TEST_DEPS})
+endif()
+
+add_executable(test_sig_stfl_mem test_sig_stfl_mem.c)
+target_link_libraries(test_sig_stfl_mem PRIVATE ${API_TEST_DEPS})
+
+# add_executable(kat_sig_stfl kat_sig_stfl.c)
+# target_link_libraries(kat_sig_stfl PRIVATE ${API_TEST_DEPS})
+
```
dstebila, line 116:
```suggestion
```

tests/example_sig_stfl.c
```diff
@@ -0,0 +1,165 @@
...
+                  OQS_SIG_STFL *sig);
+
+/* This function gives an example of the signing operations
+ * using only compile-time macros and allocating variables
+ * statically on the stack, calling a specific algorithm's functions
+ * directly.
+ *
+ * The macros OQS_SIG_dilithium_2_length_* and the functions OQS_SIG_dilithium_2_*
+ * are only defined if the algorithm dilithium_2 was enabled at compile-time
+ * which must be checked using the OQS_ENABLE_SIG_dilithium_2 macro.
```
dstebila, line 32: Comments need to be updated

tests/example_sig_stfl.c
```diff
@@ -0,0 +1,165 @@
...
+ * was compiled.
+ */
+static OQS_STATUS example_stack(void) {
+
+#ifdef OQS_ENABLE_SIG_STFL_XMSS_SHA256_H16
+
+	OQS_STATUS rc;
+
+	uint8_t public_key[OQS_SIG_STFL_alg_xmss_sha256_h16_length_pk];
+	OQS_SECRET_KEY *secret_key = OQS_SECRET_KEY_new(OQS_SIG_STFL_alg_xmss_sha256_h16);
```
dstebila, line 45: Check for NULL (malloc failure)

src/sig_stateful/lms/sig_stfl_lms.h
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+
+#ifndef OQS_SIG_STFL_LMS_H
+#define OQS_SIG_STFL_LMS_H
+
+#include "../sig_stfl.h"
+
+
+#endif /* OQS_SIG_STFL_LMS_H */
```
dstebila, line 9: This file shouldn't be here.

tests/example_sig_stfl.c
```diff
@@ -0,0 +1,165 @@
...
+	OQS_SIG_STFL *sig = NULL;
+	uint8_t *public_key = NULL;
+	OQS_SECRET_KEY *secret_key = NULL;
+	uint8_t *message = NULL;
+	uint8_t *signature = NULL;
+	size_t message_len = MESSAGE_LEN;
+	size_t signature_len;
+	OQS_STATUS rc;
+
+	sig = OQS_SIG_STFL_new(OQS_SIG_alg_dilithium_2);
```
dstebila, line 104: Wrong algorithm identifier. This example code should have failed -- did you test it?

tests/example_sig_stfl.c
```diff
@@ -0,0 +1,165 @@
...
+		return EXIT_SUCCESS;
+	} else {
+		return EXIT_FAILURE;
+	}
+}
+
+void cleanup_heap(uint8_t *public_key, OQS_SECRET_KEY *secret_key,
+                  uint8_t *message, uint8_t *signature,
+                  OQS_SIG_STFL *sig) {
+	if (sig != NULL) {
```
dstebila, line 158: Why are you testing if `sig` is NULL before the `secret_key` free?

tests/logs/problem2_Mon 25 Jul 2022 05:01:24 PM EDT.log
```diff
@@ -0,0 +1,1385 @@
+============================= test session starts ==============================
```
dstebila, line 1: All of these log files shouldn't be included

tests/test_code_conventions.py
```diff
@@ -26,6 +26,17 @@ def test_datasheet_sig(sig_name):
...
+
+# Ensure every stateful signature algorithm in the code
+# is mentioned in the documentation.
+# @helpers.filtered_test
+# @pytest.mark.skipif(sys.platform.startswith("win"), reason="Not needed on Windows")
+# @pytest.mark.parametrize('sig_stfl_name', helpers.available_sig_stfls_by_name())
+# def test_datasheet_sig_stfl(sig_stfl_name):
+#     helpers.run_subprocess(
+#         ['grep', '-r', sig_stfl_name, 'docs/algorithms']
+#     )
```
dstebila, line 38: Need to be uncommented

tests/test_kat.py
```diff
@@ -35,6 +35,20 @@ def test_sig(sig_name):
...
+#     if not(helpers.is_sig_stfl_enabled_by_name(sig_stfl_name)): pytest.skip('Not enabled')
+#     output = helpers.run_subprocess(
+#         [helpers.path_to_executable('kat_sig_stfl'), sig_stfl_name],
+#     )
+#     output = output.replace("\r\n", "\n")
+#     h256 = sha256()
+#     h256.update(output.encode())
+
+#     assert(kats[sig_stfl_name] == h256.hexdigest())
+
```
dstebila, line 51: We don't have KATs for XMSS so you can delete this code.

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,271 @@
...
+
+#include "system_info.c"
+
+typedef struct magic_s {
+	uint8_t val[31];
+} magic_t;
+
+
+OQS_STATUS lock_sk_key(OQS_SECRET_KEY *sk) {
+	printf("%02x", sk->secret_key[0]);
```
dstebila: Why are there print statements here and in the next two functions?

src/sig_stateful/sig_stfl.h
```diff
@@ -0,0 +1,224 @@
...
+	/** The physical secret key stored in memory as an array of bytes*/
+	uint8_t *secret_key;
+
+	unsigned long long (*sigs_total)(const OQS_SECRET_KEY *secret_key);
+
+	unsigned long long (*sigs_left)(const OQS_SECRET_KEY *secret_key);
+
+	OQS_STATUS (*lock_key)(OQS_SECRET_KEY *sk);
+	OQS_STATUS (*oqs_save_updated_sk_key)(const OQS_SECRET_KEY *sk);
```
dstebila: I don't think it's necessary for this method to start with `oqs_`. None of the others do.

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,271 @@
...
+	secret_key->release_key = release_sk_key;
+	secret_key->oqs_save_updated_sk_key = do_nothing_save;
+	if (secret_key == NULL) {
+		fprintf(stderr, "ERROR: OQS_SECRET_KEY_new failed\n");
+		goto err;
+	}
+
+	//Set the magic numbers before
+	memcpy(public_key, magic.val, sizeof(magic_t));
+	// memcpy(secret_key, magic.val, sizeof(magic_t));
```
dstebila: I think you can remove this commented-out line (and similar ones below) but add a comment why you're not using magic numbers for the secret key.

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,271 @@
...
+
+	printf("verification passes as expected\n");
+	ret = OQS_SUCCESS;
+	goto cleanup;
+
+err:
+	ret = OQS_ERROR;
+
+cleanup:
+	if (sig != NULL) {
```
dstebila, line 268: Why are you testing `sig` here when it's guarding a free on `secret_key`?

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,271 @@
...
+	struct thread_data *td = arg;
+	td->rc = sig_test_correctness(td->alg_name);
+	return NULL;
+}
+#endif
+
+int main(int argc, char **argv) {
+
+	if (argc != 2) {
+		fprintf(stderr, "Usage: test_sig algname\n");
```
dstebila: Should be test_sig_stfl

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,271 @@
...
+	// don't run Rainbow III and V in threads because of large stack usage
+	char no_thread_sig_patterns[][MAX_LEN_SIG_NAME_]  = {"Rainbow-III", "Rainbow-V"};
+	int test_in_thread = 1;
+	for (size_t i = 0 ; i < sizeof(no_thread_sig_patterns) / MAX_LEN_SIG_NAME_; ++i) {
+		if (strstr(alg_name, no_thread_sig_patterns[i]) != NULL) {
+			test_in_thread = 0;
+			break;
+		}
+	}
+	if (test_in_thread) {
```
dstebila, line 375: No need for any of these checks because Rainbow is not a stateful signature scheme.

tests/test_sig_stfl_key_save.c
```diff
@@ -0,0 +1,304 @@
...
+char filename[MAX_FILE_NAME_LEN] = "XMSS-SHA2_10_256";
+
+void prepend(char *s, const char *t) {
+	size_t len = strlen(t);
+	memmove(s + len, s, strlen(s) + 1);
+	memcpy(s, t, len);
+}
+
+OQS_STATUS lock_sk_key(OQS_SECRET_KEY *sk) {
+	printf("lk=%02x\n", sk->secret_key[0]);
```
dstebila, line 42: Why print statements?

tests/test_sig_stfl_key_save.c
```diff
@@ -0,0 +1,304 @@
...
+	printf("rk=%02x\n", sk->secret_key[0]);
+	return 0;
+}
+
+OQS_STATUS do_nothing_save(const OQS_SECRET_KEY *sk) {
+	printf("save_k=%02x\n", sk->secret_key[0]);
+	return 0;
+}
+
+OQS_STATUS sk_file_write(const OQS_SECRET_KEY *sk) {
```
dstebila, line 56: What's the difference between test_sig_stfl_key_save.c and test_sig_stfl.c?  Is it just whether the key is saved to a file or not?  Is there a reason we can't just have one test? Is this file actually called by the Python test harness?

dstebila: If we're keeping this file, please go through and apply the relevant comments from test_sig_stfl.c to this one, in case I forgot to mention things here.

tests/test_sig_stfl_key_save.c
```diff
@@ -0,0 +1,304 @@
...
+
+	printf("\nWriting to file %s\n", filename);
+	FILE *printer = fopen(filename, "w+");
+	if (printer == NULL) {
+		perror("ERROR! There is no such file. Terminating...");
+		return -1;
+	}
+
+	// Write the entire secret key byte array to the specified file.
+	for (unsigned long i = 0; i < sk->length_secret_key; i++) {
```
dstebila, line 69:
```suggestion
	for (size_t i = 0; i < sk->length_secret_key; i++) {
```

tests/test_sig_stfl_mem.c
```diff
@@ -0,0 +1,225 @@
...
+#include "tmp_store.c"
+
+typedef enum sig_ops {
+	SIG_KEYGEN      = 0,
+	SIG_STFL_SIGN   = 1,
+	SIG_STFL_VERIFY = 2
+} SIG_OPS;
+
+OQS_STATUS lock_sk_key(OQS_SECRET_KEY *sk) {
+	printf("%02x", sk->secret_key[0]);
```
dstebila, line 23: Same comment as before

tests/test_sig_stfl_mem.c
```diff
@@ -0,0 +1,225 @@
+// SPDX-License-Identifier: MIT
+
```
dstebila, line 2: I realize now this file (and the other test files) should have some comments at the top of the file explaining what this test file is for.

tests/test_sig_stfl_mem.c
```diff
@@ -0,0 +1,225 @@
...
+	if (public_key) {
+		OQS_MEM_insecure_free(public_key);
+	}
+	if (message) {
+		OQS_MEM_insecure_free(message);
+	}
+	if (signature) {
+		OQS_MEM_insecure_free(signature);
+	}
+	if (sig != NULL) {
```
dstebila, line 183: Why is free of secret_key guarded by NULL test on sig?

tests/test_sig_stfl_multithreaded.py
```diff
@@ -0,0 +1,34 @@
+# SPDX-License-Identifier: MIT
```
dstebila, line 1: What is this file for?

src/sig_stateful/sig_stfl.h
kj3moraes: I have added a SAFETY functions comment for all 3 functions to sum up their purposes.

src/sig_stateful/sig_stfl.h
kj3moraes: I have removed this field from the SIG_STFL object.

src/sig_stateful/xmss/sig_stfl_xmss_sha256.c
kj3moraes: What is the value that I need to set for `alg_version` for say "XMSS-SHA2_16_256"

src/sig_stateful/xmss/sig_stfl_xmss_sha256.c
kj3moraes: I have removed it from implementation

tests/example_sig_stfl.c
kj3moraes: No, the example code is faulty. It is not meant to be tested at this moment

src/sig_stateful/xmss/sig_stfl_xmss_sha256.c
dstebila: Is the XMSS source code versioned?  If so, use that. Otherwise you could use the RFC as the version number.

tests/test_sig_stfl_key_save.c
kj3moraes: This file was made a while back. It is very outdated and serves no purpose now that we have added the optional arguments for `test_sig_stfl`

tests/test_sig_stfl_multithreaded.py
kj3moraes: It is redundant now

src/sig_stateful/xmss/external/xmss_core.h
```diff
@@ -0,0 +1,93 @@
...
+                   const uint8_t *m, unsigned long long mlen);
+
+/**
+ * Verifies a given message signature pair under a given public key.
+ * Note that this assumes a pk without an OID, i.e. [root || PUB_SEED]
+ */
+int xmss_core_sign_open(const xmss_params *params,
+                        uint8_t *m, unsigned long long *mlen,
+                        const uint8_t *sm, unsigned long long smlen,
+                        const uint8_t *pk);
```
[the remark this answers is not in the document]
kj3moraes: No this sign open does not require the message as part of the signature. It uses the message that is passed in via the command line argument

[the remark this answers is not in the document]
dstebila: Then we should be calling it verify not sign_open.

src/sig_stateful/xmss/external/xmss.h
kj3moraes: This is the verify function. There is no message recovery taking place. Both XMSS and XMSS^MT have their verification algorithms called sign_open.

src/sig_stateful/xmss/external/xmss.h
kj3moraes: This is the verification algorithm

tests/test_sig_stfl_key_save.c
kj3moraes: The entire file has been deleted because it is now redundant.

tests/test_sig_stfl_key_save.c
kj3moraes: I have replaced it with some variable that access the secret key byte array,

tests/logs/problem2_Mon 25 Jul 2022 05:01:24 PM EDT.log
kj3moraes: I have removed then

kj3moraes:
For the documentation of several function, is there a template I need to follow or can I do something like

```
/**
 * @brief 
 * @params 
 * ... 
 */
 ```

**Comment by kj3moraes** [comment]
Would I need to change the names of the verification functions from `sign_open` to `verify` ? The sign_open functions are only used internally by the code in `external/`. I can change the `xmss_namespace.h` file. The reason I'm relunctant is because it would involve a lot of refactoring of the code.

**Comment by dstebila** [comment]
> Would I need to change the names of the verification functions from `sign_open` to `verify` ? The sign_open functions are only used internally by the code in `external/`. I can change the `xmss_namespace.h` file. The reason I'm relunctant is because it would involve a lot of refactoring of the code.

Okay, if it involves a lot of refactoring then okay to leave it as is. Maybe add a comment noting that it's called sign_open but we've changed the functionality to have detached signatures like in verify.

File: src/sig_stateful/sig_stfl.h [hunk 37]
```diff
@@ -0,0 +1,309 @@
...
+
+	/**
+	 * SAFETY FUNCTIONS
+	 * -> lock_key : "Locks" the secret key, preventing any attributes from being changed
+	 * -> save_secret_key : Saves the byte array of the secret key to a file
+	 * -> release_key : "Unlocks" the secret key, allowing attributes to be changed
+	 */
+	OQS_STATUS (*lock_key)(OQS_SECRET_KEY *sk);
+	OQS_STATUS (*save_secret_key)(const OQS_SECRET_KEY *sk);
+	OQS_STATUS (*release_key)(OQS_SECRET_KEY *sk);
```
dstebila, line 126: This will probably fail the Doxygen requirements -- there would need to be documentation for each method.

src/sig_stateful/xmss/sig_stfl_xmss_sha256.c
```diff
@@ -0,0 +1,199 @@
...
+	OQS_SIG_STFL *sig = (OQS_SIG_STFL *)malloc(sizeof(OQS_SIG_STFL));
+	if (sig == NULL) {
+		return NULL;
+	}
+	memset(sig, 0, sizeof(OQS_SIG_STFL));
+
+	sig->method_name = "XMSS-SHA2_10_256";
+	sig->alg_version = "https://datatracker.ietf.org/doc/html/rfc8391";
+
+	// Check how true this is
```
dstebila, line 37: Can remove this comment and others like it.

tests/CMakeLists.txt
```diff
@@ -5,6 +5,8 @@ option(OQS_ENABLE_TEST_CONSTANT_TIME "Build test suite with support for Valgrind
 if(CMAKE_C_COMPILER_ID STREQUAL "GNU" OR
    CMAKE_C_COMPILER_ID MATCHES "Clang")
     add_compile_options(-Wbad-function-cast)
+    add_compile_options(-g)
+    add_compile_options(-O0)
```
dstebila, line 9: Why did you add these flags?

tests/test_cmdline.py
```diff
@@ -31,6 +33,45 @@ def test_sig(sig_name):
...
+@helpers.filtered_test
+@pytest.mark.parametrize('sig_stfl_name', helpers.available_sig_stfls_by_name())
+def test_sig_stfl_save_key(sig_stfl_name):
+    directory = "../tests/saved_keys/"
+    
+    # Extract parts of the scheme name to form the filename
+    height_of_tree = sig_stfl_name[sig_stfl_name.find('_')+1:sig_stfl_name.rfind('_')].replace('/', '')
+    hash_alg = sig_stfl_name[sig_stfl_name.find('-')+1:sig_stfl_name.find('_')].lower()
+    sig_scheme = sig_stfl_name[0:sig_stfl_name.find('-')].lower()
+    hash_alg_bits = sig_stfl_name[sig_stfl_name.rfind('_')+1:]
```
dstebila, line 66: This is dependent on the current naming conventions of XMSS but wouldn't necessarily apply to LMS. Is there a reason you need to extract all these fields? Can't you just use the scheme name directly to construct the filename, without parsing?

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,416 @@
...
+		printf("Signature algorithm %s not enabled!\n", alg_name);
+		return EXIT_FAILURE;
+	}
+
+	// Extract the mode of operation from the command line arguments
+	char mode_of_operation = 'x';
+	char *filestem = NULL;
+	if (argc > 2) {
+		char argument[8];
+		strncpy(argument, argv[2], 6);
```
dstebila, line 372: Why are you using only 6 characters here?

dstebila (commented):
Getting pretty close on the source code, only a few remaining comments as far as I can tell. Still need to remove `docs/algorithms/sig_stfl/lms*`.  And then work towards getting the CI tests to pass.

baentsch:
> Would I need to change the names of the verification functions from `sign_open` to `verify` ? The sign_open functions are only used internally by the code in `external/`. I can change the `xmss_namespace.h` file. The reason I'm relunctant is because it would involve a lot of refactoring of the code.

> Okay, if it involves a lot of refactoring then okay to leave it as is. Maybe add a comment noting that it's called sign_open but we've changed the functionality to have detached signatures like in verify.

I admit I have not looked in detail at stateful hash-based signatures, so please take my comment with a grain of salt, but it strikes me as unsatisfactory to retain the name "sign" for what is (if it is) a "verify" operation: Even with a comment (if read) this is confusing. Why would re-naming call for re-factoring? Wouldn't running a simple recursive `sed` (along the lines of `-i 'g/sign_open/sign_verify/g'`) operation over all new code do the trick?

src/sig_stateful/sig_stfl.h
kj3moraes: Ok I will fix it

tests/CMakeLists.txt
kj3moraes: It was debugging purposes. Once I get circleCI tests passing, I will remove them. The 0 optimization flag was to prevent the compiler from optimizing many things out. Allowed easier debugging

tests/test_cmdline.py
kj3moraes: Yes, I should have done that when I started out saving the files. I wil mark this for removal.

kj3moraes: Once all the circleCI tests pass, I will rename the saved keys and then remove this parsing code

tests/test_sig_stfl.c
kj3moraes: `--save_key-to=` and `--read_key_from` are different lengths so I just compare the `--save` and `--read` parts which are 6 characters. The user is expected to type in the whole thing however, because the filename access will fail otherwise.

kj3moraes:
> Why would re-naming call for re-factoring? Wouldn't running a simple recursive `sed` (along the lines of `-i 'g/sign_open/sign_verify/g'`) operation over all new code do the trick?

I will try doing this. Thank you @baentsch. Should the name be changed from `sign_open` to `verify` or `sign_open` to `sign_verify` ?

dstebila:
> I will try doing this. Thank you @baentsch. Should the name be changed from `sign_open` to `verify` or `sign_open` to `sign_verify` ?

`verify`

kj3moraes:
I'm having trouble configuring the CircleCI tests to only use the `read_key_from=` functionality. How can I configure the config.yml in .circleCI to omit some tests ?

 `test_sig_stfl` allows for 2 parameters : 
* \<alg-name\>
* **One of** `--save_key_to=<path to filename(without file extension)>` **or** `--read_key_from=<path to filename(without file extension)>`

I have manually generated all the keys for the variants and they are stored in `tests/saved_keys`. **I only want the circleCI tests to read keys**. 

The `test_cmdline.py` file has the [respective functions to allow for this functionality.](https://github.com/open-quantum-safe/liboqs/blob/2f1e1dc7d6a0159333665d19915499c130013812/tests/test_cmdline.py#L36)

I would also like to emphasize that [all the tests pass when I use the read-key functionality.](https://github.com/open-quantum-safe/liboqs/blob/590971dada17c5258ef18f03f38c014a30e7a7f3/tests/tmp/tests.out)`

baentsch:
How and where can one read up on these parameters (actually, best, on the whole implementation concept)? It's not quite usual to find keys stored in a repository... Is that because keygen is too slow (for CI)?

> How can I configure the config.yml in .circleCI to omit some tests

That depends on how they are triggered -- via .py files (using helpers.py)? If so, set the environment variable(s) SKIP_ALGS and/or SKIP_TESTS suitably.

kj3moraes:
> How and where can one read up on these parameters (actually, best, on the whole implementation concept)? It's not quite usual to find keys stored in a repository... Is that because keygen is too slow (for CI)?

Yes, the keys are stored because keygen is too slow and the CI tests would time out. So the keys are stored and we are only running signing and verification. You can read up on the XMSS paramters on its [RFC page](https://datatracker.ietf.org/doc/html/rfc8391)

> That depends on how they are triggered -- via .py files (using helpers.py)? If so, set the environment variable(s) SKIP_ALGS and/or SKIP_TESTS suitably.

Yes, the tests are triggered via .py files. The `test_cmdline.py` fie is supposed to call `test_sig_stfl_new_key`, `test_sig_stfl_read_key` (or) `test_sig_stfl_save_key`. Currently new_key and save_key are commented out because I want the CI tests to pass but I need to uncomment them somewhere down the line.

kj3moraes:
Furthermore, I want to point out that every single one of the [XMSS variants fails the ubuntu-bionic-i386 tests](https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/2224/workflows/14e39fa9-6a00-40f7-b7d6-4134d638213b/jobs/18221/tests) for some reason. The other circleCI tests and are being fed the same parameters. Could someone please look into this ? I am confused as to why it is taking place.

tests/test_sig_stfl.c
Martyrshot: This may be why we are getting weird behaviour on 32 bit ubuntu. Right now you only copy 6 characters, but c strings need to include the null terminator.

From the strncpy man page:
```
strncpy() functions copy at most len characters from src into dst.  If src is less than len characters long, the remainder of dst is filled with ‘\0’ characters.  Otherwise, dst is not terminated.
```

So right now I think `argument` isn't null terminated so strcmp is probably going further along than you think it is thus causing all comparisons to be false.

Martyrshot:
> Furthermore, I want to point out that every single one of the [XMSS variants fails the ubuntu-bionic-i386](https://github.com/open-quantum-safe/liboqs/runs/7842758019?check_suite_focus=true) and [linux-arm emulated tests](https://github.com/open-quantum-safe/liboqs/runs/7842758019?check_suite_focus=true) for some reason. The other circleCI tests and are being fed the same parameters. Could someone please look into this ? I am confused as to why it is taking place.

I commented on a pre-existing thread with what I think the issue is.

baentsch:
> Could someone please look into this ? I am confused as to why it is taking place.

Honestly, I don't understand why this code works on any platform: https://github.com/open-quantum-safe/liboqs/blob/a603f1ab562feaacd11ce3a1fd0e47a95677e5ed/tests/test_sig_stfl.c#L365-L369

This copies 6 bytes but you need 7 to successfully do `strcmp`: Can you really safely assume the NULL string terminator has been properly initialized into the "argument" string? Wouldn't using `strncmp` be more prudent?

> I commented on a pre-existing thread with what I think the issue is.

Could you please point to where? As I didn't find that, I'm posting my thought above (regardless).

tests/test_sig_stfl.c
dstebila: I think this would all be resolved if you do away with this intermediate `argument` variable and just do strncmp directly on argv[2].  Here's some sample code that I think would work, unless I'm missing something.

```c
#include <stdio.h>
#include <string.h>

int main(void) {
    char *a = "--read_key_from=abc";
    printf("%d\n", strncmp(a, "--read_key_from=", 16));
    printf("%d\n", strncmp(a, "--save_key_to=", 14));
    char *b = "--save_key_to=abc";
    printf("%d\n", strncmp(b, "--read_key_from=", 16));
    printf("%d\n", strncmp(b, "--save_key_to=", 14));
}
```

Martyrshot:
> Could you please point to where? As I didn't find that, I'm posting my thought above (regardless).
Sorry @baentsch, I should have linked it, but I said the exact same thing you did. :)

baentsch:
OK, the simple `strncmp` bug is gone. Now it's something to do with the actual logic:

```
. > ./build/tests/test_sig_stfl XMSSMT-SHA2_60/12_256 --read_key_from=./tests/saved_keys/tmp_xmssmt6012_sha256
ERROR: OQS_SIG_STFL_verify failed
```

When digging into this, it turns out `public_key` is NULL going into verify; after adding:
```
+       printf("Going into signing with pubkey %p\n", public_key);
        rc = OQS_SIG_STFL_sign(sig, signature, &signature_len, message, message_len, secret_key);
+       printf("After signing pubkey  = %p\n", public_key);
```

 the output became:
```
root@5c8026f19df1:~/project# ./build/tests/test_sig_stfl XMSSMT-SHA2_60/12_256 --read_key_from=./tests/saved_keys/tmp_xmssmt6012_sha256
Configuration info
==================
Target platform:  i586-Linux-5.13.0-1023-aws
Compiler:         gcc (7.5.0)
[...]
Going into signing with pubkey 0xf700077f
After signing pubkey  = (nil)
```

This points to a massive pointer arithmetic problem within the `sign` function (destroying variables not even passed). And I don't have time to "dive deep" into those right now, sorry. 

FWIW, I did all these tests on the CCI test VM via the SSH access function ("Re-Run Job with SSH").

kj3moraes:
> FWIW, I did all these tests on the CCI test VM via the SSH access function ("Re-Run Job with SSH").
Thanks for pointing out the problem @baentsch. Unfortunately I cant SSH into the box. I have followed the instructions on the Enable SSH dropdown but it keeps saying "Permission denied".  How did you manage to SSH into the CircleCI box ?

baentsch:
> How did you manage to SSH into the CircleCI box ?

I authorized CCI to access my github account _and_ I'm part of the "open-quantum-safe" Organization at CCI, which is why I assume CCI can check my public key registered with github (if it does that at all :-). Unfortunately I cannot check those things for your account, but  @dstebila probably knows.

dstebila:
> > How did you manage to SSH into the CircleCI box ?
> 
> I authorized CCI to access my github account _and_ I'm part of the "open-quantum-safe" Organization at CCI, which is why I assume CCI can check my public key registered with github (if it does that at all :-). Unfortunately I cannot check those things for your account, but @dstebila probably knows.

I've added @kj3moraes to the OQS organization at OQS which I think is what CCI uses for its access control.

dstebila:
@christianpaquin One of the failures we're getting on Windows with msvc2019 is due to variable length arrays. Do you know if any newer versions of msvc support variable length arrays?  A quick search seems to [suggest not](https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/) but if you have inside connections that might uncover something we don't know about, that would be helpful.

christianpaquin:
> @christianpaquin One of the failures we're getting on Windows with msvc2019 is due to variable length arrays. Do you know if any newer versions of msvc support variable length arrays?

I don't know; I'll ask around. I don't see this failure in the CI tests; do you have something I can point people to?

dstebila:
> > @christianpaquin One of the failures we're getting on Windows with msvc2019 is due to variable length arrays. Do you know if any newer versions of msvc support variable length arrays?
> 
> I don't know; I'll ask around. I don't see this failure in the CI tests; do you have something I can point people to?

There build failures occur near the end of https://ci.appveyor.com/project/dstebila/liboqs/builds/44510385/job/aj5ed4qj5dxxxa4v, which are pointing for example to https://github.com/open-quantum-safe/liboqs/blob/km-stateful-hashbased-signatures/src/sig_stateful/xmss/external/xmss_core_fast.c#L259

christianpaquin:
Following up, doesn't look like modern msvc support variable length arrays. I've seen discussions on using a different approach (vectors on C++, not sure what would be best for C). Anyway, we can certainly disable msvc support for v1 to unblock this, and address this later.

bhess:
Since no VLA are available on msvc, using malloca would be an alternative with this toolchain (if the structure fits on the stack).

dstebila:
> Since no VLA are available on msvc, using malloca would be an alternative with this toolchain (if the structure fits on the stack).

Yes, it looks like that will be the way we have to go.  Unfortunately it does mean changing source code.

kj3moraes:
> Since no VLA are available on msvc, using malloca would be an alternative with this toolchain (if the structure fits on the stack).

I have used this method to patch several locations in the code that use VLAs. A preprocessor macro to identiy the compiler _MSC_VER is used to decide whether to use _malloca or plain VLA.

kj3moraes:
@dstebila , should I merge master into this branch (and resolve any conflicts) to incorporate all the changes that took place in the latest 0.7.2 release ?

baentsch:
> should I merge master into this branch

From experience such approach is "more painful" (for you) and "chatty" (for the git log) than rebasing your contribution to the current master/main: Please consider this sequence instead:
a) Run `git rebase -i main` in this branch, ensuring all your commits are listed as "squashed" (and conflicts are resolved) -- except the one you want to retain to document what's this all about. This cleans up the git commit log.
b) If not already done, run `git config --global push.default simple`
c) `git push -f`

-> This should clean up your contribution ("collapse" all commits into one) and move this branch to a clean/mergeable state on top of `main`.

baentsch:
Can I ask about the status of this code? There's some interest to add this to the IETF interop testing.

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,417 @@
...
+#define DIRECTORY_PLUS_PREFIX "../tests/saved_keys/tmp_"
+
+#define PRIVATE_KEY_EXT ".prv"
+#define PUBLIC_KEY_EXT ".pub"
+
+typedef struct magic_s {
+	uint8_t val[31];
+} magic_t;
+
+void prepend(char *s, const char *t) {
```
dstebila, line 41: Could have some risky memory management and string functions

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,417 @@
...
+	return sk != NULL ? OQS_SUCCESS : OQS_ERROR;
+}
+
+OQS_STATUS do_nothing_save(const OQS_SECRET_KEY *sk) {
+	return sk != NULL ? OQS_SUCCESS : OQS_ERROR;
+}
+
+OQS_STATUS sk_file_write(const OQS_SECRET_KEY *sk) {
+	char filename[64];
+	strcpy(filename, (char *)sk->data);
```
dstebila, line 61: Possible buffer overflow

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,417 @@
...
+	signature = malloc(sig->length_signature + 2 * sizeof(magic_t));
+	if ((public_key == NULL) || (message == NULL) || (signature == NULL)) {
+		fprintf(stderr, "ERROR: malloc failed\n");
+		goto err;
+	}
+
+	// Define the secret key object
+	secret_key = OQS_SECRET_KEY_new(method_name);
+	secret_key->data = (void *)filestem;
+	secret_key->lock_key = lock_sk_key;
```
dstebila, line 127: Should this be `secret_key->lock_key = &lock_sk_key;` (and similarly for other function pointers in lines below)?  I don't remember precisely how function pointers work...

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,417 @@
...
+		}
+
+		size_t result = fread(public_key, 1, sig->length_public_key, pub_key);
+		if (result != sig->length_public_key) {
+			goto err;
+		}
+		fclose(pub_key);
+
+		// Private Key
+		filename[strlen(filename) - strlen(PUBLIC_KEY_EXT)] = '\0';
```
dstebila, line 182: Messy

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,417 @@
...
+	memcpy(message + message_len, magic.val, sizeof(magic_t));
+	memcpy(signature + sig->length_signature, magic.val, sizeof(magic_t));
+
+	// =================================== KEYPAIR / FILE READ =========================================
+
+	OQS_randombytes(message, message_len);
+	OQS_TEST_CT_DECLASSIFY(message, message_len);
+
+	if (mode == READING) {
+		strcpy(filename, filestem);
```
dstebila, line 166: Possible buffer overflow

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,417 @@
...
+
+		// Public Key
+		strcat(filename, PUBLIC_KEY_EXT);
```
dstebila, line 169: Possibly buffer overflow

tests/test_sig_stfl.c
```diff
@@ -0,0 +1,417 @@
...
+		}
+		fclose(pub_key);
+	}
+
+	// ===========================================================================================
+
+
+	// ====================================== SIGNING =============================================
+
+	rc = OQS_SIG_STFL_sign(sig, signature, &signature_len, message, message_len, secret_key);
```
dstebila, line 235: Just for debugging purposes, could check here to be sure that the output `signature_len` is no bigger than the maximum signature length we allocated.

tests/test_sig_stfl.c
dstebila: @xvzcf This thread might be relevant to debugging the problem

dstebila:
Here is the repository that the code was based on: https://github.com/mkannwischer/xmssfs

baentsch:
@xvzcf It's been awfully silent on this issue: Are you still looking into this? Any new problem areas identified (that someone may help with)?

xvzcf:
> @xvzcf It's been awfully silent on this issue: Are you still looking into this? Any new problem areas identified (that someone may help with)?

We're thinking of reworking this PR from the ground up and we're currently trying to determine what the best source to take the implementation from is; I can elaborate in the weekly meeting.

dstebila:
After a lot of debugging attempts, we haven't been able to isolate the problem.  Our current plan is to start with a fresh copy of the [XMSS reference code](https://github.com/XMSS/xmss-reference) and gradually re-apply the features added in this pull request.  As such we're going to close this pull request for now and start a new branch to develop on.

PR #1281, opened 2022-08-05
Update release notes

dstebila:
As suggested in https://github.com/open-quantum-safe/liboqs/discussions/1276#discussioncomment-3334829


baentsch (approved):
Thanks!

PR #1282, opened 2022-08-05
Add option to all pytests to skip particular algorithms

dstebila:
We have `SKIP_ALGS` available on `test_constant_time.py` but nowhere else.  This adds it to the general test filter alongside `SKIP_TESTS`.  (The method I've used to add it is somewhat brutish, but seems to do the job.)

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -75,7 +75,7 @@ jobs:
       - name: Run tests
         timeout-minutes: 60
         env:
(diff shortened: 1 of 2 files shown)


baentsch (changes requested):
Basically LGTM -- it would be good practice to add a test using this _somewhere_. What about skipping some Sphincs+ variant somewhere in CI? Little risk, tests the mechanism, speeds up CI and saves CO2 emissions :-)

baentsch (approved):

PR #1294, opened 2022-08-22
Ensure build without an executable stack (fixes #1285)

### LINKED ISSUE #1285: `test_non_executable_stack` fails on mips* and hppa architectures
I attempted building liboqs 0.7.2~rc1-1 on Debian unstable as part of Debian packaging effort. Test `test_non_executable_stack` is failing on Debian mips64el, mipsel and hppa architectures (full matrix [here](https://buildd.debian.org/status/package.php?p=liboqs)). Failures look like this:
```
=================================== FAILURES ===================================
__________________________ test_non_executable_stack ___________________________

    @helpers.filtered_test
    @pytest.mark.skipif(not(sys.platform.startswith("linux")), reason="Only supported on Linux")
    @pytest.mark.skipif(not(os.path.exists(helpers.get_current_build_dir_name()+'/lib/liboqs.so')), reason="Only supported on builds with a shared library")
    def test_non_executable_stack():
        liboqs = helpers.get_current_build_dir_name()+'/lib/liboqs.so'
        out = helpers.run_subprocess(
            ['readelf', '--wide', '--segments', liboqs]
        )
        lines = out.strip().split("\n")
        for line in lines:
            if "GNU_STACK" in line:
                chunks = line.strip().split()
                flags = chunks[6]
>               assert(flags == 'RW')
E               AssertionError: assert 'RWE' == 'RW'
E                 - RW
E                 + RWE
E                 ?   +
```

sebastinas:
Until it is clear why the shared library on mips64el and hppa is built with the executable bit set for the stack, build with both `-Wa,--noexecstack` (for the assembler) and `-Wl,-z,--noexecstack` (for the linker).

I can confirm that with this change liboqs 0.7.2 builds on mips64el.


baentsch (approved):
LGTM. Particularly like the effort to support older `cmake` versions. Thanks!

PR #1300, opened 2022-08-31
Fallback code for `aligned_alloc` and use of `explicit_bzero`

sebastinas:
There are two parts to this PR. The first one adds checks for `aligned_alloc` and friends and uses whatever is available. The second part fixes the check for `memset_s` (via cmake and Annex K support) and also uses `explicit_bzero` if available. 

### DIFF
--- a/src/common/common.c
+++ b/src/common/common.c
@@ -5,7 +5,12 @@
 #include <errno.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
+
+#if !defined(OQS_HAVE_POSIX_MEMALIGN) || defined(__MINGW32__) || defined(__MINGW64__) || defined(_MSC_VER)
+#include <malloc.h>
+#endif
 
 #if defined(_WIN32)
 #include <windows.h>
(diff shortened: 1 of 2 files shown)


src/common/common.c
```diff
@@ -233,31 +245,69 @@ OQS_API void OQS_MEM_insecure_free(void *ptr) {
...
+#else
+	// Check alignment (power of 2) and size (multiple of alignment)
+	if (alignment & (alignment - 1) || size & (alignment - 1)) {
+		errno = EINVAL;
+		return NULL;
+	}
+
+#if defined(OQS_HAVE_POSIX_MEMALIGN)
+	// Check alignment (needs to be >= sizeof(void*))
+	if (alignment < sizeof(void *)) {
```
baentsch, line 259: Shouldn't this condition rather result in an immediate error as this violates the [precondition](https://github.com/open-quantum-safe/liboqs/blob/9ee96d803ef8fb822a7d9b32572276b47d9575f4/src/common/common.h#L199-L205)? Also, would callers of the function not consider `alignment` immutable (by the function)?

baentsch (commented):
Except for the one question, all else LGTM. Thanks for the contribution! What's missing from your side to move this PR out of Draft mode?

src/common/common.c
sebastinas: Interesting points.

1) The standard requires for `aligned_alloc` that the size if a multiple of the alignment and the alignment to be a power of 2. Implementations can define additional preconditions, i.e. that the alignment is a multiple of `sizeof(void*)` as it's the case in glibc. So my question is: should the OQS provided function be close to the requirements outlined by the standard or should it mimic glib'c `aligned_alloc`? 

2) Increasing the alignment is always safe (as long as `sizeof(void*)` of 2). The returned pointer satisfies the callers alignment requirements. But after looking at the code again, size might need to be adjusted for the new alignment.

Depending on the answer to 1), I'll update the code accordingly. I have no strong opinion. In Picnic I was trying to go for loosing as many implementation-defined requirements as possible. But if all OQS internal users satisfy the `sizeof(void*)` requirement anyway, I can also enforce that in the compat code.

sebastinas:
> What's missing from your side to move this PR out of Draft mode?

I was waiting for the CI to complete.

baentsch:
> So my question is: should the OQS provided function be close to the requirements outlined by the standard or should it mimic glib'c aligned_alloc?

No strong preference from my side. @jschanck : The comment was added by you: Any preference from your side?

src/common/common.c
sebastinas: I've changed it to fail if `alignment < sizeof(void*)`. If there is interest to relax the requirement, we can do that later.

src/common/common.c
```diff
@@ -233,31 +245,64 @@ OQS_API void OQS_MEM_insecure_free(void *ptr) {
...
-#else // musl, maybe others.
-	return aligned_alloc(alignment, size);
+#elif defined(OQS_HAVE_MEMALIGN)
+	return memalign(alignment, size);
+#elif defined(__MINGW32__) || defined(__MINGW64__)
+	return __mingw_aligned_malloc(size, alignment);
+#elif defined(_MSC_VER)
+	return _aligned_malloc(size, alignment);
+#else
+	if (!size) {
```
dstebila, line 272: Thanks Sebastian! Would you mind adding a few comments here and in the corresponding part of `OQS_MEM_aligned_free` explaining what's going on with the pointer arithmetic?  Just so we have a record in case someone else needs to debug a future problem here.

sebastinas: Done.

dstebila (approved):

dstebila:
Thanks Sebastian!

Issue #1303, opened 2022-09-13
Signature length miscalculation of Falcon & Picnic Families - Python Wrapper


Solymanx7:
I found this issue while I was working on a project that compares between different PQC algorithms regarding one of my subjects in master, and wanted to notify you about it.

Kindly check attached image, I compared 5 algorithms together the only one that has mismatching values of signature length is Falcon-512.

Here is code snippet I used to calculate signature size which I believe it gives accurate value 
`print("Signature size: ",len(signature)," Bytes")`

![2022-09-13_21h34_09](https://user-images.githubusercontent.com/29024788/189993705-1f676d57-e825-4068-a7aa-dbd7865ada3f.png)

baentsch:
Where do you take the value "655" from? Manually from the specs? Further, this deviation does not occur with Falcon-1024? If so, this were a duplicate of https://github.com/open-quantum-safe/openssl/discussions/395; Issue is already visible upstream in https://github.com/PQClean/PQClean/blob/d3fcc0fa8d868835a92545fbd08c17497a787102/crypto_sign/falcon-512/META.yml#L6

Solymanx7:
Yes from details function : `signer.details`, but problem still occurs with Falcon-1024 where actual signature size = 1279  Bytes , while details function return 1330.
Also all Picnic_* algorithms have same issue , kindly check attached image for example.
![2022-09-14_08h32_30](https://user-images.githubusercontent.com/29024788/190077259-639f2a7b-30cc-4f98-89a4-8ae059c32349.png)

Solymanx7:
I have changed the title so that it matches perfectly the content of the issue.

baentsch:
> Yes from details function 

The origin of value "690" in the case of falcon-512 is understood. I was wondering where you take the "Signature Size" values in your output from? Both for falcon as well as for picnic ("655" in the case of falcon-512).

Solymanx7:
> > Yes from details function 
> 
> The origin of value "690" in the case of falcon-512 is understood. I was wondering where you take the "Signature Size" values in your output from? Both for falcon as well as for picnic ("655" in the case of falcon-512).

I used len() function from py on the signature itself

Solymanx7:
Thanks for the cooperative team and for the quick response and fix, Thank you baentsch for your effort !
I'm happy for contributing to this library 
If no one minds, I'm closing this issue.

PR #1306, opened 2022-09-14
addressing sig length questions

baentsch:
Explains #1303 

### DIFF
--- a/src/sig/sig.h
+++ b/src/sig/sig.h
@@ -242,7 +242,7 @@ typedef struct OQS_SIG {
 	 * compile-time macros `OQS_SIG_*_length_*`.
 	 *
 	 * @param[out] signature The signature on the message represented as a byte string.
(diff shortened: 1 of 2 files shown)


src/sig/sig.h
```diff
@@ -242,7 +242,7 @@ typedef struct OQS_SIG {
 	 * compile-time macros `OQS_SIG_*_length_*`.
 	 *
 	 * @param[out] signature The signature on the message represented as a byte string.
-	 * @param[out] signature_len The length of the signature.
+	 * @param[out] signature_len The actual length of the signature. May be smaller than `length_signature` for some algorithms.
```
dstebila, line 245:
```suggestion
	 * @param[out] signature_len The actual length of the signature. May be smaller than `length_signature` for some algorithms since some algorithms have variable length signatures.
```

dstebila (approved):

Issue #1317, opened 2022-11-23
Remove NTRU-Prime, Saber, Rainbow, and Picnic.


xvzcf:
They're no longer in the NIST PQC competition and we lack other compelling reasons to keep them.

baentsch:
@xvzcf FYI, Here are all downstream PRs readily accessible in one place.

xvzcf:
Accomplished by PRs https://github.com/open-quantum-safe/liboqs/pull/1325, https://github.com/open-quantum-safe/liboqs/pull/1326, https://github.com/open-quantum-safe/liboqs/pull/1321, https://github.com/open-quantum-safe/liboqs/pull/1323

PR #1321, opened 2022-11-24
Remove rainbow

xvzcf:
Removes the rainbow signature scheme as per https://github.com/open-quantum-safe/liboqs/issues/1317. I'll submit a PR per algorithm since removing each one is not totally mechanical and straightforward. Once all the algorithms are removed I'll move on to the downstream projects.

* [x] Does this PR change the the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in [oqs-provider](https://github.com/open-quantum-safe/oqs-provider), [OQS-OpenSSL](https://github.com/open-quantum-safe/openssl), [OQS-BoringSSL](https://github.com/open-quantum-safe/boringssl), and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)


baentsch (commented):
Piecemeal removal is OK -- but is the assumption that you merge each PR before moving on to the next? If so, I'd strongly suggest you work "backwards", i.e., first remove the algs from downstream use. Otherwise, this approach will generate many failed CI runs. If not, i.e., you intend to merge only when all algs are done and then move on "downstream", the failing downstream CI runs will not be so numerous ... maybe I'll install an oqs-github-mailstorm-filter on my Inbox :)

xvzcf:
> Piecemeal removal is OK -- but is the assumption that you merge each PR before moving on to the next? If so, I'd strongly suggest you work "backwards", i.e., first remove the algs from downstream use. Otherwise, this approach will generate many failed CI runs. If not, i.e., you intend to merge only when all algs are done and then move on "downstream", the failing downstream CI runs will not be so numerous ... maybe I'll install an oqs-github-mailstorm-filter on my Inbox :)

If we let the downstream CIs fail for a few days, once all the algorithms are removed in liboqs it'll just be one PR to each downstream as opposed to the 4 PRs per downstream if we update each time we remove an algorithm here. Granted, it would be ideal if each PR in liboqs gets reviewed and merged quickly to minimize the number of emails sent.

baentsch:
> If we let the downstream CIs fail for a few days, once all the algorithms are removed in liboqs it'll just be one PR to each downstream as opposed to the 4 PRs per downstream if we update each time we remove an algorithm here.

The above does not argue for (4) downstream PR's per algorithm removal, but for first removing _all_ algs from the downstream projects in one go: The downstream projects can build OK with the algorithms-to-be-removed-in-liboqs eliminated/not referenced any more -- but not vice versa. Therefore PRs for the downstreams have just been created: If those would be OK for you (and get merged), removal of the algs in `liboqs` should be possible to proceed piecemeal without any CI breakages.

dstebila:
> > If we let the downstream CIs fail for a few days, once all the algorithms are removed in liboqs it'll just be one PR to each downstream as opposed to the 4 PRs per downstream if we update each time we remove an algorithm here.
> 
> The above does not argue for (4) downstream PR's per algorithm removal, but for first removing _all_ algs from the downstream projects in one go: The downstream projects can build OK with the algorithms-to-be-removed-in-liboqs eliminated/not referenced any more -- but not vice versa. Therefore PRs for the downstreams have just been created: If those would be OK for you (and get merged), removal of the algs in `liboqs` should be possible to proceed piecemeal without any CI breakages.

Ah, very clever!

baentsch (approved):

xvzcf:
I did, perhaps by the time the other algorithms are removed this will have been merged in.

PR #1328, opened 2022-12-12
Add ntruprime

ryndia:
The purpose of the pr is to add the ntruprime key exchange algorithm.

* [ ] 
Yes this pr add ntruprime to the available algorithm for key exchange algorithm

Thank you.


ryndia:
Hello, I removed all other ntruprime algorithm except the sntrup761 as requested and sorry for the last 3 commits, make some typo error while editing the code.

Thank you

baentsch:
Thanks for the PR. However, the majority of the OQS core team is not convinced that it is a good idea to retain an algorithm that is not supported by its authors any more. Therefore, please let us know 
a) who/how many people need this algorithm in active ("main" branch) support in liboqs and
b) whether there would be the problem using [the last "round 3" release of liboqs supporting this algorithm](https://github.com/open-quantum-safe/liboqs/releases/tag/0.7.2) until [a NIST-approved algorithm is added to openssh](https://github.com/open-quantum-safe/liboqs/issues/1329#issuecomment-1349225890).

loganaden:
We did not see any statement about the status of ntru prime from the authors. Going through https://ntruprime.cr.yp.to/warnings.html doesn't mention that it was abandoned by the authors. Can someone please clarify ?

OpenSSH hasn't indicated that they will remove ntru prime in favour of kyber. They could also support both right ?

dstebila:
> We did not see any statement about the status of ntru prime from the authors. Going through https://ntruprime.cr.yp.to/warnings.html doesn't mention that it was abandoned by the authors. Can someone please clarify ?

That's true.  But equally we're not aware of any further push for standardization of NTRU Prime in some other venue.

> OpenSSH hasn't indicated that they will remove ntru prime in favour of kyber. They could also support both right ?

Certainly they could, but we think it is best for the community to focus on standards.

djmdjm:
OpenSSH will continue to support NTRUPrime but will likely add hybrid Kyber/x25519 soon too. Barring a weakness discovered in NTRUPrime, I doubt that we would remove it before 2027 at the earliest (i.e. we'll keep it for at least five years)

baentsch:
> OpenSSH will continue to support NTRUPrime but will likely add hybrid Kyber/x25519 soon too. Barring a weakness discovered in NTRUPrime, I doubt that we would remove it before 2027 at the earliest (i.e. we'll keep it for at least five years)

Thanks for this truly long-term statement, @djmdjm : I'd thus think it would be prudent to have sntrup761 in the main branch, particularly considering liboqs and its integrations may change APIs over such time horizon, making past `liboqs` releases containing sntrup761 probably useless eventually.

> Certainly they could, but we think it is best for the community to focus on standards.

This argument taken to the extreme would call for the outright removal of all code not on [the "STD" and "NIST_R4" lists in #1333](https://github.com/open-quantum-safe/liboqs/blob/f947520848704706777cedf591fcdc9221c453f6/.CMake/alg_support.cmake#L565-L568): I don't think we want to do that, right?

What about this proposal then: This PR needs to re-base, add a statement to the algorithm documentation that it's retained only for openssh interop testing and/or until a weakness in the algorithm is detected and is not listed in the "STD" or "NIST_R4" `liboqs` build option (i.e., the "experimental" list of algorithms). Objections, @xvzcf @dstebila @christianpaquin ?

bbbrumley:
> retain an algorithm that is not supported by its authors any more

Please don't mistake lack of changes to the NTRU Prime specification as lack of support :) You should take it a a sign of stability.

dstebila:
Alright, hearing enough demand for sntrup761, let's keep it in; I think appropriate to put it under the "All" flag rather than "STD" or "NIST_R4".

thomwiggers:
There's also been some discussion about this on the CFRG mailing list: https://mailarchive.ietf.org/arch/msg/cfrg/AX9WGbiNpNmRpe9KJ1DEc7vi6ss/

baentsch:
> There's also been some discussion about this on the CFRG mailing list: https://mailarchive.ietf.org/arch/msg/cfrg/AX9WGbiNpNmRpe9KJ1DEc7vi6ss/

Thanks for the pointer. A key statement in https://mailarchive.ietf.org/arch/msg/cfrg/exMg3HsGkSyI8S17UgutQww6u_g/:

> - Don’t ever use non-standardized algorithms like SIDH.

But then again, `liboqs` supported SIDH until "known-broken" -- and then swiftly removed it. The proposal above is the same for `sntrup761`: Keep until broken (or not needed).

So, @ryndia, please re-base so this gets properly reviewed.

ryndia:
@baentsch I update the pull request, I hope its good.
Thank you

.CMake/alg_support.cmake
```diff
@@ -216,6 +216,20 @@ if(OQS_DIST_X86_64_BUILD OR (OQS_USE_AES_INSTRUCTIONS AND OQS_USE_AVX2_INSTRUCTI
 endif()
 endif()
 
+option(OQS_ENABLE_KEM_NTRUPRIME "Enable ntruprime algorithm family" ON)
+cmake_dependent_option(OQS_ENABLE_KEM_ntruprime_ntrulpr653 "" ON "OQS_ENABLE_KEM_NTRUPRIME" OFF)
```
baentsch, line 220: Why is this option here? We only want to enable sntrup761, right?

baentsch (changes requested):
sntrup761 added to testing and successfully tested: Good. Code updates also good, just one  one "spurious" reference to ntrulpr653 in a `cmake` file -- weird... Please remove that and we're good as far as I'm concerned.

baentsch:
One more thought: Please add an explicit statement that "sntrup761" is only included until it is not needed for interop testing any more. The statement that it shall be removed if broken already is implicit in the language at https://github.com/open-quantum-safe/liboqs#limitations-and-security -- so maybe it might be suitable to add the "needed-for-interop" statement also there. Agreed, @dstebila ?
Finally, please feel free to add yourself to https://github.com/open-quantum-safe/liboqs/blob/main/CONTRIBUTORS, @ryndia

baentsch (approved):

bbbrumley:
Very nice catches @baentsch ! LGTM. (Admittedly I skipped most of the `pq_clean_*` additions, since it looks like these are coming from some upstream.)

PR #1333, opened 2022-12-14
Enable algorithm filtering

### LINKED ISSUE #1331: Algorithm enablement
Following up on https://github.com/open-quantum-safe/liboqs/issues/1245#issuecomment-1338878620 this is to propose implementing a `cmake` build option labelled "OQS_ALGS_ENABLED" with the following possible values and effects:
- STD: Enabling Kyber512,768,1024 (no 90s), Dilithium2,3,5 (no AES), SPHINCS+ (not robust), Falcon
- NIST_R4: Enabling All algs explicitly listed for round 4 of the NIST competition
- EXPERIMENTAL: Enabling all algs not removed from liboqs; this would also activate for example Dilithium5AES or "robust" SPHINCS+ (and possibly NTRUPrime algs along the lines of [this discussion](https://github.com/open-quantum-safe/liboqs/issues/1245#issuecomment-1346632256)).

Proposed default value: "STD" (also to be activated if "OQS_ALGS_ENABLED" is not explicitly set).

Comments/alternative proposals welcome before implementing as described.

baentsch:
Fixes #1331 

* [ No] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [ No] Does this PR change the the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in [oqs-provider](https://github.com/open-quantum-safe/oqs-provider), [OQS-OpenSSL](https://github.com/open-quantum-safe/openssl), [OQS-BoringSSL](https://github.com/open-quantum-safe/boringssl), and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)


.CMake/alg_support.cmake
```diff
@@ -2,6 +2,48 @@
...
+      if(_var MATCHES "^OQS_ENABLE_..._" AND NOT _var MATCHES "_AVAILABLE$")
+          set(${_var} OFF PARENT_SCOPE)
+          # Case 1, family name
+	  foreach (_alg ${ARGV0})
+             string(TOUPPER ${_alg} upalg)
+	     if("OQS_ENABLE_${upalg}i" MATCHES "^${_var}")
+                 set(${_var} ON PARENT_SCOPE)
+             endif()
+          endforeach()
+          # Case 2, exact match
```
xvzcf, line 24: Rather than this sort of finicky parsing, wouldn't it be preferable to have additional `{nist-r4|std}.cmake` files where we have a bunch of `option(OQS_ENABLE_{KEM|SIG}_{ALG} "Enable {ALG} algorithm family" ON)` lines (these could be auto-generated).

baentsch: Why create more code if existing code does the job? Besides, this code until now successfully drove our "OQS_MINIMAL_BUILD" logic, so arguably is proven to work OK.

baentsch:
@dstebila FYI, additional testing incl. AppVeyor/Windows added as discussed yesterday, so this is now ready from my side.

dstebila (changes requested):
Code looks good to me. But if we're going to change the default compilation behaviour to omit some algorithms, then I think that should also update README.md to indicate this.

baentsch:
> Code looks good to me. But if we're going to change the default compilation behaviour to omit some algorithms, then I think that should also update README.md to indicate this.

So done in https://github.com/open-quantum-safe/liboqs/commit/4eb3567da561fb012a037413cdbfac2fac39bb35

baentsch:
Just re-considered the implications of merging this: Downstream projects must be prepared for the potential reduction of algorithms supported (or if no optimized build for some downstream projects are desired, must build `liboqs` with the "All/experimental algs" setting). So this PR shall not be merged before those PRs are available.

README.md
```diff
@@ -36,6 +36,8 @@ More information on OQS can be found [here](https://openquantumsafe.org/) and in
 
 Details on each supported algorithm can be found in the [docs/algorithms](https://github.com/open-quantum-safe/liboqs/tree/main/docs/algorithms) folder.
 
+The list below references all available algorithms. Recommended for further use are the ones either already standardized or in the process of standardization by NIST. To this end, `liboqs` by default is built only with support for "Kyber" (excluding the "-90s" variants), Dilithium (excluding the "aes" variants), Falcon, SPHINCS+ (excluding the "robust" variants). Other, e.g., NIST round 4 or experimental, algorithms can be activated by setting the [OQS_ALGS_ENABLED](CONFIGURE.md#oqs_algs_enabled) build configuration variable to suitable values other than "STD".
```
dstebila, line 39:
```suggestion
The list below indicates all algorithms supported by liboqs, but not all those algorithms are available in the default build configuration.  The default build configuration includes only those algorithms that have been selected for standardization by NIST, specifically Kyber (excluding the "-90s" variants), Dilithium (excluding the "-AES" variants), Falcon, and SPHINCS+ (excluding the "robust" variants). Other algorithms, such as those included NIST round 4 or otherwise made available for experimental purposes,  can be activated by setting the [OQS_ALGS_ENABLED](CONFIGURE.md#oqs_algs_enabled) build configuration variable to `NIST_R4` or `All`. 
```

tests/example_kem.c
```diff
@@ -42,7 +42,7 @@ static OQS_STATUS example_stack(void) {
 #ifndef OQS_ENABLE_KEM_frodokem_640_aes // if FrodoKEM-640-AES was not enabled at compile-time
 	printf("[example_stack] OQS_KEM_frodokem_640_aes was not enabled at "
 	       "compile-time.\n");
-	return OQS_ERROR;
+	return OQS_SUCCESS; // nothing done successfully ;-)
```
dstebila, line 45: Was there a reason you changed this?  I would think we'd want these to return failures so that we know the test wasn't configured properly.

baentsch: Yes. Only with this change the test passes OK when Frodo640AES is not configured in (for "STD" and "NIST_R4" configs). The logic for this change was that if Frodo is not active there's no error in the library/configuration or the test. It's simply that Frodo is not there.

The alternative would be to re-write the whole test to "dynamically" select an algorithm that is available (e.g., by introducing yet another C-level config variable -- that could be dependent on the new `OQS_ALGS_ENABLED` `cmake`-level define). This would change the focus in this example code  from the simple "this is how to use one specific algorithm" to "this is how to configure liboqs for use with different algorithms", though: Would that be OK from your perspective?

dstebila: Ah, I just noticed this was an example file, not a test file.  And as far as I can tell none of the test suites actually execute this, so it's fine.  Although at this point it would probably be appropriate to switch it over to Kyber -- which I have done in a new commit.

baentsch: > And as far as I can tell none of the test suites actually execute this, so it's fine.

Not quite right: 
The examples _are_ part of testing: https://github.com/open-quantum-safe/liboqs/blob/a7e1b8a32f482f3b93b7d08d686d62942e5674df/tests/test_cmdline.py#L8-L13

> it would probably be appropriate to switch it over to Kyber -- which I have done in a new commit.

Thanks -- makes sense from a "standardization point of view" (and resolves the issue for config option "STD").

But this doesn't resolve it for config "NIST_R4" -- which does not include Kyber: CI would have failed if you had reverted the return code change that started the whole discussion (to fail on alg not available).

So three options now:
1) leave as-is now
2) make example fail on alg-not-available and add Kyber to "NIST_R4"
3)  make example fail on alg-not-available and create mechanism to select available algorithm (e.g., by new "OQS_ALGS_ENABLED-derived" C-level define as suggested above)

baentsch (commented):
One more thought: Please add an explicit statement that "sntrup761" is only included until it is not needed for interop testing any more. The statement that it shall be removed if broken already is implicit in the language at https://github.com/open-quantum-safe/liboqs#limitations-and-security -- so maybe it might be suitable to add the "needed-for-interop" statement also there. Agreed, @dstebila ?
Finally, please feel free to add yourself to https://github.com/open-quantum-safe/liboqs/blob/main/CONTRIBUTORS, @ryndia

baentsch:
Thanks @dstebila for marking the conversation resolved. For some reason, there's still a blocking "Changes requested" marker, though... Did I overlook something?

dstebila (approved):

PR #1350, opened 2023-01-11
llvm15 update

baentsch:
This follows https://github.com/open-quantum-safe/ci-containers/pull/65 and unblocks CI failures.

Suggested to merge after #1351 


xvzcf:
Could you fix the conflict and then we can merge it in?

Issue #1358, opened 2023-01-13
Mismatch between markdown and YAML documentation for some algorithms


xvzcf:
- In [classic_mceliece.md](https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/kem/classic_mceliece.md): The line `Note: This algorithm is known to fail memory leak testing on x86_64.` is not in the corresponding YAML.
- In [hqc.md](https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/kem/hqc.md): The `Security advisory` section is not in the corresponding YAML.
- In [sphincs.md](https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/sig/sphincs.md): The `Test limitation` section is not in the corresponding YAML.

The mismatch causes these lines/sections to be removed when `update_upstream_alg_docs.py` is run.

PR #1359, opened 2023-01-13
Updated PQClean commit in `copy_from_upstream.yml`

xvzcf:


src/oqsconfig.h.cmake
```diff
@@ -61,6 +61,13 @@
 #cmakedefine OQS_ENABLE_KEM_frodokem_976_shake 1
 #cmakedefine OQS_ENABLE_KEM_frodokem_1344_aes 1
 #cmakedefine OQS_ENABLE_KEM_frodokem_1344_shake 1
+#cmakedefine OQS_ENABLE_KEM_NTRUPRIME 1
+#cmakedefine OQS_ENABLE_KEM_ntruprime_sntrup761 1
+#cmakedefine OQS_ENABLE_KEM_ntruprime_sntrup761_avx2 1
+
```
baentsch, line 67: Code duplication?

baentsch (changes requested):
LGTM except one set of apparently duplicated define's. However, I'd suggest to either manually add back the documentation items lost as per #1358 (that's what I did after running the script previously) or to first fix #1358 permanently.

dstebila (approved):

src/oqsconfig.h.cmake
xvzcf: Fixed.

xvzcf:
> LGTM except one set of apparently duplicated define's. However, I'd suggest to either manually add back the documentation items lost as per #1358 (that's what I did after running the script previously) or to first fix #1358 permanently.

I've manually added the documentation items back for now.

baentsch (approved):

PR #1384, opened 2023-02-08
Fix rendering error in Markdown

dstebila:
The Markdown for "ancestors of primary source" wasn't rendering properly, this fixes it.


baentsch (approved):

baentsch:
Just a nit: Is it intentional that the order "Source" and "Implementation license" changes with this PR (for "Optimized Implementation Sources") , i.e., no longer is in line with the ordering for "Primary Source"? See for example with "kyber".

PR #1386, opened 2023-02-09
Update Falcon to 20230207 

### LINKED ISSUE #1315: Malleable Falcon signatures, may need upgrade to v1.2 spec
Hi there,

The clean/avx2 implementations of Falcon-512 and Falcon-1024 seem to be unaligned with some changes introduced in v1.2 of the spec.

In particular, the implementation of the signature verification algorithm currently in libOQS accepts mauled signatures due to how `comp_decode` in `codec.c` is implemented. I attached an example program for ease of inspection.

Following Section 3.11.2 and Algorithms 17-18 in the Falcon spec v1.2, I believe that in order to enforce a unique encoding of compressed signatures, the decode implementation should not accept 100000001 ("-0") as a valid encoding of 0, and should check that signatures are padded using zeroes.

Kind Regards,
Giacomo Fenzi

Attached bug: [bug.c](https://gist.github.com/WizardOfMenlo/59e6e0457f94bf5caf13909f319007de)

dstebila:
PQClean commit 96dfee95cc56207d1ec5e3a2df306d6614ad7c6c

Fixes #1315

* [No] Does this PR change the the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in [oqs-provider](https://github.com/open-quantum-safe/oqs-provider), [OQS-OpenSSL](https://github.com/open-quantum-safe/openssl), [OQS-BoringSSL](https://github.com/open-quantum-safe/boringssl), and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)


baentsch (approved):
LGTM. As KATs change I presume this also requires updates to  OIDs/sig numbers downstream?

dstebila:
> LGTM. As KATs change I presume this also requires updates to OIDs/sig numbers downstream?

Yes. What order do you want to proceed in for merging?  Merge this and then do number changes downstream, or prepare number changes downstream first?  Note that there should not be any downstream build failures due to merging this, as the set of algorithms available remains the same.

baentsch:
> there should not be any downstream build failures due to merging

Agreed -- hence, I'd say merge this now and we update the downstreams afterwards. Shall we do this now already (for Falcon and BIKE) or wait for HQC & McEliece? I'd opt for the former if these two algs are not "close to availability".

dstebila:
> > there should not be any downstream build failures due to merging
> 
> Agreed -- hence, I'd say merge this now and we update the downstreams afterwards. Shall we do this now already (for Falcon and BIKE) or wait for HQC & McEliece? I'd opt for the former if these two algs are not "close to availability".

I think we can go with the BIKE and Falcon downstream updates now.  McEliece may not be too far away, but HQC will definitely be further off.

baentsch:
@dstebila Same question as with BIKE: Is it really correct that the Falcon version didn't change (still "1.2")? This ID would still be used downstream -- but with different code points -- that doesn't feel right.

dstebila:
You're right, should be bumped. I'm not sure what the current version identifier is -- trying to track it down.

Issue #1390, opened 2023-02-14
Falcon tests failing


dstebila:
https://github.com/open-quantum-safe/liboqs/actions/runs/4154153928/jobs/7186334170#step:6:3093

baentsch:
It's more than that: Falcon's also the reason that [openssl tests began failing](https://app.circleci.com/pipelines/github/open-quantum-safe/openssl/1119/workflows/1f7c0a91-f914-4be5-a388-9b6a41a451b5/jobs/5353)

baentsch:
Update: Falcon 1024 reliably fails in oqs-provider, too -- when doing many (>10) repetitions of the same (keygen-sign-verify) cycle --> There's got to be some "randomness" issue that makes signatures "bad"/unverifiable. The same problem cannot be reproduced with falcon512. Based on this, a `liboqs`-based reproducer has been created with #1391.

baentsch:
"One more thing" -- when running the new "full cycle" speed test on OSX, the falcon1024 problem does not appear. It also does not appear on Windows when using MSVC. Using GNU tools, the problem does appear also under Windows, though. Fascinating. My guess is the quality of the randomness generators differs between the platforms and that has an impact on how Falcon-1024 operates (or not).
Long story short: Until someone really understands what's going on here, I'd suggest to revert #1386.

PR #1391, opened 2023-02-16
Add full-cycle speed test

baentsch:
Running `./tests/speed_sig Falcon-1024` with this new code reliably reproduces #1390 and could unearth similar such errors in the future -- ~when~ now added to CI: ~Feedback as to whether we should do this welcome: Would complete this PR with KEM-equivalent and github CI tests then~.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -117,7 +117,7 @@ jobs:
       - run:
           name: Run tests
           no_output_timeout: 1h
-          command: mkdir -p tmp && python3 -m pytest --verbose --ignore=tests/test_code_conventions.py --junitxml=build/test-results/pytest/test-results.xml << parameters.PYTEST_ARGS >>
+          command: mkdir -p tmp && python3 -m pytest --verbose --ignore=tests/test_speed.py --ignore=tests/test_code_conventions.py --junitxml=build/test-results/pytest/test-results.xml << parameters.PYTEST_ARGS >>
           environment:
              SKIP_ALGS: << parameters.SKIP_ALGS >>
       - store_test_results: # Note that this command will fail when running CircleCI locally, that is expected behaviour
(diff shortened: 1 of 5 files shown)


baentsch:
Failures mostly "as expected" (triggered by the Falcon-1024 bug). @Martyrshot : Can you see a reason why this code SIGSEGVs on the ARM emulator? It works perfectly on the M1 (??).

Martyrshot:
My first thought is that maybe an optimized implementation is being used when it shouldn't be. I wonder if the new falcon implementation checks feature support differently than the previous version did.

baentsch:
@Martyrshot Sorry, I wasn't clear in my question: It's only the new "fullcycletest" (keygen-sign-verify) that fails on ARM emulator -- and that for all algorithms (see test log), not just falcon.

Martyrshot:
@baentsch Can I get access to that environment? Do these happen only when `fullcycle` is `true`? Or do these illegal instructions happen regardless?

baentsch:
> @baentsch Can I get access to that environment? 

Most definitely: It's "simply" a set of docker images. All code is visible [here](https://github.com/open-quantum-safe/liboqs/blob/40b01fdbb270f8614fde30e65d30e9da18c02393/.github/workflows/linux.yml#L94-L136). I'm right now trying to reproduce myself...

dstebila (approved):

thomwiggers:
See also https://github.com/PQClean/PQClean/issues/477

baentsch:
> if there is some data width issues somewhere

Maybe -- but definitely not in the code added as part of this PR: Just confirmed that the `speed_kem/sig` programs _always_ crash on this platform -- in other words, it seems like repeated API calls fail there. Created #1393 to track -- in case anyone has a keen interest in that old platform.

PR #1395, opened 2023-02-20
Update Falcon implementation

### LINKED ISSUE #1390: Falcon tests failing

thomwiggers:
Updates Falcon to the latest version of the code.

* See also #1386 (version number is now better specified, instead of just the day I got the code from Thomas Pornin).

Closes #1390 and #1315


docs/algorithms/sig/falcon.yml
```diff
@@ -17,7 +18,7 @@ website: https://falcon-sign.info
 nist-round: 3
 spec-version: v1.2
```
baentsch, line 19: KATs changed -- so would a spec version update not be called for?

docs/algorithms/sig/falcon.yml
```diff
@@ -17,7 +18,7 @@ website: https://falcon-sign.info
...
 primary-upstream:
-  source: https://github.com/PQClean/PQClean/commit/33bceb17eb06a40fbdc72251f533734e8d869615
+  source: https://github.com/PQClean/PQClean/commit/245c95cd1ec326f8f38e26cb17a28832701ba17b
   spdx-license-identifier: CC0-1.0
   upstream-ancestors:
```
baentsch, line 23: Is this license identifier still correct given the newly documented patent "caveat"?

thomwiggers: This license identifier was never correct, as Falcon's implementation is MIT-licensed.

baentsch: :-) :-/ An opportunity to correct this then, no?

baentsch:
FWIW, the "randomness" error is gone/repeated speed tests pass -- Thanks, @thomwiggers! However, the constant_time tests fail on my machine.

docs/algorithms/sig/falcon.yml
thomwiggers: The spec of Falcon is a complicated story. The Falcon 1.2 PDF specifies different lengths of signatures than what the implementation has used. This PR switches over to the shorter signature lengths, so actually we're bringing it in spec with 1.2?

On the other hand, there's no way we're compatible with the NIST KATs now, because those still use the 690-byte signatures.

baentsch: Is there any way to get input from the authors of this algorithm? How do they refer to the different algorithm versions, if not by version ID? If this implementation is not in line with the NIST KATs, does reference to NIST make any sense for this code version? How is this code base then to be viewed relative to the NIST competition? "between rounds"? "unrelated"?

baentsch (changes requested):
- Clarity to be obtained as to how this code relates to the NIST competition
- license update
- version ID update
- constant time checks should pass (or algorithm should be documented not having this property/excluding it from testing, see #1397 as example for the latter)

thomwiggers:
The implementation should be constant-time; and there aren't too many differences with the implementation that was here before. The main difference seems the checks to resolve #1315. Maybe the existing suppressions need to be reconstructed because a bunch of the code moved back from c files into .h files

docs/algorithms/sig/falcon.yml
thomwiggers: I don't think they're being that consistent about it. The implementation simply supports a bunch of different encodings of signatures. I'm not sure why the signatures were still set at the longer lengths, in the NIST submission, but I don't think it's necessary anymore.

baentsch: So, how do we conclude this then?

baentsch:
> the existing suppressions need to be reconstructed because a bunch of the code moved back from c files into .h files

Will you add this to this PR then or do you want to do a separate PR as soon as weekly CI constant time testing fails (maybe also a "solution" for #1400)?

thomwiggers:
I have no idea how to do that

baentsch:
> I have no idea how to do that

This is documented in https://github.com/open-quantum-safe/liboqs/blob/main/tests/test_constant_time.py .

thomwiggers:
I'm not sure if I will have time in the coming month to write these valgrind suppression files, I'm afraid, so I would appreciate it if someone could take over this PR.

dstebila:
> I'm not sure if I will have time in the coming month to write these valgrind suppression files, I'm afraid, so I would appreciate it if someone could take over this PR.

I've updated the constant-time suppression files.

baentsch:
> @baentsch What do you mean by "license update"?

This has been adressed in the mean time by https://github.com/open-quantum-safe/liboqs/pull/1395/commits/1d9315858fd800ce27b55a24f00f3f670b73da29

baentsch:
> I've updated the constant-time suppression files.

Thanks very much. They're looking good to me (and the tests pass on my machine). That indeed only leaves the version ID question open to close this.

dstebila:
> > I've updated the constant-time suppression files.
> 
> Thanks very much. They're looking good to me (and the tests pass on my machine). That indeed only leaves the version ID question open to close this.

Version ID question is now resolved: an email from Thomas Pornin to me and Thom says that he would call this version "20211101", so I've labelled it as such.

dstebila:
Any objections to merging?

baentsch (approved):

baentsch:
> Any objections to merging?

Nope. Thanks @thomwiggers for the PR and @dstebila for addressing all issues (I had :)

PR #1397, opened 2023-02-21
document BIKE lacks constant time protection

### LINKED ISSUE #1396: bike-l5 constant time test failing
https://github.com/open-quantum-safe/liboqs/actions/runs/4214023415/jobs/7314192772

We could simply add a line to  the [the suppression file](https://github.com/open-quantum-safe/liboqs/blob/49164467b6456a217b09c5d1a6c830abb64fc1a6/tests/constant_time/kem/issues.json#L2-L3) -- but that BIKE is not constant time should also be suitably documented, e.g., along the same lines in that we document that HQC is not providing constant time protection, right?

baentsch:
Fixes #1396 . Tagging @dkostic @crockeea for fact check.

* [ no Does this PR change the the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in [oqs-provider](https://github.com/open-quantum-safe/oqs-provider), [OQS-OpenSSL](https://github.com/open-quantum-safe/openssl), [OQS-BoringSSL](https://github.com/open-quantum-safe/boringssl), and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -16,12 +16,12 @@ jobs:
             container: openquantumsafe/ci-ubuntu-focal-x86_64:latest
             CMAKE_ARGS: -DOQS_OPT_TARGET=generic -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON
             PYTEST_ARGS: --numprocesses=auto -k 'test_constant_time'
-            SKIP_ALGS: 'SPHINCS\+-SHA*,Classic-McEliece-6(.)*'
+            SKIP_ALGS: 'SPHINCS\+-SHA*,Classic-McEliece-6(.)*,BIKE*'
           - name: extensions
             container: openquantumsafe/ci-ubuntu-focal-x86_64:latest
             CMAKE_ARGS: -DOQS_OPT_TARGET=haswell -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON
             PYTEST_ARGS: --numprocesses=auto -k 'test_constant_time'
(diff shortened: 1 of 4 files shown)


dkostic:
BIKE should be constant-time, at least we tried to do so. I'd like to go over the logs and fix any potential non-ct behavior. In the meantime, feel free to mark it as non-ct.

baentsch:
> BIKE should be constant-time, at least we tried to do so. I'd like to go over the logs and fix any potential non-ct behavior. In the meantime, feel free to mark it as non-ct.

Nope -- if you think the BIKE code is c-t, by all means, please check our tests and correct them (or the code). Will not move this PR beyond Draft until hearing back from you: Thanks for looking into this.

dkostic:
I opened #1400 to address one of the issues reported by Valgrind. I believe the other issues are false positives but I'd appreciate another set of eyes on it to confirm that:
```
==7198== Use of uninitialised value of size 1
==7198==    at 0x114055: _mm256_slli_epi64 (avx2intrin.h:698)
==7198==    by 0x114055: OQS_KEM_bike_l5_secure_set_bits_avx2 (sampling_avx2.c:67)
...
==7198== Use of uninitialised value of size 1
==7198==    at 0x1AA559: _mm256_srli_epi64 (avx2intrin.h:789)
==7198==    by 0x1AA559: rotate256_small (decode_avx2.c:88)
```
^^Both values in those function calls are initialized afaict.

```
==7198== Conditional jump or move depends on uninitialised value(s)
==7198==    at 0x1A6AF1: get_threshold (decode.c:103)
```
^^There is no conditional jump in this function and all the variables are initialized before use.

baentsch:
Obviated by #1400

PR #1399, opened 2023-02-22
correct free in test_kem/sig

### LINKED ISSUE #1398: Cleanup in test_sig.c seems wrong
The cleanup section in `tests/test_sig.c` has a check if `sig != NULL`, in which case it will free the `secret_key` buffer. But a bunch of other buffers are also only `malloc`d after `sig` is initialized.  This would not be a problem, because `OQS_MEM_insecure_free` handles `NULL` correctly, however, `sizeof(magic_t)` is subtracted from each of these pointers.

This should probably be handled differently; just moving the checks into the `sig != NULL` if is probably not sufficient as there is another case where they may end up there with some buffers partially initialized; if one of the `malloc`s fails.

baentsch:
Fixes #1398 

### DIFF
--- a/tests/test_kem.c
+++ b/tests/test_kem.c
@@ -162,16 +162,14 @@ err:
 	ret = OQS_ERROR;
 
 cleanup:
-	if (kem != NULL) {
-		if (secret_key) {
-			OQS_MEM_secure_free(secret_key - sizeof(magic_t), kem->length_secret_key + 2 * sizeof(magic_t));
-		}
-		if (shared_secret_e) {
-			OQS_MEM_secure_free(shared_secret_e - sizeof(magic_t), kem->length_shared_secret + 2 * sizeof(magic_t));
-		}
-		if (shared_secret_d) {
-			OQS_MEM_secure_free(shared_secret_d - sizeof(magic_t), kem->length_shared_secret + 2 * sizeof(magic_t));
-		}
+	if (secret_key) {
+		OQS_MEM_secure_free(secret_key - sizeof(magic_t), kem->length_secret_key + 2 * sizeof(magic_t));
+	}
+	if (shared_secret_e) {
+		OQS_MEM_secure_free(shared_secret_e - sizeof(magic_t), kem->length_shared_secret + 2 * sizeof(magic_t));
+	}
+	if (shared_secret_d) {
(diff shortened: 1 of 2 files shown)


tests/test_kem.c
```diff
@@ -163,12 +163,22 @@ static OQS_STATUS kem_test_correctness(const char *method_name) {
...
+		if (secret_key) {
+			OQS_MEM_secure_free(secret_key - sizeof(magic_t), kem->length_secret_key + 2 * sizeof(magic_t));
+		}
+		if (shared_secret_e) {
+			OQS_MEM_secure_free(shared_secret_e - sizeof(magic_t), kem->length_shared_secret + 2 * sizeof(magic_t));
+		}
+		if (shared_secret_d) {
+			OQS_MEM_secure_free(shared_secret_d - sizeof(magic_t), kem->length_shared_secret + 2 * sizeof(magic_t));
+		}
+	}
```
thomwiggers, line 175: Do we even still need `kem != NULL`?

tests/test_sig.c
```diff
@@ -143,11 +143,19 @@ static OQS_STATUS sig_test_correctness(const char *method_name) {
 
 cleanup:
 	if (sig != NULL) {
-		OQS_MEM_secure_free(secret_key - sizeof(magic_t), sig->length_secret_key + 2 * sizeof(magic_t));
+		if (secret_key) {
+			OQS_MEM_secure_free(secret_key - sizeof(magic_t), sig->length_secret_key + 2 * sizeof(magic_t));
+		}
```
thomwiggers, line 148: do we still need `sig != NULL` here?

thomwiggers (changes requested):
Looks pretty good, I just have one question about `sig` and `kem`. Are these pointers `NULL`-initialized?

thomwiggers:
(Yes, they are `NULL`-initialized)

tests/test_sig.c
baentsch: no

tests/test_kem.c
baentsch: no

dstebila (approved):

baentsch:
@thomwiggers Please re-review to remove the merge blocker.

thomwiggers (approved):

PR #1400, opened 2023-02-23
Make BIKE decode function void to avoid ct issues

dkostic:
The `decode` function in BIKE was made void to avoid any potential constant-time issues.

Partially addresses #1396 


baentsch (changes requested):
Please be sure to update https://github.com/open-quantum-safe/liboqs/blob/49164467b6456a217b09c5d1a6c830abb64fc1a6/tests/constant_time/kem/issues.json#L2-L3 Documentation in https://github.com/open-quantum-safe/liboqs/blob/main/tests/test_constant_time.py

dstebila (approved):

baentsch:
The latest commit removed the BIKE CT-error suppression without adding a curated one and in consequence now [constant time tests](https://github.com/open-quantum-safe/liboqs/blob/1f62f66a4722e34cf82a69adbd953786d087b4f0/.github/workflows/weekly.yml#L15-L24) fail for all BIKE variants. 
@thomwiggers @dstebila : Are your approvals indicating your intention to scrap the `valgrind` testing then or have you given it without consideration of the (weekly run) tests by @jschanck ?

thomwiggers:
I have not run this code, but it looks decent. I was expecting such problems to be caught by CI though.

baentsch:
> I have not run this code, but it looks decent. 

Well, removing suppression (as done here) means all (also false positive) `valgrind` findings are triggered and ct testing fails. As part of my review, I did run this code through "test_constant_time.py" to see whether it indeed is flawless as the PR implies by eliminating suppression entirely -- but it failed. See the documentation in https://github.com/open-quantum-safe/liboqs/blob/main/tests/test_constant_time.py.

> I was expecting such problems to be caught by CI though.

It is -- but only after the merge as we moved these very compute-intensive tests to weekly testing in an effort to have faster CI cycles and be more environmentally and economically sensitive: This decision possibly needs to be reverted as it  creates a false belief about this particular code property if the author does not run the test locally. The latter is necessary to check the `valgrind` report and do a suitable suppression file - manually checking & documenting each and every `valgrind` finding for correctness. Again, see the documentation in https://github.com/open-quantum-safe/liboqs/blob/main/tests/test_constant_time.py. Now also added a statement to the [Contributing Wiki](https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide#contributing-new-quantum-safe-algorithms).

thomwiggers:
Not running these tests on each push probably makes sense, but not automatically running them at all before pushing is indeed maybe also inconvenient and error-prone. I'm not sure if this can easily be set up in Github Actions or CircleCI, but it would be nice if it would be possible to run a second layer of CI tests "pre-merge". I think I've seen such stuff in the Rust project with e.g. their bot Bors (https://github.com/rust-lang/homu), which can also roll up a bunch of changes and subject them to the large test suite together. Checking if a commit is a "roll-up" should at least be easy.

thomwiggers:
I'm not sure if I will have time in the coming month to write these valgrind suppression files, I'm afraid, so I would appreciate it if someone could take over this PR.

tests/constant_time/kem/issues.json
```diff
@@ -1,6 +1,7 @@
 {
-  "BIKE-L1": ["bike_has_no_timing_protections"],
-  "BIKE-L3": ["bike_has_no_timing_protections"],
+  "BIKE-L1": [],
```
dstebila, line 2: As Michael points out, removing this means that constant-time tests are failing.

dstebila (changes requested):

dstebila:
> The latest commit removed the BIKE CT-error suppression without adding a curated one and in consequence now [constant time tests](https://github.com/open-quantum-safe/liboqs/blob/1f62f66a4722e34cf82a69adbd953786d087b4f0/.github/workflows/weekly.yml#L15-L24) fail for all BIKE variants. @thomwiggers @dstebila : Are your approvals indicating your intention to scrap the `valgrind` testing then or have you given it without consideration of the (weekly run) tests by @jschanck ?

No I hadn't given consideration of the constant-time feature, thanks for catching that, Michael.

@dkostic, is the current BIKE implementation intended to be constant-time?  If so, then there are potential non-constant-time lines that will need to be investigated, and either fixed or manually flagged as not a problem via suppression files.  If it's not intended to be constant-time, then the lines in the issues.json file need to be restored.

If you want to see what's being detected, here is how to do it.  

```bash
# use the OQS Docker CI image
docker pull openquantumsafe/ci-ubuntu-focal-x86_64:latest
docker run -dti openquantumsafe/ci-ubuntu-focal-x86_64:latest /bin/bash
docker attach <identifier output by previous command>
# clone your repo
git clone git@github.com:dkostic/liboqs
cd liboqs
git checkout bike-fix-ct-issues
# build with constant-time test enabled
cmake -GNinja -DOQS_OPT_TARGET=generic -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON ..
ninja
# run Valgrind on BIKE-L1
valgrind --tool=memcheck --gen-suppressions=all --error-exitcode=1 --max-stackframe=20480000 --num-callers=20 tests/test_kem BIKE-L1
```

baentsch:
@dstebila : Thanks for the "reproducer script". Slightly simplified and corrected below:
```
docker run -it openquantumsafe/ci-ubuntu-focal-x86_64 bash
```

In the `docker` image now running:

```
git clone --depth 1 --branch bike-fix-ct-issues https://github.com/dkostic/liboqs.git
cd liboqs && mkdir build && cd build
cmake -GNinja -DOQS_OPT_TARGET=generic -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON .. && ninja
valgrind --tool=memcheck --gen-suppressions=all --error-exitcode=1 --max-stackframe=20480000 --num-callers=20 tests/test_kem BIKE-L1
```

@dkostic : Task remaining for a person who understands the code inside-out: Review suppression reports output as to whether they contain truly "dubious" code behaviour -- if not, give each suppression a name explaining why the code is no problem and then check them into the "tests/constant_time/kem/passes" folder with a suitable name (referenced in the passes.json file read be the `liboqs` ct test script).

dkostic:
As I commented in #1397, this PR addresses one of the issues reported by Valgrind that is valid. I believe the other issues are false positives but I'd appreciate another set of eyes on it to confirm that. Sorry, I should have left the comment on this PR instead. I can wait until tomorrow and if no-one objects until then I'll add the suppression files.

Valgrind reports these two errors that I think are false positives:
```
==7198== Use of uninitialised value of size 1
==7198==    at 0x114055: _mm256_slli_epi64 (avx2intrin.h:698)
==7198==    by 0x114055: OQS_KEM_bike_l5_secure_set_bits_avx2 (sampling_avx2.c:67)
...
==7198== Use of uninitialised value of size 1
==7198==    at 0x1AA559: _mm256_srli_epi64 (avx2intrin.h:789)
==7198==    by 0x1AA559: rotate256_small (decode_avx2.c:88)
```
^^Both values in those function calls are initialized afaict.

```
==7198== Conditional jump or move depends on uninitialised value(s)
==7198==    at 0x1A6AF1: get_threshold (decode.c:103)
```
^^There is no conditional jump in this function and all the variables are initialized before use.

dstebila:
> Valgrind reports these two errors that I think are false positives:
> 
> ```
> ==7198== Use of uninitialised value of size 1
> ==7198==    at 0x114055: _mm256_slli_epi64 (avx2intrin.h:698)
> ==7198==    by 0x114055: OQS_KEM_bike_l5_secure_set_bits_avx2 (sampling_avx2.c:67)
> ```

Maybe this is being triggered by the `w` value that has been set from `wlist` on line sampling_avx2.c:65?  Does `wlist` contain data derived from secret data?

> ```
> ==7198== Use of uninitialised value of size 1
> ==7198==    at 0x1AA559: _mm256_srli_epi64 (avx2intrin.h:789)
> ==7198==    by 0x1AA559: rotate256_small (decode_avx2.c:88)
> ```

I can't tell from reading the code.

> ```
> ==7198== Conditional jump or move depends on uninitialised value(s)
> ==7198==    at 0x1A6AF1: get_threshold (decode.c:103)
> ```
> 
> ^^There is no conditional jump in this function and all the variables are initialized before use.

Weird... am I reading correctly that this seems to be about the `DMSG` macro which prints out debugging information?

dkostic:
>==7198== Use of uninitialised value of size 1
==7198==    at 0x114055: _mm256_slli_epi64 (avx2intrin.h:698)
==7198==    by 0x114055: OQS_KEM_bike_l5_secure_set_bits_avx2 (sampling_avx2.c:67)
Maybe this is being triggered by the w value that has been set from wlist on line sampling_avx2.c:65? Does wlist contain data derived from secret data?

yes, `wlist` is part of the secret key but I don't see any "non safe" operations with `w`, or non-ct accesses to `wlist`.

>==7198== Conditional jump or move depends on uninitialised value(s)
==7198==    at 0x1A6AF1: get_threshold (decode.c:103)
Weird... am I reading correctly that this seems to be about the DMSG macro which prints out debugging information?

This is line 103: https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/bike/additional_r4/decode.c#L103.
I don't think it's about DMSG, that's a no-op macro (https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/bike/additional_r4/defs.h#L106).

dstebila:
> > ==7198== Use of uninitialised value of size 1
> > ==7198==    at 0x114055: _mm256_slli_epi64 (avx2intrin.h:698)
> > ==7198==    by 0x114055: OQS_KEM_bike_l5_secure_set_bits_avx2 (sampling_avx2.c:67)
> > Maybe this is being triggered by the w value that has been set from wlist on line sampling_avx2.c:65? Does wlist contain data derived from secret data?
> 
> yes, `wlist` is part of the secret key but I don't see any "non safe" operations with `w`, or non-ct accesses to `wlist`.

Indeed, it seems like `w` is just being used as the offset in the shift operation, so unless `_mm256_slli_epi64` has some weird non constant-time behaviour in its second argument, I can't see anything.

> > ==7198== Conditional jump or move depends on uninitialised value(s)
> > ==7198==    at 0x1A6AF1: get_threshold (decode.c:103)
> > Weird... am I reading correctly that this seems to be about the DMSG macro which prints out debugging information?
> 
> This is line 103: https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/bike/additional_r4/decode.c#L103. I don't think it's about DMSG, that's a no-op macro (https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/bike/additional_r4/defs.h#L106).

Wouldn't it be https://github.com/dkostic/liboqs/blob/bike-fix-ct-issues/src/kem/bike/additional_r4/decode.c#L103 ?  Which version of the code are you running Valgrind on?

dkostic:
>Indeed, it seems like w is just being used as the offset in the shift operation, so unless _mm256_slli_epi64 has some weird non constant-time behaviour in its second argument, I can't see anything.

ok, I'll add that in a suppression file.

>Wouldn't it be https://github.com/dkostic/liboqs/blob/bike-fix-ct-issues/src/kem/bike/additional_r4/decode.c#L103 ? Which version of the code are you running Valgrind on?

nope, the Valgrind error message is coming from the code in the main branch, so https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/bike/additional_r4/decode.c#L103

dstebila:
> nope, the Valgrind error message is coming from the code in the main branch, so https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/bike/additional_r4/decode.c#L103

Okay. I can't see any reason why decode.c#L103 would raise that warning -- it's just uint32_t addition and multiplication.

thomwiggers:
Briefly returning to the issue of maybe batch-running certain tests before merges to throw this in: it seems Github CI's new merge queue feature may be helpful https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue

tests/constant_time/kem/passes.json
```diff
@@ -1,6 +1,7 @@
 {
-  "BIKE-L1": [],
-  "BIKE-L3": [],
+  "BIKE-L1": ["bike"],
+  "BIKE-L3": ["bike"],
+  "BIKE-L3": ["bike"],
```
baentsch, line 4: This should read "BIKE-L5". Tests otherwise fail.

baentsch (changes requested):
Thanks for the update. Except for one typo (I presume) this addresses all my previous comments. Thanks. Accordingly closed #1397.

tests/constant_time/kem/passes.json
dkostic: right, fixed it.

baentsch (approved):

jschanck:
> Okay. I can't see any reason why decode.c#L103 would raise that warning -- it's just uint32_t addition and multiplication.

The constants there (`THRESHOLD_COEFF0` and `THRESHOLD_COEFF1`) are floating point values. Looking at the assembly from gcc 10 on my AVX2 machine, I do see a branch inserted for converting a `uint64_t` into a `double` using the `VCVTSI2SD` instruction. The instruction expects a signed integer; the branch is there to handle the case where the high bit of the unsigned integer happens to be set.

I put a breakpoint on the branch in gdb and didn't hit it, and from the C code it doesn't look like the high bit of the integer should ever be set. Still, this doesn't look like a false positive to me. If it is, it might be a bug in valgrind's emulation of a nearby instruction. More likely, especially given the warnings about unintialized values, there's some UB that makes it hard to reason about the compiled program from the C description.

dkostic:
>The constants there (THRESHOLD_COEFF0 and THRESHOLD_COEFF1) are floating point values. Looking at the assembly from gcc 10 on my AVX2 machine, I do see a branch inserted for converting a uint64_t into a double using the VCVTSI2SD instruction. The instruction expects a signed integer; the branch is there to handle the case where the high bit of the unsigned integer happens to be set.
I put a breakpoint on the branch in gdb and didn't hit it, and from the C code it doesn't look like the high bit of the integer should ever be set. Still, this doesn't look like a false positive to me. If it is, it might be a bug in valgrind's emulation of a nearby instruction. More likely, especially given the warnings about unintialized values, there's some UB that makes it hard to reason about the compiled program from the C description.

That's a very good point, thanks! I'll look into ways to handle this appropriately by doing integer arithmetic instead of floating point (and avoiding division by multiplying and shifting)...

dkostic:
I pushed another change that should fix the issue pointed out by John. I also removed the suppression for the issue from the constant-time test suppression files. Tested with Docker instructions from above -- looks like it's fixed.

src/kem/bike/additional_r4/decode.c
```diff
@@ -93,11 +93,26 @@ _INLINE_ uint8_t get_threshold(IN const syndrome_t *s)
...
+  //   T0' = 1353000000 and T1' = 697220.
+  // Then, instead of computing the threshold by:
+  //   T0 + T1*S,
+  // we compute:
+  //   (T0' + T1'*S)/10^8,
+  // where S is the syndrome weight.
+  // All relatively modern compilers transform the division by 10^8 to
+  // a multiplication and a right shift (both constant-time instructions).
+  uint64_t thr  = THRESHOLD_COEFF0 + (THRESHOLD_COEFF1 * syndrome_weight);
+  thr /= THRESHOLD_DIV_CONST;
```
jschanck, line 113: Why not write this explicitly as a multiplication and right shift? Are you sure that *every* compiler with *every* combination of optimization flags does the right thing?

Personally, I would ask the submission team to revise the constants so that they can be represented exactly in binary fixed point. The constants were obtained through a numerical optimization process---there's no reason to choose constants that complicate the implementation.

dstebila: Would revising the constants be an algorithmic change (resulting in different I/O behaviour), or just an implementation change?

dkostic: Hi folks, sorry for such a long delay, covid put me out of work for a few weeks.

@jschanck I replaced the division with mul+shift as you suggested. When/if BIKE team revise the constants we can change the implementation.

dkostic: @dstebila you requested a change before merging, is the latest change satisfying your requirements?

dstebila (approved):

PR #1414, opened 2023-03-06
add merge group

baentsch:
Adds the use of Github action merge groups as per https://github.com/open-quantum-safe/liboqs/pull/1400#issuecomment-1453292004

May completely replace weekly testing if everyone agrees this is a good approach (and it works as advertised :).

### DIFF
--- a/.github/workflows/mergegroup.yml
+++ b/.github/workflows/mergegroup.yml
@@ -1,5 +1,6 @@
-name: Weekly constant time tests
+name: Constant time tests in merge queue
 
+# should only trigger on merge queue as per https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue
 on:
   pull_request:
   merge_group:
--- a/.github/workflows/mergegroup.yml
+++ b/.github/workflows/mergegroup.yml
@@ -1,8 +1,8 @@
 name: Weekly constant time tests
 
 on:
-  schedule:
-  - cron: "5 0 * * 0"
+  pull_request:
+  merge_group:
 
 jobs:
 
@@ -12,9 +12,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: extensions
+          - name: generic
             container: openquantumsafe/ci-ubuntu-focal-x86_64:latest
-            CMAKE_ARGS: -DOQS_OPT_TARGET=haswell -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON
+            CMAKE_ARGS: -DOQS_OPT_TARGET=generic -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON
             PYTEST_ARGS: --numprocesses=auto -k 'test_constant_time'
             SKIP_ALGS: 'SPHINCS\+-SHA*,Classic-McEliece-6(.)*'
     container:


baentsch:
@dstebila Could you please see whether you can add/enable "Require merge queue" in the project settings as per https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#managing-a-merge-queue ? This then should enable this "run-only-once-before-merge" feature if I get it right.

dstebila:
> @dstebila Could you please see whether you can add/enable "Require merge queue" in the project settings as per https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#managing-a-merge-queue ? This then should enable this "run-only-once-before-merge" feature if I get it right.

I believe I've enabled it; there were a variety of options to set which I left at the default.

baentsch:
> I believe I've enabled it; there were a variety of options to set which I left at the default.

Thanks. Let's then see whether a merge enables this new feature. Provided we merge -- it seems Falcon has ct-problems (?!?).

baentsch:
> Let's then see whether a merge enables this new feature. 

@dstebila FYI, your changes did enable a new way to merge ("tested" with #1405), BUT
- this makes [AppVeyor fail](https://ci.appveyor.com/project/dstebila/liboqs/builds/46427899)
- there's no option to edit the squash commit message (handy to fix up longer PR "history" for better merge commit message), so

> there were a variety of options to set which I left at the default.

Any of those options that may improve the above?

dstebila:
> @dstebila FYI, your changes did enable a new way to merge ("tested" with #1405), BUT
> 
>  * this makes [AppVeyor fail](https://ci.appveyor.com/project/dstebila/liboqs/builds/46427899)
>  * there's no option to edit the squash commit message (handy to fix up longer PR "history" for better merge commit message), so
> 
> > there were a variety of options to set which I left at the default.
> 
> Any of those options that may improve the above?

I don't think any of the options will help.  Here's a screenshot of the available options, none of them look like obvious helpers to me.

image

As for the AppVeyor fail, I don't know why the requested branch is not available from AppVeyor's perspective.  One could probably set up the AppVeyor configuration to be skipped on branches matching this naming convention, but that's a workaround, not a solution.

baentsch:
> I don't think any of the options will help. Here's a screenshot of the available options, none of them look like obvious helpers to me.

Well, the first one shows it's using Squash-and-merge by default (just without the ability to edit the combined merge message): Is there a "Squash-and-merge-with-edit" option? If not, maybe we should suggest that feature -- and/or drop our use of merge queues. What's your opinion? At the core, lots of effort to save committers from themselves (having to fix bugs emerging in the weekly run that would be obvious if they'd run that single test locally).

> As for the AppVeyor fail, I don't know why the requested branch is not available from AppVeyor's perspective. One could probably set up the AppVeyor configuration to be skipped on branches matching this naming convention, but that's a workaround, not a solution.

Agree. "Funny" thing: After the merge was done, AppVeyor ran again on `main` -- and ran OK. So the "intermediate AppVeyor failure" isn't really one (i.e., the last merge CI run is OK as it should be). [The documentation](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-other-ci-providers) suggests creating such "work-around", but then again, why? If AppVeyor doesn't properly run on these temporary branches, I'd call it goodness: It _should_ run only once (on `main`) -- and that it does. So I'd consider this not a problem.

dstebila:
> > I don't think any of the options will help. Here's a screenshot of the available options, none of them look like obvious helpers to me.
> 
> Well, the first one shows it's using Squash-and-merge by default (just without the ability to edit the combined merge message): Is there a "Squash-and-merge-with-edit" option? If not, maybe we should suggest that feature -- and/or drop our use of merge queues. What's your opinion? At the core, lots of effort to save committers from themselves (having to fix bugs emerging in the weekly run that would be obvious if they'd run that single test locally).

The options are "Merge commit", "Squash and merge", or "Rebase and merge".  

If we want to edit the main merge commit title, I think we can do that by editing the Pull Request title first, and it will just use that.  But as for the longer merge commit comment, I don't think there'll be any way to clean that up.

> > As for the AppVeyor fail, I don't know why the requested branch is not available from AppVeyor's perspective. One could probably set up the AppVeyor configuration to be skipped on branches matching this naming convention, but that's a workaround, not a solution.
> 
> Agree. "Funny" thing: After the merge was done, AppVeyor ran again on `main` -- and ran OK. So the "intermediate AppVeyor failure" isn't really one (i.e., the last merge CI run is OK as it should be). [The documentation](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-other-ci-providers) suggests creating such "work-around", but then again, why? If AppVeyor doesn't properly run on these temporary branches, I'd call it goodness: It _should_ run only once (on `main`) -- and that it does. So I'd consider this not a problem.

Okay, so we'd need to implement this workaround to tell AppVeyor not to run on these temporary branches?

baentsch:
> But as for the longer merge commit comment, I don't think there'll be any way to clean that up.

I'd then consider this feature to create an unduly unreadable commit log and suggest to disable merge groups again. Closing this PR in turn.

dstebila:
> > But as for the longer merge commit comment, I don't think there'll be any way to clean that up.
> 
> I'd then consider this feature to create an unduly unreadable commit log and suggest to disable merge groups again. Closing this PR in turn.

Okay, I've disabled merge queues.

PR #1415, opened 2023-03-09
Fix constant time failure for Falcon AVX2

dstebila:
Constant-time tests for Falcon on AVX2 are failing as per https://github.com/open-quantum-safe/liboqs/actions/runs/4333882416/jobs/7567377140

This is due to line numbers changing slightly since the last revision.

* [No] Does this PR change the the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in [oqs-provider](https://github.com/open-quantum-safe/oqs-provider), [OQS-OpenSSL](https://github.com/open-quantum-safe/openssl), [OQS-BoringSSL](https://github.com/open-quantum-safe/boringssl), and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)

### DIFF
--- a/tests/constant_time/sig/passes/falcon_keygen
+++ b/tests/constant_time/sig/passes/falcon_keygen
@@ -26,6 +26,12 @@
 {
    Rejection sampling for small (f,g) norm
    Memcheck:Cond
+   src:keygen.c:4175 # fun:PQCLEAN_FALCON*_*_keygen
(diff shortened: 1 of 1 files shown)


baentsch (approved):
Tests passed OK -- this time on an AVX2 machine...

Issue #1245, opened 2022-07-03
Classify algorithm status


baentsch:
A concrete NIST announcement date, hurray: https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/7yLIZcFOMF0/m/vn43l1tQAQAJ

Bets still accepted.... :-) Eliminating McEliece and SPHINCS+ would be a boon for our CI runtime (and the world's power-consumption-induced CO2 emissions).

baentsch:
Well, (combatting) global warming seems to have been no major decision criterion :-(

On the bright side it seems we now can claim/document to have all [selected](https://csrc.nist.gov/Projects/post-quantum-cryptography/selected-algorithms-2022) and all [round 4](https://csrc.nist.gov/Projects/post-quantum-cryptography/round-4-submissions) algorithms. Suggestion thus to prune (post 0.7.2)

- [ ] Frodo
- [ ] NTRU
- [ ] NTRU prime
- [ ] Rainbow

jschanck:
I suggest keeping NTRU for now. There's a footnote on page 18 of the [round 3 status report](https://nvlpubs.nist.gov/nistpubs/ir/2022/NIST.IR.8413.pdf) which says that NTRU is the backup if the patent agreements for Kyber are not finalized by end-of-year.

dstebila:
> I suggest keeping NTRU for now. There's a footnote on page 18 of the [round 3 status report](https://nvlpubs.nist.gov/nistpubs/ir/2022/NIST.IR.8413.pdf) which says that NTRU is the backup if the patent agreements for Kyber are not finalized by end-of-year.

Yes, I had in mind that we'd probably keep NTRU around for the time being due to that footnote. In today's status meeting we didn't make any firm decisions about what to prune, instead wanting to think more over the next couple of weeks and see whether there will be a future for any of the algorithms outside of NIST, e.g. resubmission to Round 4 or interest by other governments. We might provide some configure-time options to help people select certain bundles of algorithms.

baentsch:
Agree. So the goal for the resolution of this issue is to define and create a mechanism to build `liboqs` such as to differentiate at configure time (at least) between "standardized", "round4" and "research" algorithms, leading to different levels of (CI) testing and default inclusion in library builds. A final class comprises algorithms to be completely pruned. 

So here's a first proposal:

"standard": kyber, dilithium, falcon, sphincs
"round4": bike, classic_mceliece, hqc, sike
[...]

PR #1420, opened 2023-03-18
Update SPHINCS+

### LINKED ISSUE #1249: Pull SPHINCS+ update
https://github.com/sphincs/sphincsplus/pull/28 and https://github.com/PQClean/PQClean/issues/446.  Note that this changes algorithm specification, so if we include this in liboqs 0.7.2 our policy would actually bump the version number to 0.8.0. Alternatively we could leave this until after the 0.7.2 release cycle.

dstebila:
Fixes #1249.

* [Yes] Does this PR change the the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in [oqs-provider](https://github.com/open-quantum-safe/oqs-provider), [OQS-OpenSSL](https://github.com/open-quantum-safe/openssl), [OQS-BoringSSL](https://github.com/open-quantum-safe/boringssl), and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)
* [ ] TODO: Ensure all references to the removed variants (Haraka) are removed.


.CMake/alg_support.cmake
```diff
@@ -468,61 +468,133 @@ if(OQS_DIST_X86_64_BUILD OR (OQS_USE_AVX2_INSTRUCTIONS))
     cmake_dependent_option(OQS_ENABLE_SIG_sphincs_shake256_128f_robust_avx2 "" ON "OQS_ENABLE_SIG_sphincs_shake256_128f_robust" OFF)
 endif()
 
+if(CMAKE_SYSTEM_NAME MATCHES "Linux|Darwin")
+if(OQS_DIST_ARM64_V8_BUILD OR (OQS_USE_ARM_NEON_INSTRUCTIONS AND OQS_USE_SHA3_INSTRUCTIONS))
```
baentsch, line 472: This looks _very_ fishy: Should this rather read "OQS_USE_ARM_SHA3_INSTRUCTIONS"?

baentsch (changes requested):
Didn't we decide to drop the "robust" variants as per NIST's recommendation (https://github.com/open-quantum-safe/liboqs/issues/1245#issuecomment-1337566477 )?

.CMake/alg_support.cmake
baentsch: Possibly fixed by https://github.com/open-quantum-safe/liboqs/pull/1421

dstebila: Thanks!

dstebila:
> Didn't we decide to drop the "robust" variants as per NIST's recommendation ([#1245 (comment)](https://github.com/open-quantum-safe/liboqs/issues/1245#issuecomment-1337566477) )?

Yes, you're right.  I've removed the robust variants as well as the Haraka variants (NIST's email also says they only plan to move forward with SHA2 and SHAKE variants).

baentsch:
> removed the robust variants as well as the Haraka variants

Compile runs down from 1900 to 1300 .... The environment says Thank you :) 

But the compile errors still confuse me: I locally have missing symbols

```
/usr/bin/ld: thash_shake_robustx4.c:(.text.PQCLEAN_SPHINCSSHAKE128FROBUST_AVX2_thashx4+0x2a3): undefined reference to `KeccakP1600times4_PermuteAll_24rounds'
/usr/bin/ld: lib/liboqs.a(hash_shakex4.c.o): in function `PQCLEAN_SPHINCSSHAKE128FSIMPLE_AVX2_prf_addrx4':
hash_shakex4.c:(.text.PQCLEAN_SPHINCSSHAKE128FSIMPLE_AVX2_prf_addrx4+0x17e): undefined reference to `KeccakP1600times4_PermuteAll_24rounds'
/usr/bin/ld: lib/liboqs.a(thash_shake_simplex4.c.o): in function `PQCLEAN_SPHINCSSHAKE128FSIMPLE_AVX2_thashx4':
thash_shake_simplex4.c:(.text.PQCLEAN_SPHINCSSHAKE128FSIMPLE_AVX2_thashx4+0x237): undefined reference to `KeccakP1600times4_PermuteAll_24rounds'
/usr/bin/ld: lib/liboqs.a(hash_shakex4.c.o):hash_shakex4.c:(.text.PQCLEAN_SPHINCSSHAKE128SROBUST_AVX2_prf_addrx4+0x17e): more undefined references to `KeccakP1600times4_PermuteAll_24rounds' follow
collect2: error: ld returned 1 exit status
```
and the CI runs have even weirder problems. Maybe a fresh look tomorrow resolves these....

dstebila:
> > removed the robust variants as well as the Haraka variants
> 
> Compile runs down from 1900 to 1300 .... The environment says Thank you :)
> 
> But the compile errors still confuse me: I locally have missing symbols
> 
> ```
> /usr/bin/ld: thash_shake_robustx4.c:(.text.PQCLEAN_SPHINCSSHAKE128FROBUST_AVX2_thashx4+0x2a3): undefined reference to `KeccakP1600times4_PermuteAll_24rounds'
> /usr/bin/ld: lib/liboqs.a(hash_shakex4.c.o): in function `PQCLEAN_SPHINCSSHAKE128FSIMPLE_AVX2_prf_addrx4':
> hash_shakex4.c:(.text.PQCLEAN_SPHINCSSHAKE128FSIMPLE_AVX2_prf_addrx4+0x17e): undefined reference to `KeccakP1600times4_PermuteAll_24rounds'
> /usr/bin/ld: lib/liboqs.a(thash_shake_simplex4.c.o): in function `PQCLEAN_SPHINCSSHAKE128FSIMPLE_AVX2_thashx4':
> thash_shake_simplex4.c:(.text.PQCLEAN_SPHINCSSHAKE128FSIMPLE_AVX2_thashx4+0x237): undefined reference to `KeccakP1600times4_PermuteAll_24rounds'
> /usr/bin/ld: lib/liboqs.a(hash_shakex4.c.o):hash_shakex4.c:(.text.PQCLEAN_SPHINCSSHAKE128SROBUST_AVX2_prf_addrx4+0x17e): more undefined references to `KeccakP1600times4_PermuteAll_24rounds' follow
> collect2: error: ld returned 1 exit status
> ```
> 
> and the CI runs have even weirder problems. Maybe a fresh look tomorrow resolves these....

I did a quick pull without looking much into the changes.  I recall there was a change involving removal of a Keccak-related file from the PQClean directories, but I didn't look into it.  I don't think I'll get a chance to do so over the next few days, unfortunately.

baentsch:
>  I recall there was a change involving removal of a Keccak-related file from the PQClean directories, but I didn't look into it. 

Yup -- before this PR there was a `copy_from_upstream` patch file for sphincs removing `fips202x4.[c|h]`. Seems to be gone for some reason (?). Manually doing these removals only unearthed missing includes like for `immintrin.h`, so some more work required...

> I don't think I'll get a chance to do so over the next few days, unfortunately.

NP; so focusing just on functionality (sphincs+ update) for the IETF hackathon I opted to simply remove the AVX2 optimizations from this PR (see https://github.com/open-quantum-safe/liboqs/pull/1422). This solves most build problems, but for example https://github.com/open-quantum-safe/liboqs/actions/runs/4465669616/jobs/7842995392, confirms the same results on my local machine and the (optimized) M1, namely incorrect KATs for 4 algorithms:

```
FAILED tests/test_kat.py::test_sig[SPHINCS+-SHA256-192f-simple] - AssertionError: assert 'fd4e301339b2...2b414f962f44c' == '5302cb39e848...b4f38caa7b99d'
FAILED tests/test_kat.py::test_sig[SPHINCS+-SHA256-256s-simple] - AssertionError: assert '05d15a742539...fa529071b0748' == 'db9b817748aa...652e1c98cdbf7'
FAILED tests/test_kat.py::test_sig[SPHINCS+-SHA256-192s-simple] - AssertionError: assert '0fa07f3f7775...3924ba2713839' == '9ecda7889d92...0226ebb06daf2'
FAILED tests/test_kat.py::test_sig[SPHINCS+-SHA256-256f-simple] - AssertionError: assert 'bd88b4945316...ae1370ed13480' == '38206d5db048...e76d4ce9fca23'
```

--> It seems the KATs introduced with this PR are not all correct. Thus, I'm afraid we cannot use this Sphincs+ update in the IETF hackathon. Agreed, @dstebila @xvzcf ? Good ideas how to facilitate that regardless very welcome.

baentsch:
> @baentsch @dstebila I think at least part of the problem lies with our OpenSSL SHA2 code. In the [mb-remove-avx2-from-sphincs](https://github.com/open-quantum-safe/liboqs/tree/mb-remove-avx2-from-sphincs) branch, compare the `focal-std-openssl` job triggered by [my commit](https://github.com/open-quantum-safe/liboqs/commit/723d7ababd8d794b7dd8c4c2f00d2bbb49974e05) (the job transcript can be found [here](https://github.com/open-quantum-safe/liboqs/actions/runs/4855089057/jobs/8653279229)) and [the job from the previous commit](https://github.com/open-quantum-safe/liboqs/actions/runs/4465669616/jobs/7842995392).

Interesting, indeed: Just by setting OQS_OPENSSL_OFF the test passes. It looks like our self tests then are also insufficient/not equivalent to the use of the API in `liboqs`. Not good.

baentsch:
I now started to look a bit into this issue and concluded that the use of the `OQS_SHA2_sha512_inc` API causes the problem, right? When trying to understand the API, I wonder why [these comments](https://github.com/open-quantum-safe/liboqs/blob/d704da0c5fd0e6d22d2a2f4212910cf383d8c345/src/common/sha2/sha2.h#L210C41-L211) are as they are (referencing SHA3 in SHA2 include comments)? Cut-copy-paste error? I didn't find similar ones in the code, but it might be worth while fixing this (@dstebila?) as part of this PR. Now keeping looking further...

xvzcf:
> I now started to look a bit into this issue and concluded that the use of the `OQS_SHA2_sha512_inc` API causes the problem, right?

I've been looking into it as well, and yes it's this API that seems to be changing the KAT values.

Martyrshot:
>I wonder why [these comments](https://github.com/open-quantum-safe/liboqs/blob/d704da0c5fd0e6d22d2a2f4212910cf383d8c345/src/common/sha2/sha2.h#L210C41-L211) are as they are (referencing SHA3 in SHA2 include comments)?

I believe that was a fat finger on my part. Oops! Sorry!

baentsch:
> > I wonder why [these comments](https://github.com/open-quantum-safe/liboqs/blob/d704da0c5fd0e6d22d2a2f4212910cf383d8c345/src/common/sha2/sha2.h#L210C41-L211) are as they are (referencing SHA3 in SHA2 include comments)?
> 
> I believe that was a fat finger on my part. Oops! Sorry!

Good to know :) Will you do a PR fixing this? 

Next question: Am I reading correctly that our code is double-freeing MD contexts: https://github.com/open-quantum-safe/liboqs/blob/d704da0c5fd0e6d22d2a2f4212910cf383d8c345/src/common/sha2/sha2_ossl.c#L125-L136

(when `OQS_SHA2_sha512_inc_finalize` and `OQS_SHA2_sha512_inc_ctx_release` are called in succession)? Is this intentional? (see https://www.openssl.org/docs/man3.1/man3/EVP_MD_CTX_free.html:

> The EVP_MD_CTX_create() and EVP_MD_CTX_destroy() functions were renamed to EVP_MD_CTX_new() and EVP_MD_CTX_free() in OpenSSL 1.1.0, respectively.

baentsch:
The more I read this code, the less happy I am: _Many_ OpenSSL API calls are made without checking their return values. This is not good, considering for example

> Ignoring failure returns of EVP_DigestInit_ex(), EVP_DigestInit_ex2(), or EVP_DigestInit() can lead to undefined behavior on subsequent calls updating or finalizing the EVP_MD_CTX such as the EVP_DigestUpdate() or EVP_DigestFinal() functions. 

(from https://www.openssl.org/docs/man3.0/man3/EVP_DigestInit_ex.html).

Shall I do a PR for this or does anyone else volunteer?

dstebila:
> > Next question: Am I reading correctly that our code is double-freeing MD contexts:
> 
> Our API expects that you call one or the other of `_inc_finalize`, or `_inc_ctx_release`, but not both. Whether that is obeyed is a different question, of course.

I suppose we could put some null sets/checks (`state->ctx = null` on release, and `if state->ctx == null` before release) to mitigate any damage if they are called twice.

dstebila:
> The more I read this code, the less happy I am: _Many_ OpenSSL API calls are made without checking their return values. This is not good, considering for example
> 
> > Ignoring failure returns of EVP_DigestInit_ex(), EVP_DigestInit_ex2(), or EVP_DigestInit() can lead to undefined behavior on subsequent calls updating or finalizing the EVP_MD_CTX such as the EVP_DigestUpdate() or EVP_DigestFinal() functions.
> 
> (from https://www.openssl.org/docs/man3.0/man3/EVP_DigestInit_ex.html).
> 
> Shall I do a PR for this or does anyone else volunteer?

Since most of our hashing API has no return value (void), would you then be putting in something similar to `OQS_EXIT_IF_NULL`?

baentsch:
And another logical question: Why is SHA2_BLOCK_SIZE defined as 64 in the OpenSSL wrapper (https://github.com/open-quantum-safe/liboqs/blob/d704da0c5fd0e6d22d2a2f4212910cf383d8c345/src/common/sha2/sha2_ossl.c#L45 and https://github.com/open-quantum-safe/liboqs/blob/d704da0c5fd0e6d22d2a2f4212910cf383d8c345/src/common/sha2/sha2_ossl.c#L121-L123) while 128 bytes seem to be the length of a block used in the C implementation: https://github.com/open-quantum-safe/liboqs/blob/d704da0c5fd0e6d22d2a2f4212910cf383d8c345/src/common/sha2/sha2_c.c#L624-L625 Or am I misunderstanding the C code here (didn't dive into it)?

baentsch:
The last question indeed resolves the problem with the KATs: This diff resolves the issue:

```
diff --git a/src/common/sha2/sha2_ossl.c b/src/common/sha2/sha2_ossl.c
index b8e71cf9..e5b69e0d 100644
-       EVP_DigestUpdate((EVP_MD_CTX *) state->ctx, in, inblocks * SHA2_BLOCK_SIZE);
+       EVP_DigestUpdate((EVP_MD_CTX *) state->ctx, in, inblocks * 2 * SHA2_BLOCK_SIZE);
```

dstebila:
> The last question indeed resolves the problem with the KATs: This diff resolves the issue:
> 
> ```
> diff --git a/src/common/sha2/sha2_ossl.c b/src/common/sha2/sha2_ossl.c
> index b8e71cf9..e5b69e0d 100644
> -       EVP_DigestUpdate((EVP_MD_CTX *) state->ctx, in, inblocks * SHA2_BLOCK_SIZE);
> +       EVP_DigestUpdate((EVP_MD_CTX *) state->ctx, in, inblocks * 2 * SHA2_BLOCK_SIZE);
> ```

Hold on, isn't the SHA384 block size also 128 bytes?

dstebila:
Before we get too excited, about declaring this resolved, I think we should also strengthen the OQS test programs to detect this flaw.

baentsch:
> we should also strengthen the OQS test programs to detect this flaw.

Most definitely. And strengthen the way we use the OpenSSL APIs. Let's talk about this... now :)

baentsch:
> would you then be putting in something similar to OQS_EXIT_IF_NULL

Yes.

dstebila:
> > would you then be putting in something similar to OQS_EXIT_IF_NULL
> 
> Yes.

Okay, makes sense.

docs/algorithms/sig/sphincs.yml
```diff
@@ -141,7 +142,7 @@ parameter-sets:
     no-secret-dependent-branching-checked-by-valgrind: true
     large-stack-usage: false
 - name: SPHINCS+-Haraka-192f-robust
-  claimed-nist-level: 3
+  claimed-nist-level: 2
```
baentsch, line 145: @dstebila Why is there still a reference to a Haraka version? And why did the claimed NIST level change? Would it be more appropriate to remove all references to all no-longer supported SPHINCS+ variants from this YML file? FWIW, I did the same for Kyber90s and DilithiumAES as `copy_from_upstream` did not have the automation for this.

dstebila: I didn't realize that copy_from_upstream wouldn't clean that all up.  I'll make a TODO to search for references to the removed algorithms.

xvzcf:
> In terms of sequencing here, how should we proceed?

How about:

1. Merge https://github.com/open-quantum-safe/liboqs/pull/1454
2. Start a fresh branch that adds just the clean variants and merge that in
3. Add the AVX2 variants later

> Should we change our naming as well? This would have a lot of downstream effects.

I think we should.

baentsch:
> > In terms of sequencing here, how should we proceed?
> 
> How about:
> 
>     1. Merge [Fixing OpenSSL SHA2 incremental API integration #1454](https://github.com/open-quantum-safe/liboqs/pull/1454)
> 
>     2. Start a fresh branch that adds just the clean variants and merge that in
> 
>     3. Add the AVX2 variants later
> 
Fine with me.
> 
> > Should we change our naming as well? This would have a lot of downstream effects.
> 
> I think we should.

Name changes in `liboqs` will cause headaches -- but I'm afraid using a naming different to the one used by the authors will cause even more headaches in the long run. We already now have the big issue how to name/designate this variant of Sphincs. I'm already scratching my head how to broach this with the IETF-Interop team... There, new names actually may be beneficial to show this is a new algorithm variant.

baentsch:
> 3. Add the AVX2 variants later

Second thought on this: Is this _really_ necessary? It seems the AVX2 code does pass on some platforms (see CI). So what about simply disabling it for the time being on those where it doesn't pass? Or fix the problems? ;-)

docs/algorithms/sig/sphincs.yml
baentsch: > I didn't realize that copy_from_upstream wouldn't clean that all up. 

I didn't either when doing https://github.com/open-quantum-safe/liboqs/pull/1455. All it needed was removal of the no longer needed algs from https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/sig/sphincs.yml (in this case) and re-running `copy_from_upstream`.

xvzcf:
@baentsch to avoid duplicating work I thought I'd note here I'm going to work on:

1. Merging in the clean variants of SPHINCS into liboqs in with the name changes
2. Fixing the downstream breakages
3. Adding in the AVX2 versions to liboqs

dstebila:
> @baentsch to avoid duplicating work I thought I'd note here I'm going to work on:
> 
>     1. Merging in the clean variants of SPHINCS into liboqs in with the name changes
> 
>     2. Fixing the downstream breakages
> 
>     3. Adding in the AVX2 versions to liboqs

Well I just read #1460 and I think I understand what's going on there.  And between #1460 and #1420, that would accomplish everything on your list except for the algorithm name changes.  So maybe it would save work to follow Michael's plan of merging #1460 into #1420 and then #1420 into main?  The only question would then be whether we do the SPHINCS+ name changes before merging #1420 into main or afterwards.  In the interests of keeping the commits smaller (already #1460+1420 will be pretty big) I'd say we do the name changes after merging 1420.

baentsch:
> The only question would then be whether we do the SPHINCS+ name changes before merging https://github.com/open-quantum-safe/liboqs/pull/1420 into main or afterwards. In the interests of keeping the commits smaller (already https://github.com/open-quantum-safe/liboqs/pull/1460+1420 will be pretty big) I'd say we do the name changes after merging 1420.

Good for me (big surprise :). This will allow https://github.com/open-quantum-safe/oqs-provider/pull/158 to run right away (doing a full downstream test).

xvzcf:
Sure, sounds good, on it.

dstebila:
> > The only question would then be whether we do the SPHINCS+ name changes before merging #1420 into main or afterwards. In the interests of keeping the commits smaller (already #1460 will be pretty big) I'd say we do the name changes after merging 1420.
> 
> Good for me (big surprise :). This will allow [open-quantum-safe/oqs-provider#158](https://github.com/open-quantum-safe/oqs-provider/pull/158) to run right away (doing a full downstream test).

@xvzcf how about you? I know that's a little different than what we talked about earlier (and what you wrote in https://github.com/open-quantum-safe/liboqs/pull/1420#issuecomment-1548121960) before we read #1460.

xvzcf:
If I can get a once over and 1-2 approvals on this it should be good to merge

baentsch (approved):

baentsch:
> If I can get a once over and 1-2 approvals on this it should be good to merge

Done. But it doesn't feel quite right that I give an approval on this as all changes done to get this to "green" were done by me in #1460. But if that got a good review, I think we're good to move forward.

baentsch:
Thanks for this merge, @xvzcf . Are you going to look into the downstreams now? For oqs-openssl111 it should be as simple as copying over generate.yml from oqs-provider (if https://github.com/open-quantum-safe/oqs-provider/pull/158 still passes OK after the #1420 merge).

xvzcf:
> Thanks for this merge, @xvzcf . Are you going to look into the downstreams now? For oqs-openssl111 it should be as simple as copying over generate.yml from oqs-provider (if [open-quantum-safe/oqs-provider#158](https://github.com/open-quantum-safe/oqs-provider/pull/158) still passes OK after the #1420 merge).

I'm working on it, I'm updating the names in liboqs first and then I'll update the downstreams

PR #1423, opened 2023-03-20
patching SPHINCS again

baentsch:
This may be a better approach to correcting the AVX2 integration problem wrt common code: This adds a patch again removing the project-specific `fips202x4.h/c` files.

HOWEVER, changes were necessary to the common code declarations to permit this to compile: See src/common/pqclean_shims/fips202x4.h below.

This clearly is not quite right given that the tests don't pass -- but that makes me wonder: Is the common code implementation of `KeccakP1600times4_PermuteAll_24rounds` possibly not quite what the new Sphincs+ code base expects/has built-in? If @jschanck would have time to glance this over, this would be great, given that his PR https://github.com/open-quantum-safe/liboqs/pull/948 did things correctly the last time.


thomwiggers:
@bwesterb changed how SPHINCS+ works by directly integrating the Keccak permutation in a few more places than just `fips202x4`. This means that the way that liboqs has offered its shims to SPHINCS+ is no longer compatible. 

I had a lot of trouble getting this to work in my quick-and-dirty integration of SPHINCS+ [here](https://github.com/thomwiggers/liboqs/tree/dirty-oqs-thesis-benchmarks). This only compiles with very specific flags to CMake, as the namespacing of `KeccakP1600` also seems to change depending on them.

thomwiggers:
For example, note this line https://github.com/open-quantum-safe/liboqs/blob/ds-update-pqclean-20230317/src/sig/sphincs/pqclean_sphincs-shake-192f-simple_avx2/hash_shakex4.c#L10

baentsch:
Thanks for the heads-up. Will (need to) leave this to @dstebila the next few weeks.

PR #1430, opened 2023-04-06
Initial commit of Cisco's open source LMS

### LINKED ISSUE #1098: Add stateful hash-based signatures
[LMS](https://datatracker.ietf.org/doc/html/rfc8554) and [XMSS](https://datatracker.ietf.org/doc/html/rfc8391) as standardized by [CFRG](https://irtf.org/cfrg) may be of use to Verisign and others.

ashman-p:
This is the initial commit. The following commits should have more weighty code additions.

Fixes #1098.

* [Y] Does this PR change the the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in [oqs-provider](https://github.com/open-quantum-safe/oqs-provider), [OQS-OpenSSL](https://github.com/open-quantum-safe/openssl), [OQS-BoringSSL](https://github.com/open-quantum-safe/boringssl), and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)


baentsch (changes requested):
> This is the initial commit. The following commits should have more weighty code additions.

I'm not sure I like this approach: Are you saying you want to keep adding to a PR during development? This will generate _loads_ of unnecessary notification emails to the whole team (which in turn will lead to all emails about your work to be disregarded); it further isn't particularly helpful (e.g., for auto-filtering out those emails) that this PR is not marked Draft. 

My preference would be for a PR to be created in OQS only when it's basically done -- and to be marked Ready for Review only when its CI passes.

ashman-p:
No worries, i get it. I most likely misunderstood the discussion concerning the initial commit of the LMS source and preferences on code sizes for reviews.

Issue #1427, opened 2023-04-04
Improve EVP_sha calls when using OpenSSL3


baentsch:
I'm aware of the following antipattern - in 3.0. using old methods like EVP_sha3_256() instead of fetching a digest object via EVP_MD_fetch and using a fetched object gives a performance penalty. See https://github.com/openssl/openssl/pull/20354/files#diff-86fbd44f96c9f78b9c4a27d8b5fb9d83aa787b9a7d92c6774d9172c5952b791f for example. If it is the case (and looks like it is), it's worth fixing

_Originally posted by @beldmit in https://github.com/open-quantum-safe/liboqs/issues/1426#issuecomment-1494591453_

baentsch:
Reference to AES added as per https://github.com/open-quantum-safe/liboqs/issues/1428#issuecomment-1496135182

baentsch:
@beldmit Are you aware of how much performance this costs? If I get it right, it would mean adding "state" to (re)using Digests and AES (to stop OpenSSL re-searching and re-creating the corresponding EVP objects), right? Somewhat along the lines of what we [already do for XKCP](https://github.com/open-quantum-safe/liboqs/blob/d61d81c526da8bb62e363f5a75191689572151cb/src/common/sha3/sha3.h#L41-L45).

beldmit:
Right,it just eliminates search/recreate. Didn't measure it myself, but it's a relatively expensive operation

baentsch:
> it's a relatively expensive operation

I can't help it, but the scientist in me wants numbers. So here we go:

![grafik](https://user-images.githubusercontent.com/57787676/230383429-1c82a683-43fb-4250-a6fd-5fe6e3637c85.png)

If this is only due to the difference in the way algorithms are fetched, this indeed is a bummer (aka an issue worth while tackling).

beldmit:
Speaking frankly, I'd run a test under `valgrind --callgrind` in some more or less realistic scenario.

baentsch:
> run a test under valgrind --callgrind in some more or less realistic scenario.

ACK. Just wanted to quickly get beyond 

> Didn't measure it 

Contributions very welcome. Or are you indeed not at liberty to contribute?

beldmit:
I don't have any problems with contribution beyond the capacity ones :).

beldmit:
Is there a initializing and deinitializing function to place fetching in?

baentsch:
[...]

PR #1431, opened 2023-04-07
Strawman version of one-time fetching MD objects from OpenSSL

beldmit:
We need init them and free them in one place to avoid threading issues.

### DIFF
--- a/src/common/aes/aes_ossl.c
+++ b/src/common/aes/aes_ossl.c
@@ -9,6 +9,36 @@
 
 #include <openssl/evp.h>
 
+EVP_CIPHER *_aes_128_ecb() {
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+	static EVP_CIPHER *ptr;
+	ptr = ptr ? ptr : EVP_CIPHER_fetch(NULL, "AES-128-ECB", NULL);
+	return ptr;
+#else
+	return EVP_aes_128_ecb();
+#endif
+}
+
+EVP_CIPHER *_aes_256_ecb() {
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+	static EVP_CIPHER *ptr;
+	ptr = ptr ? ptr : EVP_CIPHER_fetch(NULL, "AES-256-ECB", NULL);
+	return ptr;
+#else
+	return EVP_aes_256_ecb();
+#endif
+}
+
+EVP_CIPHER *_aes_256_ctr() {
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+	static EVP_CIPHER *ptr;
+	ptr = ptr ? ptr : EVP_CIPHER_fetch(NULL, "AES-256-CTR", NULL);
+	return ptr;
+#else
+	return EVP_aes_256_ctr();
+#endif
+}
+
 struct key_schedule {
 	int for_ECB;
 	EVP_CIPHER_CTX *ctx;
@@ -34,7 +64,7 @@ void OQS_AES128_ECB_load_schedule(const uint8_t *key, void **schedule) {
 	ks->for_ECB = 1;
 	ks->ctx = EVP_CIPHER_CTX_new();
 	OQS_EXIT_IF_NULLPTR(ks->ctx);
-	OQS_OPENSSL_GUARD(EVP_EncryptInit_ex(ks->ctx, EVP_aes_128_ecb(), NULL, key, NULL));
+	OQS_OPENSSL_GUARD(EVP_EncryptInit_ex(ks->ctx, _aes_128_ecb(), NULL, key, NULL));
 	EVP_CIPHER_CTX_set_padding(ks->ctx, 0);
 }
 
@@ -73,7 +103,7 @@ void OQS_AES256_ECB_load_schedule(const uint8_t *key, void **schedule) {
 	ks->for_ECB = 1;
 	ks->ctx = EVP_CIPHER_CTX_new();
 	OQS_EXIT_IF_NULLPTR(ks->ctx);
-	OQS_OPENSSL_GUARD(EVP_EncryptInit_ex(ks->ctx, EVP_aes_256_ecb(), NULL, key, NULL));
+	OQS_OPENSSL_GUARD(EVP_EncryptInit_ex(ks->ctx, _aes_256_ecb(), NULL, key, NULL));
 	EVP_CIPHER_CTX_set_padding(ks->ctx, 0);
 }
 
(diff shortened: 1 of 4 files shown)


beldmit:
Related: #1427

baentsch (commented):
> We need init them and free them in one place to avoid threading issues. 

What about using https://github.com/open-quantum-safe/liboqs/blob/b1d42d61f63aa61ce007ada7939e326e0d6e896c/src/common/common.c#L197 for the former and adding a function for the latter? Probably easier than the alternative of adding `init` and `close` to all OQS operations.

beldmit:
Thanks! It's quite reasonable

baentsch:
> Thanks! It's quite reasonable

But then again -- is this _really_ sufficient? Are these EVP objects thread-safe? What if someone calls in different threads different OpenSSL APIs, all reaching into `oqsprovider` which in turn calls back into OpenSSL in this way?

[The documentation](https://www.openssl.org/docs/manmaster/man7/crypto.html) states

> However most OpenSSL data structures are not thread-safe.

beldmit:
EVP_MD and EVP_CIPHER objects should be thread-safe - they are just sets of callbacks and flags. The problem may happen if the provider providing them is unloaded.

baentsch:
> The problem may happen if the provider providing them is unloaded.

Good. I don't see how this should be facilitated by an application that is using `liboqs` directly: `liboqs` doesn't make available (doesn't know about :) `OSSL_PROVIDER` or `OSSL_LIB_CTX`, so how could an unload happen?

If `oqsprovider` in turn triggers usage of `liboqs` what about checking there that a suitable provider for SHA/AES is loaded/available (in the same `OSSL_LIB_CTX`)? But yikes, then the proposed implementation passing `NULL` to the `EVP_fetch` routines doesn't cut it.... :-(

beldmit:
I groomed the code according my understanding of how it goes.
The tests work for me. 

Consequence: `OQS_init` call becomes mandatory

beldmit:
At this point I'm waiting for your feedback whether we should follow this direction as a whole or not

baentsch (commented):
> At this point I'm waiting for your feedback whether we should follow this direction as a whole or not

As far as I'm concerned, it does. The one thing not looking good (also as per CI) is that the ossl_helper functions should be namespaced: Proposal would be "oqs_ossl_xyz" (instead of just "_xyz")

beldmit:
I will adjust namespace, but also init and free functions must be documented as mandatory...

src/common/ossl_helpers.c
```diff
@@ -0,0 +1,118 @@
+// SPDX-License-Identifier: MIT
+#include <oqs/oqs.h>
+#if defined(OQS_USE_OPENSSL)
+#include <openssl/evp.h>
+#include "ossl_helpers.h"
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+static EVP_MD *sha256_ptr, *sha384_ptr, *sha512_ptr,
```
baentsch, line 7: Would it be sensible to initialize them all to NULL....

src/common/ossl_helpers.c
```diff
@@ -0,0 +1,118 @@
...
+
+	aes128_ecb_ptr = EVP_CIPHER_fetch(NULL, "AES-128-ECB", NULL);
+	aes256_ecb_ptr = EVP_CIPHER_fetch(NULL, "AES-256-ECB", NULL);
+	aes256_ctr_ptr = EVP_CIPHER_fetch(NULL, "AES-256-CTR", NULL);
+#endif
+}
+
+const EVP_MD *oqs_sha256(void) {
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+	return sha256_ptr;
```
baentsch, line 33: ...and then check here whether they are NULL here (meaning OQS_init hasn't been called)? Or may even _then_ "automatically" call `OQS_init`?

src/common/ossl_helpers.c
```diff
@@ -0,0 +1,118 @@
...
+#else
+	return EVP_sha256();
+#endif
+}
+
+const EVP_MD *oqs_sha384(void) {
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+	return sha384_ptr;
```
baentsch, line 41: ...same thought here and at all of the functions below.

baentsch (approved):
Thank you very much again for your contribution: Very much appreciated! Don't feel stopped to address/comment on the single comments I just made -- but if you have more urgent things to do, we'll follow up accordingly.

src/common/ossl_helpers.c
beldmit: `static` keyword does it for us

src/common/ossl_helpers.c
beldmit: First, in theory, they still may be NULL even if OQS_init is called - so checking OpenSSL DigestInit result is a good idea.

Second and worse, we will get a threading problem here.

beldmit:
I'm going to add OQS_destroy function and call it in the tests. I'd like to finalize this issue if we have an agreement what should be done

src/common/ossl_helpers.c
baentsch: > Second and worse, we will get a threading problem here.

ACK. Comment withdrawn.

baentsch:
> I'm going to add OQS_destroy function and call it in the tests. I'd like to finalize this issue if we have an agreement what should be done

Good for me -- provided the comment [here](https://github.com/open-quantum-safe/liboqs/blob/b1d42d61f63aa61ce007ada7939e326e0d6e896c/src/common/common.h#L140-L145), gets updated (making the invocation of `OQS_init` a requirement for correct use of the library). Thanks again.

beldmit:
So. I finished what I understood was worth doing. Could you please specify what are the next steps to deal with it? Is there anything to be done by me just now or it can be merged as is and we could move forward?

src/common/common.h
```diff
@@ -138,11 +138,16 @@ typedef enum {
 OQS_API int OQS_CPU_has_extension(OQS_CPU_EXT ext);
 
 /**
- * This currently only sets the values in the OQS_CPU_EXTENSIONS,
- * and so has effect only when OQS_DIST_BUILD is set.
+ * This currently sets the values in the OQS_CPU_EXTENSIONS
+ * and prefetches the OpenSSL obejcts if necessary.
```
baentsch, line 142: nit: typo. Can deal with when doing next doc review. Would also recommend phrasing more strongly along the lines of "Needs to be invoked to ensure proper operation of library."

baentsch:
> Is there anything to be done by me just now 

Not really. Thanks again for the contribution.

> or it can be merged as is and we could move forward?

Yes from my perspective but I leave the merge to another team member (@dstebila : Back again?)

sahanaprasad07:
Hi @beldmit , would it make sense to replace EVP_aes_256_ecb() in src/common/rand/rand_nist.c?

beldmit:
Probably yes, if this code is used multiple times.

src/common/common.h
dstebila, line 142:
```suggestion
 * and prefetches the OpenSSL objects if necessary.
```

dstebila (approved):

dstebila:
> > Hi @beldmit , would it make sense to replace EVP_aes_256_ecb() in src/common/rand/rand_nist.c?
> 
> Probably yes, if this code is used multiple times.

Okay, done in 87780d2. Will merge once CI passes.

PR #1447, opened 2023-05-02
src/common/common.c (set_available_cpu_extensions): Cope without `HWCAP_SHA3'

distorted-mdw:
Introduced in Linux 4.15, which, I admit, was a while ago.


baentsch (approved):
Enhances "compilability" on old Linux's, so LGTM. Thanks for the contribution.

dstebila (approved):

PR #1448, opened 2023-05-02
src/kem/kyber/.../*.S, src/sig/dilithium/.../*.S: Return using `ret'

distorted-mdw:
Partly, `lr` isn't a register alias under older versions of GNU `as`. But, mostly, `br` doesn't mean the same thing to the branch predictor, so it will cause unnecessary misprediction stalls.

### DIFF
--- a/src/kem/kyber/pqclean_kyber1024_aarch64/__asm_NTT.S
+++ b/src/kem/kyber/pqclean_kyber1024_aarch64/__asm_NTT.S
@@ -167,7 +167,7 @@ _PQCLEAN_KYBER1024_AARCH64_asm_ntt_SIMD_top:
     .unreq    counter
     pop_all
 
-    br lr
+    ret
 
 
 .align 2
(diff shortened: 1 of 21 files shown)


baentsch (commented):
Just a comment regarding the placement of this PR: The code you are suggesting to change is automatically imported to `liboqs` from [PQClean](https://github.com/PQClean/PQClean), so it would be more suitable to propose the code changes there instead of in this project.

vincentvbh:
This should be integrated into https://github.com/PQClean/PQClean/pull/488 since the code base is the same.

distorted-mdw:
Yes.  The reason I hadn't marked it for review is that I haven't yet prepared the MR for PQClean.

baentsch:
@distorted-mdw Do you still plan on doing a PR to PQClean for this? Our latest update from PQClean (#1512) didn't pull this in....

baentsch:
Closing due to lack of feedback & stale-ness.

PR #1449, opened 2023-05-02
src/common/pqclean_shims/compat.h: Fix polyfill version check sense

distorted-mdw:
We need to define the types if GCC is too /old/.

### DIFF
--- a/src/common/pqclean_shims/compat.h
+++ b/src/common/pqclean_shims/compat.h
@@ -22,10 +22,10 @@
 
 #if defined(__GNUC__) && !defined(__clang__)
 
-#if ((__GNUC__ << 16) + __GNUC_MINOR__ >= ((7) << 16) + (1)) // at least GCC 7.1
+#if ((__GNUC__ << 16) + __GNUC_MINOR__ < ((7) << 16) + (1)) // at least GCC 7.1
 /* Versions of the GCC pre-7.1 don't have __m256*_u types */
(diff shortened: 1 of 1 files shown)


baentsch (commented):
Thanks for the contribution. Is (possibly not correctly) supporting compilers this old really an issue, though? General question: Does `liboqs` really want to support compilers this old or rather add documentation and compiler version checks (for a certain minimum version) to minimize our efforts on outdated software? We also check for minimum `cmake` version, for example... Thoughts @dstebila @xvzcf @christianpaquin ?

Add: One of the [CI failures in the Sphincs+ update](https://github.com/open-quantum-safe/liboqs/actions/runs/4462022915/jobs/7836295816?pr=1420) are also related to a `gcc > 7.1` version check (failing: It seems McElice code _will not_ compile for gcc<7.1).

dstebila:
I don't have a particular minimum gcc target in mind, but am agreeable to raising it unless we hear a strong demand and commitment to support.

baentsch:
As agreed in our call, we rather prefer documenting and enforcing via CMakeLists.txt a minimum compiler version of 7.1, thus making this PR moot.

@distorted-mdw Please let us know if you know of a use case (for `liboqs`) where an older compiler would need to be supported.

baentsch:
As discussed above, this PR is becoming superfluous when #1451 is merged and will be closed then.

Issue #1452, opened 2023-05-09
Bug in OSX integration of OpenSSL


mouse07410:
**Describe the bug**
On a newly-prepared platform , I'm getting a ton of errors trying to run test-suite:
```text
.  .  .
tests/test_speed.py:21: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

command = ['/Users/ur20980/src/liboqs/build/tests/speed_sig', 'SPHINCS+-SHAKE256-256s-robust', '-f']
working_dir = '.'
env = {'ANT_OPTS': '-Dhttp.proxyHost=llproxy.llan.ll.mit.edu -Dhttp.proxyPort=8080', 'AS': 'yasm -m amd64 -f macho64 -p gas -X gnu ', 'AS_INTEGRATED_ASSEMBLER': '1', 'BASH_SILENCE_DEPRECATION_WARNING': '1', ...}
expected_returncode = 0, input = None, ignore_returncode = False

    def run_subprocess(command, working_dir='.', env=None, expected_returncode=0, input=None, ignore_returncode=False):
        """
        Helper function to run a shell command and report success/failure
        depending on the exit status of the shell command.
        """
        env_ = os.environ.copy()
        if env is not None:
            env_.update(env)
        env = env_
    
        # Note we need to capture stdout/stderr from the subprocess,
        # then print it, which pytest will then capture and
        # buffer appropriately
        print(working_dir + " > " + " ".join(command))
    
        result = subprocess.run(
                command,
                input=input,
                stdout=subprocess.PIPE,
                stderr=subprocess.STDOUT,
                cwd=working_dir,
                env=env,
            )
    
        if not(ignore_returncode) and (result.returncode != expected_returncode):
            print(result.stdout.decode('utf-8'))
>           assert False, "Got unexpected return code {}".format(result.returncode)
E           AssertionError: Got unexpected return code -4

tests/helpers.py:42: AssertionError
-------------------------------------------- Captured stdout call --------------------------------------------
. > /Users/ur20980/src/liboqs/build/tests/speed_sig SPHINCS+-SHAKE256-256s-robust -f

__________________________________ test_sig[SPHINCS+-SHAKE256-256f-simple] ___________________________________
[...]

PR #1453, opened 2023-05-11
disable OpenSSL if neither AES,SHA2, or SHA3 are OSSL-provided

baentsch:
Fix to https://github.com/open-quantum-safe/liboqs/issues/1452#issuecomment-1542743269: No OpenSSL logic must be available if none of the "piece part" OQS_OPENSSL-enablement defines are set.

### DIFF
--- a/.CMake/alg_support.cmake
+++ b/.CMake/alg_support.cmake
@@ -66,6 +66,9 @@ cmake_dependent_option(OQS_USE_SHA2_OPENSSL "" ON "OQS_USE_OPENSSL" OFF)
 # enough to support our incremental API.
 cmake_dependent_option(OQS_USE_SHA3_OPENSSL "" OFF "OQS_USE_OPENSSL" OFF)
 
+# sanity check: Disable OpenSSL if not a single OpenSSL component define is on
+cmake_dependent_option(OQS_USE_OPENSSL "" ON "OQS_USE_AES_OPENSSL OR OQS_USE_SHA2_OPENSSL OR OQS_USE_SHA3_OPENSSL" OFF)
+
 if(CMAKE_SYSTEM_NAME MATCHES "Linux|Darwin")
 if(OQS_DIST_X86_64_BUILD OR OQS_USE_AVX2_INSTRUCTIONS)
     cmake_dependent_option(OQS_ENABLE_SHA3_xkcp_low_avx2 "" ON "NOT OQS_USE_SHA3_OPENSSL" OFF)


.CMake/alg_support.cmake
```diff
@@ -66,6 +66,9 @@ cmake_dependent_option(OQS_USE_SHA2_OPENSSL "" ON "OQS_USE_OPENSSL" OFF)
 # enough to support our incremental API.
 cmake_dependent_option(OQS_USE_SHA3_OPENSSL "" OFF "OQS_USE_OPENSSL" OFF)
```
mouse07410, line 67: I am no certain about this line: by default `OQS_USE_SHA3_OPENSSL` is `OFF`, **and** if `OQS_USE_OPENSSL=ON` **also** turns it `OFF`?

baentsch: Hmm .. somewhat surprising instruction, indeed. I presume this had been `ON` until the incremental API got added and @jschanck simply toggled the original "ON" when adding the comment above. I'm inclined to let it remain as-is as a reminder that we may want to improve this as OpenSSL's APIs change to make the comment incorrect (i.e., to allow use of OpenSSL for SHA3_inc operations).

However, given that the incremental API _is_ implemented also for SHA3 and OpenSSL, maybe @jschanck could state more explicitly here what he actually meant by the comment (and whether my assumption above is correct)?

baentsch: > > I'm inclined to let it remain as-is  .  .  .
> 
> Leaving it as-is completely prevents use of OpenSSL SHA3 - there'd be no way to enable it. Is that what you want?

That might have been the goal of @jschanck :) but it definitely isn't the case: Passing `-DOQS_USE_SHA3_OPENSSL=ON` to `cmake` enables `liboqs` to use OpenSSLs SHA3:

```
$ ./tests/test_sha3 
Configuration info
==================
Target platform:  x86_64-Linux-5.4.0-146-generic
Compiler:         gcc (7.5.0)
OQS version:      0.8.0-dev
Git commit:       d704da0c5fd0e6d22d2a2f4212910cf383d8c345
OpenSSL enabled:  Yes (OpenSSL 1.1.1  11 Sep 2018)
AES:              NI
SHA-2:            OpenSSL
SHA-3:            OpenSSL
CPU exts active:  AES AVX PCLMULQDQ POPCNT SSE SSE2 SSE3
Success! passed sha3-256 known answer tests 
Success! passed sha3-512 known answer tests 
Success! passed shake-128 known answer tests 
Success! passed shake-256 known answer tests 
Success! passed four-way parallel shake-128 known answer tests 
Success! passed four-way parallel shake-256 known answer tests 

```

dstebila (approved):

.CMake/alg_support.cmake
mouse07410: > but it definitely isn't the case: Passing `-DOQS_USE_SHA3_OPENSSL=ON` to `cmake` enables `liboqs` to use OpenSSLs SHA3

Why doesn't that line disable `OQS_USE_SHA3_OPENSSL` when it sees that `OQS_USE_OPENSSL` is set?

mouse07410: >  is that this command after all is/declares/initializes an _option_, i.e., a variable that can be ultimately set on the command line

But this value _is_ an option - so, the only question is whether setting a value in the command line overrides what's the `CMakeLists.txt` content says. Your example seems to show that passing via CLI does override, but I still have my doubts.

PR #1463, opened 2023-05-16
relax OpenSSL initialization error handling

### LINKED ISSUE #1462: OpenSSL3 strict error checking causes downstream problems
https://github.com/open-quantum-safe/liboqs/pull/1454 introduced strict error checking e.g., here: https://github.com/open-quantum-safe/liboqs/blob/24c7f4f07896252fe82a042633cf1d9e893153e4/src/common/ossl_helpers.c#L17 

Under specific preconditions (not activating `default` provider before `oqsprovider`) this causes https://github.com/open-quantum-safe/oqs-provider/issues/159.

Completely disregarding errors (as in the original code version https://github.com/open-quantum-safe/liboqs/pull/1431) seems suboptimal, too as it would lead to a "surprising" performance degradation in `liboqs` (depending on which provider is listed first), right @beldmit ?

Any suggestions how to resolve this quagmire _very_ welcome. Without better proposals, I tend to change [oqs_fetch_ossl_objects](https://github.com/open-quantum-safe/liboqs/blob/24c7f4f07896252fe82a042633cf1d9e893153e4/src/common/ossl_helpers.c#LL14C6-L14C28) to collect error messages and just output a warning if any have been encountered -- but not exit.

baentsch:
Fixes #1462

### DIFF
--- a/src/common/ossl_helpers.c
+++ b/src/common/ossl_helpers.c
@@ -14,29 +14,23 @@ static EVP_CIPHER *aes128_ecb_ptr, *aes256_ecb_ptr, *aes256_ctr_ptr;
 void oqs_fetch_ossl_objects(void) {
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
 	sha256_ptr = EVP_MD_fetch(NULL, "SHA256", NULL);
-	OQS_EXIT_IF_NULLPTR(sha256_ptr, "OpenSSL");
 	sha384_ptr = EVP_MD_fetch(NULL, "SHA384", NULL);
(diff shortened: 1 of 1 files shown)


beldmit:
I like this approach but probably it's still worth grouping sha2, sha3, and aes, instead of everything

baentsch:
> I like this approach but probably it's still worth grouping sha2, sha3, and aes, instead of everything

I thought about this -- but this generates simply 3 error lines instead of 1 -- with no additional value: It's simply the default provider missing after all....

dstebila (approved):

paulidale:
Nice solution.

PR #1469, opened 2023-05-17
once OSSL init

baentsch:
Straw man proposal for improving #1463.

### DIFF
--- a/src/common/ossl_helpers.c
+++ b/src/common/ossl_helpers.c
@@ -11,7 +11,9 @@ static EVP_MD *sha256_ptr, *sha384_ptr, *sha512_ptr,
 static EVP_CIPHER *aes128_ecb_ptr, *aes256_ecb_ptr, *aes256_ctr_ptr;
 #endif
 
-void oqs_fetch_ossl_objects(void) {
+CRYPTO_ONCE OQS_ONCE_STATIC_INIT;
+
+static void oqs_fetch_ossl_objects(void) {
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
 	sha256_ptr = EVP_MD_fetch(NULL, "SHA256", NULL);
 	sha384_ptr = EVP_MD_fetch(NULL, "SHA384", NULL);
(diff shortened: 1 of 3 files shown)


src/common/ossl_helpers.c
```diff
@@ -53,6 +55,11 @@ void oqs_free_ossl_objects(void) {
 
 const EVP_MD *oqs_sha256(void) {
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
+	if (!sha256_ptr) {
```
t8m, line 58: Strictly speaking this depends on access to the pointer being atomic. Other than that this should work fine.

baentsch: > Strictly speaking this depends on writes to the pointer being atomic. 

There's only a single function where this (or any of these) pointer is written: https://github.com/open-quantum-safe/liboqs/blob/e7bc261841a459631e6a389b0377e572b9e63b78/src/common/ossl_helpers.c#L16-L30 

And _that_ should be guarded via/accessible by only one thread via `CRYPTO_THREAD_run_once`, right?

So changing this to Ready For Review. @beldmit : As author of https://github.com/open-quantum-safe/liboqs/pull/1431 any further input from your side?

beldmit: Sorry for unclarity, I mean I wanted to avoid the synchronization/thread-safety stuff. The approach LGTM

t8m: > There's only a single function where this (or any of these) pointer is written:
> 
> https://github.com/open-quantum-safe/liboqs/blob/e7bc261841a459631e6a389b0377e572b9e63b78/src/common/ossl_helpers.c#L16-L30

If the pointer is written non-atomically the check could load some half-written pointer which is non-null but also invalid, the invalid (non-NULL) pointer would be then returned from the function.

In reality I do not think it can happen apart from some truly exotic hardware/platforms.

paulidale (commented):
This is a good solution IMO.
We rely on the run once heavily throughout OpenSSL.

PR #1459, opened 2023-05-15
Update McEliece (& Falcon)

baentsch:
Would fix #1314.

HOWEVER, this is meant to remain a Draft PR as it doesn't pull from PQClean "master" and only meant for answering this question of @praveksharma 

> I'm new to both PQClean and liboqs so I'm not sure what to make of this. Does this PR require to include these other implementations before it can be merged?

There were some not quite obvious quirks before the contents of the McEliece update in https://github.com/PQClean/PQClean/pull/469 could be integrated to `liboqs`: Please see the changes particularly in `docs/algorithms/kem/classic_mceliece.yml`, `scripts/copy_from_upstream/copy_from_upstream.py`, `scripts/copy_from_upstream/copy_from_upstream.yml` and the addition of `src/common/pqclean_shims/crypto_declassify.h`.

Particularly on the latter file, asking @dstebila as to whether we really need this -- and whether the addition of the SPDX line is appropriate (wasn't there): Shall this be done upstream before https://github.com/PQClean/PQClean/pull/469 is merged?

Final question: This inadvertently also changes Falcon -- and there, the KATs. Why is this so? Is the McEliece branch in PQClean older than https://github.com/PQClean/PQClean/pull/476 ? One more reason to not merge this. But at least CI will show us which impact merging https://github.com/PQClean/PQClean/pull/469 would have on `liboqs`.

* [Yes] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [No] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in [oqs-provider](https://github.com/open-quantum-safe/oqs-provider), [OQS-OpenSSL](https://github.com/open-quantum-safe/openssl), [OQS-BoringSSL](https://github.com/open-quantum-safe/boringssl), and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)

Issue #1437, opened 2023-04-17
CC0 license is an obstacle


beldmit:
Dear colleagues, 

we are planning to package liboqs for Fedora. We build liboqs with `-DOQS_ALGS_ENABLED=STD` to minimize support of non-standardized algorithms.

We found that the folders
```
./src/kem/kyber/pqclean_kyber*_aarch64
./src/sig/dilithium/pqclean_dilithium*_aarch64
./src/sig/sphincs/pqclean_sphincs*
```
contain CC0 license which is not acceptable for Fedora by default (and AFAIU, other Linux distributions).

I believe that for aarch64 we could fallback to a generic (slower) implementation in the compile time, but it's obviously not the best possible option.

Is it possible to get in contact with the authors of these implementation to get an agreement about double-licensing of these implementations? MIT, Apache2 or smth else is fine for our purpose.

Many thanks in advance!

dstebila:
I've raised issues with the SPHINCS+ and neon-ntt team who did the aarch64 implementations for Kyber and Dilithium.

baentsch:
@beldmit As this "big company license madness" (pardon the quip from someone believing in all benefits of open source) apparently isn't sth easily resolved what about this proposal: Would it make sense/resolve your issue to add a (`cmake`) build define that ensures only MIT code gets included in a build? Or do you(r downstream :) need _all code_ (incl. unbuild pieces) to be "license clean" from your perspective?

beldmit:
I'll ask my colleagues but I have a feeling it's acceptable.

baentsch:
@beldmit @simo5 Any update on the above? Would it help if we'd introduce an "OQS_BUILD_MIT_ONLY" flag, for example? Or more generically an "OQS_PERMITTED_LICENSES" string to be set to suit any taste?

dstebila:
With https://github.com/PQClean/PQClean/pull/488 landed in PQClean, once we update liboq, the aarch64 implementations of Kyber and Dilithium in liboqs should be acceptably licensed.  That just leaves SPHINCS+; I've pinged them again on https://github.com/sphincs/sphincsplus/issues/49.

beldmit:
> @beldmit @simo5 Any update on the above? Would it help if we'd introduce an "OQS_BUILD_MIT_ONLY" flag, for example? Or more generically an "OQS_PERMITTED_LICENSES" string to be set to suit any taste?

[...]

PR #1470, opened 2023-05-17
Update Classic McEliece

### LINKED ISSUE #1314: Upgrade Classic McEliece to NIST round 4 specification
Hi! Are there any plans to upgrade the Classic McEliece KEM to the [round 4 specification](https://classic.mceliece.org/nist.html)?

The Rust implementation `classic-mceliece-rust` is doing that here: https://github.com/Colfenor/classic-mceliece-rust/pull/33

praveksharma:
Fixes #1314 and takes into account discussion from #1459. 

Changes SPDX line for `src\common\pqclean_shims\crypto_declassify.h` from MIT to CC0.


baentsch (changes requested):
Thanks for taking this on, @praveksharma ! Basically LGTM; it would be better if the old "*_vec" and "_avx" source files also already can be removed with this PR, though -- so that @dstebila doesn't have to do another "cleanup PR" again (as I failed to do this with Kyber90s :-/ :-)

praveksharma:
@baentsch, I've made a new commit which removes the old vec/avx directories.

baentsch (approved):

baentsch:
Hmm -- @dstebila: do the remaining errors require a manual review of the [suppression files](https://github.com/open-quantum-safe/liboqs/tree/main/tests/constant_time/kem) or is it just a case of changed line numbers? If the former, who can do this? How long would that take? I wouldn't be particularly happy about holding 0.8.0 just for this. But having to do an 0.9.0 in a few days later because of the KAT changes also isn't appealing....

praveksharma:
The commit 8d16512ffe7ddae7895fd5c8d0115724a949ccb2 fixes the CI errors. 

The `scan_build` issue is fixed by patching the Classic McEliece source code before copying it from PQClean. `scan-build` was complaining about an error that could've occurred if a variable went outside the range the algorithm allows it go. An assert statement telling `scan-build` that this is not possible gets it to stop complaining.

The memory leak issue is fixed by building Classic McEliece with lower optimisation, `-01` instead of `-03`.

src/kem/classic_mceliece/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ if(OQS_ENABLE_KEM_classic_mceliece_348864_avx2)
...
     target_include_directories(classic_mceliece_348864_avx2 PRIVATE ${CMAKE_CURRENT_LIST_DIR}/pqclean_mceliece348864_avx2)
     target_include_directories(classic_mceliece_348864_avx2 PRIVATE ${PROJECT_SOURCE_DIR}/src/common/pqclean_shims)
-    target_compile_options(classic_mceliece_348864_avx2 PRIVATE  -mavx2  -mpopcnt )
+    target_compile_options(classic_mceliece_348864_avx2 PRIVATE  -mavx2  -mpopcnt -O1)
```
baentsch, line 22: This reduces compiler optimization under all circumstances -- not just for memory leak testing. Is this intentional/desirable?

tests/test_leaks.py
```diff
@@ -10,9 +10,9 @@
 def test_kem_leak(kem_name):
     if not(helpers.is_kem_enabled_by_name(kem_name)): pytest.skip('Not enabled')
     if sys.platform != "linux" or os.system("grep ubuntu /etc/os-release") != 0 or os.system("uname -a | grep x86_64") != 0: pytest.skip('Leak testing not supported on this platform')
-    if kem_name == "Classic-McEliece-8192128": pytest.skip("Classic-McEliece-8192128 known to fail memory leak testing")
+    # if kem_name == "Classic-McEliece-8192128": pytest.skip("Classic-McEliece-8192128 known to fail memory leak testing")
```
baentsch, line 13: Suggest to completely remove this (now commented) code as well as the corresponding statement at https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/kem/classic_mceliece.md#advisories

baentsch:
Thanks for these updates getting everything "green", @praveksharma ! 

> The memory leak issue is fixed by building Classic McEliece with lower optimisation, -01 instead of -03.

This IMO begs one change to https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/kem/classic_mceliece.md#advisories: Something along the lines "`clang` reports memory leaks for all algorithms in this family when compiled with an optimization level greater than '-O1', so care is advised using this algorithm at higher optimization levels".

baentsch (dismissed):
One more change request on this PR: The liboqs version number must be moved to "0.9.0-dev" in the toplevel `CMakeLists.txt`. Arguably also the "SOVERSION" (in `src/CMakeLists.txt`) -- but I'm not sure about that: Did we bump that when going to "0.8.0", @dstebila ?

src/kem/classic_mceliece/CMakeLists.txt
praveksharma: This is intended. Unless we reduce the compiler optimisation for all circumstances the leak would persist for `clang` builds outside of testing.

praveksharma:
d5d8ccd216207f34515f7157d0b03953a353da05 fixes the version number and the issues regarding the comment. That leaves us with deciding whether or not to bump the SOVERSION.

dstebila:
Let's increment the SOVERSION.

src/common/pqclean_shims/crypto_declassify.h
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: CC0
```
baentsch, line 1: Is CC0 really right/necessary here? This adds (complication) to #1437...

docs/algorithms/kem/classic_mceliece.md
```diff
@@ -4,38 +4,38 @@
...
+  - **Source**: https://github.com/PQClean/PQClean/commit/d742438e5c541958bfd58070cd8668d757d88e07
   - **Implementation license (SPDX-Identifier)**: Public domain
 - **Ancestors of primary source**:
-  - SUPERCOP-20191221 "vec" and "avx" implementations
+  - SUPERCOP-20221025 "clean" and "avx2" implementations
 
 ## Advisories
 
-- The implementation of the Classic-McEliece-8192128 parameter set is known to fail memory leak testing on x86-64.
+- Classic-McEliece-460896, Classic-McEliece-460896f, Classic-McEliece-6960119, and Classic-McEliece-6960119f parameter sets fail memory leak testing on x86-64 when building with ``clang`` using optimization level ``-O2`` and ``-O3``. Care is advised when using the algorithm at higher optimization levels.
```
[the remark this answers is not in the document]
praveksharma: All other variants do pass memory leak testing, and are included in the CI. We also include `gcc` bases tests in the CI and those pass the leak tests too. However, the CI doesn't include tests for some architectures (PowerPC and some ARM variants) so updating the advisory is a good call. c046b7553f4839ffb0ee745606792bddc203cbf5 does this.

src/common/pqclean_shims/crypto_declassify.h
praveksharma: c046b7553f4839ffb0ee745606792bddc203cbf5 replaces `crypto_declassify.h` with a MIT version.

docs/algorithms/kem/classic_mceliece.md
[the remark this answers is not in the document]
baentsch: Thanks for this update. I'm afraid though I'll need to be a joykill again: Please see/run https://github.com/open-quantum-safe/liboqs/blob/f032c209613c0a75fbabd7ca6da560f9a3a600a6/.github/workflows/linux.yml#L35 and see whether something changes... I'm afraid just updating the .md file doesn't cut it...

scripts/copy_from_upstream/src/kem/family/CMakeLists.txt
praveksharma, line 1: Jinja was missing this line because the implementation name changed from `avx` to `avx2`. I've included the `-O1` option here for proper upstream code integration.

.CMake/compiler_opts.cmake
praveksharma, line 1: Removed this option from here because it's already included in build. See [this](https://github.com/open-quantum-safe/liboqs/pull/1470/commits/639977a0feb4bf139af94cd3d4553bad42296a07#r1264582142) comment.

.CMake/compiler_opts.cmake
```diff
@@ -96,7 +96,6 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang")
     add_compile_options(-Wextra)
     add_compile_options(-Wpedantic)
     add_compile_options(-Wno-unused-command-line-argument)
-    add_compile_options(-Wno-language-extension-token)
```
praveksharma: Removed this option from here because it's already included in build. See [this](https://github.com/open-quantum-safe/liboqs/pull/1470/commits/639977a0feb4bf139af94cd3d4553bad42296a07#r1264582142) comment.

scripts/copy_from_upstream/src/kem/family/CMakeLists.txt
```diff
@@ -49,8 +49,8 @@ if(OQS_ENABLE_KEM_{{ family }}_{{ scheme['scheme_c'] }}_{{ impl['name'] }})
         {%- if impl['compile_opts'] %}
     target_compile_options({{ family }}_{{ scheme['scheme'] }}_{{ impl['name'] }} PUBLIC {{ impl['compile_opts'] }})
         {%- endif %}
-        {%- if family == 'classic_mceliece' and impl['name'] == 'avx' %}
```
praveksharma, line 52: Jinja was missing this line because the implementation name changed from `avx` to `avx2`. I've included the `-O1` option here for proper upstream code integration.

docs/algorithms/kem/classic_mceliece.md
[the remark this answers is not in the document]
praveksharma: Thanks for pointing this out Michael! Code copied from upstream should integrate properly with [639977a](https://github.com/open-quantum-safe/liboqs/pull/1470/commits/639977a0feb4bf139af94cd3d4553bad42296a07).

baentsch (approved):
Thanks for working through all these issues, @praveksharma ! Now LGTM.

PR #1482, opened 2023-05-25
Add XMSS parameter xmss_sha256_h10 

ducnguyen-sb:
# Add XMSS parameter xmss_sha256_h10

This version of XMSS:

1. Original XMSS has signature format: `(sm || m)` (signature concatenates with message). This PR separates `sm` and `m`. Minimal change to internal of XMSS. See: 
a. https://github.com/ducnguyen-sb/xmss-reference/commit/954308800a0ba47357c881838e8ce585a9cd3d97#diff-c79859c458945f7449e50f5ca7e44857716c35763142802cba881cdfd0a61df7
b. https://github.com/ducnguyen-sb/xmss-reference/commit/954308800a0ba47357c881838e8ce585a9cd3d97#diff-2089b4bfa66835b4b7b1d2b4c54c9622288b9149badc7dfd92167fc194f8961a

3. Original XMSS has hash calls like this:

```c
static int core_hash(const xmss_params *params,
                     unsigned char *out,
                     const unsigned char *in, unsigned long long inlen)
{
    unsigned char buf[32];

    if (params->n == 24 && params->func == XMSS_SHA2) {
        OQS_SHA2_sha256(buf, in, inlen);
        memcpy(out, buf, 24);
    }
    else if (params->n == 24 && params->func == XMSS_SHAKE256) {
        OQS_SHA3_shake256(out, 24, in, inlen);
    }   
    else if (params->n == 32 && params->func == XMSS_SHA2) {
        OQS_SHA2_sha256(out, in, inlen);
    }
    else if (params->n == 32 && params->func == XMSS_SHAKE128) {
        OQS_SHA3_shake128(out, 32, in, inlen);
    }
    else if (params->n == 32 && params->func == XMSS_SHAKE256) {
        OQS_SHA3_shake256(out, 32, in, inlen);
    }
    else if (params->n == 64 && params->func == XMSS_SHA2) {
        OQS_SHA2_sha512(out, in, inlen);
    }
    else if (params->n == 64 && params->func == XMSS_SHAKE256) {
        OQS_SHA3_shake256(out, 64, in, inlen);
    }
    else {
        return -1;
    }
    return 0;
}
```
This PR optimizes hash call via define macro at compile time. The goal is to write just enough code to serve the hash call, and avoid branch jump during execution. Example for parameter `xmss_sha256_h10`:

```c
#include <oqs/sha2.h>
#include "hash.h"

int core_hash(const xmss_params *params,
              unsigned char *out,
              const unsigned char *in, unsigned long long inlen)
{
    (void)params;
    OQS_SHA2_sha256(out, in, inlen);
    return 0;
}

```
4. Share code: This PR has plan to share code between parameters using define and namespace. The differences between parameters at compile time is handled by carefully craft `CMakefile.txt`

5. Additional robust check for unsigned int and memory initialization error: 

https://github.com/ducnguyen-sb/xmss-reference/pull/17/files

6. A way to generate NIST-KAT: 

https://github.com/ducnguyen-sb/xmss-reference/pull/16/files#diff-b8143f018bcb227109839d76aa73c408742fa1fd56b6c4013833274583321c4d

All the changes above are propagated to this PR. 


ducnguyen-sb:
@xvzcf Can you point me how can I figure out the test on [Github Action failed](https://github.com/open-quantum-safe/liboqs/actions/runs/5086010544/jobs/9140093292?pr=1482)

The verbose debug is [here](https://github.com/open-quantum-safe/liboqs/pull/1482/files#diff-56ef09d4e72bc6507f9a58e8425f498ba23e63d2adc4fc8982fcff419d219b17R125)

ducnguyen-sb:
Thanks to @ashman-p and @malbert1 : I was able to fix the XMSS casting bugs.

src/sig_stfl/xmss/external/Makefile
dstebila, line 1: This Makefile isn't needed inside liboqs.

src/sig_stfl/xmss/external/core_hash_n32_sha2.c
```diff
@@ -0,0 +1,11 @@
+#include <oqs/sha2.h>
+#include "hash.h"
+
+int core_hash(const xmss_params *params,
```
dstebila, line 4: How will this work when there are multiple variants using different hash functions?

src/sig_stfl/xmss/external/namespace.h
```diff
@@ -0,0 +1,14 @@
...
+#define XMSS_NAMESPACE_H
+
+
+#define XMSS_PARAMS oqs_sig_stfl_xmss
+#define XMSS_PARAMS_INNER oqs_sig_stfl_xmss_inner
+
+#define XMSS_NAMESPACE(funcname) XMSS_(XMSS_PARAMS, funcname)
+#define XMSS_INNER_NAMESPACE(funcname) XMSS_(XMSS_PARAMS_INNER, funcname)
+#define XMSS_(prefix, funcname) XMSS__(prefix, funcname)
+#define XMSS__(prefix, funcname) prefix ## _ ## funcname
```
dstebila, line 11: If I understand this correctly, this namespaces everything with `oqs_sig_stfl_xmss` or `oqs_sig_stfl_xmss_inner`, but has no variant-specific namespacing.

tests/kat_sig_stfl.c
dstebila, line 1: Will need to add a test in `tests/test_kat.py` to run `kat_sig_stfl` as part of the test suite.

dstebila (changes requested):
Some changes and discussion needed as per comments inline.

src/sig_stfl/xmss/external/Makefile
ducnguyen-sb: fixed in https://github.com/open-quantum-safe/liboqs/pull/1482/commits/7e83c2c04696ac69687cf8ff0f185854aac80a11

src/sig_stfl/xmss/external/core_hash_n32_sha2.c
ducnguyen-sb: I think I will use define macro as inside the file and cmakelist https://github.com/open-quantum-safe/liboqs/pull/1482/commits/6712094df456542d5d1386189b3eb49c4b9f59f8 and https://github.com/open-quantum-safe/liboqs/pull/1482/commits/3677849a77ab5b098a672e17734e939e85cc54ae

tests/kat_sig_stfl.c
ducnguyen-sb: yup, added in https://github.com/open-quantum-safe/liboqs/pull/1482/commits/9d6cb0b4bb0b84482ccc479e00b8173e4285200a

src/sig_stfl/xmss/external/namespace.h
ducnguyen-sb: Yup, I intended to share the code between parameters, so there is no plan for parameter namespace. The `_inner` prefix is mainly to indicate the internal functions of `xmss` so it's easy to distinguish them while debugging.

Issue #1206, opened 2022-04-27
Adding a DeriveKeyPair functionality


tedeaton:
When a KEM keypair is derived in liboqs, it relies on randomness using [rand.h](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/rand/rand.h). In some situations, it's useful to generate keypairs deterministically from a seed. When [RFC9180](https://www.rfc-editor.org/rfc/rfc9180#name-cryptographic-dependencies) defines a KEM, they include a `DeriveKeyPair` option to generate keypairs deterministically from a seed. As an example use case, [Messaging Layer Security](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html) makes extensive use of functionality to share multiple keypairs by communicating seeds. I think it would be nice to match RFC9180 and add a functionality to deterministically derive a keypair from a seed.

This is kind of possible in liboqs currently. You can set the randomness to use `OQS_RAND_alg_nist_kat`, and derive all the randomness used from a seed, but this is a somewhat imperfect solution for a few reasons:

- The seed must be exactly 48 bytes
- It references a global variable, resulting in concurrency issues
- Modifying the randomness used is (to my knowledge) not available with any of the liboqs language wrappers

I'm curious if others think this is a functionality worth adding and if so, what the proper way to tackle it might be. Some schemes, like Kyber, are already constructed in such a way that everything is derived from a seed, but liboqs provides limited ability to set that seed. I suspect the situation for other schemes is much more complex.

baentsch:
> I'm curious if others think this is a functionality worth adding and if so, what the proper way to tackle it might be. 

For reproducibility of any tests, IMO this is a functionality worth having. But then again, this is what `OQS_randombytes_nist_kat` provides. If this is too limited (48 bytes, concurrency issues), couldn't one simply call `OQS_randombytes_custom_algorithm` with an algorithm of personal choice to get around those issues?

Finally, if such new RFC-compliant and multicore-capable rand provider is needed (?) and available, why not discuss adding it to the pre-existing rand-sources with a new name and thus make it available to all language-wrappers (assuming `OQS_randombytes_switch_algorithm` is available in all language wrappers)?

tedeaton:
[...]

Issue #1081, opened 2021-08-29
Shared library versioning scheme (SOVERSION)


vt-alt:
As pointed there https://github.com/open-quantum-safe/openssh/issues/111#issuecomment-907751920 `liboqs` would benefit from proper _library versioning_. (Sometimes called _API version_ or _SOVERSION_). This is different thing than usual project versioning. It's purpose to maintain binary (ABI) compatibility between executable binaries and shared libraries. Currently, liboqs just copy project version into SOVERSION which does not reflect backward incompatible ABI changes (which may cause obscure program errors when binary that is linked to a newer library used on system with older library — current versioning allows this, because major version number (that signifying backward-incompatible ABI change) is not incremented (while library does have backward incompatible ABI changes).

References:
* General explanation https://stackoverflow.com/questions/12637841/what-is-the-soname-option-for-building-shared-libraries-for
* GNU tooling https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html (Note that their `-version-info` arguments order is different from what will actually appear in the library suffix - they say `current:revision:age`, but library will be named like `liboqs.so..current.age.revision`).
* Cmake property https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html

vt-alt:
I found you currently using `SOVERSION 0` in `src/CMakeLists.txt` for `oqs` target. Perhaps, you just need to increment it sometimes. I tried already — it looked unusual in compare to what libtool does, but seems valid too. I tried to set `SOVERSION` to `1`. It creates [SONAME](https://en.wikipedia.org/wiki/Soname) `liboqs.so.1` with symlink to `liboqs.so.0.7.0`. In that case binaries will link to `liboqs.so.1` which is good.

vt-alt:
I think Dilithium ABI (that was caused heap corruption) is changed with changes like this: https://github.com/open-quantum-safe/liboqs/commits/main/src/sig/dilithium/sig_dilithium.h, for example when any `_length_` define is increased. `abipkgdiff` cannot detect this, because (I think) it is change in `define` and buffers are passed using generic `uint8_t *` pointer of unspecified length.

So, any release after such change should have incremented SOVERSION.

baentsch:
Thanks again for the explanation and tests. 

> So, any release after such change should have incremented SOVERSION.

[...]

PR #1491, opened 2023-06-07
Derive keypair

tedeaton:
This Pull Request is intended to add deterministic key derivation to liboqs, specifically with Kyber. It adds a new compile flag, "OQS_HAZARDOUS_ENABLE_DERIVE_KEYPAIR", which exposes deterministic key derivation. This is intended to be used with RFC 9180 for hybrid public key encryption, which defines KEMs to have a "DeriveKeyPair" algorithm (in addition to key generation, encapsulation, and decapsulation).

Adds functionality requested in #1206 

This adds a "derive_keypair" to the KEM API when the compile flag OQS_HAZARDOUS_ENABLE_DERIVE_KEYPAIR is enabled.

I tried to have it so that if the OQS_HAZARDOUS_ENABLE_DERIVE_KEYPAIR is not defined, then no changes happen to liboqs. This results in some code replication. A smarter way to do it would be to have regular key generation just call deterministic key generation with a random seed, but this would mean altering how regular key generation works so I stayed away from that (for now). It should be an easy change everywhere else if that's eventually modified.

As well, I've added a test_derive_keypair script, which just derives a keypair twice from the same, random seed and compares them to check that the same secret and public key is derived each time.

There should be some sort of check to make sure that everything other than kyber is disabled, since this only adds deterministic key derivation for kyber, but I wasn't sure about the best way to add that.


bhess:
The changes to `src/kem/kyber/...` would probably be overwritten after pulling updates with copy_from_upstream. There's a patch mechanism to apply changes after copying from upstream.

I wonder if there is a more generic way to add this API to libOQS without patching the upstream sources. All KEMs would otherwise also need to be patched to enable this.
It seems that this API basically replaces the calls to `randombytes` by providing fixed randomness. If randombytes was configurable (e.g. either using system randomness, or a deterministic PRNG), wouldn't this have the same effect?

dstebila:
> The changes to `src/kem/kyber/...` would probably be overwritten after pulling updates with copy_from_upstream. There's a patch mechanism to apply changes after copying from upstream.

Yes this is problematic, either the deterministic API needs to be added upstream or added here via the patch mechanism.

> I wonder if there is a more generic way to add this API to libOQS without patching the upstream sources. All KEMs would otherwise also need to be patched to enable this. It seems that this API basically replaces the calls to `randombytes` by providing fixed randomness. If randombytes was configurable (e.g. either using system randomness, or a deterministic PRNG), wouldn't this have the same effect?

Well, `OQS_randombytes` is user-configurable, but not in a local way, only globally, so it wouldn't be thread-safe.

dstebila:
> Agreed -- and in any case, this changes the `liboqs` ABI, so needs an update to the library version (see #1497 and #1081) -- and having a library version dependent on a compile flag is not really desirable, I'd think.

Adding a function to the ABI doesn't require an SOVERSION bump, only changing the behaviour of the existing ABI.

dstebila:
Before merging this, we'd also need some documentation updates explaining this new functionality (in the README and may be examples) as well as adding it to the algorithm data sheets.

And as pointed out in https://github.com/open-quantum-safe/liboqs/pull/1491#issuecomment-1582499077, also need to either make changes upstream or add them to our patch mechanism so they don't get wiped out next time we do a copy-from-upstream.

baentsch:
@tedeaton Just checking the status of this PR: Do you still want to move this forward? Any particular timeline you're working towards? Any open questions from your side regarding the feedback above?

baentsch:
Suggest to close this PR: It's pretty stale (lots of conflicts) and @tedeaton has never chimed in to this discussion.

dstebila:
Agreed. It's still a desirable feature but this PR is stale at this point.

PR #1493, opened 2023-06-08
Re-add OQS_SECRET_KEY

dstebila:
Introduce stateful key data object to allow tracking variant specific keys.

### DIFF
--- a/src/sig_stfl/sig_stfl.h
+++ b/src/sig_stfl/sig_stfl.h
@@ -151,8 +151,6 @@ typedef struct OQS_SIG_STFL {
 typedef struct OQS_SIG_STFL_SECRET_KEY OQS_SIG_STFL_SECRET_KEY;
 
 typedef struct OQS_SIG_STFL_SECRET_KEY {
-	/** A local ordinal representing the LMS parameter of the signature scheme. */
-	uint32_t oid;
 
 	/** Associated signature object */
 	OQS_SIG_STFL *sig;
--- a/src/sig_stfl/sig_stfl.h
+++ b/src/sig_stfl/sig_stfl.h
@@ -163,9 +163,6 @@ typedef struct OQS_SIG_STFL_SECRET_KEY {
 	/* The variant specific secret key data */
 	void *secret_key_data;
 
-	/* User defined data that may be used for the SAFETY functions */
-	void *data;
-
 	/* Function that returns the total number of signatures for the secret key */
 	unsigned long long (*sigs_total)(const OQS_SIG_STFL_SECRET_KEY *secret_key);
 
@@ -200,6 +197,14 @@ typedef struct OQS_SIG_STFL_SECRET_KEY {
 	 */
 	OQS_STATUS (*lock_key)(OQS_SIG_STFL_SECRET_KEY *sk);
 
+    /**
+     * Secret Key Unlocking / Releasing Function
+     *
+     * @param[in] sk The secret key represented as OQS_SIG_STFL_SECRET_KEY object
+     * @return OQS_SUCCESS or OQS_ERROR
+     */
+    OQS_STATUS (*unlock_key)(OQS_SIG_STFL_SECRET_KEY *sk);
+
 	/**
 	 * Secret Key Saving Function
 	 *
@@ -207,14 +212,6 @@ typedef struct OQS_SIG_STFL_SECRET_KEY {
 	 * @return OQS_SUCCESS or OQS_ERROR
 	 */
 	OQS_STATUS (*save_secret_key)(const OQS_SIG_STFL_SECRET_KEY *sk);
-
-	/**
-	 * Secret Key Unlocking / Releasing Function
-	 *
-	 * @param[in] sk The secret key represented as OQS_SIG_STFL_SECRET_KEY object
-	 * @return OQS_SUCCESS or OQS_ERROR
-	 */
-	OQS_STATUS (*release_key)(OQS_SIG_STFL_SECRET_KEY *sk);
 
 	/**
 	 * Secret Key free internal variant specific data
--- a/src/sig_stfl/sig_stfl.h
+++ b/src/sig_stfl/sig_stfl.h
@@ -55,6 +55,74 @@ OQS_API int OQS_SIG_STFL_alg_count(void);
 OQS_API int OQS_SIG_STFL_alg_is_enabled(const char *method_name);
 
 /**
+ * @brief OQS_SECRET_KEY object for stateful signature schemes
+ */
+typedef struct OQS_SECRET_KEY OQS_SECRET_KEY;
+
+typedef struct OQS_SECRET_KEY {
+    /** A local ordinal representing the LMS parameter of the signature scheme. */
+    uint32_t oid;
+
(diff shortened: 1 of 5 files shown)


src/sig_stfl/sig_stfl.h
```diff
@@ -54,6 +54,74 @@ OQS_API int OQS_SIG_STFL_alg_count(void);
...
+ * @brief OQS_SECRET_KEY object for stateful signature schemes
+ */
+typedef struct OQS_SECRET_KEY OQS_SECRET_KEY;
+
+typedef struct OQS_SECRET_KEY {
+    /** A local ordinal representing the LMS parameter of the signature scheme. */
+    uint32_t oid;
+
+    /** Printable string representing the name of the signature scheme. */
+    const char *method_name;
```
dstebila, line 67: Let's replace this with a point to the OQS_SIG_STFL object, which contains the oid and method_name.

src/sig_stfl/sig_stfl.h
```diff
@@ -54,6 +54,74 @@ OQS_API int OQS_SIG_STFL_alg_count(void);
...
+    /* Function that returns the number of signatures left for the secret key */
+    unsigned long long (*sigs_left)(const OQS_SECRET_KEY *secret_key);
+
+    /**
+     * Secret Key retrieval Function
+     *
+     * @param[in] sk The secret key represented as OQS_SECRET_KEY object
+     * @return byte pointer or null
+     */
+    uint8_t * (*get_key)(OQS_SECRET_KEY *sk);
```
dstebila, line 90: Just to be sure about the convention here, the idea is that the function will malloc a new byte array of the appropriate size (presumably no bigger than length_secret_key up above) and then return a pointer to that?  Which the caller is responsible for freeing?  I think the API documentation could be clearer including indicating that the caller is responsible for freeing the result.  One thing that's unclear though is how the caller is told what the length of the allocated byte array is.

I think `get_key` is a bit unclear as a method name. Maybe either `key_to_byte_string` or `serialize_key`.

Also will need to add something that deserializes the key from a byte string back to an OQS_SECRET_KEY object.  Whether it's a static method of the OQS_SECRET_KEY object or a method of OQS_SIG_STFL.

src/sig_stfl/sig_stfl.h
```diff
@@ -54,6 +54,74 @@ OQS_API int OQS_SIG_STFL_alg_count(void);
...
+     */
+    OQS_STATUS (*save_secret_key)(const OQS_SECRET_KEY *sk);
+
+    /**
+     * Secret Key Unlocking / Releasing Function
+     *
+     * @param[in] sk The secret key represented as OQS_SECRET_KEY object
+     * @return OQS_SUCCESS or OQS_ERROR
+     */
+    OQS_STATUS (*release_key)(OQS_SECRET_KEY *sk);
```
dstebila, line 114: I know we had `save_secret_key` and `release_key` in the earlier API, but now I'm wondering about the difference between those two.  For a calling application, could it just incorporate whatever action it would take during `save_secret_key` into `release_key`?  When would these ever need to be separated?

src/sig_stfl/sig_stfl.h
ashman-p: Done.

src/sig_stfl/sig_stfl.h
```diff
@@ -54,6 +54,74 @@ OQS_API int OQS_SIG_STFL_alg_count(void);
...
+    const char *method_name;
+
+    /* The (maximum) length, in bytes, of secret keys for this signature scheme. */
+    size_t length_secret_key;
+
+    /* The variant specific secret key data */
+    void *secret_key_data;
+
+    /* User defined data that may be used for the SAFETY functions */
+    void *data;
```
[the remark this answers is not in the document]
ashman-p: Removed "data" member.

src/sig_stfl/sig_stfl.h
ashman-p: renamed 'release_key' to 'unlock_key'. I tok 'save_secret_key' to meant to persist the the key (file write).

src/sig_stfl/sig_stfl.h
```diff
@@ -54,6 +54,74 @@ OQS_API int OQS_SIG_STFL_alg_count(void);
  */
 OQS_API int OQS_SIG_STFL_alg_is_enabled(const char *method_name);
 
+/**
+ * @brief OQS_SECRET_KEY object for stateful signature schemes
+ */
+typedef struct OQS_SECRET_KEY OQS_SECRET_KEY;
+
+typedef struct OQS_SECRET_KEY {
```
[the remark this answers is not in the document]
ashman-p: Done.

src/sig_stfl/sig_stfl.h
ashman-p: When do you think this pointer would get set?

src/sig_stfl/sig_stfl.h
ashman-p: Changing to "serialize_key" and "deserializes". Will update and make clearer memory mgmt responsibilities.

   /**
     * Secret Key retrieval Function
     *
     * @param[in] sk The secret key represented as OQS_SECRET_KEY object
     * @param[out] key_len length of the returned byte string
     * @return byte pointer or null
     */
    uint8_t * (* serialize_key)(OQS_SECRET_KEY *sk, size_t *key_len);

src/sig_stfl/sig_stfl.h
```diff
@@ -145,6 +145,86 @@ typedef struct OQS_SIG_STFL {
...
+	OQS_SIG_STFL *sig;
+
+	/* The (maximum) length, in bytes, of secret keys for this signature scheme. */
+	size_t length_secret_key;
+
+	/* The variant specific secret key data */
+	void *secret_key_data;
+
+	/* User defined data that may be used for the SAFETY functions */
+	void *data;
```
dstebila, line 167: Still got an extra field here.

src/sig_stfl/sig_stfl.h
dstebila: Good question. I guess it would be helpful to see some test code doing these things in sequence, and then I guess it would be clear when to set it...

src/sig_stfl/sig_stfl.h
ashman-p: oops.

src/sig_stfl/sig_stfl.h
```diff
@@ -145,6 +145,83 @@ typedef struct OQS_SIG_STFL {
...
 } OQS_SIG_STFL;
 
+/**
+ * @brief OQS_SIG_STFL_SECRET_KEY object for stateful signature schemes
+ */
+typedef struct OQS_SIG_STFL_SECRET_KEY OQS_SIG_STFL_SECRET_KEY;
+
+typedef struct OQS_SIG_STFL_SECRET_KEY {
+	/** A local ordinal representing the LMS parameter of the signature scheme. */
+	uint32_t oid;
```
dstebila, line 155:
```suggestion
```

Issue #1495, opened 2023-06-11
liboqs.pc: incorrect libdir value


vt-alt:
**Describe the bug**
```
builder@x86_64:/.in$ grep libdir /usr/lib64/pkgconfig/liboqs.pc
libdir=${prefix}/lib
Libs: -L${libdir} -loqs
```
I think (for ALT) proper `libdir` value should be `/usr/lib64` for x86_64. Usually it should have `lib64` for 64-bit systems and `lib` for 32-bit systems. Except for Debian based systems where is different convention — there it will be `lib/x86_64-linux-gnu` perhaps.)

This difference, though, does not produce incorrect flags (on ALT):
```
builder@x86_64:/.in$ pkg-config liboqs --libs
-loqs
```
This is because `-l/usr/lib` is stripped by default by `pkg-config` (based on (default) value of `PKG_CONFIG_SYSTEM_LIBRARY_PATH` containing `/usr/lib`.

So this is **not a major** bug, but better be fixed for correctness in future versions.

BTW. It's possible to force `pkg-config` to not strip system dirs:
```
builder@x86_64:/.in$ PKG_CONFIG_ALLOW_SYSTEM_LIBS=1 pkg-config liboqs --libs
-L/usr/lib -loqs
```
So this could in theory produce erroneous linking for some very non-standrad build environments.

**Environment (please complete the following information):**
 - OS: ALT
 - liboqs version 0.8.0

vt-alt:
Change like this seems to produce correct values:
```diff
-libdir=${prefix}/lib
-includedir=${prefix}/include
+libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
+includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
```

baentsch:
> Change like this seems to produce correct values:

Thanks for letting us know. Would you want to contribute the above via PR?

vt-alt:
I will send it later, when I test it on Debian.

dstebila:
Fixed by #1496.

PR #1496, opened 2023-06-11
Fix libdir value in liboqs.pc

vt-alt:
Libdir is not always '/usr/lib' and its value is different between distributions and architectures. Set it from CMAKE_INSTALL_LIBDIR. Also, set includedir in a similar way.

Currently, this incorrect setting did not produce incorrect output by pkg-config --libs because pkg-config strips 'system library paths' form the output. But in non-standard build environments this may cause incorrect linking.

Issue: https://github.com/open-quantum-safe/liboqs/issues/1495

### DIFF
--- a/src/liboqs.pc.in
+++ b/src/liboqs.pc.in
@@ -1,6 +1,6 @@
 prefix=@CMAKE_INSTALL_PREFIX@
-libdir=${prefix}/lib
-includedir=${prefix}/include
+libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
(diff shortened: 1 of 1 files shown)


vt-alt:
I tested on ALT and Debian and new values appear to me correct.
[Debian builds with](https://buildd.debian.org/status/fetch.php?pkg=liboqs&arch=amd64&ver=0.7.2-2&stamp=1661924965&raw=0) cmake setting `-DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu` and on ALT with default setting (determined by cmake).

dstebila (approved):

PR #1498, opened 2023-06-14
update version and remove CCI triggers

baentsch:
- Update release information following release
- Remove CCI triggers from projects no longer supporting CCI


dstebila (approved):

dstebila:
I think it should be possible to trigger a Github Action in a downstream repo using a similar approach.  Even if the parent job is in CircleCI, it could make a REST API call to trigger a job in Github Actions as long as it has the relevant access tokens.  Maybe something like this would help: https://dev.to/rikurouvila/how-to-trigger-a-github-action-with-an-htt-request-545

PR #1501, opened 2023-06-22
create deb package and retain as artifact

baentsch:
In pursuit of https://github.com/open-quantum-safe/oqs-provider/issues/94 this PR prepares (for) binary releases (to be added to release packages, for example).

The binary is intentionally created to
- only contain the standardized algorithms ("STD")
- only use OpenSSL3 APIs (so it can seamlessly be integrated to ´oqsprovider´)
- be built shared (such as to permit independent `dpkg -i`(installs))

Suggestions for other setups welcome. Suggestions where (else) to store the artifact also welcome (www.openquantumsafe.org?)

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -80,13 +80,13 @@ jobs:
             container: openquantumsafe/ci-alpine-amd64:latest
             CMAKE_ARGS: -DOQS_STRICT_WARNINGS=ON -DOQS_USE_OPENSSL=OFF
             PYTEST_ARGS: --ignore=tests/test_alg_info.py
-          - name: focal-std-openssl
+          - name: focal-nistr4-openssl
             container: openquantumsafe/ci-ubuntu-focal-x86_64:latest
-            CMAKE_ARGS: -DOQS_STRICT_WARNINGS=ON -DOQS_ALGS_ENABLED=STD
-            PYTEST_ARGS: --ignore=tests/test_leaks.py
-          - name: jammy-nistr4-openssl3
-            container: openquantumsafe/ci-ubuntu-jammy:latest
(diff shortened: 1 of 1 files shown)


dstebila:
I'm not familiar with this feature of Github. What does it mean to preserve it as an artifact?  Is it turned into a release?  Or is there some other place it's available?

baentsch:
> I'm not familiar with this feature of Github. What does it mean to preserve it as an artifact? Is it turned into a release? Or is there some other place it's available?

Github artifacts are documented [here](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts). They are retained ([and can be downloaded](https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts?tool=webui)) for some time after the build, e.g., if they "stood the test of time" to then be made (either manually or via specific code: download & store to some place) available wherever one sees fit (e.g., in binaries linked to releases). This PR does not contain such "automated deploy code": We'd first need to decide where to send it to -- and under which conditions.

baentsch:
For further clarification: With https://github.com/open-quantum-safe/oqs-provider/pull/196 and this PR landing, deployment of `liboqs` on an x64 Linux with OpenSSL3 present then simply consists of running ` dpkg -i liboqs-0.8.1-dev-Linux.deb`. This step of course is not necessary if `liboqs` has already been integrated to & installed via the package management system of the distribution used.

Subsequent installation of `oqsprovider` then is simply `dpkg -i oqs-provider-0.5.1-dev-Linux.deb`. After this, one can operate `openssl` with all OQS alg combinations (again, assuming they have `openssl` >= 3.0 installed and activated `oqsprovider`). Again, `dpkg -i` would not be necessary any more if/when `oqsprovider` is integrated to the distribution's package management system.

PR #1508, opened 2023-07-13
[NFCI] Move Keccak rhotates tables to rodata

aaupov:
rhotates tables are placed to .text section which confuses tools such as BOLT. Move them to rodata to unbreak and avoid polluting icache/iTLB with data.

Sync with OpenSSL https://github.com/openssl/openssl/pull/21440

### DIFF
--- a/src/common/sha3/xkcp_low/KeccakP-1600/avx2/KeccakP-1600-AVX2.S
+++ b/src/common/sha3/xkcp_low/KeccakP-1600/avx2/KeccakP-1600-AVX2.S
@@ -1042,6 +1042,7 @@ KeccakP1600_12rounds_FastLoop_Absorb_LanesAddLoop:
 .size   KeccakP1600_12rounds_FastLoop_Absorb,.-KeccakP1600_12rounds_FastLoop_Absorb
 #endif
 
+.section .rodata
 .equ    ALLON,        0xFFFFFFFFFFFFFFFF
 
 .balign 64


aaupov:
Darwin enables `old_gas_syntax` variable, and maintains the set of patches. Dropping this PR for maintainers to figure out the proper way to include this change.

dstebila:
Hi Amir, we have very little experience with these types of matters, so any guidance from you other experts would be appreciated.

aaupov:
> Hi Amir, we have very little experience with these types of matters, so any guidance from you other experts would be appreciated.

There's this patch mechanism that I don't understand enough but which unfortunately interferes with this change: https://github.com/open-quantum-safe/liboqs/blob/main/scripts/copy_from_xkcp/patches/lib_low_KeccakP-1600_AVX2_KeccakP-1600-AVX2.s

It turns on this flag `old-gas-syntax` which breaks macos builds. I don't have the context of why it was introduced so I can't just turn it off as part of PR.

Moreover, my understanding is that there are two options to bring this upstream openssl change down to liboqs: 
1. re-generate the upstream assembly source and deal with the possible breakage of patches,
2. incorporate the upstream change into patches, ensuring they will break the future updated code.

The upstream change itself is trivial, but I want to avoid doing the maintainer work of either of the two approaches above.

baentsch:
> Dropping this PR for maintainers to figure out the proper way to include openssl change.

> there are two options to bring this upstream openssl change down 

I'm somewhat confused by these statements: The code this PR touches is not related to openssl, but its upstream is [XKCP](https://github.com/XKCP/XKCP). And in that project I didn't find the change done in https://github.com/openssl/openssl/pull/21440 so the proposed update options from upstream won't quite solve the issue. 

Without having this change in upstream (XKCP) the proposed change may be added to the upstream patch -- but I don't know either the rationale for enabling "old-gas-syntax" in #1173, so asking its authors @jschanck and @dstebila for advice. 

Without such rationale, we could add the proposed change to https://github.com/open-quantum-safe/liboqs/tree/main/scripts/copy_from_xkcp/patches guarding for Darwin (e.g. with a surrounding "#ifndef old_gas_syntax"). Would that serve your purpose @aaupov ?

baentsch:
Suggest closing without further feedback by @aaupov . OK @dstebila ?

Issue #1320, opened 2022-11-23
Update Kyber and Dilithium ARM code.


xvzcf:

xvzcf:
As previously noted, our implementations are from PQClean, which in turn takes it from the [neon-ntt](https://github.com/neon-ntt/neon-ntt) project.

baentsch:
> As previously noted, our implementations are from PQClean, which in turn takes it from the [neon-ntt](https://github.com/neon-ntt/neon-ntt) project.

Interesting to note -- what's the implication of this for us? Did PQClean not integrate the latest code from neon-ntt yet? Is neon-ntt itself not up-to-date compared to the ref implementation?

dstebila:
PQClean doesn't have the latest code from neon-ntt.

baentsch:
> PQClean doesn't have the latest code from neon-ntt.

Is this still the case? Or would/should a fresh `copy_from_upstream` resolving #1506 also resolve this issue?

mczraf:
> > PQClean doesn't have the latest code from neon-ntt.
> 
> Is this still the case? Or would/should a fresh `copy_from_upstream` resolving #1506 also resolve this issue?

I believe it is still the case. I've checked [PQClean repo](https://github.com/PQClean/PQClean/commits/master/crypto_kem/kyber768/aarch64/neon_poly.c), and they are still using the original Neon implementation from Sep/2021.  

It might be worth adding that upgrading this implementation to a newer version may lead to some additional benefits. For example, LibOQS Kyber's implementation is not marked as passing all Valgrind branch-on-secrets tests ([see here](https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/kem/kyber.md)). There's a chance this could be automatically fixed by simply resorting to the newer Neon implementation. PS: At the moment, I can't dedicate cycles to this effort myself unfortunately but I thought it would be helpful to comment here to reinforce that this seems like an interesting move for LibOQS. 

Thank you all for this amazing project :)

Rafael Misoczki (speaking for myself)

baentsch:
Thanks for the feedback, @mczraf ! https://github.com/PQClean/PQClean/pull/493 seems to confirm --and be work to resolve-- this.

SWilson4:
Fixed by #1512.

PR #1512, opened 2023-07-19
kyber/dilithium aarch64 pull from pqclean + patches

### LINKED ISSUE #1506: macOS M1: dilithium/aarch64 not compiling with gcc
**Describe the bug**
Compiling fails on macOS (M1) when using gcc.

```
...
[ 63%] Building ASM object src/sig/dilithium/CMakeFiles/dilithium_2_aarch64.dir/pqclean_dilithium2_aarch64/__asm_iNTT.S.o
/Users/bhe/git/liboqs/src/sig/dilithium/pqclean_dilithium2_aarch64/__asm_iNTT.S:8:1: error: unknown directive
.type PQCLEAN_DILITHIUM2_AARCH64_asm_intt_SIMD_top, %function
^
/Users/bhe/git/liboqs/src/sig/dilithium/pqclean_dilithium2_aarch64/__asm_iNTT.S:418:1: error: unknown directive
.type PQCLEAN_DILITHIUM2_AARCH64_asm_intt_SIMD_bot, %function
^
...
```

**To Reproduce**
```
cmake -DCMAKE_C_COMPILER=gcc-13 ..
make
```

 - OS: macOS 13.4.1
 - OpenSSL version 1.1.1t
 - Compiler version used: gcc-13 (Homebrew GCC 13.1.0) 13.1.0
 - Build variables used "-DCMAKE_C_COMPILER=gcc-13"
 - liboqs version: main

**Additional context**
The respective ".type <fun>, %function" statements in the aarch64 assembly code are excluded for clang but not for gcc:
https://github.com/open-quantum-safe/liboqs/blob/93e784725e3be0aaa2691a557456ff786155a7b2/src/sig/dilithium/pqclean_dilithium2_aarch64/__asm_iNTT.S#L7-L9
PQClean upstream appears to have removed the ".type <fun>, %function" statements in the [meanwhile](https://github.com/PQClean/PQClean/commit/0b8db0f8323734b60f638c97f026ac8b66b42378), so a fresh copy_from_upstream might help.

bhess:
- Runs copy_from_upstream with pqclean code, updating kyber and dilithium for aarch64
- Updates upstream patches & adds a few more fixes 
- Fixes a compiler warning in the arm aes code (https://github.com/open-quantum-safe/liboqs/issues/1506#issuecomment-1641645941)
- Fixes #1506 #1320
- Updates a few pqclean upstream licenses. 
- adds a falcon update from pqclean upstream
- updated dilithium suppression files (changed line numbers) -> CT tests passing using Fedora/aarch64

TODOs: 

- [x] KATs of 2 out of 3 Kyber versions (768, 1024) seem to fail with the updated upstream code.
@Martyrshot did you face any similar issues when including the previous code?
- [x] Update license infos


mczraf:
@bhess Thank you for working on the KAT mismatch issue. Quick question: would you have any rough time estimate for the integration of this pull-request? I suspect that the KAT mismatch needs to be fixed first but is there any other blocking issues for this integration?

src/kem/kyber/pqclean_kyber1024_aarch64/NTT_params.h
```diff
@@ -1,3 +1,30 @@
+
+/*
+ * CC0 1.0 Universal or the following MIT License
+ *
+ * MIT License
```
baentsch, line 5: This (dual license) is good(ness) in light of #1437. But then the contents of the PQClean aarch upstream algorithm YML seem to need updating (to also list MIT), no?

src/kem/kyber/pqclean_kyber1024_aarch64/fips202x2.h
```diff
@@ -1,10 +1,17 @@
+
+/*
+ * This file is licensed
+ * under Apache 2.0 (https://www.apache.org/licenses/LICENSE-2.0.html)
```
baentsch, line 4: Another license for the YML...

src/kem/kyber/pqclean_kyber1024_aarch64/feat.S
```diff
@@ -0,0 +1,168 @@
+
+/*
+MIT License
```
bhess, line 3: This is MIT only

src/kem/kyber/pqclean_kyber1024_aarch64/__asm_poly.S
```diff
@@ -1,14 +1,37 @@
 
+/*
+ * CC0 1.0 Universal or the following MIT License
```
bhess, line 3: CC0 or MIT

src/kem/kyber/pqclean_kyber1024_aarch64/fips202x2.h
bhess: I read here Apache-2.0 or CC0 (for the SPDX-id)

bhess:
KATs are now matching and the license infos updated, so the PR seems ready from my side @mczraf . The constant-time tests mentioned in https://github.com/open-quantum-safe/liboqs/issues/1320#issuecomment-1640641574 are now also passing (tested locally, there isn't a CI CT-run for arm at the moment).

The update appears to include the changes noted in https://github.com/open-quantum-safe/liboqs/issues/1245#issuecomment-1319218926 (i.e. feat.s), so it should close #1320.

@dstebila There's a small Falcon update as part in this pqclean-pull you wanted to check if it is ok to include.

baentsch:
LGTM. Thanks!

src/kem/kyber/pqclean_kyber1024_aarch64/Makefile
dstebila, line 1: We should be able to delete this file, the PQClean Makefiles are not needed nor used in liboqs.

dstebila (approved):

src/kem/kyber/pqclean_kyber1024_aarch64/Makefile
dstebila: As well as the other Makefiles in the PQClean directories.

dstebila (changes requested):
Please remove superfluous Makefiles

src/kem/kyber/pqclean_kyber1024_aarch64/Makefile
bhess: Added a commit that removes them.

bhess:
> Please remove superfluous Makefiles

The makefiles are removed now, updated the copy_from_upstream script to handle this case when there are arch-specific upstreams.

dstebila (approved):

docs/algorithms/kem/kyber.yml
```diff
@@ -22,9 +22,9 @@ primary-upstream:
   spdx-license-identifier: CC0-1.0 or Apache-2.0
 optimized-upstreams:
   pqclean-aarch64:
-    source: https://github.com/PQClean/PQClean/commit/d742438e5c541958bfd58070cd8668d757d88e07
+    source: https://github.com/PQClean/PQClean/commit/66e50172055aaf1b9a16d8f35fe03b0807f2723e
       with copy_from_upstream patches
-    spdx-license-identifier: CC0-1.0
+    spdx-license-identifier: (CC0-1.0 or Apache-2.0) and (CC0-1.0 or MIT) and MIT
```
baentsch, line 27: A question regarding this SPDX documentation: Am I assuming right that this has been manually created from the single source code LICENSE files? In addition, is it intentional (and legal) to not include the actual LICENSE file(s) from PQClean? When looking at the contents of those (example: https://github.com/PQClean/PQClean/blob/master/crypto_kem/kyber1024/aarch64/LICENSE) the documented upstream license does not seem to be in line with the SPDX statement in this PR -- or am I overlooking where Apache-2.0 is granted for the whole of the `aarch64` kyber code base? Or did https://github.com/PQClean/PQClean/pull/488 simply fail to update the main LICENSE file?

docs/algorithms/sig/sphincs.yml
```diff
@@ -26,7 +26,7 @@ nist-round: 3
 spec-version: NIST Round 3 submission, v3.1 (June 10, 2022)
 spdx-license-identifier: CC0-1.0
 primary-upstream:
-  source: https://github.com/PQClean/PQClean/commit/d742438e5c541958bfd58070cd8668d757d88e07
+  source: https://github.com/PQClean/PQClean/commit/66e50172055aaf1b9a16d8f35fe03b0807f2723e
```
baentsch, line 29: Similar (license documentation) question here: This states CC0 -- which is textually correct looking at the immediate [LICENSE upstream in PQClean](https://github.com/PQClean/PQClean/blob/master/crypto_sign/sphincs-sha2-256s-robust/clean/LICENSE). When looking at the ultimate origin code base [Sphincs+ is multi licensed](https://github.com/sphincs/sphincsplus/blob/master/LICENSE). If we manually "curate" license statements (see comment on kyber above) why not do the same with Sphincs+? This would resolve #1437 without having to push LICENSE statements "through" PQClean.

docs/algorithms/kem/kyber.yml
bhess: The spdx-license-identifier was manually curated looking at the license information in the headers of the individual files. It would be nicer if there was a single LICENSE or NOTICE file upstream containing all this information. The main LICENSE indeed seems out of sync, I don't know if this is intentional. @dstebila, since you reviewed https://github.com/PQClean/PQClean/pull/488, do you have an insight here?

docs/algorithms/sig/sphincs.yml
bhess: Not sure if this is possible if PQClean only lists CC0.

baentsch: In that case [manual curation](https://github.com/open-quantum-safe/liboqs/pull/1512#discussion_r1271902542) in general should stop and we ought to move to full automation (see #1514).

bhess: Agreed. However, there seems to be a nuance when curating the information (no matter if manual or automatic):
(1) Kyber/aarch64 case: PQClean LICENSE file lists one license and the PQClean source files contain more/other licenses.
(2) SPHNCS+ case: PQClean code base lists a single license. The code's ancestor contains more/other licenses.

baentsch: "Manual" curation allows (arguably/hopefully legally informed) updates, automated license collection does not. In either case, work is required (whether already in this PR or only when/if resolving #1514, I leave to the reviewers of this PR).

docs/algorithms/kem/kyber.yml
dstebila: No insight; the PR only modified the individual files, rather than touching the related licence file.  Probably an oversight.

docs/algorithms/sig/dilithium.md
```diff
@@ -12,7 +12,7 @@
 - **Optimized Implementation sources**: https://github.com/pq-crystals/dilithium/commit/3e9b9f1412f6c7435dbeb4e10692ea58f181ee51 with copy_from_upstream patches
   - **pqclean-aarch64**:<a name="pqclean-aarch64"></a>
       - **Source**: https://github.com/PQClean/PQClean/commit/66e50172055aaf1b9a16d8f35fe03b0807f2723e with copy_from_upstream patches
-      - **Implementation license (SPDX-Identifier)**: (CC0-1.0 or Apache-2.0) and (CC0-1.0 or MIT) and MIT
+      - **Implementation license (SPDX-Identifier)**: CC0-1.0 and (CC0-1.0 or Apache-2.0) and (CC0-1.0 or MIT) and MIT
```
baentsch, line 15: Do I understand this logical statement to now mean that CC0 applies for PQClean code for Dilithium `aarch64` (as well as --the commercially less problematic-- MIT license)? In other words, from the perspective of #1437 an automated license check permitting only MIT and Apache code now has to exclude Dilithium `aarch64` code (which was not the case before this commit), right?

bhess: I'd personally agree with this interpretation, but tagging @beldmit for his feedback.
The CC0 was added in the commit because the main LICENSE file mentions only CC0, I've added an issue upstream to clarify if this is an oversight or intentional: https://github.com/PQClean/PQClean/issues/495.

dstebila (approved):

PR #1531, opened 2023-08-22
ci: add CI for android

res0nance:
Adds CI for the android platform for all ABIs.


dstebila:
Thanks!

I've pushed to our repository to actually trigger the builds and it seems to be working.  But for now it just does the build without any tests -- are there any meaningful tests we can do?

res0nance:
> Thanks!
> 
> I've pushed to our repository to actually trigger the builds and it seems to be working. But for now it just does the build without any tests -- are there any meaningful tests we can do?

I don't think there is any meaningful tests we can do here outside of compile testing. I added these builds to ensure that android doesn't accidentally get broken.

dstebila (approved):

PR #1533, opened 2023-08-24
LMS updated to use new SK API

ashman-p:

### DIFF
--- a/src/sig_stfl/lms/sig_stfl_lms_functions.c
+++ b/src/sig_stfl/lms/sig_stfl_lms_functions.c
@@ -18,7 +18,7 @@
 typedef struct OQS_LMS_KEY_DATA {
 
 	/* Tree levels. */
-	unsigned levels;
+	uint32_t levels;
 
 	/* Array, 8 levels max, of LMS types */
 	param_set_t lm_type[8];
@@ -27,18 +27,18 @@ typedef struct OQS_LMS_KEY_DATA {
 	param_set_t lm_ots_type[8];
 
 	/* LMS public key */
-	unsigned char public_key[60];
+	uint8_t public_key[60];
 
 	/* Length of aux data */
 	size_t len_aux_data;
 	/* internal nodes info of the Merkle tree */
-	unsigned char *aux_data;
+	uint8_t *aux_data;
 
 	/* Length of sec_key */
 	size_t len_sec_key;
 
 	/* secret key data */
-	unsigned char *sec_key;
+	uint8_t *sec_key;
 
 	/* User defined data that may be used for the SAFETY functions */
 	void *data;
@@ -144,6 +144,8 @@ int oqs_sig_stfl_lms_keypair(uint8_t *pk, OQS_SIG_STFL_SECRET_KEY *sk, const uin
 				return -1;
 			}
 		} else {
+			OQS_MEM_insecure_free(oqs_key_data);
+			oqs_key_data = NULL;
 			return -1;
 		}
 
@@ -247,7 +249,7 @@ int oqs_sig_stfl_lms_sign(OQS_SIG_STFL_SECRET_KEY *sk,
 
 	size_t sig_len;
 	bool status;
-	unsigned char *sig = NULL;
+	uint8_t *sig = NULL;
 	uint8_t *priv_key = NULL;
 	oqs_lms_key_data *oqs_key_data = NULL;
 	struct hss_working_key *w = NULL;
@@ -381,14 +383,14 @@ void oqs_secret_lms_key_free(OQS_SIG_STFL_SECRET_KEY *sk) {
  * Convert LMS secret key object to byte string
  * Writes secret key + aux data if present
  */
-size_t oqs_serialize_lms_key(OQS_SIG_STFL_SECRET_KEY *sk,  uint8_t **sk_key) {
+size_t oqs_serialize_lms_key(const OQS_SIG_STFL_SECRET_KEY *sk,  uint8_t **sk_key) {
 
 	oqs_lms_key_data *lms_key_data = NULL;
 	size_t key_len = 0;
 	if (sk) {
 		uint8_t *sk_key_buf = NULL;
 		lms_key_data = sk->secret_key_data;
-		if (lms_key_data->sec_key) {
+		if (lms_key_data && lms_key_data->sec_key) {
 			size_t buf_size_needed = lms_key_data->len_aux_data + lms_key_data->len_sec_key;
 			key_len = buf_size_needed;
 			/* pass back serialized data */
@@ -397,6 +399,10 @@ size_t oqs_serialize_lms_key(OQS_SIG_STFL_SECRET_KEY *sk,  uint8_t **sk_key) {
 					sk_key_buf = malloc(buf_size_needed * sizeof(uint8_t));
 					if (sk_key_buf) {
 
+						/*
+						 * Serialized data is sec_key followed by aux data
+						 * So aux data begins after buffer top + sec_key length
+						 */
 						if (lms_key_data->len_sec_key) {
 							memcpy(sk_key_buf, lms_key_data->sec_key, lms_key_data->len_sec_key);
 						}
@@ -407,8 +413,6 @@ size_t oqs_serialize_lms_key(OQS_SIG_STFL_SECRET_KEY *sk,  uint8_t **sk_key) {
 
 						*sk_key = sk_key_buf;
 						key_len = sk->length_secret_key + lms_key_data->len_aux_data;
-					} else {
-						OQS_MEM_insecure_free(sk_key_buf);
 					}
 				}
 			} //sk_key
@@ -418,11 +422,11 @@ size_t oqs_serialize_lms_key(OQS_SIG_STFL_SECRET_KEY *sk,  uint8_t **sk_key) {
 }
 
 /*
- * Convert LMS secret key object to byte string
+ * Convert LMS byte string to secret key object
  * Writes secret key + aux data if present
  * key_len is priv key length + aux length
  */
-int oqs_deserialize_lms_key(OQS_SIG_STFL_SECRET_KEY *sk, size_t key_len, uint8_t *sk_buf) {
+int oqs_deserialize_lms_key(OQS_SIG_STFL_SECRET_KEY *sk, size_t key_len, const uint8_t *sk_buf) {
 	int oqs_status = -1;
 	oqs_lms_key_data *lms_key_data = NULL;
 	uint8_t priv_ky_len = hss_get_private_key_len((unsigned )(1), NULL, NULL);
@@ -484,5 +488,5 @@ int oqs_deserialize_lms_key(OQS_SIG_STFL_SECRET_KEY *sk, size_t key_len, uint8_t
 		sk->secret_key_data = lms_key_data;
 		oqs_status = 0;
 	}
-	return oqs_status;;
+	return oqs_status;
 }
--- a/src/sig_stfl/lms/sig_stfl_lms_functions.c
+++ b/src/sig_stfl/lms/sig_stfl_lms_functions.c
@@ -9,8 +9,44 @@
 #include "sig_stfl_lms_wrap.h"
 #include <stdio.h>
 
+#define DEFAULT_AUX_DATA 10916   /* Use 10+k of aux data (which works well */
+/* with the above default parameter set)  */
+/**
+ * @brief OQS_LMS_KEY object for HSS key pair
+ */
+
+typedef struct OQS_LMS_KEY_DATA {
+
+	/* Tree levels. */
+	unsigned levels;
+
+	/* Array, 8 levels max, of LMS types */
+	param_set_t lm_type[8];
+
+	/* Array, 8 levels max, of LM OTS types */
+	param_set_t lm_ots_type[8];
+
+	/* LMS public key */
+	unsigned char public_key[60];
+
+	/* Length of aux data */
+	size_t len_aux_data;
+	/* internal nodes info of the Merkle tree */
+	unsigned char *aux_data;
+
+	/* Length of sec_key */
+	size_t len_sec_key;
+
+	/* secret key data */
+	unsigned char *sec_key;
+
+	/* User defined data that may be used for the SAFETY functions */
+	void *data;
+
+} oqs_lms_key_data;
+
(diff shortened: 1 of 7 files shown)


ashman-p:
Looks like variant XMSS-SHAKE_20_256 is hanging. This should be addressed after things settle with all the API changes.

src/sig_stfl/lms/sig_stfl_lms_functions.c
```diff
@@ -264,3 +350,139 @@ int oqs_sig_stfl_lms_verify(const uint8_t *m, size_t mlen,
...
+ * Writes secret key + aux data if present
+ */
+size_t oqs_serialize_lms_key(OQS_SIG_STFL_SECRET_KEY *sk,  uint8_t **sk_key) {
+
+	oqs_lms_key_data *lms_key_data = NULL;
+	size_t key_len = 0;
+	if (sk) {
+		uint8_t *sk_key_buf = NULL;
+		lms_key_data = sk->secret_key_data;
+		if (lms_key_data->sec_key) {
```
SWilson4, line 391:
```suggestion
		if (lms_key_data && lms_key_data->sec_key) {
```

SWilson4: Should probably check that lms_key_data is not null before dereferencing it. Or is that guaranteed at this point in the code?

src/sig_stfl/lms/sig_stfl_lms_functions.c
```diff
@@ -9,8 +9,44 @@
...
+	/* Length of aux data */
+	size_t len_aux_data;
+	/* internal nodes info of the Merkle tree */
+	unsigned char *aux_data;
+
+	/* Length of sec_key */
+	size_t len_sec_key;
+
+	/* secret key data */
+	unsigned char *sec_key;
```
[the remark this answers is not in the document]
SWilson4: This value is checked against `NULL` below, so we should probably make sure it's `NULL` to begin with.

src/sig_stfl/lms/sig_stfl_lms_functions.c
```diff
@@ -264,3 +350,139 @@ int oqs_sig_stfl_lms_verify(const uint8_t *m, size_t mlen,
...
+				if (buf_size_needed) {
+					sk_key_buf = malloc(buf_size_needed * sizeof(uint8_t));
+					if (sk_key_buf) {
+
+						if (lms_key_data->len_sec_key) {
+							memcpy(sk_key_buf, lms_key_data->sec_key, lms_key_data->len_sec_key);
+						}
+
+						if (lms_key_data->len_aux_data) {
+							memcpy(sk_key_buf + lms_key_data->len_sec_key, lms_key_data->sec_key, lms_key_data->len_aux_data);
```
SWilson4, line 405:
```suggestion
							memcpy(sk_key_buf + lms_key_data->len_sec_key, lms_key_data->aux_data, lms_key_data->len_aux_data);
```

tests/test_sig_stfl.c
```diff
@@ -379,22 +379,45 @@ static OQS_STATUS sig_stfl_test_secret_key(const char *method_name) {
...
 	if (sk == NULL) {
 		fprintf(stderr, "ERROR: OQS_SECRET_KEY_new failed\n");
 		goto err;
 	}
 
 	printf("================================================================================\n");
-	printf("Create stateful Secret Key  %s\n", method_name);
+	printf("Generate keypair  %s\n", method_name);
 	printf("================================================================================\n");
+	rc = OQS_SIG_STFL_keypair(sig_obj, public_key, sk);
```
ducnguyen-sb, line 410: Ah, you have to replace this function with `sig_stfl_KATs_keygen`, similar to line 308. 
Reason: at some parameter of XMSS, key generation takes about hours to finish. 
I think some variants of LMS also have very long key generation time. 
So that's why in function `sig_stfl_KATs_keygen`, it decides to generate keypair from the actual keypair function, or just simply load the key from KATs.

src/sig_stfl/lms/sig_stfl_lms_functions.c
```diff
@@ -264,3 +350,139 @@ int oqs_sig_stfl_lms_verify(const uint8_t *m, size_t mlen,
...
+						}
+
+						*sk_key = sk_key_buf;
+						key_len = sk->length_secret_key + lms_key_data->len_aux_data;
+					} else {
+						OQS_MEM_insecure_free(sk_key_buf);
```
SWilson4, line 411: I don't think it's necessary to free `sk_key_buff` here, right? If this branch is reached, `sk_key_buf` is set to null.

tests/test_sig_stfl.c
ducnguyen-sb: I hope this will fix problems with XMSS. @ashman-p

src/sig_stfl/lms/sig_stfl_lms_functions.c
SWilson4, line 411:
```suggestion
						key_len = 0;
```

src/sig_stfl/lms/sig_stfl_lms_functions.c
```diff
@@ -264,3 +350,139 @@ int oqs_sig_stfl_lms_verify(const uint8_t *m, size_t mlen,
...
+				OQS_MEM_insecure_free(lms_key_data);
+				OQS_MEM_insecure_free(lms_sk);
+				return oqs_status;
+			}
+		}
+
+		sk->secret_key_data = lms_key_data;
+		oqs_status = 0;
+	}
+	return oqs_status;;
```
SWilson4, line 487:
```suggestion
	return oqs_status;
```

src/sig_stfl/lms/sig_stfl_lms_functions.c
```diff
@@ -264,3 +350,139 @@ int oqs_sig_stfl_lms_verify(const uint8_t *m, size_t mlen,
...
+					}
+				}
+			} //sk_key
+		}
+	} //sk
+	return key_len;
+}
+
+/*
+ * Convert LMS secret key object to byte string
```
SWilson4, line 421:
```suggestion
 * Convert LMS byte string to secret key object
```

src/sig_stfl/sig_stfl.h
```diff
@@ -206,21 +206,23 @@ typedef struct OQS_SIG_STFL_SECRET_KEY {
...
 	 * @param[in] sk The secret key represented as OQS_SIG_STFL_SECRET_KEY object
-	 * @param[out] key_len length of the returned byte string
-	 * @returns newly created pointer to ley byte string if none-zero length. Caller
-	 * deletes the buffer.
+	 * @param[out] sk_buf private key data as a byte stream
+	 * @returns length of key material data available
+	 * Caller deletes the buffer if memory was allocated.
 	 */
-	uint8_t *(*serialize_key)(OQS_SIG_STFL_SECRET_KEY *sk, size_t key_len);
+	size_t (*serialize_key)(OQS_SIG_STFL_SECRET_KEY *sk,  uint8_t **sk_buf);
```
SWilson4, line 213:
```suggestion
	size_t (*serialize_key)(const OQS_SIG_STFL_SECRET_KEY *sk,  uint8_t **sk_buf);
```

src/sig_stfl/lms/sig_stfl_lms_functions.c
```diff
@@ -264,3 +350,139 @@ int oqs_sig_stfl_lms_verify(const uint8_t *m, size_t mlen,
...
+		OQS_MEM_insecure_free(key_data);
+		sk->secret_key_data = NULL;
+	}
+}
+
+/*
+ * Convert LMS secret key object to byte string
+ * Writes secret key + aux data if present
+ */
+size_t oqs_serialize_lms_key(OQS_SIG_STFL_SECRET_KEY *sk,  uint8_t **sk_key) {
```
SWilson4, line 384:
```suggestion
size_t oqs_serialize_lms_key(const OQS_SIG_STFL_SECRET_KEY *sk,  uint8_t **sk_key) {
```

Seems like we can make this `const`.

src/sig_stfl/lms/sig_stfl_lms_functions.c
```diff
@@ -264,3 +350,139 @@ int oqs_sig_stfl_lms_verify(const uint8_t *m, size_t mlen,
...
+	} //sk
+	return key_len;
+}
+
+/*
+ * Convert LMS secret key object to byte string
+ * Writes secret key + aux data if present
+ * key_len is priv key length + aux length
+ */
+int oqs_deserialize_lms_key(OQS_SIG_STFL_SECRET_KEY *sk, size_t key_len, uint8_t *sk_buf) {
```
SWilson4, line 425:
```suggestion
int oqs_deserialize_lms_key(OQS_SIG_STFL_SECRET_KEY *sk, size_t key_len, const uint8_t *sk_buf) {
```

src/sig_stfl/sig_stfl.h
```diff
@@ -206,21 +206,23 @@ typedef struct OQS_SIG_STFL_SECRET_KEY {
...
-	 * deletes the buffer.
+	 * @param[in]  sk OQS_SIG_STFL_SECRET_KEY object
+	 * @param[in]  key_len length of the returned byte string
+	 * @param[in]  sk_key The secret key represented as OQS_SIG_STFL_SECRET_KEY object
+	 * @param[in]  key_len length of the returned byte string
+	 * @returns  status of the operation populated with key material none-zero length. Caller
+	 * deletes the buffer. if sk_buf is NULL the function returns the length
 	 */
-	uint8_t *(*deserialize_key)(OQS_SIG_STFL_SECRET_KEY *sk, size_t key_len, uint8_t *sk_key);
+	OQS_STATUS (*deserialize_key)(OQS_SIG_STFL_SECRET_KEY *sk, size_t key_len, uint8_t *sk_buf);
```
SWilson4, line 225:
```suggestion
	OQS_STATUS (*deserialize_key)(OQS_SIG_STFL_SECRET_KEY *sk, size_t key_len, const uint8_t *sk_buf);
```

SWilson4 (commented):
A few small suggestions from me, but overall everything looks good.

src/sig_stfl/lms/sig_stfl_lms_functions.c
```diff
@@ -264,3 +350,139 @@ int oqs_sig_stfl_lms_verify(const uint8_t *m, size_t mlen,
...
+	if (sk->secret_key_data) {
+		oqs_lms_key_data *key_data = (oqs_lms_key_data *)sk->secret_key_data;
+		if (key_data) {
+			OQS_MEM_secure_free(key_data->sec_key, key_data->len_sec_key);
+			key_data->sec_key = NULL;
+
+			OQS_MEM_secure_free(key_data->aux_data, key_data->len_aux_data);
+		}
+
+		OQS_MEM_insecure_free(key_data->data);
```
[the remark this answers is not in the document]
ashman-p: This is intended to be a variant specific deep cleaning function with knowledge of the internals.

src/sig_stfl/lms/sig_stfl_lms_functions.c
ashman-p: Added this comment here...
					    /* 
					     * Serialized data is sec_key followed by aux data
					     * So aux data begins after buffer top + sec_key length
					     */

src/sig_stfl/lms/sig_stfl_lms_functions.c
ashman-p: Thought was find out the size of the serialized data without actually doing the work.

tests/test_sig_stfl.c
ashman-p: Added temp logic to exclude algs that take too long to generate keys or otherwise don't have a KAT file.

src/sig_stfl/lms/sig_stfl_lms_functions.c
SWilson4: Right, but aux_data isn't actually being copied here, sec_key is being copied twice.

ashman-p: Yep. Got it.

ashman-p:
Will look at the macOS build in a subsequent commit since the issue seems not related to the current commit.

ashman-p:
Will look at the macOS build in a subsequent commit since the issue seems not related to the current commit.

src/sig_stfl/lms/sig_stfl_lms_functions.c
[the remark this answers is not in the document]
ashman-p: Let me know if you are ok with this?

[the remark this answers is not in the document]
SWilson4: My concern was that, since the `data` field is described as user-defined, the user could (say) set `data` to point to stack memory, in which case freeing `data` would cause serious problems. Also, the `data` field is also not initialized to `NULL` by default, so freeing it could cause problems unless it's already been `malloc`ed, and it's not `malloc`ed anywhere that I can see.

That said, I'm a little shaky on what the purpose of the `data` pointer is, so I might be missing something.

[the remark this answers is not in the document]
SWilson4: If this isn't actually a concern and/or will be addressed in a future PR, feel free to resolve this thread.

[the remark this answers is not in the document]
ashman-p: I will just remove it until we hit/need code for a use case.

src/sig_stfl/lms/sig_stfl_lms_functions.c
SWilson4: Should `lms_key_data->len_sec_key` be used instead of `sk->length_secret_key`, then? (Since that's the value that was passed to `memcpy`.) Or will those values always be identical?

SWilson4: (I'm assuming here that `key_len` is meant to return the length of the serialized string; if it's meant to be an upper bound or some other value let me know.)

ashman-p: > 

Yes, the length of the serialized string.

SWilson4 (approved):

PR #1537, opened 2023-08-30
Pull FIPS draft versions of Kyber/Dilithium

### LINKED ISSUE #1521: Update Kyber and Dilithium to FIPS versions
Once the FIPS drafts are released, we'll need to update Kyber and Dilithium based on the tweaks added by NIST.  

Upstream already has started on this:
- https://github.com/pq-crystals/kyber/tree/standard
- https://github.com/pq-crystals/dilithium/tree/standard

bhess:
Pulls standard-version of pq-crystals repo (fips203/204 drafts).

TODOs:

- [ ] currently depends on a few fixes & yml updates upstream (https://github.com/pq-crystals/dilithium/pull/76, https://github.com/pq-crystals/kyber/pull/62) -> point to standard branch before merging
- [x] pull kyber upstream (fips203 draft)
- [x] pull dilithium upstream (fips204 draft)
- [ ] currently doesn't include optimized aarch64 versions: either wait for it to land in pqclean or merge without
- [x] decide on naming (~~could~~ stay with kyber/dilithium until FIPS graduates from draft status)
- [x] update CT-files

Fixes #1521 


baentsch:
> stay with kyber/dilithium until FIPS graduates from draft status

+1

baentsch:
> doesn't include optimized aarch64 versions: either wait for it to land in pqclean or merge without

"merge without" would be my preference to move forward with oqsprovider & interop testing

dstebila:
Regarding the aarch64 implementation, @mkannwischer is working on that and can provide an update on his path to that.

As for naming, yes, I guess it makes sense to stick with the name Kyber for now and then switch to ML-KEM once the FIPS documents are finalized.  @cryptojedi, any opinion?

cryptojedi:
Douglas Stebila ***@***.***> wrote:
> Regarding the aarch64 implementation, @mkannwischer is working on that
> and can provide an update on his path to that.
> 
> As for naming, yes, I guess it makes sense to stick with the name
> Kyber for now and then switch to ML-KEM once the FIPS documents are
> finalized.  @cryptojedi, any opinion?

Sounds good to me. Could also wait until the name is fixed for TLS
and/or SSH, right?

baentsch:
> Could also wait until the name is fixed for TLS and/or SSH, right?

As well as the OIDs, yes. (Only) The reference code (with old names/proprietary OIDs) is waited for (integration) e.g., for [interop testing](https://github.com/IETF-Hackathon/pqc-certificates).

baentsch:
What about merging them into a(n equally named) "standard" branch in `liboqs`? That would allow us to ("test-")integrate them in downstream projects (most notably `oqsprovider` to make progress in interop testing): Arguably, (optimized ARM or actually any) performance doesn't matter there, just functionality. Integration to "main" could be done as and when the upstream projects catch up.

bhess:
Sounds good to me @baentsch.

For the full PR there are two open points:
- aarch64 optimization
- the dependency on two PRs in pq-crystals upstream (https://github.com/pq-crystals/dilithium/pull/76, https://github.com/pq-crystals/kyber/pull/62). @cryptojedi any feedback on them?

baentsch:
> Sounds good to me @baentsch.

Alright then -- can you re-target this PR to a different branch ("standard") then? Or just create a separate PR (or branch) for that (that we then maintain until this PR closes for good with the upstream updates merged)?

bhess:
> Alright then -- can you re-target this PR to a different branch ("standard") then? Or just create a separate PR (or branch) for that (that we then maintain until this PR closes for good with the upstream updates merged)?

See https://github.com/open-quantum-safe/liboqs/pull/1543. I think the PR here could stay open until completed.

dstebila:
Thanks Basil for getting this ready so quickly. I guess this means we should re-question whether to include this in 0.9.0 or save for 0.10.0.  I think it would be nice to get upstream approval of the fixes in the pq-crystals repository, but hopefully that's straightforward.  So that would leave us in the position of deciding whether to push this 0.10.0 due the lack of aarch64 optimized implementation.  Thoughts?  Tagging @cryptojedi and @mkannwischer for input as well.

baentsch:
>  So that would leave us in the position of deciding whether to push this 0.10.0 due the lack of aarch64 optimized implementation. Thoughts?

In the interest to have release "0.9.0" support all "final" algorithms (and use that in interop testing with other implementations [scheduled for end September](https://github.com/IETF-Hackathon/pqc-certificates/discussions/70)) I would very much support including this in 0.9.0 (assuming we do this before end September) -- even with the consequence of dropping `aarch64` optimizations for now: 
a) We still support `aarch64` (via ref) 
b) This would substantially reduce effort in several projects: Maintaining a separate branch in `liboqs` (see #1543) as well as all downstreams interested in using this code base having to introduce the same "standard" branch maintenance duplicates effort and 
c) creates confusion: "Why are standard algorithms not the default in `liboqs`?"

What worries me more in this PR though, is that the upstreams do not seem to have a notion of version. This has been updated only locally in https://github.com/open-quantum-safe/liboqs/pull/1543: I'd think this either needs to be added also to this PR (with `liboqs` assuming responsibility for algorithm versioning) -- or ideally upstream.

baentsch:
@SWilson4 Please discuss in the next team meeting whether to heed the above (include in 0.9.0 --properly versioned-- or not).

dstebila:
> > I'd think this either needs to be added also to this PR (with `liboqs` assuming responsibility for algorithm versioning)
> 
> This PR and #1543 have the same source branch, the update is also applied here.
> 
> Vadim posted a [message](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/EkeX7jk9LRI/m/OY3jYaTTAwAJ?utm_medium=email&utm_source=footer) that highlights a few inconsistencies/typos in the FIPS 204 draft, compared to the pq-crystals/standard implementation. The algorithms appear to match apart from one minor difference, which can lead to different KAT and signatures in a "literal" implementation of the draft. This might be relevant for interop. I'd therefore tend to state, for example, "FIPS 204 (Draft, August 24, 2023), implemented by pq-crystals/standard" (for spec-version, if we do this in liboqs). Any thoughts?

Do we know if NIST is expected to make this change?

baentsch:
>  I'd therefore tend to state, for example, "FIPS 204 (Draft, August 24, 2023), implemented by pq-crystals/standard" (for spec-version, if we do this in liboqs). Any thoughts?

I'm personally no big fan of version strings with more letters than numbers, let alone some code exegesis. I also wouldn't refer to FIPS if there is a possibility that the FIPS standard could be implemented differently. If pqcrystals doesn't assign a version, what about simply "pqcrystals 200823"?

dstebila:
I talked with @cryptojedi this afternoon, and he tells me that there are incompatibilities between the current FIPS drafts and PQCrystals standard branch for both Kyber and Dilithium. He says that they've told NIST of the changes that they think should be made in the FIPS drafts, but there's no confirmation yet whether those changes will be made.  And for Dilithium, there's still a change that he thinks should be made that isn't yet in the PQCrystals standard branch.

His suggestion was that it would be better not to include the Kyber and Dilithium updates in the 0.9.0 release, and instead keep them in a separate branch for now which can be tagged for those wanting to use it in the IETF PQC PKI hackathon.

baentsch:
> His suggestion was that it would be better not to include the Kyber and Dilithium updates in the 0.9.0 release, and instead keep them in a separate branch for now which can be tagged for those wanting to use it in the IETF PQC PKI hackathon.

OK -- then this PR may be kept in Draft until the discussion with NIST concludes. Given there's no date for this and the PR's anyway already in need of re-base, closing it seems the prudent thing to do. Please re-open if you disagree, @bhess .

#1543 just merged to separate (arguably somewhat inaccurately named) "standard" branch. Will follow up with a PR for `oqsprovider` using that so that can be used in interop testing.

Issue #1540, opened 2023-09-07
Environment-specific Classic McEliece constant-time leaks


praveksharma:
**Describe the bug**
The testing suite is reporting [potential memory leaks](https://github.com/open-quantum-safe/liboqs/pull/1527#issuecomment-1701305585) ([test_constant_time.txt](https://github.com/open-quantum-safe/liboqs/files/12553409/test_constant_time.txt)) for Classic McEliece which are not documented in the suppression files.

**To Reproduce**
I am unable to reproduce this issue locally. Additionally, these errors aren't being caught by the CI during weekly constant time tests. One might try to reproduce this with following steps:
```
docker run -dti openquantumsafe/ci-ubuntu-focal-x86_64 /bin/bash
docker attach (output of docker run command)
cd
git clone https://github.com/open-quantum-safe/liboqs
cd liboqs
mkdir build
cd build
cmake -GNinja -DOQS_OPT_TARGET=generic -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON ..
ninja
cd ..
SKIP_ALGS='BIKE*,Frodo*,HQC*,Kyber*,NTRU*,stru*,ntru*,LightSaber,Saber,FireSaber,Dilithium*,Falcon*,,SPHINCS*' python3 tests/test_constant_time.py  --verbose
```

Thank you for sharing this issue @bhess! Could you please share information about your environment: 
**Environment (please complete the following information):**
 - OS: [e.g. Ubuntu 20]
 - OpenSSL version [e.g., 3.0.2]
 - Compiler version used [e.g., clang 9.0.0]
 - Build variables used [e.g., "-DOQS_ALGS_ENABLED=STD"]
 - liboqs version [e.g. 0.7.2 or main branch]

praveksharma:
Addressing the (potential) leaks documented in this issue doesn't fix the larger issue of the current Classic McEliece implementation (potentially) not being constant time in certain environments. Aside from documenting this in the advisories (as in #1541) should we be taking additional measures for the final 0.9.0 release?

bhess:
Thank you for picking this up @praveksharma !

The environment is:

OS: Ubuntu 22.04.3 LTS
OpenSSL version 3.0.2
Compiler version used: GCC 11.4.0
Build variables used: -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON
liboqs version: main

[...]

PR #1541, opened 2023-09-07
Update Classic McEliece suppression files

praveksharma:
Updates suppression files to include potential leaks document in #1540.
Updates Classic McEliece advisories to include information about environment-specific constant time leaks.
Removes Classic-McEliece-6XX from weekly constant time test skip list. 


praveksharma:
@bhess I was hoping you could help me work on this issue, since I can't reproduce these errors locally. Could you check if Valgrind still complains about potential constant time leaks? 

It might be a good idea to add `--error-limit=no` to valgrind's [options list](https://github.com/open-quantum-safe/liboqs/blob/0dafd4f02c5a8fc000502ae51005a9962b671860/tests/test_constant_time.py#L193) to ensure valgrind reports all errors.

baentsch (commented):
Thanks for keeping updating the suppression files for this algorithm family.

As this PR adds this (IMO very correct) advisory:

> Current implementation of the algorithm may not be constant-time

and looking at the length/extent of the suppression files and the above advisory I wonder whether it wouldn't be simpler to _completely_ exclude McEliece from constant time testing until someone states interest in giving McEliece the CT property (?) We have something similar [for BIKE](https://github.com/open-quantum-safe/liboqs/blob/main/tests/constant_time/kem/issues/bike_has_no_timing_protections) -- although I wonder whether this is (still) correct given there's no corresponding advisory for BIKE (???).

bhess:
> @bhess I was hoping you could help me work on this issue, since I can't reproduce these errors locally. Could you check if Valgrind still complains about potential constant time leaks?

I ran the tests and there is one error remaining. Seems to be all fine when adding one more rule.
```
{
   This implementation of Classic McEliece may not be constant time.
   Memcheck:Value8
   src:pk_gen.c:322
   # fun:PQCLEAN_MCELIECE348864_AVX2_pk_gen
   fun:PQCLEAN_MCELIECE348864_AVX2_crypto_kem_keypair
}
```

praveksharma:
> looking at the length/extent of the suppression files and the above advisory I wonder whether it wouldn't be simpler to _completely_ exclude McEliece from constant time testing until someone states interest in giving McEliece the CT property (?) 

> We have something similar [for BIKE](https://github.com/open-quantum-safe/liboqs/blob/main/tests/constant_time/kem/issues/bike_has_no_timing_protections) -- although I wonder whether this is (still) correct given there's no corresponding advisory for BIKE (???).

Looking through the [file history](https://github.com/open-quantum-safe/liboqs/commits/main/tests/constant_time/kem/issues/bike_has_no_timing_protections) for the BIKE suppression file, it looks like the advisories weren't updated when the suppression file was created. I've added a similar advisory for BIKE. 

I've updated the suppression file with the remaining error, we should be good to go.

bhess (commented):
The McEliece ct-run looks ok on my test machine. Thanks for the update!

However, I didn't notice any BIKE errors. There also seem to be no issues registered in issues.json. I think [this file](https://github.com/open-quantum-safe/liboqs/blob/main/tests/constant_time/kem/issues/bike_has_no_timing_protections) could be removed since ct-fixes to BIKE were made in https://github.com/open-quantum-safe/liboqs/pull/1400.

.github/workflows/weekly.yml
```diff
@@ -16,7 +16,7 @@ jobs:
             container: openquantumsafe/ci-ubuntu-focal-x86_64:latest
             CMAKE_ARGS: -DOQS_OPT_TARGET=generic -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON
             PYTEST_ARGS: --numprocesses=auto -k 'test_constant_time'
-            SKIP_ALGS: 'SPHINCS\+-SHA*,Classic-McEliece-6(.)*'
+            SKIP_ALGS: 'SPHINCS\+-SHA*, Classic-McEliece-(.)*, BIKE(.)*'
```
baentsch, line 19: Skipping BIKE seems incorrect as per #1400.

docs/algorithms/kem/bike.md
```diff
@@ -11,6 +11,9 @@
 - **Ancestors of primary source**:
   - https://bikesuite.org/files/v5.0/Reference_Implementation.2022.10.04.1.zip
 
+## Advisories
+This implementation of BIKE is not constant time.
```
baentsch, line 15: Changing this only in the .MD and not the .YML will not survive the next execution of `copy_from_upstream.py`.

docs/algorithms/kem/classic_mceliece.md
```diff
@@ -14,6 +14,7 @@
 ## Advisories
 
 - Classic-McEliece-460896, Classic-McEliece-460896f, Classic-McEliece-6960119, and Classic-McEliece-6960119f parameter sets fail memory leak testing on x86-64 when building with ``clang`` using optimization level ``-O2`` and ``-O3``. Care is advised when using the algorithm at higher optimization levels, and any other compiler and architecture.
+- Current implementation of the algorithm may not be constant-time. Additionally, environment specific constant-time leaks may not be documented; please report potential constant-time leaks when found.
```
baentsch, line 17: Same here. This ought to be in the YML (too) from which the MD is generated.

baentsch (changes requested):
Beyond the single comment made I concur with @bhess that 

> [this file](https://github.com/open-quantum-safe/liboqs/blob/main/tests/constant_time/kem/issues/bike_has_no_timing_protections) could be removed since ct-fixes to BIKE were made in https://github.com/open-quantum-safe/liboqs/pull/1400.

baentsch (approved):

PR #1542, opened 2023-09-07
Add XMSS Serialize/Deserialize 

ducnguyen-sb:
# LMS 

This PR slightly modify #1533 
- Return OQS_STATUS
- Flatten control-flow

I tried to keep the same logic as possible. Please let me know if there is something wrong @ashman-p 

# XMSS

Add serialize and deserialize function to XMSS. 


src/sig_stfl/xmss/sig_stfl_xmss_sha256_h20.c
```diff
@@ -56,6 +56,10 @@ OQS_SIG_STFL_SECRET_KEY *OQS_SECRET_KEY_XMSS_SHA256_H20_new(void) {
 	sk->sigs_left = NULL;
 	sk->sigs_total = NULL;
 
+	// Secret serialize/deserialize function
+	sk->serialize_key = OQS_SECRET_KEY_XMSS_serialize_key;
+	sk->deserialize_key = OQS_SECRET_KEY_XMSS_deserialize_key;
+
 	// Initialize the key with length_secret_key amount of bytes.
 	sk->secret_key_data = (uint8_t *)malloc(sk->length_secret_key * sizeof(uint8_t));
```
ashman-p, line 64: Check for NULL? Other places need this too.

src/sig_stfl/xmss/sig_stfl_xmss_sha256_h10.c
```diff
@@ -56,6 +56,10 @@ OQS_SIG_STFL_SECRET_KEY *OQS_SECRET_KEY_XMSS_SHA256_H10_new(void) {
 	sk->sigs_left = NULL;
 	sk->sigs_total = NULL;
 
+	// Secret serialize/deserialize function
+	sk->serialize_key = OQS_SECRET_KEY_XMSS_serialize_key;
+	sk->deserialize_key = OQS_SECRET_KEY_XMSS_deserialize_key;
+
 	// Initialize the key with length_secret_key amount of bytes.
 	sk->secret_key_data = (uint8_t *)malloc(sk->length_secret_key * sizeof(uint8_t));
```
ashman-p, line 64: Check for NULL after malloc.

src/sig_stfl/xmss/sig_stfl_xmss_sha256_h20.c
ducnguyen-sb: thanks @ashman-p , I fixed in https://github.com/open-quantum-safe/liboqs/pull/1542/commits/67d3aebf4c0574faa7a754373cb21129a3a37073

src/sig_stfl/xmss/sig_stfl_xmss_sha256_h10.c
ducnguyen-sb: fixed in https://github.com/open-quantum-safe/liboqs/pull/1542/commits/67d3aebf4c0574faa7a754373cb21129a3a37073

ashman-p (approved):

ashman-p (approved):

src/sig_stfl/xmss/sig_stfl_xmss_secret_key_functions.c
```diff
@@ -0,0 +1,42 @@
...
+
+	if (sk->secret_key_data) {
+		// Key data already present
+		// We dont want to trample over data
+		return OQS_ERROR;
+	}
+
+	memcpy(sk->secret_key_data, sk_buf, sk_len);
+
+	return OQS_SUCCESS;
```
ducnguyen-sb, line 41: So I have this piece of code that we discuss in the meeting: 
So we check if the address of array `secret_key_data` is either NULL or not. If it's not NULL, we assume it's already initialized, thus we return error. 
To reach line 39, `sk->secret_key_data` must be NULL. Which make the `memcpy` is SEGFAULT. 

I'm thinking of creating `static once = 1` as static variable to make sure the initialization happen only once. 
But this lead to:

- We will have 2 ways to allocate memory for `sk->secret_key_data`, either in `deserialize`, and in key generation.
- If I use `static` variable, I don't know in the case we multithread/fork process, would it become complicate to make sure the `sk->secret_key_data` happen only once with `static` variable.

ducnguyen-sb: After a while I change the logic of the code to:

```c
/* Deserialize XMSS byte string into an XMSS secret key data */
OQS_STATUS OQS_SECRET_KEY_XMSS_deserialize_key(OQS_SIG_STFL_SECRET_KEY *sk, const size_t sk_len, const uint8_t *sk_buf) {
	if (sk == NULL || sk_buf == NULL || (sk_len != sk->length_secret_key)) {
		return OQS_ERROR;
	}

	if (sk->secret_key_data != NULL) {
		// Key data already present
		// We dont want to trample over data
		return OQS_ERROR;
	}

	sk->secret_key_data = malloc(sk_len);

	if (sk->secret_key_data == NULL) {
		return OQS_ERROR;
	}

	memcpy(sk->secret_key_data, sk_buf, sk_len);

	return OQS_SUCCESS;
}
```

What do you think @ashman-p @dstebila ?

ducnguyen-sb: Commited in https://github.com/open-quantum-safe/liboqs/pull/1542/commits/7e4f2c93ddd90900320e1f74163efbe23938fc4a

ashman-p: This looks fine to me.
It might be good to walk through the use-case where it is necessary to call "deserialize' on a working secret key object. Those use-cases would help determine the desired behaviors for each expected outcome.

ducnguyen-sb:
This looks good to me. I'll merge this PR.

PR #1543, opened 2023-09-08
Kyber/Dilithium fips versions -> standard branch

bhess:
PR to merge the Kyber/Dilithium (fips draft versions) to a standard branch. May be obsolete once #1537 is resolved.


docs/algorithms/sig/dilithium.yml
```diff
@@ -15,22 +15,16 @@ website: https://pq-crystals.org/dilithium/
 nist-round: 3
 spec-version: 3.1
```
baentsch, line 16: Looks fishy to me: No change to spec version but different KATs?

docs/algorithms/kem/kyber.yml
```diff
@@ -17,15 +17,9 @@ website: https://pq-crystals.org/
 nist-round: 3
 spec-version: NIST Round 3 submission
```
baentsch, line 18: Looks fishy to me: No change to spec version but different KATs?

baentsch (changes requested):
Can I ask for a change to the version ID of the algs, please? Otherwise https://github.com/open-quantum-safe/oqs-provider/blob/main/oqs-template/oqs-sig-info.md for example will show the same algorithm "Implementation Version" for a different OID: Would be IMO pretty confusing.

docs/algorithms/sig/dilithium.yml
bhess: updated in 743c15ceea63ac88631d1ec16bab1ea6061eb6f0

docs/algorithms/kem/kyber.yml
bhess: updated in 743c15ceea63ac88631d1ec16bab1ea6061eb6f0

baentsch (approved):
Thanks! LGTM now.

baentsch:
Thanks again for PR and updates -- I'd suggest to wait with merge until Sep 13 to see whether we _really_ have to go this route (or a straight merge as per https://github.com/open-quantum-safe/liboqs/pull/1537#issuecomment-1712127229 is preferred (@dstebila @cryptojedi ?) and we can avoid maintaining separate branches). If we need to do it, I'll then set up a separate branch in `oqsprovider` with the new OIDs you provided to BouncyCastle so you don't have to bother with that too.

PR #1544, opened 2023-09-08
Prefer arc4random on Apple platforms

res0nance:
On Apple platforms arc4random is always available and it never fails. This removes a possible failure mode of liboqs on Apple platforms with `getentropy()`.

The man page for `getentropy()` on MacOSX also discourages its use

```
However, it should be noted that getentropy() is primarily intended for use in the construction and seeding of userspace PRNGs like
     arc4random(3) or CC_crypto(3).  Clients who simply require random data should use arc4random(3), CCRandomGenerateBytes() from
     CC_crypto(3), or SecRandomCopyBytes() from the Security framework instead of getentropy() or random(4)
```

On iOS we currently use SecRandomCopyBytes() this requires linking to the Security framework and the call can fail. We have run into issues with this call failing so we propose using `arc4random_buf()` as it can never fail.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -60,10 +60,6 @@ endif()
 if(${OQS_USE_OPENSSL})
     target_link_libraries(oqs PRIVATE ${OPENSSL_CRYPTO_LIBRARY})
 endif()
-if(IOS)
-    find_library(IOS_SECURITY Security)
-    target_link_libraries(oqs PUBLIC ${IOS_SECURITY})
-endif()
 
 target_include_directories(oqs
                            PUBLIC
--- a/src/common/rand/rand.c
+++ b/src/common/rand/rand.c
@@ -8,14 +8,7 @@
 #else
 #include <unistd.h>
 #include <strings.h>
-#if defined(__APPLE__)
-#include <TargetConditionals.h>
-#if TARGET_OS_IPHONE || TARGET_IPHONE_SIMULATOR
-#include <Security/SecRandom.h>
-#else
-#include <sys/random.h>
-#endif
-#else
+#if !defined(__APPLE__)
 #include <unistd.h>
 #endif
 #endif
@@ -65,6 +58,11 @@ OQS_API void OQS_randombytes(uint8_t *random_array, size_t bytes_to_read) {
 }
 
 #if !defined(_WIN32)
+#if defined(__APPLE__)
+void OQS_randombytes_system(uint8_t *random_array, size_t bytes_to_read) {
+	arc4random_buf(random_array, bytes_to_read);
+}
+#else
 #if defined(OQS_HAVE_GETENTROPY)
 void OQS_randombytes_system(uint8_t *random_array, size_t bytes_to_read) {
 	while (bytes_to_read > 256) {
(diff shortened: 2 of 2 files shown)


dstebila:
We don't have an iOS test platform; have you been able to test it there to confirm it builds?

res0nance:
> We don't have an iOS test platform; have you been able to test it there to confirm it builds?

I have tested it on tvOS, iOS, macOS m1. They all compile fine.

Maybe we can use the file in https://github.com/leetal/ios-cmake to setup builds for the iOS / tvOS platforms? I'm assuming the licensing is fine. I can set that up in a different PR.

baentsch (changes requested):
Thanks for the suggestion(s) and this PR to improve building on those platforms!

One concern, though

> so we propose using arc4random_buf() as it can never fail.

Can you please link to documentation confirming that `arc4random_buf` is at least as secure as `SecRandomCopyBytes`? Something with "RC4" in its name doesn't create warm and fuzzy feelings about its security properties :-)

Otherwise the PR LGTM, particularly if you confirm

> tested it on tvOS, iOS, macOS m1. They all compile fine.

res0nance:
> Thanks for the suggestion(s) and this PR to improve building on those platforms!
> 
> One concern, though
> 
> > so we propose using arc4random_buf() as it can never fail.
> 
> Can you please link to documentation confirming that `arc4random_buf` is at least as secure as `SecRandomCopyBytes`? Something with "RC4" in its name doesn't create warm and fuzzy feelings about its security properties :-)
> 
> Otherwise the PR LGTM, particularly if you confirm
> 
> > tested it on tvOS, iOS, macOS m1. They all compile fine.

Ah the "RC4" name is from the old days, from the manpage

```
The original version of this random number generator used the RC4
    (also known as ARC4) algorithm. In OS X 10.12 it was replaced with the
    NIST-approved AES cipher, and it may be replaced again in the future as
    cryptographic techniques advance
```

On the topic of being secure, from the manpage

```
These functions use a cryptographic pseudo-random number generator
    to generate high quality random bytes very quickly. One data pool is used
    for all consumers in a process, so that consumption under program flow can
    act as additional stirring. The subsystem is re-seeded from the kernel
    random number subsystem on a regular basis
    ```
    
It seems like the intent of apple here is to provide a simple to use cryptographically-secure random  number generator that they will update as technology progresses.

dstebila:
> > so we propose using arc4random_buf() as it can never fail.
> 
> Can you please link to documentation confirming that `arc4random_buf` is at least as secure as `SecRandomCopyBytes`? Something with "RC4" in its name doesn't create warm and fuzzy feelings about its security properties :-)

I was worried about that too, but also found the passage that res0nance quoted saying that it's no longer specifically using RC4. So I think it's okay.

dstebila (approved):

baentsch (approved):

baentsch:
> Maybe we can use the file in https://github.com/leetal/ios-cmake to setup builds for the iOS / tvOS platforms? I'm assuming the licensing is fine. I can set that up in a different PR.

Sounds like a nice addition to avoid regression, too. Looking forward to the PR. Any chance to _run_ the result in CI, too (in an emulator)?

PR #1546, opened 2023-09-11
ci: add CI for apple mobile platforms

res0nance:
Adds CI for Apple mobile platforms.

Unfortunately I'm not sure how to effectively run the tests within the emulator. Plus it will likely be incredibly flaky. I think we should compile test this until we can figure out how to better run the full suite of tests.


baentsch (approved):
LGTM. Brings build-testing to the same level as for android (#1531). Thanks for the contribution!

dstebila (approved):

PR #1549, opened 2023-09-11
Call Keccak_(X4_)Dispatch with pthread_once

### LINKED ISSUE #1548: Keccak_(X4_)Dispatch Initialize_ptr assignment is thread unsafe
`Keccak_(X4_)Dispatch` `Initialize_ptr` assignment is thread unsafe due to race condition. `Keccak_Dispatch` in [xkcp_sha3.c](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/sha3/xkcp_sha3.c#L42) and `Keccak_X4_Dispatch` in [xkcp_sha3x4.c](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/sha3/xkcp_sha3x4.c#L36).

Example execution:
* Thread 0 executes Keccak_X4_Dispatch until and including [this line](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/sha3/xkcp_sha3x4.c#L36), setting `Keccak_X4_Initialize_ptr = &KeccakP1600times4_InitializeAll_avx2;`. Pauses here for the rest of this example.
* Thread 1 calls [`gen_matrix()`](https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/kyber/pqcrystals-kyber_kyber768_avx2/indcpa.c#L173) -> [`shake128x4_inc_init()`](https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/kyber/pqcrystals-kyber_kyber768_avx2/indcpa.c#L245) -> [`keccak_x4_inc_reset()`](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/sha3/xkcp_sha3x4.c#L166) -> [`(*Keccak_X4_Initialize_ptr)()`](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/sha3/xkcp_sha3x4.c#L67).
  * Here, `Keccak_X4_Initialize_ptr` is [`&KeccakP1600times4_InitializeAll_avx2`](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/sha3/xkcp_low/KeccakP-1600times4/avx2/KeccakP-1600-times4-SIMD256.c#L108), which just `memset`s without issue and returns.
* Thread 1 goes back up to `gen_matrix()`, next line is to call [`shake128x4_absorb_once()`](https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/kyber/pqcrystals-kyber_kyber768_avx2/indcpa.c#L246) -> [`shake128x4_inc_absorb()`](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/pqclean_shims/fips202x4.c#L7), which is a [macro](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/pqclean_shims/fips202x4.h#L20) for [`OQS_SHA3_shake128_x4_inc_absorb`](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/sha3/xkcp_sha3x4.c#L169) and a tail call for [`keccak_x4_inc_absorb()`](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/sha3/xkcp_sha3x4.c#L170), which tries to call [`(*Keccak_X4_AddBytes_ptr)()`](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/sha3/xkcp_sha3x4.c#L102).
  * `Keccak_X4_AddBytes_ptr` is still `NULL` because Thread 0 paused before setting it! Hence, calls 0x0 and segfaults.

Made a fix using `pthread_once` in #1549 but not cross-platform, open to other fixes

**Environment:**
 - OS: CentOS Stream 9 
 - OpenSSL 1.1.1
 - Clang 12
 - liboqs version 0.8.0

zxjtan:
Fixes #1548.

Call Keccak_(X4_)Dispatch using pthread_once to ensure setting of global function pointers is atomic and done only once.

* [x] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  NO
* [x] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? NO

### DIFF
--- a/src/common/sha3/xkcp_sha3x4.c
+++ b/src/common/sha3/xkcp_sha3x4.c
@@ -8,6 +8,9 @@
 #include <oqs/common.h>
 #include <oqs/oqsconfig.h>
 
+#if CMAKE_USE_PTHREADS_INIT
+#include <pthread.h>
+#endif
 #include <stddef.h>
 #include <stdint.h>
 #include <stdlib.h>
@@ -18,17 +21,17 @@
 #define KECCAK_X4_CTX_BYTES (KECCAK_X4_CTX_ALIGNMENT * \
   ((_KECCAK_X4_CTX_BYTES + KECCAK_X4_CTX_ALIGNMENT - 1)/KECCAK_X4_CTX_ALIGNMENT))
 
-/* The first call to Keccak_Initialize will be routed through dispatch, which
- * updates all of the function pointers used below.
- */
-static KeccakX4InitFn Keccak_X4_Dispatch;
-static KeccakX4InitFn *Keccak_X4_Initialize_ptr = &Keccak_X4_Dispatch;
+#if CMAKE_USE_PTHREADS_INIT
+static pthread_once_t dispatch_once_control = PTHREAD_ONCE_INIT;
+#endif
+
+static KeccakX4InitFn *Keccak_X4_Initialize_ptr = NULL;
 static KeccakX4AddByteFn *Keccak_X4_AddByte_ptr = NULL;
 static KeccakX4AddBytesFn *Keccak_X4_AddBytes_ptr = NULL;
 static KeccakX4PermuteFn *Keccak_X4_Permute_ptr = NULL;
 static KeccakX4ExtractBytesFn *Keccak_X4_ExtractBytes_ptr = NULL;
 
-static void Keccak_X4_Dispatch(void *state) {
+static void Keccak_X4_Dispatch(void) {
 // TODO: Simplify this when we have a Windows-compatible AVX2 implementation of SHA3
 #if defined(OQS_DIST_X86_64_BUILD)
 #if defined(OQS_ENABLE_SHA3_xkcp_low_avx2)
@@ -59,11 +62,16 @@ static void Keccak_X4_Dispatch(void *state) {
 	Keccak_X4_Permute_ptr = &KeccakP1600times4_PermuteAll_24rounds;
 	Keccak_X4_ExtractBytes_ptr = &KeccakP1600times4_ExtractBytes;
 #endif
-
-	(*Keccak_X4_Initialize_ptr)(state);
 }
 
 static void keccak_x4_inc_reset(uint64_t *s) {
+#if CMAKE_USE_PTHREADS_INIT
+	pthread_once(&dispatch_once_control, Keccak_X4_Dispatch);
+#else
+	if (Keccak_X4_Initialize_ptr == NULL) {
+		Keccak_X4_Dispatch();
+	}
+#endif
 	(*Keccak_X4_Initialize_ptr)(s);
 	s[100] = 0;
 }
@@ -234,4 +242,3 @@ void OQS_SHA3_shake256_x4_inc_ctx_release(OQS_SHA3_shake256_x4_inc_ctx *state) {
 void OQS_SHA3_shake256_x4_inc_ctx_reset(OQS_SHA3_shake256_x4_inc_ctx *state) {
 	keccak_x4_inc_reset((uint64_t *)state->ctx);
 }
-
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -30,8 +30,11 @@ if(NOT WIN32)
     else()
         set(INTERNAL_TEST_DEPS ${LIBM})
     endif()
+    if(CMAKE_USE_PTHREADS_INIT)
+        set(INTERNAL_TEST_DEPS ${INTERNAL_TEST_DEPS} Threads::Threads)
+    endif()
     if(DEFINED SANITIZER_LD_FLAGS)
-        set(INTERNAL_TEST_DEPS "${INTERNAL_TEST_DEPS} ${SANITIZER_LD_FLAGS}")
+        set(INTERNAL_TEST_DEPS ${INTERNAL_TEST_DEPS} ${SANITIZER_LD_FLAGS})
     endif()
     execute_process(COMMAND ${PROJECT_SOURCE_DIR}/scripts/git_commit.sh OUTPUT_VARIABLE GIT_COMMIT)
     add_definitions(-DOQS_COMPILE_GIT_COMMIT="${GIT_COMMIT}")
@@ -59,6 +62,9 @@ else()
 endif()
 
 set(API_TEST_DEPS oqs ${LIBM})
+if(CMAKE_USE_PTHREADS_INIT)
+    set(API_TEST_DEPS ${API_TEST_DEPS} Threads::Threads)
+endif()
 
 # KEM API tests
 add_executable(example_kem example_kem.c)
@@ -68,11 +74,7 @@ add_executable(kat_kem kat_kem.c)
 target_link_libraries(kat_kem PRIVATE ${API_TEST_DEPS})
 
 add_executable(test_kem test_kem.c)
-if((CMAKE_C_COMPILER_ID MATCHES "Clang") OR (CMAKE_C_COMPILER_ID STREQUAL "GNU"))
-    target_link_libraries(test_kem PRIVATE ${API_TEST_DEPS} Threads::Threads)
-else ()
-    target_link_libraries(test_kem PRIVATE ${API_TEST_DEPS})
-endif()
+target_link_libraries(test_kem PRIVATE ${API_TEST_DEPS})
 
 add_executable(test_kem_mem test_kem_mem.c)
 target_link_libraries(test_kem_mem PRIVATE ${API_TEST_DEPS})
@@ -88,11 +90,7 @@ add_executable(kat_sig kat_sig.c)
 target_link_libraries(kat_sig PRIVATE ${API_TEST_DEPS})
 
 add_executable(test_sig test_sig.c)
-if((CMAKE_C_COMPILER_ID MATCHES "Clang") OR (CMAKE_C_COMPILER_ID STREQUAL "GNU"))
-    target_link_libraries(test_sig PRIVATE ${API_TEST_DEPS} Threads::Threads)
-else ()
-    target_link_libraries(test_sig PRIVATE ${API_TEST_DEPS})
-endif()
+target_link_libraries(test_sig PRIVATE ${API_TEST_DEPS})
 
 add_executable(test_sig_mem test_sig_mem.c)
 target_link_libraries(test_sig_mem PRIVATE ${API_TEST_DEPS})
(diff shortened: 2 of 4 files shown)


dstebila:
@zxjtan will you be continuing to work on this?

zxjtan:
@dstebila I have no experience writing Win32 code, but I can try my hand at making it cross-platform if there are reviewers who can check that I am writing it idiomatically.

baentsch:
@zxjtan OQS already has code parts that are not optimized for Win32. In other words, it'd be beneficial to see `liboqs` improve "just" for pthread environments (as is the goal with this PR) rather than wait for a good Win32 solution to the problem. If a Windows user runs into #1548 we can revisit. What should happen first/now, though, is that this PR passes CI. Do you have time to get this resolved?

zxjtan:
@baentsch Sure, I can gate the pthread functionality and pass the CI quite soon.

zxjtan:
CI still runs into `undefined reference to symbol 'call_once@@GLIBC_2.28'`, any tips on how to proceed?

baentsch:
> CI still runs into `undefined reference to symbol 'call_once@@GLIBC_2.28'`, any tips on how to proceed?

Well, that depends on which lib is supposed to hold this symbol: It seems, glibc2.28 doesn't/isn't C11 compliant. What about using pthreads (i.e., [pthread_once](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_once.html) instead? Then linking in `pthreads` lib should get this to work.

zxjtan:
> > CI still runs into `undefined reference to symbol 'call_once@@GLIBC_2.28'`, any tips on how to proceed?
> 
> Well, that depends on which lib is supposed to hold this symbol: It seems, glibc2.28 doesn't/isn't C11 compliant. What about using pthreads (i.e., [pthread_once](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_once.html) instead? Then linking in `pthreads` lib should get this to work.

No dice, build still fails with `undefined reference to symbol 'pthread_once@@GLIBC_2.2.5'`. However, in both the pthread_once and call_once cases, this is caused by `error adding symbols: DSO missing from command line`. Do I have to change something in the CMake files to make it link?

baentsch:
> Do I have to change something in the CMake files to make it link?

I'd say so: That's what I meant with 

> Then linking in pthreads lib should get this to work.

Sorry for being imprecise.

So, I'd think one option may be to lift the "pthreads-availability" test logic you correctly built in to the source code to `cmake`variable level (such as to then add the lib there -- or not, if unavailable).

.CMake/compiler_opts.cmake
```diff
@@ -108,6 +108,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang")
         set(THREADS_PREFER_PTHREAD_FLAG ON)
         find_package(Threads REQUIRED)
         set(OQS_USE_PTHREADS_IN_TESTS 1)
+        set(OQS_USE_PTHREADS_IN_LIBOQS 1)
```
baentsch, line 111: Are you 100% sure pthreads is available if `clang` is the compiler?

.CMake/compiler_opts.cmake
```diff
@@ -170,6 +171,7 @@ elseif(CMAKE_C_COMPILER_ID STREQUAL "GNU")
...
```
baentsch, line 174: Same question wrt `gcc`. What about rather checking for presence of `pthreads` (e.g., as per https://stackoverflow.com/questions/1620918/cmake-and-libpthread)?

.CMake/compiler_opts.cmake
zxjtan: I am using the same logic as test_kem and test_sig for [`#include`ing `<pthread.h>`](https://github.com/open-quantum-safe/liboqs/blob/a80fdf8ee48ea979b5b90ceba026610338608ce0/tests/test_kem.c#L10) and [linking the library](https://github.com/open-quantum-safe/liboqs/blob/a80fdf8ee48ea979b5b90ceba026610338608ce0/tests/CMakeLists.txt#L72), so I am perplexed about why it's not working here. Besides, the `cmake` step should fail if pthreads is not found since the `find_package` is called with `REQUIRED`.

baentsch: > I am perplexed about why it's not working here

Me as well -- but there is a conceptual difference in this PR: It adds the pthreads dependency to the library. In testing, it's with the executable directly (where the executable linker may resolve the dependency automatically). As to why your explicit reference to "Threads::Threads" isn't honoured, I cannot comment on. [SO comments that `-lpthread` needs to be added](https://stackoverflow.com/questions/59508259/in-depth-explanation-for-why-we-need-pthread-in-linker-option-for-gcc).

> Besides, the cmake step should fail if pthreads is not found since the find_package is called with REQUIRED.

This is exactly what I don't find very appealing: This will break builds of `liboqs` if "pthreads" is not available. Wouldn't it be better to not break (but behave as it does right now)? But then again, the current behaviour isn't right -- so if we need to add pthreads to make it operate correctly with threading, so be it. Community opinions solicited.

zxjtan: > This will break builds of liboqs if "pthreads" is not available.
The `find_package(Threads REQUIRED)` call was added by @dstebila in 4cc0398. Can defer to him but I think it can be made not-required without breaking anything, and then conditioning pthread usage on `Threads_FOUND`.

The above is separate from making pthreads link, which I am still working on.

src/common/sha3/xkcp_sha3x4.c
```diff
@@ -59,11 +62,16 @@ static void Keccak_X4_Dispatch(void *state) {
 	Keccak_X4_Permute_ptr = &KeccakP1600times4_PermuteAll_24rounds;
 	Keccak_X4_ExtractBytes_ptr = &KeccakP1600times4_ExtractBytes;
 #endif
-
-	(*Keccak_X4_Initialize_ptr)(state);
 }
 
 static void keccak_x4_inc_reset(uint64_t *s) {
+#if OQS_USE_PTHREADS_IN_LIBOQS
```
baentsch, line 68: Why is this guard of "dispatch_once_control" using a different #define than "CMAKE_USE_PTHREADS_INIT"?

zxjtan: Because I made a mistake. Thanks for the catch!

baentsch: UW. Seems to have resolved the last CI problem. Please feel free to move this to Ready for Review when CI passes.

zxjtan: I could mark Ready for Review, but I'm not sure if I should wait for #1576 to be merged first as a separate logical unit

tests/CMakeLists.txt
```diff
@@ -88,11 +90,7 @@ add_executable(kat_sig kat_sig.c)
 target_link_libraries(kat_sig PRIVATE ${API_TEST_DEPS})
 
 add_executable(test_sig test_sig.c)
-if((CMAKE_C_COMPILER_ID MATCHES "Clang") OR (CMAKE_C_COMPILER_ID STREQUAL "GNU"))
-    target_link_libraries(test_sig PRIVATE ${API_TEST_DEPS} Threads::Threads)
-else ()
-    target_link_libraries(test_sig PRIVATE ${API_TEST_DEPS})
-endif()
+target_link_libraries(test_sig PRIVATE ${API_TEST_DEPS})
```
baentsch, line 93: Nice simplification: If [FindThreads](http://www.devdoc.net/linux/cmake-3.9.6/module/FindThreads.html) sets `CMAKE_USE_PTHREADS_INIT` this automatically gets activated also for testing and thus doesn't need compiler (version) checks any more.

baentsch (commented):
Code LGTM. The only slight concern I have is that this seems to add pthreads as a requirement to `liboqs` (and hence, its users). Have you tested that this new code works OK if used by a downstream project, @zxjtan ? 

Thinking that through, what about adding CI to `liboqs` automatically validating this property with downstream projects we consider essential, @dstebila ? Or would this create a hen-and-egg situation where downstream projects' status (of not being ready for an `liboqs` update) can break `liboqs` CI and hence stop the whole procedure? Let me create a separate discussion for this...

zxjtan:
> The only slight concern I have is that this seems to add pthreads as a requirement to `liboqs`
I don't think it does. I think of it not as a requirement but an environment fix: if you are running on a platform with threading, and if that threading model is `pthreads`, this PR provides a fix. Otherwise the code will not `#include <pthread.h>` because you're on something else like Win32 and you can submit a fix yourself, or you have no threading at all and there is nothing to fix.

baentsch (approved):
LGTM. Thanks for the contribution! @dstebila This is one more item to be added to the "Windows ToDo" list (as and when we start it per https://github.com/orgs/open-quantum-safe/discussions/1583.

Martyrshot (approved):
This looks good to me too! Taking a brief look at the Win32 API the fix will likely look very similar, but it seems the biggest obstacle will be finding someone with a Win32 machine to implement the fix :-)

dstebila (approved):

baentsch:
@zxjtan Thanks again for the contribution; it turned out it causes a failure in one downstream integration: https://app.circleci.com/pipelines/github/open-quantum-safe/oqs-demos/1046/workflows/e474db0c-43e2-4628-b634-23b9f3b5d95a/jobs/5979. Does this ring a bell? At first glance this should work as `-lpthread` is clearly configured in....

zxjtan:
> @zxjtan Thanks again for the contribution; it turned out it causes a failure in one downstream integration: https://app.circleci.com/pipelines/github/open-quantum-safe/oqs-demos/1046/workflows/e474db0c-43e2-4628-b634-23b9f3b5d95a/jobs/5979. Does this ring a bell? At first glance this should work as `-lpthread` is clearly configured in.... Edit/correct: `-lpthread` is requested to be configured in but the final `openssl` link command doesn't contain that instruction... Hmm.....

The `sed` in [this line](https://github.com/open-quantum-safe/oqs-demos/blob/a6f60e5fc1c28f90a14e6780e40f94d1db38d229/quic/Dockerfile-server#L27) might have something to do with it?

baentsch:
> The `sed` in [this line](https://github.com/open-quantum-safe/oqs-demos/blob/a6f60e5fc1c28f90a14e6780e40f94d1db38d229/quic/Dockerfile-server#L27) might have something to do with it?

I don't think so: This doesn't change (presence of) `-lpthread` but just adds `liboqs`. HOWEVER, looking further into the build log, [this line](https://app.circleci.com/pipelines/github/open-quantum-safe/oqs-demos/1046/workflows/e474db0c-43e2-4628-b634-23b9f3b5d95a/jobs/5979?invite=true#step-104-9581_85) is key, I'd gather: 

> ./config --prefix=/opt/oqs-openssl-quic/.openssl no-shared no-threads 

This configures the OpenSSL installation within `quic-nginx` to be built without pthreads. Then, in turn, the open symbols in `liboqs` cannot resolve as `liboqs` is (needed to be) linked into `openssl111`.

I see two ways out of this:
1) Upgrade the nginx-quic demo to use OpenSSL3 (as per https://github.com/open-quantum-safe/oqs-demos/issues/182)
2) Disable pthreads usage in the `liboqs` build for this demo

For 1) I am out of my depth (having to ask the original contributor @igorbarshteyn for help/advice); 2) would require a `cmake` variable in `liboqs` to disable the use of pthreads.

Any other options? We could of course also completely discontinue support of the QUIC demo as it is based on unsupported software (OpenSSL111). The more I think about this, the more I like it. Several people have voiced willingness to help on https://github.com/open-quantum-safe/oqs-demos/issues/182 but nothing has been forthcoming. And OpenSSL111 by now is EOL. Thus I think we should sunset this PQ-OpenSSL1-QUIC integration.

PR #1552, opened 2023-09-14
Document Falcon constant time errors

praveksharma:
Documents Falcon constant time errors being caught by CI.
Adds updates to Classic McEliece docs that were missed in #1541.

[No] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in [oqs-provider](https://github.com/open-quantum-safe/oqs-provider), [OQS-OpenSSL](https://github.com/open-quantum-safe/openssl), [OQS-BoringSSL](https://github.com/open-quantum-safe/boringssl), and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)


SWilson4:
Tacking on documentation for the recently added Falcon ARM implementation.

baentsch (approved):
The contents of the PR look OK and logical. But it seems to document a regression in Falcon and the `liboqs`-built-in SHA3 code (not being CT any more), do I understand this right? Is Falcon indeed the only algorithm affected by this common code CT-failure?

baentsch:
> The contents of the PR look OK and logical. But it seems to document a regression in Falcon and the `liboqs`-built-in SHA3 code (not being CT any more), do I understand this right? Is Falcon indeed the only algorithm affected by this common code CT-failure?

Correcting myself: I mis-read the YML indentation level documenting the CT error (falsely thinking common SHA3 is the problem). Assuming I now read correctly: Is this Falcon failure only in the AVX2 optimization? If so, what is the reason that all Falcon code is now considered to have secret dependent branching (assuming I read the YML right this time...)?

praveksharma:
> Is this Falcon failure only in the AVX2 optimization? If so, what is the reason that all Falcon code is now considered to have secret dependent branching?

The errors are in fact only only in the AVX2 optimization. I didn't realise that default behaviour was to enable AVX2 optmisations even supplying the `-DOQS_USE_AVX2_INSTRUCTIONS=OFF`  option [when building on x86_64](https://github.com/open-quantum-safe/liboqs/blob/b8a0bc72dd951c881dbeef8db70f4944c6f21df1/.CMake/alg_support.cmake#L296); I also missed the AVX2 tag in the suppression files, hence the incorrect documentation. I've made the relevant corrections and will merge the changes.

PR #1560, opened 2023-09-26
Test against all 100 KAT values

### LINKED ISSUE #1418: Expand test coverage to all 100 NIST KAT values
Currently our KAT tests only test against the first KAT, but most NIST submissions include (if memory serves) 100 values. We could do all of these.

SWilson4:
Add functionality to test against all 100 NIST KAT values.

- [x] Extend the `kat_kem` and `kat_sig` programs to generate all 100 KAT values
- [x] Collect SHA256 hashes for the extended KAT values
- [x] Create a Python script analogous to `test_nistkat.py` to run all 100 tests
- [x] Integrate the tests into a CI workflow

Fixes #1418 


SWilson4:
While working on this, I noticed that our implementation of Falcon-1024 differs from the upstream on a single KAT, number 82. I put together a demonstration here: https://github.com/SWilson4/falcon-diff

This seems to be explained by the signature buffer size used for the KATs: our implementation uses the signature size of 1280 bytes while the upstream code passes a larger buffer of 1330 bytes (even though the signature size upstream is still defined to be 1280 bytes). I'm not sure yet why this is done, but redefining the signature size to be 1280 in liboqs does make all 100 KATs pass.

.circleci/config.yml
```diff
@@ -353,47 +355,78 @@ workflows:
...
+      and:
+        - equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
+        - equal: [ "NIST KATs", << pipeline.schedule.name >> ]
+        - equal: [ sw-full-kat, << pipeline.git.branch >> ]
+    jobs:
+      - linux_oqs:
+          name: nistkat-generic
+          context: openquantumsafe
+          CONTAINER: openquantumsafe/ci-ubuntu-focal-x86_64:latest
+          CMAKE_ARGS: -DOQS_OPT_TARGET=generic
```
baentsch, line 421: This arg alone shouldn't work as OQS_DIST_BUILD=ON by default: See https://github.com/open-quantum-safe/liboqs/blob/main/CONFIGURE.md#oqs_opt_target

.circleci/config.yml
```diff
@@ -353,47 +355,78 @@ workflows:
...
+          PYTEST_ARGS: --numprocesses=auto -k 'test_kat_all'
+      - linux_oqs:
+          name: nistkat-avx2
+          context: openquantumsafe
+          CONTAINER: openquantumsafe/ci-ubuntu-focal-x86_64:latest
+          CMAKE_ARGS: -DOQS_DIST_BUILD=OFF
+          PYTEST_ARGS: --numprocesses=auto -k 'test_kat_all'
+      - arm_machine:
+          name: nistkat-aarch64
+          CMAKE_ARGS: -DOQS_DIST_BUILD=OFF
```
baentsch, line 431: This should then only test the "auto" config on `aarch64`. Is this what we want? I'd rather suggest a "generic"/ref-code and a "neon"/optimized run.

SWilson4: Right, so the generic run would have `-DOQS_DIST_BUILD=OFF -DOQS_OPT_TARGET=generic` and the optimized run would have `-DOQS_DIST_BUILD=OFF` (with `OQS_OPT_TARGET` defaulting to `native`)?

.circleci/config.yml
SWilson4: Good catch, thanks! I'll change it to set `-DOQS_DIST_BUILD=OFF`.

.circleci/config.yml
baentsch: Sounds right -- but I'm not knowledgeable enough about ARM to authoritatively confirm this. But then again, if it works for `x64` why should it be wrong for `aarch64` :-)

baentsch:
> redefining the signature size to be 1330 in liboqs does make all 100 KATs pass.

This also speaks for your option 3 in https://github.com/open-quantum-safe/liboqs/issues/1561#issuecomment-1791559434. Yes, @dstebila this may be a deviation from the "sig size" number given in the Falcon spec, but a) that number seems to be an "arbitrary choice" and b) aren't KATs there to "tie down" any spec ambiguities?

src/common/rand/rand.h
```diff
@@ -66,6 +66,16 @@ OQS_API void OQS_randombytes(uint8_t *random_array, size_t bytes_to_read);
  */
 OQS_API void OQS_randombytes_nist_kat_init_256bit(const uint8_t *entropy_input, const uint8_t *personalization_string);
 
+/**
+ * Saves the state of the NIST DRBG, allowing it to be recovered later.
+ */
+OQS_API void OQS_randombytes_nist_kat_save_state(void);
```
baentsch, line 72: I'm absolutely not sure I like this API design: It doesn't seem idempotent (what if someone calls this twice, before calling `restore`?) and creating state in the library. What about returning here an opaque structure/pointer and taking this as parameter in the restore below instead?

src/common/rand/rand_nist.c
```diff
@@ -38,6 +38,7 @@ typedef struct {
 } AES256_CTR_DRBG_struct;
 
 static AES256_CTR_DRBG_struct DRBG_ctx;
+static AES256_CTR_DRBG_struct DRBG_ctx_backup;
```
baentsch, line 41: aha -- here's the "duplicate state" -- but again, it only holds a single "second" state, so at the very least it should be stated at the API that calling `save` twice may result in "surprising" behaviour.

tests/kat_kem.c
```diff
@@ -42,6 +43,19 @@ static void fprintBstr(FILE *fp, const char *S, const uint8_t *A, size_t L) {
 	fprintf(fp, "\n");
 }
 
+static int is_mceliece(const char *method_name) {
+	return ( !strcmp(method_name, OQS_KEM_alg_classic_mceliece_348864)
```
baentsch, line 47: Hmm -- this looks "tedious" and prone to future errors if McEliece changes names or parameters. Is there really no other common "McEliece" designator?

baentsch (commented):
Conceptually LGTM. I'd approve if there'd be a good rationale for the API (or a change to "externalize" the state :)

The environment says Thanks for your diligent exclusion of this long-running test from the standard CI suite. I just wonder whether there'd be a more future-proof way to do this, e.g., some sort of "weekly_only" test list that can be summarily excluded from all PR and push Ci tests (?) Or should we create a separate issue for this? Something separating the .py files into "WEEKLY_TESTS", "PR_ONLY_TESTS" and "ALWAYS_TESTS"...

src/common/rand/rand.h
SWilson4: One could do that, but I believe it would require either (1) "exporting" the `AES256_CTR_DRBG_ctx` structure so that it can be used by the test programs or (2) duplicating the definition of said structure in each test program that requires it (currently only `kat_kem` and `kat_sig`). Both of those are OK, I guess, but I don't think that the required "carving up" of the NIST DRBG code would be worth it: I doubt there's any use case that will require successive "save" calls without a "restore", and there is already state stored to enable the NIST DRBG.

I'd personally rather not even make the "save" and "restore" functions part of the API, but it seems to be the only clean(ish) way to make them accessible to our test programs...

baentsch: > I'd personally rather not even make the "save" and "restore" functions part of the API, 

Then it's two of us...

> but it seems to be the only clean(ish) way to make them accessible to our test programs...

...so then we probably do something wrong: Things only to be used by `liboqs` testing ought to have an internal API. But it seems we only have one `liboqs` API, namely a very public one. My understanding of public APIs is that they need to be maintained (and maintainable). Adding something to such public API should be carefully considered. I'd strongly suggest moving features we only need internally into a truly internal API, maybe something like "test_helpers.h/c" in the "tests" directory?

SWilson4: How about this approach:

- a `NIST_KAT_DRBG_STATE` struct
- `void nist_kat_get_state(NIST_KAT_DRBG_STATE *)`
- `void nist_kat_set_state(const NIST_KAT_DRBG_STATE *)`

The "save" and "restore" functionality can be implemented using these functions in the new "test_helpers" files.

This seems to me to be the least disruptive option: the NIST DRBG continues to function as it does currently (library state), and `get_state` and `set_state` provide the minimal functionality required to implement save and restore. It's also clear what they do: no issues with idempotence etc.

I realize this sounds very much like option (1) which I objected to above, but using the get/set paradigm instead of save/restore makes it more palatable imo.

baentsch: Is it necessary to expose the struct? Couldn't we make do with a generic `void*`? Otherwise the proposal looks good (even though I'd have preferred an option to completely leave the API untouched -- but given we already expose the NIST RNG (incl. state) I also don't see a way around this :(

SWilson4: > Is it necessary to expose the struct? Couldn't we make do with a generic `void*`?

Sure, that should be fine. I'll get to work on it.

SWilson4:
I'm going to mark this PR as draft while I tinker with the internal API configuration. Will reopen for review when I'm ready for feedback.

SWilson4:
Once (if) #1667 lands, I'll rebase this branch and reopen this PR.

SWilson4:
This has been rebased to use the new internal API and is once again ready for review. Pardon the long and messy commit history.

tests/test_helpers.c
```diff
@@ -0,0 +1,157 @@
...
+	       || (0 == strcmp(method_name, OQS_KEM_alg_hqc_192))
+	       || (0 == strcmp(method_name, OQS_KEM_alg_hqc_256));
+}
+
+/* OQS_KAT_PRNG interface implementation */
+
+OQS_KAT_PRNG *OQS_KAT_PRNG_new(const char *method_name) {
+	OQS_KAT_PRNG *prng = malloc(sizeof(OQS_KAT_PRNG));
+	if (prng != NULL) {
+		prng->max_kats = is_mceliece(method_name) ? 10 : 100;
```
baentsch, line 84: I thought "all" means 100 KATs? Why only 10 for McEliece?

.github/workflows/windows.yml
```diff
@@ -29,4 +29,4 @@ jobs:
         run: pip.exe install pytest pytest-xdist pyyaml
       - name: Run tests
         run: |
-          python -m pytest --numprocesses=auto -vv --maxfail=10 --ignore=tests/test_code_conventions.py --junitxml=build\test-results\pytest\test-results.xml
+          python -m pytest --numprocesses=auto -vv --maxfail=10 --ignore=tests/test_code_conventions.py --ignore=tests/test_kat_all.py --junitxml=build\test-results\pytest\test-results.xml
```
baentsch, line 32: I know we didn't test KATs before on Windows, but would that be a sensible addition while we're at this? Or rather a separate issue (and addition to the weekly runs)?

tests/test_helpers.c
SWilson4: The McEliece NIST submission only provides 10 KATs.

.github/workflows/windows.yml
SWilson4: We do run the single KAT tests on Windows; this `--ignore` flag is just to exclude the "full" KAT runs, as on all platforms. But it would probably be a good idea to expand the weekly "full" KAT testing to other platforms than Linux on x86_64. I'll open an issue for it.

dstebila (approved):

PR #1572, opened 2023-10-04
Secret Key Query 

ashman-p:
Query the number of signing operations a private key can handle in total as well as the number
of signing operations that has already been done..


ducnguyen-sb:
Thanks @ashman-p , I only have one question. Other than that the PR LGTM!

ashman-p:
> Thanks @ashman-p , I only have one question. Other than that the PR LGTM!

Thank you. Also, you will see that some of the tests only address LMS but that will change once the relevant callback features have been added to XMSS.

PR #1578, opened 2023-10-13
Add section to CONFIGURE.md link

iyanmv:
This is a very small change so that `ninja gen_docs` works as expected.

Without this change, doxygen (at least, version 1.9.8) gives the following error:
> unable to resolve reference to '/liboqs/CONFIGURE.md' for \ref command

This makes `scripts/run_doxygen.sh` return exit code 1 instead of 0, and causes `ninja gen_docs` to fail

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ In order to optimize support effort,
 		cmake -GNinja ..
 		ninja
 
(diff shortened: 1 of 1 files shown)


baentsch (approved):
LGTM. Thanks for the contribution!

dstebila (approved):

PR #1585, opened 2023-10-19
Pull new HQC implementation from upstream

### LINKED ISSUE #995: Known non-constant time behaviour in HQC
Our constant time test raises three issues with the "clean" implementation of the HQC code. I've reviewed these and I am of the opinion that they are true instances of non-constant time behaviour.

I've copied a description of each issue and a link to the issue file below.

---
[hqc_gf_mul](https://github.com/open-quantum-safe/liboqs/blob/main/tests/constant_time/kem/issues/hqc_gf_mul)

Secret-dependent indexing into the "gf_exp" array in line 20 of gf.c:
https://github.com/open-quantum-safe/liboqs/blob/8981b7f33749abb027e9fe907c2f0e811c5c9048/src/kem/hqc/pqclean_hqc-rmrs-128_clean/gf.c#L16-L21

---

[hqc_vect_set_random_fixed_weight](https://github.com/open-quantum-safe/liboqs/blob/main/tests/constant_time/kem/issues/hqc_vect_set_random_fixed_weight)

Line 90 of vector.c reveals partial information about the location of non-zero bits in a secret fixed weight vector
https://github.com/open-quantum-safe/liboqs/blob/8981b7f33749abb027e9fe907c2f0e811c5c9048/src/kem/hqc/pqclean_hqc-rmrs-128_clean/vector.c#L82-L92

---

[hqc_fast_convolution_mult](https://github.com/open-quantum-safe/liboqs/blob/main/tests/constant_time/kem/issues/hqc_fast_convolution_mult)

The "fast_convolution_mult" routine makes secret dependent accesses to an array. It obfuscates the pattern of memory accesses by applying some permutations, but I'm not convinced that this is an adequate countermeasure.
https://github.com/open-quantum-safe/liboqs/blob/8981b7f33749abb027e9fe907c2f0e811c5c9048/src/kem/hqc/pqclean_hqc-rmrs-128_clean/gf2x.c#L70

### LINKED ISSUE #1319: Update HQC.
The implementation taken from PQClean [is based on the 2020-10-01 specification](https://github.com/PQClean/PQClean/blob/master/crypto_kem/hqc-rmrs-128/META.yml#L25), and the latest version is [2021/06/06](https://pqc-hqc.org/implementation.html). The latest specification uses KECCAK for domain separation and randomness.

SWilson4:
Run copy_from_upstream to pull the [recently merged](https://github.com/PQClean/PQClean/pull/512) HQC updates from PQClean. Ignore Kyber and Dilithium updates as we don't yet want to move to the "standard" version.

This is unfortunately a bit of a "kitchen sink" PR: it contains
1. a new version of HQC from https://github.com/PQClean/PQClean/pull/512,
2. updates to Falcon and Sphincs+ which landed in PQClean before the HQC code was merged, and
3. support for the HQC custom PRNG for KAT checking.

Notably, this PR does _not_ contain the latest [Kyber](https://github.com/PQClean/PQClean/pull/504) and [Dilithium](https://github.com/PQClean/PQClean/pull/511) updates from PQClean, as this would entail updating liboqs to be in sync with the "standard" track. If/when this is merged, we will be out of sync with PQClean.

When patching HQC, I made a number of minor changes to the upstream code, mostly to address undefined or implementation-defined behaviour and endianness issues. I also wrote a new modular reduction routine to address https://github.com/PQClean/PQClean/issues/482; this can be found in `vector.c`. It might be easier to review my changes directly against the reference implementation; you can view the patches [here](https://github.com/SWilson4/package-pqclean/tree/main/hqc/patches).

Fixes #1319.
Fixes #995.


docs/algorithms/kem/classic_mceliece.md
```diff
@@ -35,8 +35,8 @@
...
 |:---------------------------------:|:-------------------------|:----------------------------|:--------------------------------|:------------------------|:-----------------------------------|:-----------------------------------------------|:----------------------|
-| [Primary Source](#primary-source) | clean                    | All                         | All                             | None                    | True                               | True                                           | True                  |
-| [Primary Source](#primary-source) | avx2                     | x86\_64                     | Linux,Darwin                    | AVX2,POPCNT             | False                              | True                                           | True                  |
+| [Primary Source](#primary-source) | clean                    | All                         | All                             | None                    | False                              | False                                          | True                  |
```
baentsch, line 38: Do we now understand these changes to McEliece?

.CMake/alg_support.cmake
```diff
@@ -176,26 +176,8 @@ endif()
 
 option(OQS_ENABLE_KEM_HQC "Enable hqc algorithm family" ON)
 cmake_dependent_option(OQS_ENABLE_KEM_hqc_128 "" ON "OQS_ENABLE_KEM_HQC" OFF)
-if(CMAKE_SYSTEM_NAME MATCHES "Linux|Darwin")
```
baentsch, line 179: Do I understand this right meaning that HQC doesn't contain AVX2 optimizations any more? Note to (at least) self: Don't be surprised about new HQC performance numbers then.

docs/algorithms/kem/hqc.md
```diff
@@ -2,11 +2,11 @@
 
 - **Algorithm type**: Key encapsulation mechanism.
 - **Main cryptographic assumption**: Syndrome decoding of structure codes (Hamming Quasi-Cyclic).
-- **Principal submitters**: Carlos Aguilar Melchor, Nicolas Aragon, Slim Bettaieb, Olivier Blazy, Jurjen Bos, Jean-Christophe Deneuville, Philippe Gaborit, Edoardo Persichetti, Jean-Marc Robert, Pascal Véron, Gilles Zémor, Loïc Bidoux.
+- **Principal submitters**: Carlos Aguilar Melchor, Nicolas Aragon, Slim Bettaieb, Loïc Bidoux, Olivier Blazy, Jurjen Bos, Jean-Christophe Deneuville, Arnaud Dion, Philippe Gaborit, Jérôme Lacan, Edoardo Persichetti, Jean-Marc Robert, Pascal Véron, Gilles Zémor.
 - **Authors' website**: https://pqc-hqc.org/
 - **Specification version**: NIST Round 3 submission.
```
baentsch, line 7: Is it intentional to not change the version identification? This looks like pretty drastically different code (sizes changing, optimizations vanishing...) and it would be prudent to be able to tell it apart from the previous version, no?

baentsch:
Oh, one more thing: This PR needs to update the top-level `CMakeLists.txt`to have "OQS_VERSION_TEXT" read "0.10.0-dev".

docs/algorithms/kem/classic_mceliece.md
SWilson4: Pinging @praveksharma to take a look at this.

SWilson4:
> The version of HQC does not seem to have been updated. Am I right in this understanding? If so, can we please fix this before merge, either by patch or upstream update?

> Oh, one more thing: This PR needs to update the top-level `CMakeLists.txt`to have "OQS_VERSION_TEXT" read "0.10.0-dev".

Both should be fixed now. I also incremented the `SOVERSION`, as we did when doing [a similar merge for McEliece](https://github.com/open-quantum-safe/liboqs/pull/1470).

docs/algorithms/kem/classic_mceliece.md
praveksharma: `classic_mceliece.md` wasn't updated along with `classic_mceliece.yml` in #1552 (presumably I forgot to run `update_upstream_alg_docs.py` at the time) so these changes are desirable.

baentsch: Will be resolved when #1589 lands and this PR rebases to it.

baentsch: @SWilson4 Update: If this PR is rebased to #1595 (assuming you approve and we land that), `copy_from_upstream` is fixed and idempotent again, pulling in HQC in the process. CI can stay as-is.

I have tested that and could do a `git push -f` on this PR to finalize this if you want/agree. Drawback: I'd most likely become co-author to this PR and cannot approve this any more. We then need two _more_ people to approve this (hint, hint, @dstebila @bhess @praveksharma :)

SWilson4: I can do the rebase/push and save people the trouble.

.CMake/alg_support.cmake
baentsch: Makes sense. Would it be reasonable then to create an issue for integrating that HQC AVX2 code in the future?

SWilson4: Yes, I'll do that right now.

src/common/rand/rand.h
```diff
@@ -66,6 +68,21 @@ OQS_API void OQS_randombytes(uint8_t *random_array, size_t bytes_to_read);
...
 
+/**
+ * Initializes the HQC PRNG with a given seed.
+ * OQS_randombytes_hqc_kat_release should be called afterwards to clean up memory.
+ *
+ * @param[in] entropy_input The seed; must be exactly 48 bytes
+ * @param[in] personalization_string An optional personalization string;
+ * may be NULL; if not NULL, must be at least 48 bytes long
+ */
+OQS_API void OQS_randombytes_hqc_kat_init(const uint8_t *entropy_input, const uint8_t *personalization_string);
```
baentsch, line 79: Is this change to/extension of the "common" `liboqs` API really necessary? It feels wrong to add an algorithm-specific API call... Couldn't this be "hidden" within the tests code? Do we assume others will need this?

SWilson4: We could define these functions within the `kat_kem.c` file and use the `OQS_randombytes_custom_algorithm` function to hook them in. I have no objection to doing that. In adding them to the API I was just following the same pattern as we use for the NIST KAT DRBG. Although that is presumably of much wider interest than the HQC version. The only use case that I can think of for someone else needing to have the HQC PRNG interface exposed is to run all 100 KATs, since our test currently only checks the first. And we'll (hopefully) be fixing that soon.

The HQC code is "public domain" so presumably the SPDX identifier for `kat_kem.c` (currently MIT) would need to be updated?

baentsch: > I have no objection to doing that. 

That would be much appreciated and keep the `liboqs` API clean.

> NIST KAT DRBG. Although that is presumably of much wider interest than the HQC version

It'd also think so.

> The HQC code is "public domain" so presumably the SPDX identifier for kat_kem.c (currently MIT) would need to be updated?

Hmm -- is a call to set up an RNG in the way HQC needs it truly HQC code / subject to its license?

docs/algorithms/kem/classic_mceliece.md
baentsch: Thanks. Saw it. Resolved.

docs/algorithms/kem/hqc.md
baentsch: Absolutely! Thanks.

src/common/rand/rand.h
baentsch: As discussed: The "MIT way" currently stated in "tests/kat_kem.c" is correct. Thanks for moving the code there. This resolves my concern.

baentsch (approved):

baentsch:
@SWilson4 Please feel free to re-request my review if you feel the changes needed to get CI to pass warrant another glance. Otherwise, Thanks again for this "deep dive"!

src/kem/hqc/pqclean_hqc-128_clean/vector.c
```diff
@@ -0,0 +1,188 @@
...
+static inline uint32_t cond_sub(uint32_t r, uint32_t n) {
+    uint32_t mask;
+    r -= n;
+    mask = 0 - (r >> 31);
+    return r + (n & mask);
+}
+
+static inline uint32_t reduce(uint32_t a, size_t i) {
+    uint32_t q, n, r;
+    q = ((uint64_t) a * m_val[i]) >> 32;
```
bhess, line 51: Just to double-check, since this is one of the constant-time patches: the index `i` isn't secret, right? Otherwise the lookup `m_val[i]` would be problematic.

SWilson4: No, it's not. This routine is only called by `vect_set_random_fixed_weight`, where `i` is a loop variable ranging from 0 to a fixed public upper bound.

bhess: ok thanks for the additional context

bhess (approved):

baentsch:
@SWilson4 FYI and FWIW, I was intrigued by the single CI failure and ran some experiments on the CCI MacOS instance via SSH: Result: The shared lib config option `-DBUILD_SHARED_LIBS=ON` reliably leads to all KEM KATs being incorrect. The following config options are irrelevant: `OQS_DIST_BUILD`, `OQS_USE_OPENSSL`; I even compiled using gcc-12 instead of clang-14: To no avail, if `liboqs` is built shared, the KEM KAT tests fail reliably on that x64 machine. I compared the exact same config on my own x64 machine (not MacOS, but also clang-14 and all CPU extensions reported identically) and all KATs are correct there. I then had the hunch that the shared lib concept of MacOS may play in this here and tested on the M1: And indeed the same config also fails there -- different architecture, different compiler, same shared library concept (MacOS). I still don't understand what's happening here.... Any MacOS experts around? (@dstebila ?)

tests/CMakeLists.txt
SWilson4, line 1: By my reasoning, this must be the change that leads to the single CI failure. Why, I don't know.

SWilson4:
> @SWilson4 FYI and FWIW, I was intrigued by the single CI failure and ran some experiments on the CCI MacOS instance via SSH: Result: The shared lib config option `-DBUILD_SHARED_LIBS=ON` reliably leads to all KEM KATs being incorrect. The following config options are irrelevant: `OQS_DIST_BUILD`, `OQS_USE_OPENSSL`; I even compiled using gcc-12 instead of clang-14: To no avail, if `liboqs` is built shared, the KEM KAT tests fail reliably on that x64 machine. I compared the exact same config on my own x64 machine (not MacOS, but also clang-14 and all CPU extensions reported identically) and all KATs are correct there. I then had the hunch that the shared lib concept of MacOS may play in this here and tested on the M1: And indeed the same config also fails there -- different architecture, different compiler, same shared library concept (MacOS). I still don't understand what's happening here.... Any MacOS experts around? (@dstebila ?)

I think I know what's happening, but I don't know why it would be caused by this particular config... if you look at the failed results, the computed KAT hashes are different at each run. This leads me to believe that either
1. the system RNG is being used instead of one of the deterministic PRNGs or
2. the deterministic PRNGs are being used but not seeded properly.

If, for instance, a deterministic PRNG were used with an uninitialized buffer it might explain the apparent randomness. However, the former possibility seems more likely to me. From the point of view of the non-HQC KATs, the only thing that's changed (other than being wrapped in an `if`/`else` block) is the build configuration for `kat_kem`: see https://github.com/open-quantum-safe/liboqs/pull/1585/files#r1381665498.

SWilson4:
> the system RNG is being used instead of one of the deterministic PRNGs

I just stepped through the code in a debugger on the M1 machine and confirmed that this is indeed the case. It's very confusing: the NIST DRBG is indeed set as the `randombytes` algorithm and used for the first call (from the `kat_kem` file to generate the seed). On subsequent calls (from the Kyber code) the OpenSSL RNG is used.

baentsch:
> On subsequent calls (from the Kyber code) the OpenSSL RNG is used.

If the code is built without OpenSSL the same issue appears. What RNG is used then? Why is this dependent on the linkage mode of the library??

tests/CMakeLists.txt
```diff
@@ -64,8 +64,9 @@ set(API_TEST_DEPS oqs ${LIBM})
 add_executable(example_kem example_kem.c)
 target_link_libraries(example_kem PRIVATE ${API_TEST_DEPS})
 
-add_executable(kat_kem kat_kem.c)
-target_link_libraries(kat_kem PRIVATE ${API_TEST_DEPS})
+# KAT KEM needs to call the internal SHA3 functions directly, hence the extra dependencies
+add_executable(kat_kem kat_kem.c ${COMMON_OBJS})
```
baentsch, line 68: Hmm -- this line makes me wonder: If one adds all COMMON_OBJS _and_ links against `liboqs` (line below, API_TEST_DEPS) which symbols "win"? The ones from the objects list or the ones from the library?

SWilson4:
> If the code is built without OpenSSL the same issue appears. What RNG is used then?

In that case the system RNG (`arc4random` for Apple) is used.

tests/CMakeLists.txt
SWilson4: > Hmm -- this line makes me wonder: If one adds all COMMON_OBJS _and_ links against `liboqs` (line below, API_TEST_DEPS) which symbols "win"? The ones from the objects list or the ones from the library?

Maybe the "winner" is picked differently on Mac vs Linux...

baentsch: What about this theory: `oqs_randombytes_algorithm` is a static function pointer in an object file; Linux keeps using that throughout the code after it has been initialized by the call in `kat_kem.c`. MacOS in turn "overloads" it with a new one when loading the library for all further invocations of library functions. I don't quite understand why the lib doesn't get loaded right at the moment `OQS_randombytes_switch_algorithm` gets called, though.... So, just a wild theory, still no idea how to fix.

SWilson4:
Thanks very much to @baentsch for getting CI to pass. I'm going to go ahead and merge into main, since there were no HQC-related changes after the approving reviews, and the changes since then are largely authored by one of the approving reviewers.

As a final check, I ran the constant time tests locally just now and HQC passed with flying colours. Hopefully this will also be confirmed in the weekly run.

PR #77, opened 2016-12-22
Enable McBits on Windows

christianpaquin:
Modified Visual Studio projects to accommodate McBits, and added instructions on how to obtain and link libsodium on Windows. I tested the integration and it worked.

PR #1588, opened 2023-10-20
Update XMSS secret key object APIs, sync with LMS 

ducnguyen-sb:
# XMSS changes

## Changelogs

- Made code compatible with MSVC in order to merge into `main` branch: Remove variable length array in `XMSS/external/` by converting them to `malloc/calloc` and `OQS_insecure_free()`
- Cleaned up unused files: `sign.c` and `sign.h`. 
- Fixed all segfaults. 
- Added `katfile` to test that takes too long to generate a key pair. 
- Removed a lot of LOC, so the code base of XMSS parameters are shorter, and easier to edit. 
- Changed in memory allocation: `sk->secret_key_data` is allocated by default. The `OQS_SECRET_KEY_XMSS_deserialize_key` is simply copy memory, instead of allocation new memory and copy. I think this should also be applied to LMS as well. I see it’s error-prone and inefficient to allocate memory each time the `deserialize` happens. 

## Tests

This code will work with either `OQS_USE_PTHREADS_IN_TESTS` on or off. I tested both cases.


src/sig_stfl/sig_stfl.h
```diff
@@ -271,26 +271,29 @@ typedef struct OQS_SIG_STFL_SECRET_KEY {
...
 	 * Secret Key retrieval Function
 	 *
 	 * @param[in] sk The secret key represented as OQS_SIG_STFL_SECRET_KEY object
 	 * @param[out] sk_len length of private key as a byte stream
 	 * @param[out] sk_buf_ptr pointer to private key data as a byte stream
 	 * @returns length of key material data available
-	 * Caller deletes the buffer if memory was allocated.
+	 * Caller is responsible to **unallocate** the pointer to buffer `sk_buf_ptr`.
 	 */
-	OQS_STATUS (*serialize_key)(const OQS_SIG_STFL_SECRET_KEY *sk, size_t *sk_len, uint8_t **sk_buf_ptr);
```
ashman-p, line 283: Any reason to not keep this a 'const'?

ashman-p (approved):

src/sig_stfl/sig_stfl.h
ducnguyen-sb: Yeah @ashman-p , because we modify the `sk->is_locked` variable, so I have to remove the `const`. what do you think?

src/sig_stfl/xmss/sig_stfl_xmss_secret_key_functions.c
```diff
@@ -27,37 +87,65 @@ OQS_STATUS OQS_SECRET_KEY_XMSS_serialize_key(const OQS_SIG_STFL_SECRET_KEY *sk,
...
+	sk->secret_key_data = NULL;
+}
 
-	return OQS_SUCCESS;
+void OQS_SECRET_KEY_XMSS_activate_lock(OQS_SIG_STFL_SECRET_KEY *sk) {
+	if (sk == NULL) {
+		return;
+	}
+
+	if (sk->is_locked == false) {
```
ashman-p, line 130: Not I understand this. If the mutex is already locked subsequent lock requests are ignored?

src/sig_stfl/sig_stfl.h
```diff
@@ -165,7 +165,7 @@ OQS_API int OQS_SIG_STFL_alg_is_enabled(const char *method_name);
  */
 typedef struct OQS_SIG_STFL {
 
-	/** A local ordinal representing the LMS parameter of the signature scheme. */
+	/** A local ordinal representing the OID parameter of the signature scheme. */
```
baentsch, line 168: Is there a place where I can read up what an "OID" is in this context? 32bits are surely not enough to represent an ASN.1 OID. When looking around, I found `sig_stfl_xmss.h`: If this header file is meant to document the externally usable API (?), it IMO needs quite a bit more documentation to be of help to any user. Please consider adding this before considering a merge to "main". Apologies for possibly jumping the gun here, but you asked me for review. Or was this unintended?

src/sig_stfl/xmss/sig_stfl_xmss_secret_key_functions.c
ducnguyen-sb: Correct. May be it's not the best way, I need to the `is_locked` variable to be thread-safe here. 
Back to the logic: 
I think if we lock the secret key, we would need only one lock. 
So if the key is locked, then the subsequent lock won't need to lock it again until the lock is released.

src/sig_stfl/sig_stfl.h
ducnguyen-sb: @baentsch , you're right. Thanks for your review, I will update the comment. 
The intended idea of "OID" is it's only for the `external/`, it should not be understood or related to ASN.1 OID.

src/sig_stfl/xmss/sig_stfl_xmss_secret_key_functions.c
ashman-p: For my understanding ... You added this logic to prevent a crash?
The way the code is currently, another thread would be allowed to access the already locked secret key. This not what we would want.
Secondly, the current thread, holding the lock, should only try to call activate_lock once unless there is some recursive calls using the same key. I would expect the system to handle that case. And if the 'activate_lock' is called by another thread, then we want it blocked.
Thats my understanding on the behavior we want. Your thoughts?

ducnguyen-sb: > For my understanding ... You added this logic to prevent a crash?

I added this to mainly prevent a hang. 

>  The way the code is currently, another thread would be allowed to access the already locked secret key. This not what we would want.

I agree totally.

> Secondly, the current thread, holding the lock, should only try to call activate_lock once unless there is some recursive calls using the same key. I would expect the system to handle that case. And if the 'activate_lock' is called by another thread, then we want it blocked.

That's right, it's my ultimate goal as well. I'm not sure the current approach is the best approach, it's mainly to avoid the hang. Feel free to edit the code, I have basic knowledge with mutex so I think I'm not doing the right job here.

src/sig_stfl/sig_stfl.h
ducnguyen-sb: Updated in https://github.com/open-quantum-safe/liboqs/pull/1588/commits/8e422a569c96a9d6254a0274d84b01ac3884af1e

src/sig_stfl/xmss/sig_stfl_xmss_secret_key_functions.c
ashman-p: Do you have a backtrace of the actual hang?

ducnguyen-sb: Hmm, I don't have a backtrace. I remember that if I comment out the `if is_lock == false` check in both activate/deactivate locks then it will hang.

ashman-p: Can you use gdb to get the traceback?

ashman-p: Can you rerun your test with but comment out the call to "activate_lock" from 	OQS_SECRET_KEY_XMSS_activate_lock(sk); at line #77  in sig_stfl_xmss_secret_key_functions.c

ducnguyen-sb: image

I'm having a mac with me right now so it's LLDB.

ducnguyen-sb: > Can you rerun your test with but comment out the call to "activate_lock" from OQS_SECRET_KEY_XMSS_activate_lock(sk); at line https://github.com/open-quantum-safe/liboqs/pull/77 in sig_stfl_xmss_secret_key_functions.c

I commented out the `activate_lock` and `activate_unlock`, it works normally.

ashman-p: Can you rebuild with debug symbols? Maybe CFLAGS="-g -O0"

ducnguyen-sb: I think it's better if you jump into the current branch and asset the hang. 
After you build, run the command below to reproduce

```
$gdb ./build/tests/test_sig_stfl XMSS-SHA2_10_256 tests/KATs/sig_stfl/xmss/XMSS-SHA2_10_256.rsp
```

ashman-p: Looks like we just need to use a recursive mutex.

ashman-p: @ducnguyen-sb can you take a look at the tests that are timing out?

ducnguyen-sb: Yup, I think I know why. Will update asap.

ducnguyen-sb: Okay, today I have time to analyze deeper this problem. 
Turn out, using RECURSIVE MUTEX increase runtime significantly. 

Benchmark all XMSS parameters in my M1 Mac.

- No thread test: 82 seconds
- With RECURSIVE MUTEX: 120 seconds (50% increase)
- Rearrange function, simple MUTEX; 120 seconds.  (50% increase)

So basically it's time out because of the use of MUTEX. 
I think the solution here is to increase timeout a little bit more.

ashman-p (approved):

ducnguyen-sb:
I don't know why the test duration increase by 50%.
I think our case is simple enough to use recursive mutex and it's nicer to do so. 
I guess the major time execution is thread state management. We can improve the our test with thread to see if it's better.

I'll think about it, if we are short on time then increase runtime is another solution.

ducnguyen-sb:
@ashman-p , I removed the recursive mutex, it works right away. 
To do so, I create an `inner_serialize` function, which is identical to `serialize` except the `acquire_lock` and `release_lock`. 

Other than that I simplify the logic code in `test_sig_stfl.c`. I think you would want to take a look at it. 
For short, let's just use a the simple mutex.

ashman-p:
> @ashman-p , I removed the recursive mutex, it works right away. To do so, I create an `inner_serialize` function, which is identical to `serialize` except it does not have the `acquire_lock` and `release_lock` functions.
> 
> Other than that I simplify the logic code in `test_sig_stfl.c`. I think you would want to take a look at it. For short, let's just use a the simple mutex.

Yep, this sounds good. Thank you.

ashman-p (commented):
Looks good. Thanks.

ashman-p (approved):

Issue #1584, opened 2023-10-18
Improve granularity of copy_from_upstream


SWilson4:
As things currently stand, the `copy_from_upstream` script attempts to integrate updates for _all_ schemes sourced from a given upstream. The update source is given by a git commit SHA for the upstream repo.

It would be nice to have the option to override this on a scheme-by-scheme basis so that integrating upstream code doesn't get messy if OQS (deliberately) goes out of sync with an upstream source on some algorithms but not others. For instance, we could specify that the PQClean aarch64 implementations of Dilithium and Kyber should remain at the latest commit before [PQClean updated their implementation to be in sync with the pq-crystals `standard` branches](https://github.com/PQClean/PQClean/pull/511) while still pulling the latest updates to HQC and Sphincs+.

baentsch:
Actually, I think I just "discovered" a "solution" for this in the code of `copy_from_upstream.py`: Just set the environment variable NOT_READY to families you don't want to copy and they're skipped. Correct @bhess? I know, documenting any feature ([see your commit adding this](https://github.com/open-quantum-safe/liboqs/commit/fc35e09a1491fe3ffaba90999c1767e65a694ac9)) is not great fun -- but not doing it decreases the code's value and causes others to re-invent the wheel (or go ballistic). In any case, not good. In that light, could you please also take a look at #1586, @bhess? Does this ring any bells in your mind? I'm now somewhat demotivated to go through this logic...

SWilson4:
I tested running `NOT_READY="kyber dilithium" python3 copy_from_upstream.py copy` on my branch, and it changed a bunch of files we wouldn't want it to, including removing Kyber and Dilithium from `CMakeLists.txt`. So unfortunately I don't think it's a quick fix. That said, perhaps it could be turned into one? I haven't looked closely at it yet.

baentsch:
> including removing Kyber and Dilithium from CMakeLists.txt. So unfortunately I don't think it's a quick fix. That said, perhaps it could be turned into one?

[...]

PR #1478, opened 2023-05-20
Patch AVX2 support: No Win for Sphincs+

baentsch:
Fixes #1476 . Also activates Sphincs testing in AppVeyor -- about time after 3 years :) Let's see how much this sets back the CI time...

* [no] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [no] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in [oqs-provider](https://github.com/open-quantum-safe/oqs-provider), [OQS-OpenSSL](https://github.com/open-quantum-safe/openssl), [OQS-BoringSSL](https://github.com/open-quantum-safe/boringssl), and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)

PR #1589, opened 2023-10-20
Run copy_from_upstream and test

### LINKED ISSUE #1586: copy_from_upstream run mistake (also in 0.9.0 release)
When looking into https://github.com/open-quantum-safe/liboqs/issues/1584 I noticed that "main" (and 0.9.0) are not idempotent to a run of `copy_from_upstream` on my machine: See the differences in https://github.com/open-quantum-safe/liboqs/blob/mb-fixcopyfromupstream: Compare for example the documentation of algorithm feature sets for Sphincs.

We have [a test protecting against such failures](https://github.com/open-quantum-safe/liboqs/blob/bac716cc91c53529e3fad88e92ff9dfd19c04577/.github/workflows/linux.yml#L21-L36), but this "silently" [failed](https://github.com/open-quantum-safe/liboqs/actions/runs/6564959643/job/17832426451) i.e., incorrectly returned OK after reporting a failure. This ought to be investigated I'd gather.

Also the [Release process documentation](https://github.com/open-quantum-safe/liboqs/wiki/Release-process) was lacking in this regard (now fixed).

Could someone else please take a look (maybe first just confirm this finding as I'm not sure my machine is not a bit misconfigured), maybe @praveksharma or @SWilson4 ? If confirmed, we might want to add wording to the release notes of 0.9.0 that the algorithm documentation isn't "quite right": At first blush, it seems upstream Sphincs claims support for all OSs while `liboqs` only documents support for Linux&Darwin for optimized Sphincs... Or is this related to the limitations of our shared code in Windows? But why then doesn't `copy_from_upstream` reconcile this?

baentsch:
Fixes #1586

### DIFF
--- a/docs/algorithms/sig/sphincs.yml
+++ b/docs/algorithms/sig/sphincs.yml
@@ -56,9 +56,6 @@ parameter-sets:
     - architecture: x86_64
       required_flags:
       - avx2
-      operating_systems:
-      - Linux
-      - Darwin
     common-crypto:
     - SHA2: liboqs
     no-secret-dependent-branching-claimed: true
@@ -86,9 +83,6 @@ parameter-sets:
     - architecture: x86_64
       required_flags:
       - avx2
-      operating_systems:
-      - Linux
-      - Darwin
     common-crypto:
     - SHA2: liboqs
     no-secret-dependent-branching-claimed: true
@@ -116,9 +110,6 @@ parameter-sets:
     - architecture: x86_64
       required_flags:
       - avx2
-      operating_systems:
-      - Linux
-      - Darwin
     common-crypto:
     - SHA2: liboqs
     no-secret-dependent-branching-claimed: true
@@ -146,9 +137,6 @@ parameter-sets:
     - architecture: x86_64
       required_flags:
       - avx2
-      operating_systems:
-      - Linux
-      - Darwin
     common-crypto:
     - SHA2: liboqs
     no-secret-dependent-branching-claimed: true
@@ -176,9 +164,6 @@ parameter-sets:
     - architecture: x86_64
       required_flags:
       - avx2
-      operating_systems:
-      - Linux
-      - Darwin
     common-crypto:
     - SHA2: liboqs
     no-secret-dependent-branching-claimed: true
@@ -206,9 +191,6 @@ parameter-sets:
     - architecture: x86_64
       required_flags:
       - avx2
-      operating_systems:
-      - Linux
-      - Darwin
     common-crypto:
     - SHA2: liboqs
     no-secret-dependent-branching-claimed: true
@@ -236,9 +218,6 @@ parameter-sets:
     - architecture: x86_64
       required_flags:
       - avx2
-      operating_systems:
-      - Linux
-      - Darwin
     common-crypto:
     - SHA3: liboqs
     no-secret-dependent-branching-claimed: true
@@ -266,9 +245,6 @@ parameter-sets:
     - architecture: x86_64
       required_flags:
       - avx2
-      operating_systems:
-      - Linux
-      - Darwin
     common-crypto:
     - SHA3: liboqs
     no-secret-dependent-branching-claimed: true
@@ -296,9 +272,6 @@ parameter-sets:
     - architecture: x86_64
       required_flags:
       - avx2
-      operating_systems:
-      - Linux
-      - Darwin
     common-crypto:
     - SHA3: liboqs
     no-secret-dependent-branching-claimed: true
@@ -326,9 +299,6 @@ parameter-sets:
     - architecture: x86_64
       required_flags:
       - avx2
-      operating_systems:
-      - Linux
-      - Darwin
     common-crypto:
     - SHA3: liboqs
     no-secret-dependent-branching-claimed: true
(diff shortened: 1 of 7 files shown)


baentsch:
@SWilson4 @praveksharma @dstebila Please carefully check this: I'm less than certain that the docs have all been generated correctly. The CI test improvement however definitely is needed.

docs/algorithms/sig/sphincs.yml
```diff
@@ -56,9 +56,6 @@ parameter-sets:
     - architecture: x86_64
       required_flags:
       - avx2
-      operating_systems:
```
SWilson4, line 59: I think we do want to document the restriction to Linux and Darwin: see [here](https://github.com/open-quantum-safe/liboqs/pull/1478).

SWilson4 (commented):
I'd like to hear @praveksharma's thoughts on the constant-time documentation changes for McEliece and Falcon; if I recall correctly he recently worked on some related issues.

baentsch:
Oops -- here's a PR I forgot to mention when we just discussed: @praveksharma @SWilson4 : Do you want to review or shall this be superseded by #1585? Would be totally fine with me. Main thing is #1586 gets fixed (incl. the CI shortcoming addressed in this PR).

SWilson4:
> Oops -- here's a PR I forgot to mention when we just discussed: @praveksharma @SWilson4 : Do you want to review or shall this be superseded by #1585? Would be totally fine with me. Main thing is #1586 gets fixed (incl. the CI shortcoming addressed in this PR).

I'd rather not try to cram the CI fix into (the already massive) #1585, so on that front I'm OK with proceeding with this PR, once we're satisfied that the documentation is indeed generated correctly. _However_, once #1585 lands we will be out of sync with PQClean, so the CI fix should break the build. (Or, if this PR lands first, #1585 will be blocked from merging.)

How ought we to deal with this? Ideally. I think we'd resolve #1584, but that might get fairly involved. Still, it will need to be done if idempotence under `copy_from_upstream.py` is a desired property. We could also do a hacky workaround like disabling the CI check for now, or adding some exceptions to it, but that seems like a recipe for future pain.

praveksharma:
> I'd like to hear @praveksharma's thoughts on the constant-time documentation changes for McEliece and Falcon; if I recall correctly he recently worked on some related issues.

@SWilson4, copying a comment from another PR over here: `classic_mceliece.md` wasn't updated along with `classic_mceliece.yml` in #1552 (presumably I forgot to run `update_upstream_alg_docs.py` at the time) so these changes are desirable. Ditto with Falcon.

baentsch:
>  if idempotence under copy_from_upstream.py is a desired property.

I'd argue idempotence under `copy_from_upstream` is essential for this project: `liboqs` is pretty much completely gutted if it cannot reliably import the algorithms. As we're not under any time pressure, I'd argue to do it "right": Fix #1584 first, maybe (just as a test) use that to re-generate #1586 (it seems we now have clarity on the McEliece docs, thanks @praveksharma ), then re-base #1585 (giving the improved copy_from_upstream a final "run for its money" -- and have it focus on HQC and not the other shenenigans).

docs/algorithms/sig/sphincs.yml
baentsch: With the latest commit this is now as desired (spec'd by the patch). "Reason": I ran `copy_from_upstream` with parameter `-k`: This way, the patches do not get thrown away as per [this code](https://github.com/open-quantum-safe/liboqs/blob/bac716cc91c53529e3fad88e92ff9dfd19c04577/scripts/copy_from_upstream/copy_from_upstream.py#L614-L615). Again, no clue why this `rmtree` statement exists... Unless I hear objections from @bhess I'll probably simply remove this line when resolving #1584 to avoid future repeats of this issue (documentation patches simply getting nixed).

So this PR is now ready for review @SWilson4 @praveksharma .

SWilson4 (approved):

docs/algorithms/sig/sphincs.yml
bhess: @baentsch the `rmtree` was originally at the end of the `def copy_from_upstream` function with the intend to delete the "repos" directory if one doesn't want to keep it. 
The `update_upstream_alg_docs.do_it(os.environ['LIBOQS_DIR'])` seems to have been added later mistakenly after removing the repos.
I'd just fix this by moving `rmtree` again to the end of the function, otherwise the -k switch doesn't have any effect.

bhess (changes requested):
see single comment above about `rmtree`.

baentsch:
@dstebila Your "code owner" flag apparently overrules 3 people agreeing that this is OK, so please review.

dstebila (approved):

dstebila:
> @dstebila Your "code owner" flag apparently overrules 3 people agreeing that this is OK, so please review.

I've approved it.  I've also changed the branch protection rules on `main` to require 2 committers to approve, rather than requiring 1 committer and a code owner, since we aren't consistently making use of the code owner functionality.

baentsch:
@SWilson4 unless I hear howls of protest I'll merge this then and re-base "sw-hqc-pull"/#1585 as I'm already using that branch to test out my ideas for resolving #1584...

PR #1595, opened 2023-10-28
Support several pqclean upstream versions

### LINKED ISSUE #1584: Improve granularity of copy_from_upstream

baentsch:
Fixes #1584.

This PR should be completed/rebased once #1589 landed.

This adds the ability to reference different pqclean versions (commits); kyber-aarch64 and dilithium-aarch64 are pinned to a new upstream name "oldpqclean". True upgrades following more current versions/commits of PQClean should be added only to the "pqclean" upstream. The logic has been tested with #1585 and seems to work OK. The proposal is to land this PR, then rebase #1585 to it. 

This PR also removes the unsupported "NOT_READY" feature from `copy_from_upstream`. 

When this lands, an update to the documentation in [the wiki](https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide#automated-inclusion) is warranted/will be done.


dstebila (approved):

baentsch:
FWIW, latest commit is a rebase-and-squash (in case you'd like to check again @dstebila )

SWilson4 (approved):
Looks good to me! I'm glad that it wasn't necessary to rewrite the entire `copy_from_upstream` script. Thanks @baentsch.

baentsch:
> I'm glad that it wasn't necessary to rewrite the entire copy_from_upstream script. 

I'm not sure about that. This script is pretty hard to understand and maintain and a re-write might be (more) beneficial in the long run.

PR #1601, opened 2023-11-04
fix kat_kem linkage to make HQC PR pass CI

baentsch:
This fixes the ["strange MacOS" CI problem](https://github.com/open-quantum-safe/liboqs/pull/1585#issuecomment-1792422996): Conceptually it isn't quite unlogical that the object file of "kem.c" should be made available to the link command of `kat_kem` (even though it should be part of the library). Let's see what CI says about the other platforms now....

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -117,47 +117,46 @@ jobs:
           name: liboqs-openssl3-shared-x64
           path: build/*.deb
 
-# Does anyone care if we delete support for this platform?
-#  linux_arm_emulated:
-#    needs: [stylecheck, buildcheck]
-#    runs-on: ubuntu-latest
-#    strategy:
-#      fail-fast: false
-#      matrix:
-#        include:
-#          - name: armhf
-#            ARCH: armhf
-#            CMAKE_ARGS: -DOQS_ENABLE_SIG_SPHINCS=OFF -DOQS_USE_OPENSSL=OFF -DOQS_OPT_TARGET=generic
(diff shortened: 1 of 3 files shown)


baentsch:
If we'd want to keep `armhf` support (???) feel free to omit the last commit, but merge https://github.com/open-quantum-safe/ci-containers/pull/68 before.

dstebila:
> If we'd want to keep `armhf` support (???) feel free to omit the last commit, but merge [open-quantum-safe/ci-containers#68](https://github.com/open-quantum-safe/ci-containers/pull/68) before.

Some Raspberry Pis are armhf, if I'm not mistaken, so it's probably still somewhat desirable as a target.

baentsch:
> Some Raspberry Pis are armhf

OK... I reverted and did [fix up CI containers](https://github.com/open-quantum-safe/ci-containers/pull/69) such as to pass this again.

Going forward, I'd really appreciate input and feedback on [Platforms to be supported](https://github.com/orgs/open-quantum-safe/discussions/1583): This way we can all decide and document which platforms are to be supported at which level. Or should we also stall this work until the LF takeover is complete and our new masters define the platforms they want commercially supported? I have a hunch that Raspberry3 and older (v4 seems to be the [first aarch64](https://www.raspberrypi.com/products/raspberry-pi-4-model-b/specifications/))'s won't be high on the list for "production level support" :)

dstebila (approved):

SWilson4:
Thank you very much for sorting out that bug, Michael! I'm going to go ahead and merge this.

PR #1603, opened 2023-11-06
Add option to dynamically load libcrypto.so.*

### LINKED ISSUE #1599: Make low-level crypto algorithms implementation switchable
Currently liboqs uses OpenSSL as the default backend for low-level support algorithms, such as AES and SHA3 if the `OQS_USE_OPENSSL` build flag is `ON`; otherwise it will use the bundled implementation of those algorithms. While integrating liboqs in GnuTLS, we found neither of them optimal: it is not desirable to bring in OpenSSL as a dependency[1], while we also don't want to have the same algorithms implemented in multiple places. Therefore I would suggest providing a way to make those algorithm implementation pluggable at either compile time or runtime.

More concretely, what I have in mind is something along the following lines:

* liboqs could provide a build option `OQS_COMMON_CRYPTO`, which takes a combo value, either `OPENSSL`, `DEFAULT`, or `NONE`. `OQS_COMMON_CRYPTO=OPENSSL` is the same as what `OQS_USE_OPENSSL=ON` currently does, `DEFAULT` is equivalent to `OQS_USE_OPENSSL=OFF`, and `NONE` means only a stub implementation (all functions return error) is compiled in
* Consumer applications (such as oqs-provider) could register implementation of those algorithms in their favorite libraries at runtime, maybe through a new API, say `OQS_AES_register_implementation`, `OQS_SHA3_register_implementation`, etc.
* To make the previous step easier, maybe liboqs could provide the default implementation as a static helper library (e.g., `liboqs_openssl.a`, `liboqs_default.a`)

This is similar to the approach taken by tpm2-tss[2].

1. https://gitlab.com/gnutls/gnutls/-/merge_requests/1786#note_1611785601
2. https://github.com/tpm2-software/tpm2-tss/pull/2365

ueno:
As discussed on #1599, this is a strawman proposal to make the common cryptographic algorithms (AES, SHA2, SHA3, and RNG) switchable to avoid OpenSSL dependency at package installation time. The applications can replace the implementation by providing custom callback functions for with `OQS_AES_set_callbacks`, `OQS_SHA2_set_callbacks`,
`OQS_SHA3_set_callbacks`, and `OQS_SHA3_x4_callbacks`.

This also adds `OQS_DLOPEN_OPENSSL` build option to use OpenSSL through dynamically loaded libcrypto.so.* with dlopen, instead of linking to the library at build time.

Fixes #1599


baentsch:
Thanks for the proposal. Conceptually, I (think I) get it, but wonder why all tests (using it, i.e., not Windows :) fail on this code: It seems this PR basically delays loading OpenSSL symbols -- but then, why does the hash (and AES) self-test `tests/test_hash.py` fail all AES and SHA2 tests? SHA3 tests pass. Also, all (PQ) algorithm tests fail. Some further explanations (or updates) would be appreciated as I don't find the time right now to debug into this.

SWilson4:
@ueno Are you interested in continuing to work on this? Just wondering whether to leave this PR open for further development or to close it. Either way, it will remain linked to the issue and can be used as a proof-of-concept / starting point for future work.

ueno:
@SWilson4 sorry for the inactivity; yes, I'm still interested in working on this, targeting the end of this quarter.

ueno:
@baentsch @SWilson4 sorry for the delay; I've managed to make this PR [pass](https://github.com/ueno/liboqs/actions/runs/8183030192) the GitHub CI on Linux and macOS. Could you take a look when you have time?

CMakeLists.txt
```diff
@@ -134,6 +134,20 @@ if(${OQS_USE_OPENSSL})
         endif()
     endif()
     find_package(OpenSSL 1.1.1 REQUIRED)
+
+    if(OQS_DLOPEN_OPENSSL)
+      find_program(OBJDUMP objdump)
```
baentsch, line 139: This seems to imply that this can only work on UNIX'ish systems, right (doc mention of this fact if I got right as per the previous comment, please)? I also infer this from absent Windows tests. Thanks in turn for adding Unix CI for this new feature.

src/common/aes/aes.h
```diff
@@ -147,6 +149,32 @@ void OQS_AES256_CTR_inc_stream_iv(const uint8_t *iv, size_t iv_len, const void *
  */
 void OQS_AES256_CTR_inc_stream_blks(void *ctx, uint8_t *out, size_t out_blks);
 
+struct OQS_AES_callbacks {
```
baentsch, line 152: We aim to be verbose in comments on the public API, so could you add some comment about (the purpose) of this struct, please?

src/common/aes/aes.h
```diff
@@ -147,6 +149,32 @@ void OQS_AES256_CTR_inc_stream_iv(const uint8_t *iv, size_t iv_len, const void *
...
+	void (*AES256_ECB_enc)(const uint8_t *plaintext, const size_t plaintext_len, const uint8_t *key, uint8_t *ciphertext);
+	void (*AES256_ECB_enc_sch)(const uint8_t *plaintext, const size_t plaintext_len, const void *schedule, uint8_t *ciphertext);
+	void (*AES256_CTR_inc_stream_iv)(const uint8_t *iv, size_t iv_len, const void *ctx, uint8_t *out, size_t out_len);
+	void (*AES256_CTR_inc_stream_blks)(void *ctx, uint8_t *out, size_t out_blks);
+};
+
+/**
+ * Set callback functions for AES operations.
+ *
+ * This function may be called at most once before OQS_init to replace
```
baentsch, line 171: Why "at most" (once callable)? What if it's never called? What if it's called more than once?

src/oqsconfig.h.cmake
```diff
@@ -22,6 +22,8 @@
 #cmakedefine OQS_USE_AES_OPENSSL 1
 #cmakedefine OQS_USE_SHA2_OPENSSL 1
 #cmakedefine OQS_USE_SHA3_OPENSSL 1
+#cmakedefine OQS_DLOPEN_OPENSSL 1
+#cmakedefine OQS_OPENSSL_CRYPTO_SONAME "@OQS_OPENSSL_CRYPTO_SONAME@"
```
baentsch, line 26: Just for my personal understanding: Where is this new config variable "OQS_OPENSSL_CRYPTO_SONAME" needed?

baentsch (changes requested):
Thanks for this contribution, @ueno ! I still think I understand how it works, let all our CI "loose" on it and have a few initial comments from a first review pass through. Just to set expectations: Final review and merge may not happen very quickly as we're trying to get a new `liboqs` release out the door and this isn't on the feature list for the release.

.CMake/alg_support.cmake
```diff
@@ -69,6 +69,8 @@ cmake_dependent_option(OQS_USE_SHA3_OPENSSL "" OFF "OQS_USE_OPENSSL" OFF)
 # sanity check: Disable OpenSSL if not a single OpenSSL component define is on
 cmake_dependent_option(OQS_USE_OPENSSL "" ON "OQS_USE_AES_OPENSSL OR OQS_USE_SHA2_OPENSSL OR OQS_USE_SHA3_OPENSSL" OFF)
 
+option(OQS_DLOPEN_OPENSSL "Enable OpenSSL with dlopen" OFF)
```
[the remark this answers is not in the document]
ueno: Done.

CMakeLists.txt
ueno: I think this should work on ELF-based Unix platforms (Linux, BSD, etc.), where objdump whose output is compatible with GNU binutils is installed.

src/common/aes/aes.h
ueno: Done.

src/oqsconfig.h.cmake
ueno: This is used as the first argument of `dlopen`.

src/common/aes/aes.h
ueno: Yes, it can be called multiple times and if it's not called, the default implementation is used. I've updated the doc comment.

CONFIGURE.md
```diff
@@ -109,6 +109,8 @@ These default choices have been made in order to optimize the default performanc
 
 When `OQS_USE_OPENSSL` is `ON`, CMake also scans the filesystem to find the minimum version of OpenSSL required by liboqs (which happens to be 1.1.1). The [OPENSSL_ROOT_DIR](https://cmake.org/cmake/help/latest/module/FindOpenSSL.html) option can be set to aid CMake in its search.
 
+By default, when `OQS_USE_OPENSSL` is `ON`, OpenSSL will be linked to liboqs at build time. This is sometimes undesirable when using liboqs from other cryptographic libraries. The separate option `OQS_DLOPEN_OPENSSL` can be used to defer loading OpenSSL at run time if it is `ON`. The default for this option is `OFF`.
```
baentsch, line 112: This is somewhat confusing (e.g., which option is by default OFF and ON?). What speaks against giving this new option it's own heading (maybe a tad deeper than OQS_USE_OPENSSL, i.e., prefixed with "###")? This also would allow directly referencing it as it'll get its own anchor. Also missing is a statement along the lines "Only available on platforms supporting `dlopen`, e.g., not on Windows" (or so).

CMakeLists.txt
baentsch: OK -- could you then please state it suitably in the config variable documentation for folks not checking out the code? I'm just wondering about MacOS (?).

src/oqsconfig.h.cmake
baentsch: Ah, thanks for the pointer. That then begs a followup question: This variable is gleaned at build time (part of `cmake` execution), right? Does this mean in turn that this facility will break when a binary containing `liboqs` built with this option is run on another machine (where the .so is located somewhere else)?

CMakeLists.txt
ueno: Elaborated that in the documentation.

macOS is a bit tricky as it's [Mach-O](https://en.wikipedia.org/wiki/Mach-O) based, where I'm not sure if SONAME is as part of the ABI, and uses different tooling (e.g., otool instead of ldd/objdump).

src/oqsconfig.h.cmake
ueno: As the variable doesn't contain path, the lookup logic is up to the dynamic linker (ld.so) on another machine. For example, on Fedora 39 the variable would end up with "libcrypto.so.3", where the corresponding shared library resides in /usr/lib64/ on x86_64. If you copy liboqs built with this dlopen feature into another distribution, such as Debian with OpenSSL 3.0, it should still be able to detect the shared library even if the location is /usr/lib/x86_64-linux-gnu/.

On the other hand, it is in theory possible to support arbitrary shared library of OpenSSL, through dlopen, as we [do](https://github.com/golang-fips/openssl/blob/576fe0d377882f8d0fd6537762ef2ff7918facc8/openssl_test.go#L22) in OpenSSL binding in Go, though this is not our current motivation behind this PR.

CONFIGURE.md
ueno: Split it in a separate subsection, and expanded the text.

src/oqsconfig.h.cmake
baentsch: > though this is not our current motivation behind this PR.

and it needn't be. Not encoding the path is good (enough). Thanks for the explanation.

baentsch:
Thanks again, @ueno  -- LGTM now. Now one "logical" question pertaining downstream use: Will this feature work without problems if someone activates the new option "OQS_DLOPEN_OPENSSL" while using `liboqs` in `oqsprovider` (running _within_ openssl) or any other integration (say, `nginx`)? Conceptually, I'd say so, but do you see any chance (well, risk, really) that this "picks up" a library different from the one operating the provider at the same time, creating any form of havoc or "strange behaviours"? Shall we add tests for this in the downstreams?

ueno:
I guess we could add such downstream integration tests through [packit](https://fedoramagazine.org/how-to-use-testing-farm-outside-of-rhel/). So far, I've created a [copr](https://copr.fedorainfracloud.org/coprs/ueno/liboqs-dlopen-openssl/builds/) repository which hosts modified liboqs package with this changeset. With the package installed on Fedora 39, I confirm nginx is working with hybrid key exchange (x25519_kyber768) following the test scenario provided by @beldmit.

ueno:
@baentsch now that 0.10.0 is out, is there anything specific we can do to move this forward? As mentioned, my local testing with oqsprovider and nginx/curl didn't show any problems.

beldmit:
I confirm I see no problems with oqsprovider. The only suspicion I have is related to multithreading use cases

baentsch (approved):
> @baentsch now that 0.10.0 is out, is there anything specific we can do to move this forward? As mentioned, my local testing with oqsprovider and nginx/curl didn't show any problems.

As stated earlier, this LGTM. With 0.10.0 out let's move this towards merge with the second review: @SWilson4 would you have time? Thanks @beldmit for your test/feedback.

SWilson4:
For some reason I'm not able to view logs for the failing CI jobs: only a message that says "this job failed". I'm going to rerun them and see if that helps.

SWilson4:
Functionality-wise, this looks OK to me. Thanks, @ueno!

It would be nice to test the `OQS_*_set_callbacks` API in CI somehow. Is this something that's feasible to do here, or is better done in another setting?

ueno:
@SWilson4 That is a good point. My plan was to add a Nettle backend and run CI [there](https://gitlab.com/gnutls/nettle), once their SHAKE interface has settled. Additionally/alternatively I could add a non-exhaustive test here, wrapping the existing callbacks.

SWilson4:
> @SWilson4 That is a good point. My plan was to add a Nettle backend and run CI [there](https://gitlab.com/gnutls/nettle), once their SHAKE interface has settled. Additionally/alternatively I could add a non-exhaustive test here, wrapping the existing callbacks.

Cool! It would be great to see another liboqs integration. I think it would be a good idea to also have some sort of testing in liboqs CI so that we don't have to monitor an external repo. Perhaps we can set up a [Docker image](https://github.com/open-quantum-safe/ci-containers) containing nettle and run some tests using it as a common code provider. Or maybe this would be better suited to [`oqs-demos`](https://github.com/open-quantum-safe/oqs-demos)---although it's sort of the reverse order of our current demos (i.e. using another project in `liboqs` as opposed to vice versa). Thoughts on this, @baentsch?

At any rate, those options are both (imo) out of scope for this PR, so for now please add the non-exhaustive test you mentioned to get it landed.

ueno:
Sure, https://github.com/open-quantum-safe/liboqs/pull/1603/commits/b60ba69b22a343f6f34c0e0f1e240b1b4076a775 is an attempt to add tests for OQS_*_set_callbacks.

SWilson4 (approved):
Looks good to me. Thank you very much for this contribution!

PR #1604, opened 2023-11-06
add uninstall support

baentsch:
This closes for good the question [how to uninstall liboqs](https://github.com/orgs/open-quantum-safe/discussions/1536).

### DIFF
--- a/.CMake/cmake_uninstall.cmake.in
+++ b/.CMake/cmake_uninstall.cmake.in
@@ -1,0 +1,24 @@
+# As per https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake
+
+if(NOT EXISTS "@CMAKE_BINARY_DIR@/install_manifest.txt")
+  message(FATAL_ERROR "Cannot find install manifest: @CMAKE_BINARY_DIR@/install_manifest.txt")
+endif()
+
+file(READ "@CMAKE_BINARY_DIR@/install_manifest.txt" files)
+string(REGEX REPLACE "\n" ";" files "${files}")
+foreach(file ${files})
+  message(STATUS "Uninstalling $ENV{DESTDIR}${file}")
+  if(IS_SYMLINK "$ENV{DESTDIR}${file}" OR EXISTS "$ENV{DESTDIR}${file}")
+    exec_program(
+      "@CMAKE_COMMAND@" ARGS "-E remove \"$ENV{DESTDIR}${file}\""
+      OUTPUT_VARIABLE rm_out
+      RETURN_VALUE rm_retval
+      )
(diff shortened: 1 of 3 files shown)


SWilson4:
Thanks for setting up this feature, Michael! Not having clear uninstallation instructions has often annoyed me when I've tinkered around with open-source code in the past, so I'm glad that this is moving forward.

It looks like there are empty directories left around after running `ninja uninstall`. On my machine:
```
$ tree <install location>
<install location>
├── include
│   └── oqs
└── lib
    ├── cmake
    │   └── liboqs
    └── pkgconfig

6 directories, 0 files
```
Is this expected?

baentsch:
> Is this expected?

Yes: The `cmake` logic clearly only removes the files it installed, not directories "indirectly" also created. Room for improvement, agreed -- but worth it?

SWilson4:
> > Is this expected?
> 
> Yes: The `cmake` logic clearly only removes the files it installed, not directories "indirectly" also created. Room for improvement, agreed -- but worth it?

Yes, definitely worth it. And I suppose that automatically removing directories is tricky and risky. Perhaps it's worth adding a note to the README suggesting that a fastidious user might want to clean up these empty directories.

SWilson4 (approved):

baentsch:
> Perhaps it's worth adding a note to the README suggesting that a fastidious user might want to clean up these empty directories.

No: Fastidious users regularly scour their hard disks for things that are out of place. I don't want to rob them of that pleasure :->>

dstebila (approved):
Just to understand how this would work.  If you run `cmake --install` and then keep that build directory around, then you can run `cmake --uninstall` afterward.  Would this also create uninstall targets for `ninja`?  And also, if you wipe out the build directory, you can't run uninstall without doing a second install (hopefully over top of the existing install, so that the paths are the same)?

baentsch:
> Just to understand how this would work. If you run `cmake --install` and then keep that build directory around, then you can run `cmake --uninstall` afterward. Would this also create uninstall targets for `ninja`? 

That's exactly what it does. As well as for any other code generator supported by `cmake`.

> And also, if you wipe out the build directory, you can't run uninstall without doing a second install (hopefully over top of the existing install, so that the paths are the same)?

Correct. [This](https://github.com/open-quantum-safe/liboqs/blob/67305dc71301e43f0fc8e521aa98f5e5ff7e2755/.CMake/cmake_uninstall.cmake.in#L13) simply iterates `cmake -E remove` over the `install_manifest.txt` generated when executing `{cmake|make|ninja|...} install`. It will delete the last install. If the user changed `CMAKE_INSTALL_PREFIX` in the same build directory without uninstalling before then tough luck: Same state as without this PR (no uninstall information around).

dstebila:
Thanks for the explanation!  Merging.

PR #1605, opened 2023-11-06
platform support documentation [skip ci]

baentsch:
Implements results of [discussion on platforms supported](https://github.com/orgs/open-quantum-safe/discussions/1583).

### DIFF
--- a/PLATFORMS.md
+++ b/PLATFORMS.md
@@ -1,0 +1,63 @@
+# Supported platforms
+
+This file documents the different platforms supported by `liboqs` and therefore defines three different support tiers:
+
+## Support tiers
+
+This classification is roughly based on the [rust platform support tier classification](https://doc.rust-lang.org/beta/rustc/platform-support.html):
+
+### Tier 1
+
+Tier 1 targets can be thought of as "guaranteed to work". The CI system builds and tests binary versions for each tier 1 target to make sure any change does not negatively affect those platforms. Platform-specific build documentation must exist.
+
+### Tier 2
+
+Tier 2 targets can be thought of as "guaranteed to build". The `liboqs` CI system contains builds for each tier 2 target; testing may or may not be available (typically depending on CI system platform availability). Therefore, tier 2 targets often work to quite a good degree and patches are always welcome! Tier 2 targets may also have known deficiencies caused by a lack of expertise to fix those on a given platform. Again, help and PRs to move platforms from tier 2 to tier 1 are always welcome.
+
+### Tier 3
+
+Tier 3 targets are those which the `liboqs` codebase has support for, but which the CI system does not build or test automatically, so they may or may not work. Platform-specific build documentation should exist.
+
+## Platform tier policy
+
+Tier 2 and tier 1 targets place work on `liboqs` core project developers as a whole, to avoid breaking the target. The broader `liboqs` community may also feel more inclined to support higher-tier targets in their work. Thus, these tiers require commensurate and ongoing efforts from the maintainers of the target, to demonstrate value and to minimize any disruptions to ongoing `liboqs` development.
+
+This policy defines the requirements for accepting a proposed target at a given level of support.
+
+Each tier builds on all the requirements from the previous tier, unless overridden by a stronger requirement.
+
+Change of tier is subject to approval by the `liboqs` technical governance team.  This team is responsible for reviewing and evaluating the target, based on these requirements and their own judgment. The tea may apply additional requirements, including subjective requirements, such as to deal with issues not foreseen by this policy. (Such requirements may subsequently motivate additions to this policy.)
+
+While these criteria attempt to document the policy, that policy still involves human judgment. Targets must fulfill the spirit of the requirements as well, as determined by the judgment of the approving team. Reviewers and team members evaluating targets and target-specific patches should always use their own best judgment regarding the quality of work, and the suitability of a target for the `liboqs` project. Neither this policy nor any decisions made regarding targets shall create any binding agreement or estoppel by any party.
+
+Before filing an issue or pull request (PR) to introduce or promote a target, the target should already meet the corresponding tier requirements. This does not preclude an existing target's maintainers using issues (on the `liboqs` repository or otherwise) to track requirements that have not yet been met, as appropriate; however, before officially proposing the introduction or promotion of a target, it should meet all of the necessary requirements. A target proposal must quote the corresponding requirements verbatim and respond to them as part of explaining how the target meets those requirements. (For the requirements that simply state that the target or the target developers must not do something, it suffices to acknowledge the requirement.)
+
+Several parts of this policy require providing target-specific documentation. Such documentation should typically appear in a subdirectory of the platform-support section of the `liboqs` manual, with a link from the target's entry in platform support.
+
+Note that a target must have already received approval for the next lower tier, and spent a reasonable amount of time at that tier, before making a proposal for promotion to the next higher tier; this is true even if a target meets the requirements for several tiers at once. This policy leaves the precise interpretation of "reasonable amount of time" up to the approving team; the team may scale the amount of time required based on their confidence in the target and its demonstrated track record at its current tier. At a minimum, multiple stable releases of `liboqs` should typically occur between promotions of a target.
+
+The availability or tier of a target in stable `liboqs` is not a hard stability guarantee about the future availability or tier of that target. Higher-level target tiers are an increasing commitment to the support of a target, and we will take that commitment and potential disruptions into account when evaluating the potential demotion or removal of a target that has been part of a stable release. The promotion or demotion of a target will not generally affect existing stable releases, only current development and future releases.
+
+In this policy, the words "must" and "must not" specify absolute requirements that a target must meet to qualify for a tier. The words "should" and "should not" specify requirements that apply in almost all cases, but for which the approving teams may grant an exception for good reason. The word "may" indicates something entirely optional, and does not indicate guidance or recommendations. This language is based on [IETF RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Platforms supported
+
+### Tier 1
+
+- x86_64/amd64/x64 for Ubuntu Linux (Focal) and MacOS (XCode 14)
+- aarch64 for Ubuntu (Focal)
+- armhf/ARM7 and aarch64 emulation on Ubuntu
+- ppc64le for Ubuntu (Focal)
+- s390x for Ubuntu (Focal)
+
+### Tier 2
+
+- x86_64/amd64/x64 for Windows 2022
+- armeabi-v7a, arm64-v8a, x86, x86_64 for Android
+- OS64, TVOS for Apple Mobile
+
+### Tier 3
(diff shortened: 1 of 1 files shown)


PLATFORMS.md
```diff
@@ -0,0 +1,63 @@
...
+- aarch64 for Ubuntu (Focal)
+- armhf/ARM7 and aarch64 emulation on Ubuntu
+- ppc64le for Ubuntu (Focal)
+- s390x for Ubuntu (Focal)
+
+### Tier 2
+
+- x86_64/amd64/x64 for Windows 2022
+- armeabi-v7a, arm64-v8a, x86, x86_64 for Android
+- OS64, TVOS for Apple Mobile
```
dstebila, line 57: I don't know what OS64 is?

PLATFORMS.md
```diff
@@ -0,0 +1,63 @@
...
+
+In this policy, the words "must" and "must not" specify absolute requirements that a target must meet to qualify for a tier. The words "should" and "should not" specify requirements that apply in almost all cases, but for which the approving teams may grant an exception for good reason. The word "may" indicates something entirely optional, and does not indicate guidance or recommendations. This language is based on [IETF RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Platforms supported
+
+### Tier 1
+
+- x86_64/amd64/x64 for Ubuntu Linux (Focal) and MacOS (XCode 14)
+- aarch64 for Ubuntu (Focal)
+- armhf/ARM7 and aarch64 emulation on Ubuntu
```
dstebila, line 49: If armhf is at Tier 1 (or will be with the ci-containers update), great!  Otherwise okay to classify it as tier 2.

PLATFORMS.md
```diff
@@ -0,0 +1,63 @@
...
+### Tier 2
+
+- x86_64/amd64/x64 for Windows 2022
+- armeabi-v7a, arm64-v8a, x86, x86_64 for Android
+- OS64, TVOS for Apple Mobile
+
+### Tier 3
+
+- x86 for Windows
+- aarch64 for MacOS
```
dstebila, line 62: aarch64 is clearly the intended future of the macOS platform, so we would definitely want to get this up to tier 1.  Is it not higher right now because we don't have any CI on this platform?

PLATFORMS.md
baentsch: Me neither. It's [this test run](https://github.com/open-quantum-safe/liboqs/actions/runs/6776860846/job/18419274742) triggered by https://github.com/open-quantum-safe/liboqs/blob/aeac3a441dd3d95821b04a26baae6f73013538d8/.github/workflows/apple.yml#L12 contributed in https://github.com/open-quantum-safe/liboqs/pull/1546. @res0nance could you clarify?.

PLATFORMS.md
baentsch: We have [a successful CI run](https://github.com/open-quantum-safe/liboqs/actions/runs/6776860848/job/18419347065) as I already did the [ci-containers update (adding doxygen)](https://github.com/open-quantum-safe/ci-containers/pull/69) -- without PR review, admittedly: I didn't want to wait another 24 h to see the HQC CI pass.

PLATFORMS.md
baentsch: >  Is it not higher right now because we don't have any CI on this platform

Correct

thomwiggers: Github now has M1 runners I believe. I learnt this when someone was complaining about how expensive they are, though, and I'm not sure if they are available for OSS yet.

baentsch: > I'm not sure if they are available for OSS yet.

They are not: See [documentation](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#machine-specs-for-larger-runners) and [worked example](https://github.com/open-quantum-safe/oqs-provider/actions/runs/7075061285/job/19256776098).

baentsch: For completeness: Error message:

```
The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings.
```

PLATFORMS.md
res0nance: OS64 is ARM64 for iphone

PLATFORMS.md
SWilson4: I agree that aarch64/Mac is correctly classified as "tier 3" but ideally would be higher. Perhaps we could look into using our M1 profiling machine as a self-hosted runner? I don't think we would want the strain of running tests on it as often as on the GitHub runners, but a scheduled daily or weekly job would be more palatable.

PLATFORMS.md
```diff
@@ -0,0 +1,63 @@
+# Supported platforms
+
+This file documents the different platforms supported by `liboqs` and therefore defines three different support tiers:
+
+## Support tiers
+
+This classification is roughly based on the [rust platform support tier classification](https://doc.rust-lang.org/beta/rustc/platform-support.html):
```
SWilson4, line 7: Might I suggest either a "Tier 0" or additional criteria for "Tier 1" which includes constant-time testing and documentation of CT issues? We currently have automated CT-testing for x86 but not aarch64. Or would this be viewed as out of scope for this document?

PLATFORMS.md
```diff
@@ -0,0 +1,63 @@
...
+In this policy, the words "must" and "must not" specify absolute requirements that a target must meet to qualify for a tier. The words "should" and "should not" specify requirements that apply in almost all cases, but for which the approving teams may grant an exception for good reason. The word "may" indicates something entirely optional, and does not indicate guidance or recommendations. This language is based on [IETF RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Platforms supported
+
+### Tier 1
+
+- x86_64/amd64/x64 for Ubuntu Linux (Focal) and MacOS (XCode 14)
+- aarch64 for Ubuntu (Focal)
+- armhf/ARM7 and aarch64 emulation on Ubuntu
+- ppc64le for Ubuntu (Focal)
```
SWilson4, line 50: I didn't recall that we had CI tests for ppc and s390x, so I dug around and found the Travis CI setup. But it looks like it hasn't been running in a while: https://app.travis-ci.com/github/open-quantum-safe/liboqs. Unless I'm missing something, we should probably knock these two down to Tier 3 pending a CI fix.

PLATFORMS.md
baentsch: Agreed. Also thought the same. But wouldn't this effort be a waste of time? The LF companies will pay hundreds of thousands -- why not use (a tiny fraction) of this money to pay for M1 runners and save us this rather tedious work? It surely would be a better use of the money than for CO2-gobbling trips of some LF executives to PQ "workshops" :)

PLATFORMS.md
baentsch: Sounds reasonable. Would you want to add proposed text straight to this PR?

PLATFORMS.md
SWilson4: > _Very_ interesting. Didn't realize that (either). Makes perfect sense -- but I suppose you mean knocking them down to Tier 2?

I had thought Tier 3 since it doesn't look to me like we're even building on those architectures anywhere at the moment (until Travis is up and running again).

baentsch:
Additional thought: "Tier 0" sounds a bit .... "close to negative" :), i.e., to me feels not properly denominating the "highest" tier. Also, what about platforms for which there shall be formally verified implementations available (that also should be CT & side-channel attack resistant & code-reviewed & fuzzed & .... etc.): Shouldn't those represent the highest tier? Slight disadvantage of such tier: It would have no entries right now...

PLATFORMS.md
dstebila: @SWilson4 will you do this update?

dstebila: Oh never mind I see that this was done, thanks.

PLATFORMS.md
dstebila: I've created issue #1630 to track this, and will close this thread as resolved.

dstebila (approved):

PLATFORMS.md
```diff
@@ -0,0 +1,64 @@
...
+
+This file documents the different platforms supported by `liboqs` and therefore defines three different support tiers:
+
+## Support tiers
+
+This classification is roughly based on the [rust platform support tier classification](https://doc.rust-lang.org/beta/rustc/platform-support.html):
+
+### Tier 1
+
+Tier 1 targets can be thought of as "guaranteed to work". The CI system builds and tests binary versions for each tier 1 target to make sure any change does not negatively affect those platforms. Platform-specific build documentation must exist. Tier 1 targets marked with an asterisk are additionally tested for constant-time behaviour. The CI system contains automated constant-time testing for each of these starred targets, and all failures are documented in the `tests/constant_time` directory. IMPORTANT: This does not mean that constant-time behaviour is guaranteed on these targets, or that non-constant-time behaviour is limited to documented exceptions. It does, however, mean that `liboqs` developers should track constant-time issues on these platforms.
```
SWilson4, line 11:
```suggestion
Tier 1 targets can be thought of as "guaranteed to work". The CI system builds and tests binary versions for each tier 1 target to make sure any change does not negatively affect those platforms. Platform-specific build documentation must exist. Tier 1 targets marked with a dagger (†) are additionally tested for constant-time behaviour. The CI system contains automated constant-time testing for each of these starred targets, and all failures are documented in the `tests/constant_time` directory. IMPORTANT: This does not mean that constant-time behaviour is guaranteed on these targets, or that non-constant-time behaviour is limited to documented exceptions. It does, however, mean that `liboqs` developers should track constant-time issues on these platforms.
```

PLATFORMS.md
```diff
@@ -0,0 +1,64 @@
...
+
+The availability or tier of a target in stable `liboqs` is not a hard stability guarantee about the future availability or tier of that target. Higher-level target tiers are an increasing commitment to the support of a target, and we will take that commitment and potential disruptions into account when evaluating the potential demotion or removal of a target that has been part of a stable release. The promotion or demotion of a target will not generally affect existing stable releases, only current development and future releases.
+
+In this policy, the words "must" and "must not" specify absolute requirements that a target must meet to qualify for a tier. The words "should" and "should not" specify requirements that apply in almost all cases, but for which the approving teams may grant an exception for good reason. The word "may" indicates something entirely optional, and does not indicate guidance or recommendations. This language is based on [IETF RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Platforms supported
+
+### Tier 1
+
+- x86_64/amd64/x64 for Ubuntu Linux (Focal)*
```
SWilson4, line 47:
```suggestion
- x86_64/amd64/x64 for Ubuntu Linux (Focal)†
```

SWilson4 (approved):
I made one (very, very minor) suggestion to change the asterisk to a dagger, as I realized that asterisks (1) don't stand out much to a reader and (2) are used for a wide variety of other meanings in documentation. Plus we already use a dagger in the README.

dstebila (approved):

Issue #1617, opened 2023-11-29
Falcon constant time tests failing


baentsch:
The [Falcon CT CI tests are failing since quite a few weeks](https://github.com/open-quantum-safe/liboqs/actions/workflows/weekly.yml) coincidentally with the last commit ([adding uninstall support](https://github.com/open-quantum-safe/liboqs/pull/1604) -- but I somehow cannot imagine how this should impact CT for Falcon). Other thoughts what could be at play welcome.

SWilson4:
I think it's more likely that #1585 somehow caused the failures, as it touched Falcon source files, and the timing is right. However, as far as I can tell the only changes were to whitespace, so I'm still a little perplexed by the sudden breakage.

At first glance, it looks like at least some of the failures are indeed non-constant-time behaviour. [The first one](https://github.com/open-quantum-safe/liboqs/actions/runs/6992637406/job/19024477997#step:6:192), for example, points to https://github.com/open-quantum-safe/liboqs/blob/78e65bf143edeceae3179c40b511e1313621ffde/src/sig/falcon/pqclean_falcon-512_avx2/keygen.c#L4153 which I believe we would consider a "pass", based on the in-line documentation.

I'll be updating the Falcon code shortly to address #1561 / https://github.com/PQClean/PQClean/issues/523 / #1608, so I suggest we delay fixing this after that's done. No sense in possibly having to update the constant-time issues / passes all over again after it lands. How does that sound @baentsch?

Not directly related to the Falcon failures, but worth noting: It seems like the "generic" run is [picking up errors in the `AVX2` functions](https://github.com/open-quantum-safe/liboqs/actions/runs/6992637406/job/19024477997#step:6:947). Is this the desired behaviour? I would have expected it to test only the "clean" implementations.

SWilson4:
> Not directly related to the Falcon failures, but worth noting: It seems like the "generic" run is [picking up errors in the `AVX2` functions](https://github.com/open-quantum-safe/liboqs/actions/runs/6992637406/job/19024477997#step:6:947). Is this the desired behaviour? I would have expected it to test only the "clean" implementations.

Ah, I see why this is happening. See #1618.

baentsch:
Bad news: CT testing now has more failures than before -- but then again, we're testing more configs now...

SWilson4:
> Bad news: CT testing now has more failures than before -- but then again, we're testing more configs now...

[...]

PR #1618, opened 2023-12-01
Ensure generic OQS_OPT_TARGET in weekly CT tests

SWilson4:
Set `-DOQS_DIST_BUILD=OFF` in the "generic" weekly constant-time test run so that the `OQS_OPT_TARGET` variable has an effect. This currently isn't being done, which means that only the `AVX2` versions are being tested (in the weekly "extensions" run). As mentioned in #1617.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -14,7 +14,7 @@ jobs:
         include:
           - name: generic
             container: openquantumsafe/ci-ubuntu-focal-x86_64:latest
-            CMAKE_ARGS: -DOQS_OPT_TARGET=generic -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON
+            CMAKE_ARGS: -DOQS_DIST_BUILD=OFF -DOQS_OPT_TARGET=generic -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON
             PYTEST_ARGS: --numprocesses=auto -k 'test_constant_time'
             SKIP_ALGS: 'SPHINCS\+-SHA*,Classic-McEliece-(.)*'
           - name: extensions


baentsch (commented):
Conceptually I agree that we should test the generic/reference/C code for CT, so this PR makes sense. But I wonder whether it should be extended to also (keep) testing the optimized code? It's running most of the time (as say AVX2 is available in most machines these days). See #1619 as a possible next step.

baentsch:
> The optimized code will still be tested via the [extensions](https://github.com/open-quantum-safe/liboqs/pull/1618/files#diff-b077dee47bccc37c2b6fef32cdd102f36acadad73ede1f410c3941f39c50cd94R22) job (I think)

Ah, Yes. Completely overlooked that one. Indeed "haswell" has the relevant extensions. Thanks (for PR & explanation)!

baentsch:
@dstebila Weekly runs are on Friday, so approval & merge before would be great. Alternatively, shall we exclude such "infrastructure" PRs from the "2 approvals needed" rule?

dstebila (approved):

dstebila:
> @dstebila Weekly runs are on Friday, so approval & merge before would be great. Alternatively, shall we exclude such "infrastructure" PRs from the "2 approvals needed" rule?

I don't know how to change the "2 approvals needed" rule to only apply to certain directories -- we can apply it only to certain branches, but I don't see directories as an option.

baentsch:
> I don't know how to change the "2 approvals needed" rule to only apply to certain directories -- we can apply it only to certain branches, but I don't see directories as an option.

Hmm -- just read the documentation and I also didn't find a way how to do that. A bit disappointing, I'd say. But what I did find is an option that demands approval by code owners. And _that_ can be set to specific directories, right? So what about the idea to carefully define code ownership(s), set this flag and revert back to just 1 review needed for PRs? Not ideal if there's a sole code owner designated for some sections and that person isn't around -- in such case the 4 eyes rule would be better again...

Issue #1490, opened 2023-06-06
Add option to exclude the compilation of OQS_randombytes_system() when a custom algorithm is specified


mabarger:
Compiling `liboqs` for systems without a (full) standard library (e.g. for embedded systems) causes problems at the moment, as `liboqs` uses multiple library functions (e.g. `fopen`), which might not be available on these systems.

Some of these calls are located in the function(s) `OQS_randombytes_system()` in [rand.c](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/rand/rand.c) and will also be compiled  when providing a custom randombytes algorithm with `OQS_randombytes_custom_algorithm()`. Unless I am missing something the compilation of `OQS_randombytes_system()` is not necessary when a custom algorithm is provided and removing them from the compilation process could ease future use and integration.

I propose adding a new configuration option, which will enable/disable compilation of these functions by wrapping them in a preprocessor macro similar to this:

```C
@@ -35,7 +35,11 @@ void OQS_randombytes_openssl(uint8_t *random_array, size_t bytes_to_read);
 // Use OpenSSL's RAND_bytes as the default PRNG
 static void (*oqs_randombytes_algorithm)(uint8_t *, size_t) = &OQS_randombytes_openssl;
 #else
+#if !defined(OQS_NO_STDLIB)
 static void (*oqs_randombytes_algorithm)(uint8_t *, size_t) = &OQS_randombytes_system;
+#else
+static void (*oqs_randombytes_algorithm)(uint8_t *, size_t) = NULL;
+#endif
 #endif
 OQS_API OQS_STATUS OQS_randombytes_switch_algorithm(const char *algorithm) {
        if (0 == strcasecmp(OQS_RAND_alg_system, algorithm)) {
@@ -64,6 +68,7 @@ OQS_API void OQS_randombytes(uint8_t *random_array, size_t bytes_to_read) {
        oqs_randombytes_algorithm(random_array, bytes_to_read);
 }
 
+#if !defined(OQS_NO_STDLIB)
 #if !defined(_WIN32)
 #if defined(OQS_HAVE_GETENTROPY)
 void OQS_randombytes_system(uint8_t *random_array, size_t bytes_to_read) {
@@ -120,6 +125,7 @@ void OQS_randombytes_system(uint8_t *random_array, size_t bytes_to_read) {
        CryptReleaseContext(hCryptProv, 0);
 }
 #endif
+#endif
 
 #ifdef OQS_USE_OPENSSL
[...]

PR #1621, opened 2023-12-04
Add support for Zephyr RTOS

Frauschi:
Hi everybody,

This PR adds support for the [Zephyr RTOS](https://github.com/zephyrproject-rtos/zephyr), so liboqs can be used as a module (see [here](https://docs.zephyrproject.org/latest/develop/modules.html)). 

The main changes to the existing library code are focussed on the CMake build environment to enable the library to be built with the Zephyr SDK. All changes should not influence existing builds outside the Zephyr environment. (However, some changes also improve the integration of liboqs directly in a CMake-based project without installing it system-wide.) 

One bigger source code change is related to random number generation. Zephyr doesn't support any of the currently present RNG interfaces. To be most flexible, I added a generic `OQS_USE_EXTERNAL_RNG` option (disabled by default). In this case, an externally defined method `OQS_randombytes_external()` must be present during link time, which is then used to obtain random data. For Zephyr, I implemented that method to generate random data using their RNG interface. 

The main Zephyr integration is all contained in the new `zephyr` folder in the root directory (common practice for Zephyr modules to have this folder). In that directory, there is a custom `CMakeLists.txt` to integrate the existing CMake infrastructure into Zephyr. Furthermore, a Kconfig file is provided to configure the build (and especially the algorithm selection). Finally, the new `OQS_randombytes_external` method is implemented in a small C source file here. 

As this is my first PR for this project, I'm not sure if the changes are upstreamable in the current form. But I'm looking forward to your feedback.

Kind regards,
Tobi


CMakeLists.txt
```diff
@@ -61,7 +68,7 @@ elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|arm64|arm64v8")
     if(${OQS_DIST_BUILD})
         set(OQS_DIST_ARM64_V8_BUILD ON)
     endif()
-elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "armel|armhf|armv7|arm32v7")
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "armel|armhf|armv7|arm32v7|arm")
```
baentsch, line 71: Is "arm" a valid (unambiguous) name for 32bit ARM? Is there a list of generally agreed-upon names for ARM architectures?

CMakeLists.txt
```diff
@@ -26,18 +26,25 @@ option(OQS_BUILD_ONLY_LIB "Build only liboqs and do not expose build targets for
...
 set(CMAKE_C_STANDARD 11)
 set(CMAKE_C_STANDARD_REQUIRED ON)
-set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_C_VISIBILITY_PRESET hidden)
 set(OQS_VERSION_TEXT "0.10.0-dev")
 set(OQS_COMPILE_BUILD_TARGET "${CMAKE_SYSTEM_PROCESSOR}-${CMAKE_HOST_SYSTEM}")
 set(OQS_MINIMAL_GCC_VERSION "7.1.0")
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+if(OQS_POSITION_INDEPENDENT_CODE)
```
baentsch, line 42: Why is this introduced -- to then disable it in the Zephyr section? Couldn't this be avoided by simply setting "CMAKE_POSITION_INDEPENDENT_CODE" to "OFF" in the Zephyr `cmake` files?

baentsch (changes requested):
Thanks for the contribution and new platform! Please provide feedback to some single comments/questions to help me understand things better. [According to the Zephyr documentation there should be a person maintaining this module](https://docs.zephyrproject.org/latest/develop/modules.html). Can we safely assume you'll do this -- also giving us a person to point questions to coming towards `liboqs` (related to this integration)? It would also be good to know that there is alignment on the platforms Zephyr and `liboqs` support: Are/have you already ascertain(ed) that?

CMakeLists.txt
Frauschi: Based on my (limited) personal experience, bare-metal and RTOS targets typically define 32bit ARM only as „arm“, 64bit ARM as „arm64“. But I will google about that to find a proper valid statement.

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
+# SPDX-License-Identifier: MIT
+
+if(CONFIG_LIBOQS)
```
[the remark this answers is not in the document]
Frauschi: All the `CONFIG_` CMake variables come from the Zephyr Kconfig build system. They are used to configure a build on a fine-grained basis without changing CMake files. We use them in the CMakeLists.txt in the Zephyr directory to incorporate user-defined settings (e.g., a user can pull liboqs into his overall local build system, but wants to disable it for a specific project or build).

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
...
+        set(CMAKE_HAVE_ALIGNED_ALLOC OFF)
+        set(CMAKE_HAVE_POSIX_MEMALIGN OFF)
+        set(CMAKE_HAVE_MEMALIGN OFF)
+        set(CMAKE_HAVE_EXPLICIT_BZERO OFF)
+        set(CMAKE_HAVE_MEMSET_S OFF)
+        set(CC_SUPPORTS_WA_NOEXECSTACK OFF)
+        set(LD_SUPPORTS_WL_Z_NOEXECSTACK OFF)
+
+        # Algorithm selection
+        if(CONFIG_LIBOQS_ENABLE_KEM_BIKE)
```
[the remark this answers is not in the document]
Frauschi: Yes, those are also set by the Kconfig system from within Zehpyr. This is the only way to configure the build for Zephyr. Hence, we cannot use the OQS defines directly, but have to „proxy“ them in this „ugly“ way.

CMakeLists.txt
Frauschi: I will look into that tomorrow and update the commit.

Frauschi:
> Thanks for the contribution and new platform! Please provide feedback to some single comments/questions to help me understand things better. [According to the Zephyr documentation there should be a person maintaining this module](https://docs.zephyrproject.org/latest/develop/modules.html). Can we safely assume you'll do this -- also giving us a person to point questions to coming towards `liboqs` (related to this integration)? It would also be good to know that there is alignment on the platforms Zephyr and `liboqs` support: Are/have you already ascertain(ed) that?

Yes, I am happy to maintain the port. 

I haven‘t completely looked into the supported platforms actually. What exactly are you looking for in that regard?

CMakeLists.txt
Frauschi: I just removed the change regarding the `POSITION_INDEPENDENT_CODE` option. It is now disabled from within the CMakeLists.txt in the zephyr directory.

CMakeLists.txt
Frauschi: “arm” is, in fact, not an unambiguous name for 32-bit ARM architectures (I found some systems using it also for 64-bit systems). Hence, I removed this option again and instead added a “workaround” for Zephyr within the CMakeLists.txt in the zephyr directory.

zephyr/CMakeLists.txt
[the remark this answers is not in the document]
baentsch: Thanks for the explanation.

Frauschi:
> Is there a 1:1 match regarding the platforms that Zephyr supports and [the ones that `liboqs` supports](https://github.com/open-quantum-safe/liboqs/pull/1605)? Or is there some deficiency that would need to be addressed in `liboqs` to get there? Or is that not necessary?

Currently, the matching platforms are 32-bit ARM, 64-bit ARM, x86 and “native build” on Linux (which is then probably one of 32/64-bit ARM, x86, or x86_64). In the latest changes of the commits, I added a dependency from within Zephyr to only enable the module if one of these supported platforms is selected. Otherwise, the liboqs module is disabled and a warning is raised in Zephyr.

To fully support all platforms of Zephyr in liboqs, we would have to add support for `arc`, `mips`, `riscv`, `sparc`, and `xtensa`. I think the most popular platforms of these are `riscv` and `xtensa` (ESP32 controllers). On these platforms, the native C code should run without problems. 

I think it doesn't make sense to add those platforms now until there is actual request (also, I don't have hardware to test liboqs on those platforms, actually).

zephyr/CMakeLists.txt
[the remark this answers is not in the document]
baentsch: Understood. That makes me wonder how (whether) we shall test this integration: Shall this be done within `liboqs` CI (which would mean we somehow have to set these configs) or do you want to rely on doing this (only) within Zephyr CI?

baentsch:
> I think it doesn't make sense to add those platforms now until there is actual request

Fully agree.

> I added a dependency from within Zephyr to only enable the module if one of these supported platforms is selected.

Good addition. Resolves my concerns around platforms: Thanks!

zephyr/CMakeLists.txt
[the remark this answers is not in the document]
Frauschi: After looking at the test infrastructure, I think it makes sense to test the Zephyr port within the `liboqs` CI. There is a docker container available from Zephyr with the SDK ready to go, which should make an integration relatively easy. 

I thought about mirroring the tests from `test_kem.c` and `test_sig.c`, one time with only the NIST selected algorithms enabled (and testing for the remaining ones to be disabled, hence, testing the ENABLE options above), and one time with all algorithms enabled. 

Do you think other tests are necessary, too? Unfortunately, I think won't have time for the tests until the end of this week, but I will add them if you confirm.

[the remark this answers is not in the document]
baentsch: > and one time with all algorithms enabled.

I'd _only_ go with this as the NIST algs are just a subset. But then again, let's see what the runtime of this will be: If it's very long, it may be better to scale things back...

> I think won't have time for the tests until the end of this week, but I will add them if you confirm.

No rush here. Please do at a pace that suits your other activities.

SWilson4:
Hi @Frauschi, thanks for the PR. Is it absolutely necessary to introduce the `OQS_USE_EXTERNAL_RNG` option? We have an API function to [set a custom RNG](https://github.com/open-quantum-safe/liboqs/blob/78e65bf143edeceae3179c40b511e1313621ffde/src/common/rand/rand.h#L44); might this suffice for your purposes?

baentsch:
Your latest comment @SWilson4 reminded me of an old-ish issue: Would it make sense to address #1490 along with this PR? What'd you say, @Frauschi ?

baentsch:
> When we disable both default OQS_randombytes_system() methods during compilation, we rely on the user to set a proper method using OQS_randombytes_custom_algorithm() during runtime. When the user forgets to do that, we cause a runtime error, as no randombytes method is available.

Well, two thoughts: 
1) Do we really want to play nanny for users changing good default configs? They should know what they do and deal with ensuing problems themselves if they activate e.g., a generic "OQS_EMBEDDED_API" option (maybe some good documentation for that option would be helpful :). We also could issue a `cmake` warning at build time if such generic embedded system config as proposed in #1490 gets activated, e.g., "no default RNG available. Be sure to call `OQS_randombytes_custom_algorithm`". This in place of an explicit flag stating that one does that (your "OQS_USE_EXTERNAL_RNG" proposal -- which arguably also would need to be amended for other thus possibly "incapacitated" calls, e.g., SHA2, etc). 

2) Or (and?) possibly more simple: What about being more defensive [here](https://github.com/open-quantum-safe/liboqs/blob/78e65bf143edeceae3179c40b511e1313621ffde/src/common/rand/rand.c#L56-L58) and output a runtime warning if `oqs_randombytes_algorithm` is NULL?

Also tagging @mabarger for input as to the propensity/time to contribute as per https://github.com/open-quantum-safe/liboqs/issues/1490#issuecomment-1596888969

zephyr/CMakeLists.txt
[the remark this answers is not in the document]
Frauschi: I added two sample applications in the `zephyr` directories for KEMs and Signatures. Both are copied and adapted from `test_kem.c` and `test_sig.c` to Zephyr. These examples are now also used in a new Github CI Action to test the two applications (execute them). I added Zephyr Version 3.4 and 3.5.

I also took a look at the CircleCI, but I even can't get those tests running locally on my side. Can you give me any hints on that side or are the Github Actions enough?

Frauschi:
> 1. Do we really want to play nanny for users changing good default configs? They should know what they do and deal with ensuing problems themselves if they activate e.g., a generic "OQS_EMBEDDED_API" option

True, I reverted my `OQS_USE_EXTERNAL_RNG` option and added a generic `OQS_EMBEDDED_BUILD` option as proposed in #1490. 

> 2\. Or (and?) possibly more simple: What about being more defensive [here](https://github.com/open-quantum-safe/liboqs/blob/78e65bf143edeceae3179c40b511e1313621ffde/src/common/rand/rand.c#L56-L58) and output a runtime warning if `oqs_randombytes_algorithm` is NULL?

Added that, too.

@baentsch @SWilson4 what do you think about the new RNG setup?

baentsch:
@Frauschi I don't quite "feel" like reviewing this so late in my day -- but will do before Monday. Independent of that could you please rebase the PR to the latest "main" to ascertain that CI passes?

zephyr/CMakeLists.txt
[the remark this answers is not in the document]
baentsch: Thanks for these tests!

> are the Github Actions enough

Absolutely; we actually prefer Github over CCI.

zephyr/samples/KEMs/sample.yaml
```diff
@@ -0,0 +1,16 @@
+sample:
+  description: liboqs KEM example
+  name: lioqs_kem_example
```
baentsch, line 3: typo

src/common/rand/rand.c
```diff
@@ -54,9 +64,14 @@ OQS_API void OQS_randombytes_custom_algorithm(void (*algorithm_ptr)(uint8_t *, s
 }
 
 OQS_API void OQS_randombytes(uint8_t *random_array, size_t bytes_to_read) {
+	if (oqs_randombytes_algorithm == NULL) {
+		fprintf(stderr, "OQS_randombytes called before algorithm was set\n");
```
baentsch, line 68: Well, possibly not before -- might also have happened because someone switched to "system" (see above), right?

src/common/rand/rand.c
```diff
@@ -107,6 +122,7 @@ void OQS_randombytes_system(uint8_t *random_array, size_t bytes_to_read) {
 	CryptReleaseContext(hCryptProv, 0);
 }
 #endif
+#endif
```
baentsch, line 125: Suggest adding a comment as to which #define(s) gets closed here

baentsch (changes requested):
Conceptually looks good. Let's see what CI says. Besides some single comments that would be good to address, I am missing documentation for the new config variable. It may be good to explain differences to [OQS_BUILD_ONLY_LIB](https://github.com/open-quantum-safe/liboqs/blob/main/CONFIGURE.md#OQS_BUILD_ONLY_LIB) and how it impacts [OQS_DIST_BUILD](https://github.com/open-quantum-safe/liboqs/blob/main/CONFIGURE.md#oqs_dist_build).

zephyr/samples/KEMs/sample.yaml
Frauschi: fixed

src/common/rand/rand.c
Frauschi: done

src/common/rand/rand.c
```diff
@@ -28,11 +30,19 @@ void OQS_randombytes_openssl(uint8_t *random_array, size_t bytes_to_read);
...
+#else
+static void (*oqs_randombytes_algorithm)(uint8_t *, size_t) = NULL;
+#endif
 #endif
 OQS_API OQS_STATUS OQS_randombytes_switch_algorithm(const char *algorithm) {
 	if (0 == strcasecmp(OQS_RAND_alg_system, algorithm)) {
+#if !defined(OQS_EMBEDDED_BUILD)
 		oqs_randombytes_algorithm = &OQS_randombytes_system;
+#else
+		oqs_randombytes_algorithm = NULL;
```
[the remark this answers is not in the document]
Frauschi: With the current implementation, you are right. 

Currently, the implementation of the `OQS_randombytes_system` function is completely disabled when `OQS_EMBEDDED_BUILD` is enabled. Hence, we cannot switch “back” to the system implementation using `OQS_randombytes_switch_algorithm`, as there is no implementation for it at all. 

As I saw it, currently  `OQS_randombytes_switch_algorithm` is only used within the test applications in the `tests` directory (which don't use `OQS_EMBEDDED_BUILD'). But I may miss some other important places.

One solution I can think of is to store the function pointer provided by `OQS_randombytes_custom_algorithm` internally in a second variable and use that as the “new” system implementation. So, when the user changed the RNG algorithm using `OQS_randombytes_switch_algorithm` and then changes it back to “system”, the method originally provided by `OQS_randombytes_custom_algorithm` is used again. This would make sure that `OQS_randombytes_switch_algorithm` works as intended all the time. However, this would permanently override the default system implementation (until reboot). 

Another solution could be to add an empty implementation of `OQS_randombytes_system` in case of an enabled `OQS_EMBEDDED_BUILD` causing a runtime error stating that one has to call `OQS_randombytes_switch_algorithm` providing a custom method. In this case, the user has to call `OQS_randombytes_custom_algorithm`  again after changing algorithms using `OQS_randombytes_switch_algorithm`. 

What do you think about these possibilities? Or do you have an idea for another solution?

src/common/rand/rand.c
Frauschi: I updated the error message to make it more “general”.

Frauschi:
> missing documentation for the new config variable

I added documentation for the new option in `CONFIGURE.md` and also added a README for the Zephyr port in the `zephyr` directory. 

> It may be good to explain differences to [OQS_BUILD_ONLY_LIB](https://github.com/open-quantum-safe/liboqs/blob/main/CONFIGURE.md#OQS_BUILD_ONLY_LIB) and how it impacts [OQS_DIST_BUILD](https://github.com/open-quantum-safe/liboqs/blob/main/CONFIGURE.md#oqs_dist_build).

I didn't mention these two explicitly, as they are not really related to the new option. With the explanation of the new option, however, I think this should also be clear to a potential user of `OQS_EMBEDDED_BUILD`.

src/common/rand/rand.c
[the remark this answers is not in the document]
baentsch: > I may miss some other important places.

Well, it's a public API so we can't know all places but have to assume (and support arbitrary) usage.

> What do you think about these possibilities? 

I very much like option 2: a) It only adds code in case "OQS_EMBEDDED_BUILD" is set, b) only impacts users calling `OQS_randombytes_switch_algorithm` with this option set (i.e., no present users) and c) adds a warning for the unwary. So would you please add this? Please also include a warning as per the above in the comment documentation of `OQS_randombytes_switch_algorithm`.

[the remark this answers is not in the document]
Frauschi: > So would you please add this? Please also include a warning as per the above in the comment documentation of `OQS_randombytes_switch_algorithm`.

I updated the commits to incorporate option 2 and added the warning. This should indeed be the best solution without changes to existing users. 

I also did a small cleanup in `rand.c` regarding all the different `OQS_randombytes_system` implementations that are present. Hope that is ok.

.CMake/compiler_opts.cmake
```diff
@@ -85,7 +85,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang|GNU")
     add_compile_options(${OQS_OPT_FLAG})
 
     # If this is not a dist build we also need to set the OQS_USE_[EXTENSION] flags
-    if(NOT ${OQS_DIST_BUILD})
+    if(NOT ${OQS_DIST_BUILD} AND NOT CMAKE_CROSSCOMPILING)
```
baentsch, line 88: For my personal understanding: Why is this change needed?

src/common/rand/rand.h
```diff
@@ -28,6 +28,10 @@ extern "C" {
 /**
  * Switches OQS_randombytes to use the specified algorithm.
  *
+ * @warning In case you have set a custom algorithm using `OQS_randombytes_custom_algorithm`
+ * before, this function will overwrite it again. Hence, you have to set your custum
```
baentsch, line 32: typo "custom"

baentsch (approved):
Apart from a nit (typo) LGTM. Thanks! Let's CI decide now. Also tagging @SWilson4 @dstebila for second review.

.CMake/compiler_opts.cmake
Frauschi: The `try_run()` command inside `gcc_clang_instrinsics.cmake`  doesn't properly use the provided compiler flags and therefore can't produce a working executable using some embedded toolchains (e.g. the Zephyr one, but also applies to other toolchains). In this case, CMake fails and compilation is impossible. 

After the `try_run()` call, all CPU features have been disabled either way for cross compilation. Hence, I disabled the execution of the file completely for cross compilation.

dstebila (approved):

baentsch:
Thanks again for your contribution, @Frauschi ! Merged.

PR #1626, opened 2023-12-12
Add ML-DSA-ipd and ML-KEM-ipd & NIST supplied test vectors

### LINKED ISSUE #1521: Update Kyber and Dilithium to FIPS versions

bhess:
Pull ML-DSA-ipd / ML-KEM-ipd from pq-crystals "standard" branch.
Adds test cases with NIST supplied test vectors: https://csrc.nist.gov/Projects/post-quantum-cryptography/post-quantum-cryptography-standardization/example-files (see also _Note on the intermediate values for ML-KEM:_ and _Note on the intermediate values for ML-DSA:_)

The NIST files are rather large (several MB) and we only use parts of the data that can be passed via the public API + randomness. So we process them and store only the data we need in `tests/PQC_Intermediate_Values` (script is included to reproduce the files)

The PR adds new test code vectors_sig.c and vectors_kem.c that for testing the NIST vectors.

The idea is to keep the current Kyber/Dilithium (round 3) versions for compatibility reasons, and add the ML-* algorithms additionally.

The naming of the algorithms is done following the IETF hackathon: ML-DSA-44-ipd, ML-DSA-65-ipd, ML-DSA-87-ipd, ML-KEM-512-ipd, ML-KEM-768-ipd, ML-KEM-1024-ipd: https://github.com/IETF-Hackathon/pqc-certificates/blob/master/docs/oid_mapping.md

TODOs:
- [x] Add NIST supplied test vectors and code to test them
- [x] Compact NIST test vectors: the files are quite large and we need only a small part of the value
- [x] Pull ML-DSA-ipd from pq-crystals "standard" branch
- [x] Pull ML-KEM-ipd from pq-crystals "standard" branch
- [x] add patches for copy_from_upstream (for renaming to ML-KEM-ipd / ML-DSA-ipd)
- [ ] ~~Pull pqclean aarch64 code (may omit for now)~~
- [x] Integrate tests with CI
- [x] Update docs
- [x] Ensure updating to latest upstream versions (e.g. with Kyberslash fixes)
- [x] Run constant-time tests on new algorithms & update valgrind files

Closes #1521.


.CMake/alg_support.cmake
```diff
@@ -396,7 +442,7 @@ if(NOT ((OQS_MINIMAL_BUILD STREQUAL "") OR (OQS_MINIMAL_BUILD STREQUAL "OFF")))
 	filter_algs("${OQS_MINIMAL_BUILD}")
 elseif (${OQS_ALGS_ENABLED} STREQUAL "STD")
 ##### OQS_COPY_FROM_UPSTREAM_FRAGMENT_LIST_STANDARDIZED_ALGS_START
-	filter_algs("KEM_kyber_512;KEM_kyber_768;KEM_kyber_1024;SIG_dilithium_2;SIG_dilithium_3;SIG_dilithium_5;SIG_falcon_512;SIG_falcon_1024;SIG_sphincs_sha2_128f_simple;SIG_sphincs_sha2_128s_simple;SIG_sphincs_sha2_192f_simple;SIG_sphincs_sha2_192s_simple;SIG_sphincs_sha2_256f_simple;SIG_sphincs_sha2_256s_simple;SIG_sphincs_shake_128f_simple;SIG_sphincs_shake_128s_simple;SIG_sphincs_shake_192f_simple;SIG_sphincs_shake_192s_simple;SIG_sphincs_shake_256f_simple;SIG_sphincs_shake_256s_simple")
+	filter_algs("KEM_kyber_512;KEM_kyber_768;KEM_kyber_1024;SIG_dilithium_2;SIG_dilithium_3;SIG_dilithium_5;SIG_ml_dsa_44;SIG_ml_dsa_65;SIG_ml_dsa_87;SIG_falcon_512;SIG_falcon_1024;SIG_sphincs_sha2_128f_simple;SIG_sphincs_sha2_128s_simple;SIG_sphincs_sha2_192f_simple;SIG_sphincs_sha2_192s_simple;SIG_sphincs_sha2_256f_simple;SIG_sphincs_sha2_256s_simple;SIG_sphincs_shake_128f_simple;SIG_sphincs_shake_128s_simple;SIG_sphincs_shake_192f_simple;SIG_sphincs_shake_192s_simple;SIG_sphincs_shake_256f_simple;SIG_sphincs_shake_256s_simple")
```
baentsch, line 445: Omission of KEM_ml_kem_512 et.al. intentional?

src/kem/kem.h
```diff
@@ -70,6 +70,12 @@ extern "C" {
 #define OQS_KEM_alg_kyber_768 "Kyber768"
 /** Algorithm identifier for Kyber1024 KEM. */
 #define OQS_KEM_alg_kyber_1024 "Kyber1024"
+/** Algorithm identifier for ML-KEM-512-ipd KEM. */
+#define OQS_KEM_alg_ml_kem_512 "ML-KEM-512-ipd"
```
baentsch, line 74: Is it necessary to put the algorithm version into its name? We never did this until now. Or am I misunderstanding the rationale for the suffix "-ipd" as denoting the FIPS (draft level)? Why not leave the suffix away and update the implementation as and when the draft matures/changes (if necessary)? [NIST FIPS 203](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.ipd.pdf) also just uses the name(s) "ML-KEM-512" etc.

baentsch (commented):
Thanks for adding these algorithms: I very much like how you use our "import machinery" to transform the "old" upstream names into the standard ones. Conceptually this all looks OK -- I'm just not happy with adding version information to the algorithm names: Those may be OK for a hackathon that is not meant to be used for more than that event, but algorithm names with version information included IMO adds an undue burden to downstream integrations: Anyone will have to update their source code as and when the "-ipd" suffix goes away (as opposed to just re-compile or only re-link iff a code change occurs during the transition to final standard).

src/kem/kem.h
dstebila: We're not certain that there won't be algorithmic changes between the current FIPS draft and the final. While we have in the past just absorbed incompatible algorithm changes within the same function name, the idea here is to use distinct names -- "Kyber" for the current NIST Round 3 version, "IPD" for the current FIPS draft, and eventually just "ML-KEM" for the final FIPS draft.  If it turns out that IPD=ML-KEM, then we can easily solve that with some macros, but if that doesn't come to pass, we're better prepared for it.

bhess:
Thanks for the review and comment @baentsch. Good point regarding naming, including the versioning in the names would indeed be a change from the current approach. I see the point of ease of re-compiling and re-deploying. However, I think there are arguments for including the versioning:
For a downstream users it seems not directly obvious if a compatibility-breaking update happened when the same name is kept (besides looking at the release notes or documentation). They in addition have to make sure that compatibility is kept, e.g. by updating OIDs for certificates. I'd argue that a name-update can help downstream to resolve ambiguities and avoid compatibility-breaking changes.
For the NIST "competition" algorithms, keeping the same name seems to be not much an issue as the experimenting/prototyping users will just update everything to the latest versions.
With the (almost) standard algorithms and practical deployments like the Kyber-r3-hybrid in Chrome it can be useful for interoperability to identify the exact version, and to be able to support different versions at the same time (r3, ipd, final ML-KEM).
A versioning scheme for algorithms is also suggested by the IETF-PQUIP [naming convention](https://github.com/ietf-wg-pquip/state-of-protocols-and-pqc?tab=readme-ov-file#algorithm-names).

src/kem/kem.h
baentsch: > If it turns out that IPD=ML-KEM, then we can easily solve that with some macros, 

True from the perspective of `liboqs` -- not so from the perspective of users of the library (and they are in significant majority): All code, documentation, write-ups and configs using "ML_KEM...-ipd" written between now and final standardization will have to be adapted as soon as the "-ipd" moniker gets dropped.

> but if that doesn't come to pass, we're better prepared for it.

In which way? If that happens, we have to change the `liboqs` code anyway. The work effort for all downstreams would be significantly eased if we use the final name already now, though: They (incl. "us" as in oqsprovider, etc.) at most need recompilation but no re-writing.

I always thought `liboqs` aims to aid in the adoption of PQC and not force all users to follow the nitty-gritty details of the standardization process. Wouldn't it be nice if people could rely on `liboqs` (as any changes during further standardization steps are kept under the hood)? I personally consider(ed) that one big advantage of using this library: All tests, use cases, documentation, papers could be retained throughout the NIST competition (for those algorithms that didn't change names or got dropped). Why drop this advantage now that final names are agreed-upon?

baentsch:
> For a downstream users it seems not directly obvious if a compatibility-breaking update happened when the same name is kept (besides looking at the release notes or documentation). They in addition have to make sure that compatibility is kept, e.g. by updating OIDs for certificates. I'd argue that a name-update can help downstream to resolve ambiguities and avoid compatibility-breaking changes.

The above has different meanings of the word "user" (of `liboqs`), so let me be precise:
Let me call "End users" those users not knowing about software development but just interested in deploying PQC -- at most writing `nginx` config files, but typically just reading documentation, writing papers or toggling algorithm name switches. Big majority by far.
Let me call "PQ integrators" those users writing code to make use of `liboqs`, e.g., entities changing code and assigning OIDs. No so numerous.

If PQ integrators need name changes to ascertain they also update their OIDs and link the right libraries, they may need to reconsider their job choice. 

The argumentation of the IETF

> However, the community wants to avoid the situation where we over-eagerly adopt the new names, then see a compatibility-breaking change in a future iteration of the FIPS draft standards, and then have multiple non-compatible algorithms with the same name.

indeed seems to indicate a lack of trust in the capabilities of software developers and interop testers to weed out possibly incompatible implementations using the same name. 

Does this mean that if there is a step after "IPD", let's call it "SPD" that everyone would need to update their software again? Good business for IT consultancies paid by the hour, bad choice for the End Users footing the bill.

> it can be useful for interoperability to identify the exact version, and to be able to support different versions at the same time (r3, ipd, final ML-KE

This is a good argument. But it is relevant only for comparative testing, i.e., development and test purposes. Should End Users be impacted by this (currently still interesting, but in the long run irrelevant) use case?

baentsch:
Re-reading this discussion, what about the idea to introduce a macro "-ipd" _now_? That way we can satisfy the IETF folks placing no trust in software developers _and_ allow people knowing what they do to minimize their long-term work effort (by taking the "risk" of using the non-ipd variant)?

src/kem/kem.h
dstebila: > > If it turns out that IPD=ML-KEM, then we can easily solve that with some macros,
> 
> True from the perspective of `liboqs` -- not so from the perspective of users of the library (and they are in significant majority): All code, documentation, write-ups and configs using "ML_KEM...-ipd" written between now and final standardization will have to be adapted as soon as the "-ipd" moniker gets dropped.

The -ipd moniker doesn't need to be dropped, it can be kept as an alias.

dstebila:
@bhess Stepping back from the discussion about -IPD naming in that one specific file to a more general question: what was your philosophy around when to use the ML-KEM-IPD naming and when to use plain ML-KEM?  

If ML-KEM-IPD != ML-KEM, then there may be a time period where we want to have both ML-KEM-IPD and ML-KEM present in the library, similar to how we're currently imagining wanting to have both Kyber and ML-KEM-IPD present in the library.  I'm not sure that would be possible with the current naming in the PR...

bhess:
> what about the idea to introduce a macro "-ipd" _now_?

I like the idea of a macro/alias. What about the other way around, i.e. "ML-KEM-512" is an alias of "ML-KEM-512-ipd"? Should there be something like an "-spd" version, the ML-KEM-512 alias can just become an alias of the -spd version. The same would then be possible for the upcoming on-ramp signatures ("OnrampSig"-R1, "OnrampSig"-R2, ..., and an alias "OnrampSig").



> what was your philosophy around when to use the ML-KEM-IPD naming and when to use plain ML-KEM?

The idea was to have a family name "ML-KEM" the individual schemes are called "ML-KEM-512-ipd", "ML-KEM-768-ipd" and "ML-KEM-1024-ipd".

> I'm not sure that would be possible with the current naming in the PR...

I think you are right, the import logic builds the internal API names using "name"_"scheme":

https://github.com/open-quantum-safe/liboqs/blob/3468dc8b7d7f1cbddb9f036560ee020450ba5069/scripts/copy_from_upstream/copy_from_upstream.yml#L140-L148

If "scheme" is named "512_ipd" instead then it should work.

I'll update the PR next week.

baentsch:
> I like the idea of a macro/alias. What about the other way around, i.e. "ML-KEM-512" is an alias of "ML-KEM-512-ipd"?

Thanks & fine with me. As long as "ML-KEM-512" (also) becomes available now. Analog for the other "ML..." (sig) algs & strengths.

README.md
```diff
@@ -46,6 +46,7 @@ The list below indicates all algorithms supported by liboqs, but not all those a
 - **FrodoKEM**: FrodoKEM-640-AES, FrodoKEM-640-SHAKE, FrodoKEM-976-AES, FrodoKEM-976-SHAKE, FrodoKEM-1344-AES, FrodoKEM-1344-SHAKE
 - **HQC**: HQC-128, HQC-192, HQC-256
 - **Kyber**: Kyber512, Kyber768, Kyber1024
+- **ML-KEM**: ML-KEM-512-ipd, ML-KEM-768-ipd, ML-KEM-1024-ipd
```
baentsch, line 49: Wouldn't it be sensible to document the aliases here, too?

docs/algorithms/kem/ml_kem.yml
```diff
@@ -0,0 +1,125 @@
...
+- Léo Ducas
+- Eike Kiltz
+- Tancrède Lepoint
+- Vadim Lyubashevsky
+- John M. Schanck
+- Gregor Seiler
+- Damien Stehlé
+crypto-assumption: Module LWE+R with base ring Z[x]/(3329, x^256+1)
+website: https://pq-crystals.org/kyber/ and https://csrc.nist.gov/pubs/fips/203/ipd
+nist-round: ipd
```
baentsch, line 17: This (string) designation causes downstream problems where "nist-round" is supposed to be an integer. Would it make sense to assign "ipd" a special new int value (sufficiently high to be above all possible future NIST rounds but also not so high such as to permit a designation for "second public draft", etc. until finalization)? Alternatively, I'll change all logic keeping algorithms in the NIST rounds together based on integer and move it to strings.

baentsch: > I'd rather move it to strings if possible

It is/done: https://github.com/open-quantum-safe/oqs-provider/pull/344

baentsch:
As a follow-up to https://github.com/orgs/open-quantum-safe/discussions/1689 this is to ask whether this PR should also change the designation of what is a ["STD" algorithm](https://github.com/open-quantum-safe/liboqs/blob/main/CONFIGURE.md#OQS_ALGS_ENABLED), at the least dropping Kyber and Dilithium.

baentsch:
Thanks for the updates to the STD algs list, @bhess. Only item left: Could you please trigger downstream testing and thus the [ML-* PR on oqs-provider](https://github.com/open-quantum-safe/oqs-provider/pull/348) as discussed (by way of adding the statement [ "[trigger downstream]"](https://github.com/open-quantum-safe/liboqs/blob/f468d212f2fde89d2442c4856c274eec694b0c14/.github/workflows/release-test.yml#L4) to a commit message)?

bhess:
> As a follow-up to https://github.com/orgs/open-quantum-safe/discussions/1689 this is to ask whether this PR should also change the designation of what is a ["STD" algorithm](https://github.com/open-quantum-safe/liboqs/blob/main/CONFIGURE.md#OQS_ALGS_ENABLED), at the least dropping Kyber and Dilithium.

Agreed, the STD algorithms should contain the standards eventually, and until the standards are finalized only the latest revisions.

> Thanks for the updates to the STD algs list, @bhess. Only item left: Could you please trigger downstream testing and thus the [ML-* PR on oqs-provider](https://github.com/open-quantum-safe/oqs-provider/pull/348) as discussed (by way of adding the statement [ "[trigger downstream]"](https://github.com/open-quantum-safe/liboqs/blob/f468d212f2fde89d2442c4856c274eec694b0c14/.github/workflows/release-test.yml#L4) to a commit message)?

Thanks, added the [trigger downstream] message now (wanted to have a passing build first before triggering downstream tests).

baentsch:
@SWilson4 I did see the trigger event going out to `oqsprovider` -- but I didn't find any indication of this having resulted in a run anywhere... Would you know where to look?

SWilson4:
> @SWilson4 I did see the trigger event going out to `oqsprovider` -- but I didn't find any indication of this having resulted in a run anywhere... Would you know where to look?

Looks like this was the triggered run: https://github.com/open-quantum-safe/oqs-provider/actions/runs/7826280145

You can find them in the "Actions" tab under "Release tests": https://github.com/open-quantum-safe/oqs-provider/actions/workflows/release.yml. It's admittedly not very clear which run corresponds to which upstream trigger; I had to go look at the logs to see which refs were checked out. Not sure how to improve that.

baentsch:
> Looks like this was the triggered run: https://github.com/open-quantum-safe/oqs-provider/actions/runs/7826280145

Thanks for the pointer. Looks like all `liboqs` & `oqsprovider` code ran and tested OK. What failed was the setup of `openssl s_server` instances... Oh dear, how I love CI debugging....

baentsch:
@SWilson4 @bhess : Downstream test [now passes](https://github.com/open-quantum-safe/oqs-provider/actions/runs/7941429013/job/21683771928) . Problem is in `openssl`: More than 45 (default) groups cause a system crash during handshake. If one of you knows `pytest`, please consider contributing a better solution than [my workaround hack for the test](https://github.com/open-quantum-safe/oqs-provider/pull/348/commits/a4bfffb820da9ad70c6c9e0766be58638c9b6243). I'll try to create a reproducer for `openssl`...

I'd thus suggest resolving the merge conflict here, @bhess, and to approve and merge this PR as well as the now successfully tested https://github.com/open-quantum-safe/oqs-provider/pull/348, fya @SWilson4 @dstebila .

dstebila (approved):

PR #1631, opened 2023-12-15
Pull latest Kyber version from upstream

bhess:
Pull latest Kyber version from upstream, which fixes a potential constant-time issue in the "ref" version as reported here:
https://groups.google.com/a/list.nist.gov/d/msgid/pqc-forum/ZXxUpck1501jEcRq%40localhost

### DIFF
--- a/docs/algorithms/kem/kyber.md
+++ b/docs/algorithms/kem/kyber.md
@@ -7,9 +7,9 @@
 - **Authors' website**: https://pq-crystals.org/
 - **Specification version**: NIST Round 3 submission.
 - **Primary Source**<a name="primary-source"></a>:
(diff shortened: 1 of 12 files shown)


baentsch (approved):

baentsch:
> fixes a potential constant-time issue in the "ref" version

I assume it's intentional that a fix to "ref" code also changes "avx2" files?!

bhess:
> I assume it's intentional that a fix to "ref" code also changes "avx2" files?!

Good point. The update includes two separate commits upstream. The avx2 change is from the earlier commit: https://github.com/pq-crystals/kyber/commit/9c32260cf9d4278864c741a82366f08440dd2f21.

PR #1632, opened 2023-12-18
Fix BIKE constant-time errors

### LINKED ISSUE #1624: BIKE constant time tests failing
> Bad news: CT testing now has more failures than before -- but then again, we're testing more configs now...

Especially given that it's no longer just Falcon failing but also BIKE. :( I haven't looked closely at the failures yet to see whether they constitute true non-CT behaviour or not. Will open an issue to track.

_Originally posted by @SWilson4 in https://github.com/open-quantum-safe/liboqs/issues/1617#issuecomment-1850243157_

SWilson4:
I have reviewed the [BIKE constant-time test failures](https://github.com/open-quantum-safe/liboqs/actions/runs/7154812712/job/19482761520) and identified the offending lines of code. They are summarized below.

https://github.com/open-quantum-safe/liboqs/blob/8449e546491e77b52ca5b2f1451086ef564e16aa/src/kem/bike/additional_r4/sampling_portable.c#L44

This appears to be triggered by the array access `pos_qw[j]`. After reviewing the function's code, I believe this to be a false positive and have documented it as a "pass". There's an existing pass for the AVX2 version of this function.

https://github.com/open-quantum-safe/liboqs/blob/8449e546491e77b52ca5b2f1451086ef564e16aa/src/kem/bike/additional_r4/decode_portable.c#L40

The issue here seems to be the "<" operator. I changed this to `!(bits >> 6)` and it now passes (on my machine).

https://github.com/open-quantum-safe/liboqs/blob/8449e546491e77b52ca5b2f1451086ef564e16aa/src/kem/bike/additional_r4/gf2x_mul_base_portable.c#L44

https://github.com/open-quantum-safe/liboqs/blob/8449e546491e77b52ca5b2f1451086ef564e16aa/src/kem/bike/additional_r4/gf2x_mul_base_portable.c#L45

https://github.com/open-quantum-safe/liboqs/blob/8449e546491e77b52ca5b2f1451086ef564e16aa/src/kem/bike/additional_r4/gf2x_mul_base_portable.c#L50

https://github.com/open-quantum-safe/liboqs/blob/8449e546491e77b52ca5b2f1451086ef564e16aa/src/kem/bike/additional_r4/gf2x_mul_base_portable.c#L51

These array accesses depend on the value of `a0`, which is one of the function arguments. Hence, I believe this to be (possible) secret-dependent behaviour. I added short loops to perform the array access in a secret-independent fashion, as is done [in HQC](https://github.com/open-quantum-safe/liboqs/blob/8449e546491e77b52ca5b2f1451086ef564e16aa/src/kem/hqc/pqclean_hqc-128_clean/gf2x.c#L49).

After these changes, the constant-time tests pass for all variants of BIKE on my machine.

Fixes #1624.


SWilson4:
Following up on discussion in today's OQS meeting: I am running into a snag when attempting to resolve the constant-time error raised for this line of code: https://github.com/open-quantum-safe/liboqs/blob/8449e546491e77b52ca5b2f1451086ef564e16aa/src/kem/bike/additional_r4/decode_portable.c#L40

My initial thought was that the error was raised because of the use of "<" operator for comparison. My "quick fix" of using a bitshift to check the `bits < 64` condition satisfied the Valgrind-powered constant-time tests but caused the Clang static analyzer to fail (see [here](https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/3078/workflows/0c99ebaf-628d-4c11-9fd3-da4de27f53fc/jobs/24608?invite=true#step-103-1624_132)). Presumably, the static analyzer doesn't recognize that `!(bits >> 6)` and `bits < 64` are equivalent and is reporting an error because the assert statement may fail. I'm not sure how to resolve this in a way that will satisfy both the static analyzer and the constant-time tests. I'm also not certain that the "<" operator was causing the test failure in the first place.

If somebody more familiar with the inner workings of BIKE could take a look and offer input on
1. whether or not this constitutes secret-dependent behaviour
2. if yes, how to resolve it in a way that keeps the static analyzer happy.

Feedback on the other constant-time errors (and my proposed fixes) would also be very welcome!

SWilson4:
Tagging @crockeea who kindly offered to pass this along to somebody who knows about BIKE.

src/kem/bike/additional_r4/gf2x_mul_base_portable.c
```diff
@@ -41,14 +41,29 @@ void gf2x_mul_base_port(OUT uint64_t *c,
   u[7] = u[6] ^ b0m;
 
   // Step 2: Multiply two elements in parallel in positions i, i+s
-  l = u[LSB3(a0)] ^ (u[LSB3(a0 >> 3)] << 3);
-  h = (u[LSB3(a0 >> 3)] >> 61);
+  for (size_t i = 0; i < 8; ++i) {
+      // use a mask for secret-independent memory access
+      mask = LSB3(a0) - i;
+      l ^= (u[i] & (uint64_t)(0 - (1 - ((uint64_t)(mask | (0 - mask)) >> 63))));
```
dkostic, line 47: maybe we can extract the mask computation to a function in `utilities.h`, something like:
```
// Returns 0xfff..ff if a == b, otherwise returns 0.
uint64_t secure_cmpeq64_mask(uint64_t a, uint64_t b) {
  uint64_t tmp = a - b;
  return (uint64_t)(0 - (1 - ((uint64_t)(tmp | (0 - tmp)) >> 63))));
}
```
and then use it here and in other places below like this:
```
l ^= u[i] & secure_cmpeq64_mask(LSB3(a0), i);
```

dkostic:
> 1. whether or not this constitutes secret-dependent behaviour

It would constitute secret-dependent behaviour if the condition could ever fail. however, we are calling the function here: https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/bike/additional_r4/decode_portable.c#L62 where we explicitly reduce the input mod 64.

> 2. if yes, how to resolve it in a way that keeps the static analyzer happy.

we should just remove the assert, it's pointless because we reduce the argument mod 64 before calling the function.

src/kem/bike/additional_r4/gf2x_mul_base_portable.c
SWilson4: Done!

SWilson4:
Thanks very much for the quick review @dkostic. I've made your suggested changes and confirmed that the constant-time tests now pass for BIKE on my machine (the CI run is a weekly scheduled job targeting `main`).

Would it be helpful if I submit a corresponding PR to the awslabs/bike-kem repo?

dkostic (approved):

dkostic:
> Thanks very much for the quick review @dkostic. I've made your suggested changes and confirmed that the constant-time tests now pass for BIKE on my machine (the CI run is a weekly scheduled job targeting `main`).
> 
> Would it be helpful if I submit a corresponding PR to the awslabs/bike-kem repo?

That'd be awesome and helpful!

SWilson4:
This has now been fixed upstream as well: https://github.com/awslabs/bike-kem/commit/96aeca01fbb4b620a58bce53ce54a29d6c11cfc5

baentsch (approved):
Thanks for bringing up- and downstream in sync (in a somewhat unusual order, but the result counts :).

PR #1633, opened 2023-12-19
pull kyber from upstream: dda29cc63af721981ee2c831cf00822e69be3220 (#1631)

dstebila:
Applies fix from #1631 to dev branch for 0.9.1 release.

* [No] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: c
 before_script:
-  - sudo apt -y install astyle cmake gcc ninja-build libssl-dev python3-pytest python3-pytest-xdist unzip xsltproc doxygen graphviz valgrind
+  - sudo apt update && sudo apt -y install astyle cmake gcc ninja-build libssl-dev python3-pytest python3-pytest-xdist unzip xsltproc doxygen graphviz valgrind
 jobs:
   include:
     - arch: ppc64le         # The IBM Power LXD container based build for OSS only
--- a/docs/algorithms/kem/kyber.md
+++ b/docs/algorithms/kem/kyber.md
@@ -7,9 +7,9 @@
 - **Authors' website**: https://pq-crystals.org/
 - **Specification version**: NIST Round 3 submission.
 - **Primary Source**<a name="primary-source"></a>:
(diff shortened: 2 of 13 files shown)


dstebila:
> It looks like the Travis build is failing due since #1629 is not merged into either branch. We know that Travis passes on the current `main` branch, which includes this patch, so I don't think it's necessary for those tests to be green on this branch.

Thanks Spencer. I've cherry-picked that commit over to this branch as well.

baentsch (approved):
LGTM. But why doesn't this target "main"?

dstebila:
> LGTM. But why doesn't this target "main"?

These have already been merged to main.  The point here is to do a quick 0.9.1 security release that takes the 0.9.0 release and just ports back the Kyber security fix (plus one extra commit that makes things build cleanly on Travis).

bhess (commented):
Note that the same issue still exists in the aarch64 code that we pull from PQClean:
https://github.com/open-quantum-safe/liboqs/blob/8449e546491e77b52ca5b2f1451086ef564e16aa/src/kem/kyber/oldpqclean_kyber1024_aarch64/poly.c#L217C1-L217C1

If you agree, I'll create two PRs with a patch: one targeting main, and one targeting the 0.9.1 branch.

dstebila:
> Note that the same issue still exists in the aarch64 code that we pull from PQClean: https://github.com/open-quantum-safe/liboqs/blob/8449e546491e77b52ca5b2f1451086ef564e16aa/src/kem/kyber/oldpqclean_kyber1024_aarch64/poly.c#L217C1-L217C1
> 
> If you agree, I'll create two PRs with a patch: one targeting main, and one targeting the 0.9.1 branch.

That would be great, thanks Basil.

bhess:
#1637 contains a patch for the aarch64 issue targeting the ds-091-kyber branch (#1636 is the same patch but targeting main, cherry-picking it wouldn't be straight-forward because of the change in #1595)

PR #1576, opened 2023-10-11
Use CMAKE_USE_PTHREADS_INIT

zxjtan:
Precursor to #1549. Replace `OQS_USE_PTHREADS_IN_TESTS` with `CMAKE_USE_PTHREADS_INIT` to clean up the Threads inclusion and linking story.

* [x] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)? NO
* [x] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? NO

PR #1653, opened 2024-01-05
find_package(Threads) regardless of BUILD_ONLY_LIB

zxjtan:
Retroactive follow-up to #1576: I realised that #1549 introduces usage of pthreads into the main library code. I also infer that the reason `find_package(Threads)` was originally gated behind `OQS BUILD_ONLY_LIB` was because only test code used pthreads. Hence, CMake should now always look for the Threads package (on non-MSVC compilers).

Just doing the above breaks Zephyr builds. Hence, this PR also adds a new CMake variable `OQS_USE_PTHREADS` that will be `true` `if (CMAKE_USE_PTHREADS_INIT AND NOT OQS_EMBEDDED_BUILD)`. It also changes code gated on `#CMAKE_USE_PTHREADS_INIT` macro to gate on `#OQS_USE_PTHREADS` instead.


SWilson4:
@zxjtan Sorry it's taken so long for us to provide feedback on this PR! It looks reasonable to me (pending a rebase on the latest `main`), but I don't have enough knowledge about `pthreads` (and threading in general) to feel comfortable approving it. @baentsch @dstebila you reviewed #1549 and #1576; would you also be able to chime in here?

baentsch (approved):
Apologies for the delay in reviewing this. I now understand the logic and tested successfully against oqs-provider. Thanks for the improvement, @zxjtan !

dstebila (approved):

PR #1654, opened 2024-01-08
Run oqs-provider release tests in CI on release candidate branches

### LINKED ISSUE #1647: (Automated) downstream release testing
This is to reduce the risk that `liboqs` RCs don't work OK with designated important downstream projects.

This issue is to discuss the approach to do this (from manual, via release documentation) to fully automated via CI.

Also, we should decide which subprojects should be subject to this treatment, i.e., which ones we _really_ support. Proposal for now is `oqsprovider` (only). Other opinions welcome.

SWilson4:
EDIT 2023-02-01: This now simply adds a trigger for functionality proposed in https://github.com/open-quantum-safe/oqs-provider/pull/345. It will function when somebody with the appropriate permissions (presumably @dstebila?) adds an `oqsbot` (classic) personal access token with `repo` and `workflow` permissions as an Actions secret with the variable name `OQSBOT_GITHUB_ACTIONS`. We already have one of these in use in CircleCI.

### Previous description (prior to 2023-02-01)

Currently, our [release process](https://github.com/open-quantum-safe/liboqs/wiki/Release-process) requires that the `oqs-provider` release tests be run manually for `liboqs` release candidate branches. This PR automates this process in CI, setting them to run automatically on pull requests for any branch whose name has the prefix `dev-`. This prefix is up for discussion; I used `dev-` because it's what we used for the latest release, and because using `rc-` would cause issues if we ever get a PR from somebody named, say, Ray Charles.

The full release tests complete in about 50 minutes in CI—see [https://github.com/open-quantum-safe/liboqs/actions/runs/7426702581](https://github.com/open-quantum-safe/liboqs/actions/runs/7426702581)—which I think is not too bad. Note that although the linked workflow says it uses a "minimal set" of algorithm combinations, it actually does run all combinations; I just didn't update the workflow documentation before pushing. The full set of tests should run again on this PR as a proof of concept; I'll revert to running them only on release candidate PRs once that workflow completes.

The CI workflow is based on the [`release-test.sh` script in the `oqs-provider` repo](https://github.com/open-quantum-safe/oqs-provider/blob/eb80076d1683121db35fabf8c1b0c6f9125c1b6d/scripts/release-test.sh). I didn't call the script directly since some aspects (checking out a new temporary branch then resetting back to main, cloning liboqs, etc) were either unnecessary in the CI setting or better handled by a built-in action.

Discussion points:
- What prefix do we want to use for release candidates?
- Is there a way to ensure that this script stays in sync with `oqs-provider` release testing requirements?
- For @baentsch specifically: what version of OpenSSL should we be checking against here? I defaulted to `master` as the [release test documentation](https://github.com/open-quantum-safe/oqs-provider/blob/eb80076d1683121db35fabf8c1b0c6f9125c1b6d/scripts/release-test.sh#L33) said to use the "latest" version.
- Do we want to also automate some of the other release testing? We could, for example, also run the full constant time tests as part of this workflow.

Fixes #1647 (assuming that `oqs-provider` is the only downstream for which we want automated release testing).

### DIFF
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -2,7 +2,7 @@ name: Release tests
 
 on:
   pull_request:
-    branches: [ 'sw-provider-**' ] # TODO: settle on standard prefix
+    branches: [ 'dev-**' ] # TODO: settle on standard prefix
 
 
 jobs:
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -23,10 +23,15 @@ jobs:
         with:
           path: liboqs
       - name: Activate all algorithms
-        run: sed -i "s/enable\: false/enable\: true/g" oqs-template/generate.yml && python3 oqs-template/generate.py
+        run: |
+          sed -i "s/enable\: false/enable\: true/g" oqs-template/generate.yml && \
+          python3 oqs-template/generate.py
       - name: Full build
         run: scripts/fullbuild.sh
       - name: Run oqs-provider tests
         run: scripts/runtests.sh
       - name: Run full TLS tests
-        run: OPENSSL_MODULES=`pwd`/_build/lib OPENSSL_CONF=`pwd`/scripts/openssl-ca.cnf python3 -m pytest --numprocesses=auto scripts/test_tls_full.py
+        env:
+          OPENSSL_MODULES: _build/lib
+          OPENSSL_CONF: scripts/openssl-ca.cnf
+        run: python3 -m pytest --numprocesses=auto scripts/test_tls_full.py
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -3,8 +3,8 @@ name: Release tests
 on: pull_request
 
 jobs:
-  if:  startsWith(github.head_ref, 'sw-provider-') # TODO: settle on standard prefix
   oqs-provider-release-test: # TODO: this is highly coupled to oqs-provider structure. Might be better downstream?
+    if:  startsWith(github.head_ref, 'sw-provider-') # TODO: settle on standard prefix
     container:
       image: openquantumsafe/ci-ubuntu-jammy:latest
     runs-on: ubuntu-latest
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -2,7 +2,7 @@ name: Release tests
 
 on:
   pull_request:
-    branches: [ 'dev-**' ] # TODO: settle on standard prefix
+    branches: [ 'sw-provider-**' ] # TODO: settle on standard prefix
 
 
 jobs:
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -1,12 +1,9 @@
 name: Release tests
 
-on:
-  pull_request:
-    branches: [ 'dev-**' ] # TODO: settle on standard prefix
-
+on: pull_request
 
 jobs:
-
+  if:  startsWith(github.head_ref, 'dev-') # TODO: settle on standard prefix
   oqs-provider-release-test: # TODO: this is highly coupled to oqs-provider structure. Might be better downstream?
     container:
       image: openquantumsafe/ci-ubuntu-jammy:latest
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -1,28 +1,31 @@
 name: Release tests
 
+# Trigger oqs-provider release tests.
+# Runs whenever a release is published, or when a commit message ends with "[trigger downstream]"
+# When triggered by a release, the liboqs release tag and the provider "<release tag>-tracker" branch are used.
+# When triggered by a commit message, the triggering liboqs branch and the provider "<liboqs branch>-tracker" branch are used.
+# If the tracker branch does not exist, the downstream pipeline should detect it and run on the main branch instead.
 on:
+  push:
   release:
     types: [published]
 
 jobs:
-  # trigger oqs-provider release tests on a tracker branch
   oqs-provider-release-test:
+    if: github.event_name == 'release' || endsWith( github.event.head_commit.message, '[trigger downstream]' )
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout release tests script
+        uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            scripts/provider-release-tests.sh
+          sparse-checkout-cone-mode: false
       - name: Trigger oqs-provider release tests
         run: |
-          curl --silent \
-               --write-out "\n%{response_code}\n" \
-               --request POST \
-               --header "Accept: application/vnd.github+json" \
-               --header "Authorization: Bearer ${{ secrets.OQSBOT_GITHUB_ACTIONS }}" \
-               --header "X-GitHub-Api-Version: 2022-11-28" \
-               --data '{
-                         "event_type": "liboqs-release",
-                         "client_payload": {
-                                             "liboqs_ref": "${{ github.event.release.tag_name }}",
-                                             "provider_ref": "${{ github.event.release.tag_name }}-tracker"
-                         }
-                       }' \
-               https://api.github.com/repos/open-quantum-safe/oqs-provider/dispatches | tee curl_out \
-            && grep -q "204" curl_out
+          CURL_FLAGS="--silent --write-out \n%{response_code}\n" \
+          ACCESS_TOKEN="${{ secrets.OQSBOT_GITHUB_ACTIONS }}" \
+          LIBOQS_REF="${{ github.ref_name }}" \
+          PROVIDER_REF="${{ github.ref_name }}-tracker" \
+          ./scripts/provider-release-tests.sh | tee curl_out \
+          && grep -q "204" curl_out
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -2,9 +2,9 @@ name: Release tests
 
 on: # TODO: do we want to run this on pull request only? push only? need to have standardized release process
   push:
-    branches: [ 'dev-**' ] # TODO: settle on prefix / suffix for release candidates
+    branches: [ 'sw-provider-**' ] # TODO: settle on prefix / suffix for release candidates
   pull_request:
-    branches: [ 'dev-**' ]
+    branches: [ 'sw-provider-**' ]
 
 
 jobs:
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -2,7 +2,7 @@ name: Release tests
 
 on:
   pull_request:
-    branches: [ 'dev-**' ]
+    branches: [ 'dev-**' ] # TODO: settle on standard prefix
 
 
 jobs:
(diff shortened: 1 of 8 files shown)


baentsch (changes requested):
I like the concept -- but this PR doesn't quite cut it yet: Please look at the [test run log](https://github.com/open-quantum-safe/liboqs/actions/runs/7449135695/job/20265074694): You see that this tests oqsprovider "main"/0.5.3-dev against `liboqs` "main"/0.10.0-dev. For a "main"/general PR test this may be OK (but do we _really_ want to run this on all PRs? Probably not, as then `oqsprovider` must be able to handle all `liboqs` changes when they appear -- unlikely to _always_ work, creating a catch-22...); it certainly isn't OK for a release (only) test. That said, by setting "LIBOQS_BRANCH" one can configure `oqsprovider` to use a downlevel `liboqs` algorithm configuration (which is necessary for the "0.9.2" release: This must force `oqsprovider` to use its "liboqs 0.9.0" algorithm configuration and not the latest (geared for 0.10.0-dev) config... Slowly wondering whether the goal is really attainable... For the time being, let's use an `oqsprovider` RC branch able to work with "0.9.x" `liboqs` RCs...

baentsch:
@SWilson4 FYI/FWIW, `oqsprovider` now has a branch called "092-tracker" that passes CI (and builds executables) tracking `liboqs` branch "dev-092". Maybe that's an acceptable approach for automation: On a "release" event, `liboqs` CI tries to trigger a downstream "release.sh" script in a branch labelled "<liboqs-release-id>-tracker" if present, otherwise "main" (as per this PR). If a "...-tracker" branch is not present but CI fails on "main", that'd be a clear indication that a downstream release is necessary/needs to be prepared for the `liboqs` release to proceed. If downstream "main" (as per your proposal here) CI passes, the `liboqs` RC is definitely good. If a "...-tracker" path exists and CI passes, that'd be a clear indication that that can be released immediately, too (possibly adapting referenced `liboqs` branch name).

SWilson4:
> @SWilson4 FYI/FWIW, `oqsprovider` now has a branch called "092-tracker" that passes CI (and builds executables) tracking `liboqs` branch "dev-092". Maybe that's an acceptable approach for automation: On a "release" event, `liboqs` CI tries to trigger a downstream "release.sh" script in a branch labelled "-tracker" if present, otherwise "main" (as per this PR). If a "...-tracker" branch is not present but CI fails on "main", that'd be a clear indication that a downstream release is necessary/needs to be prepared for the `liboqs` release to proceed. If downstream "main" (as per your proposal here) CI passes, the `liboqs` RC is definitely good. If a "...-tracker" path exists and CI passes, that'd be a clear indication that that can be released immediately, too (possibly adapting referenced `liboqs` branch name).

Sure, that approach sounds OK to me—and I like that the `release.sh` script lives downstream and hence will be kept in sync with the provider. That said, it sounds like it will require a little more work and multiple PRs, so I think I'll pause development on this feature until after we get 0.9.2 out the door. No sense in potentially delaying a security patch. How about we aim to have the process automated for the 0.10.0 release?

baentsch:
FWIW, we now have a similar case that could be used for testing proper downstream integration: #1626 changes algs and https://github.com/open-quantum-safe/oqs-provider/pull/344 is there to prove this PR to be useful... Its own PR obviously fails (using `liboqs`-main) but a new CI run triggered in #1626 targeting the branch underlying https://github.com/open-quantum-safe/oqs-provider/pull/344 would prove both code bases right before merging either. Suggestions how to facilitate that? Maybe some form of suitable branch-labelling supporting & triggering this?

SWilson4:
> FWIW, we now have a similar case that could be used for testing proper downstream integration: #1626 changes algs and [open-quantum-safe/oqs-provider#344](https://github.com/open-quantum-safe/oqs-provider/pull/344) is there to prove this PR to be useful... Its own PR obviously fails (using `liboqs`-main) but a new CI run triggered in #1626 targeting the branch underlying [open-quantum-safe/oqs-provider#344](https://github.com/open-quantum-safe/oqs-provider/pull/344) would prove both code bases right before merging either. Suggestions how to facilitate that? Maybe some form of suitable branch-labelling supporting & triggering this?

I actually just got back to working on this yesterday using the setup that you suggested above. I have a branch on the `oqs-provider` side that I'll make a PR for later today: https://github.com/open-quantum-safe/oqs-provider/tree/sw-automate-release-tests.

Once merged, the release tests would trigger automatically on releases (looking for a `-tracker` branch for the release tag) but could also be triggered via `curl` for _any_ `liboqs` GitHub ref. For instance...
```bash
curl --silent \
       --write-out "\n%{response_code}\n" \
       --request POST \
       --header "Accept: application/vnd.github+json" \
       --header "Authorization: Bearer $PERSONAL_ACCESS_TOKEN" \
       --header "X-GitHub-Api-Version: 2022-11-28" \
       --data '{ "event_type": "liboqs-release", "client_payload": { "tag": "bhe-fips-ipd" } }' \
       https://api.github.com/repos/open-quantum-safe/oqs-provider/dispatches
```
would trigger the release tests to run using the `bhe-fips-ipd` branch of `liboqs` and the `bhe-fips-ipd-tracker` branch of `oqs-provider` (if it exists, otherwise `main`).

I'll open a draft PR right now so that you can suggest any changes or extensions you might like.

SWilson4:
I've tested the functionality via swilson4/cicd-playground-upstream and swilson4/cicd-playground-downstream as best I can, but it's a little finicky, and hard to test properly without actually doing a release here.

Follow-up to this PR, perhaps: would we also like to automate the other release testing? For instance, perhaps we could run the entire constant-time test suite in CI in the same workflow.

baentsch:
> would we also like to automate the other release testing? For instance, perhaps we could run the entire constant-time test suite in CI in the same workflow.

Very good suggestion!

.github/workflows/release-test.yml
```diff
@@ -0,0 +1,28 @@
...
+  oqs-provider-release-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger oqs-provider release tests
+        run: |
+          curl --silent \
+               --write-out "\n%{response_code}\n" \
+               --request POST \
+               --header "Accept: application/vnd.github+json" \
+               --header "Authorization: Bearer ${{ secrets.OQSBOT_GITHUB_ACTIONS }}" \
```
baentsch, line 18: [This secret doesn't seem to be set yet](https://github.com/open-quantum-safe/liboqs/settings/secrets/actions).

baentsch (approved):
LGTM. I wonder whether it'd be sensible to enable testing this with another trigger condition based on the name of the branch(?). Additional advantage (beyond ascertaining that this will work on a release) is that anyone clicking Yes on the question "Does this add new algs" could trigger this logic (and downstream tests) by properly naming their branch, something along the lines of 

```
on:
  pull_request:
    branches:    
      - '**-triggerdownstream'
```

SWilson4:
> LGTM. I wonder whether it'd be sensible to enable testing this with another trigger condition based on the name of the branch(?). Additional advantage (beyond ascertaining that this will work on a release) is that anyone clicking Yes on the question "Does this add new algs" could trigger this logic (and downstream tests) by properly naming their branch, something along the lines of
> 
> ```
> on:
>   pull_request:
>     branches:    
>       - '**-triggerdownstream'
> ```

I like this idea, but I think that triggering it on `pull_request` could result in a lot of runs for a fairly expensive test. What about triggering on push and scanning the commit message (similar to `[skip ci]`)?

```
on:
  push:
    jobs:
      downstream-test:
        if: endsWith( ${{ github.event.head_commit.message }}, '[trigger downstream]' )
...
```

SWilson4:
Tacking on to my previous comment, another benefit of adding another trigger condition is that we can (more or less) test this workflow before actually doing a release...

.github/workflows/release-test.yml
SWilson4: Indeed not, I don't have sufficient permissions to add it (or view the link). @dstebila created an `oqsbot` PAT for CircleCI some time ago; we could reuse that if it's saved in cleartext anywhere. Otherwise, somebody with perms will need to generate a new one and add it.

baentsch:
> What about triggering on push and scanning the commit message (similar to [skip ci])?

_Much_ better!

scripts/provider-release-tests.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+# Trigger the oqs-provider release tests in CI.
```
baentsch, line 3: Maybe better call this "provider-test-trigger.sh"? It's not only working for releases... Also add SPDX identifier?

SWilson4: Done and done!

SWilson4:
I've just done some testing of the trigger script locally using my own personal access token, and everything seems to be working OK. See https://github.com/open-quantum-safe/oqs-provider/actions/workflows/release.yml for the results. Note that the failures are expected; they checked out older tags / branches of oqs-provider that didn't have the recently merged test script.

You can also see that the "[trigger downstream]" trigger is working as expected (again, failure is expected as the secret token hasn't yet been added): https://github.com/open-quantum-safe/liboqs/actions/workflows/release-test.yml

.github/workflows/release-test.yml
baentsch: Done. CI passes now. @bhess please review this PR and re-base #1626 after merge: When pushing a commit suitably labelled ("... [trigger downstream]"), CI should turn green on https://github.com/open-quantum-safe/oqs-provider/pull/348

dstebila:
> EDIT 2023-02-01: This now simply adds a trigger for functionality proposed in https://github.com/open-quantum-safe/oqs-provider/pull/345. It will function when somebody with the appropriate permissions (presumably @dstebila?) adds an oqsbot (classic) personal access token with repo and workflow permissions as an Actions secret with the variable name OQSBOT_GITHUB_ACTIONS. We already have one of these in use in CircleCI.

@SWilson4 I see that there was an update to the `OQSBOT_GITHUB_ACTIONS` secret in the OQS organization settings about 16 hours ago, so I'm not sure if this request has already been solved.  If not, we can meet on Monday to fix this, I'm around most of the day.

baentsch:
> > EDIT 2023-02-01: This now simply adds a trigger for functionality proposed in [open-quantum-safe/oqs-provider#345](https://github.com/open-quantum-safe/oqs-provider/pull/345). It will function when somebody with the appropriate permissions (presumably @dstebila?) adds an oqsbot (classic) personal access token with repo and workflow permissions as an Actions secret with the variable name OQSBOT_GITHUB_ACTIONS. We already have one of these in use in CircleCI.
> 
> @SWilson4 I see that there was an update to the `OQSBOT_GITHUB_ACTIONS` secret in the OQS organization settings about 16 hours ago, so I'm not sure if this request has already been solved. If not, we can meet on Monday to fix this, I'm around most of the day.

@dstebila This is resolved (for now): see above https://github.com/open-quantum-safe/liboqs/pull/1654#discussion_r1477000718 .

bhess (approved):

.github/workflows/release-test.yml
bhess: Thanks, will do after merge.

dstebila:
> @baentsch @dstebila Could you take a look at the config related to personal access tokens? [This CircleCI job](https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/3334/workflows/8f01e768-7abe-453f-b145-0ff46966d8a2/jobs/26636) that triggers GitHub Actions via a (classic) personal access token is now failing.
> 
> Edit to add: Based on the error message, I suspect that the CircleCI access token will need to be replaced with a fine-grained one.

I'll need some help doing this, pop by when you can work with me on this.

baentsch:
Please retry. Should work again: Default when activating fine-grained access tokens was disabling classic tokens. My mistake. Now changed. But we may want to consider only using fine-grained tokens in the future.

SWilson4:
> Please retry. Should work again: Default when activating fine-grained access tokens was disabling classic tokens. My mistake. Now changed. But we may want to consider only using fine-grained tokens in the future.

Concurrently I worked with @dstebila to generate an `oqs-bot` fine-grained access token which is now stored in both CircleCI GitHub Actions environments, so it sounds like it's covered on both fronts. At any rate, `main` is green again.

Following the principle of least privilege, I like the idea of only allowing fine-grained tokens. However, I could see it being useful to be able to trigger `oqs-provider` release tests on arbitrary (non `-tracker`) refs from the command line using a "classic" personal access token. Perhaps we can revisit the decision after seeing how the new functionality is used.

PR #1650, opened 2024-01-04
Add Stateful Signature (XMSS and LMS)

ashman-p:
Adds implementations of stateful signatures to liboqs for XMSS and LMS. The feature include new OQS APIs to generate key-pairs, signature generation, verification, and secret key state-management. Actual secret key storage is left up to the application.

This PR also includes an enhancement to OQS SHA2 API to allow handling updates with arbitrary length buffers.

* [ No ] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [ Yes ] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)
OQS provider support is a separate and forthcoming feature. This is separated because of the need to manage the stateful of secret keys.

Issue #1648, opened 2024-01-02
Better separate internal and public APIs


SWilson4:
We have some common code which is not intended for library users but for which `.h` files exist (for example, SHA2 and SHA3). These headers are documented as not being part of the public API, but are still placed in the system `include` directory upon install (at least on my machine running x86_64 / Linux). We ought to fix this so that only the truly public headers are installed. On the other hand, we would like this code to be available to not only the implementations in `src` but also the programs in `tests`. It seems to me that a good solution would be to separate some sort of internal API (probably via tweaking the CMake configuration) so that our implementations of SHA2 / SHA3 / etc can be made available to select test programs but not users of the library in general.

dstebila:
We do have our common symmetric code in the public headers in [CMakeLists](https://github.com/open-quantum-safe/liboqs/blob/main/CMakeLists.txt#L140-L143). On the one hand that might be a mistake, although if memory serves we might have done it so that the test programs could access those functions?  But maybe there's a way to do it so that the test programs can access them but they're still not installed for users.

baentsch:
> maybe there's a way to do it so that the test programs can access them but they're still not installed for users.

That should be possible tweaking the `cmake` install logic: Tests should rely on the build-env only while the install logic wouldn't reference those header files. It seems we need to be more selective in setting [PUBLIC_HEADERS](https://github.com/open-quantum-safe/liboqs/blob/b19697422f433b85c8e46573078226a0956d37ae/CMakeLists.txt#L137-L145) for the [install logic to do the right thing](https://github.com/open-quantum-safe/liboqs/blob/b19697422f433b85c8e46573078226a0956d37ae/src/CMakeLists.txt#L128) as per this suggestion.

SWilson4:
Here's an idea: what about building an `oqs-internal` library corresponding to the internal API (currently just `common/aes.h`, `common/sha2.h`, and `common/sha3.h`). We can toggle the visibility for this library to `default` so that test programs can link against it and access the common functions. Naturally, it wouldn't be installed at any point; it would just stay in `build/lib`.

I tinkered with this a little on the branch `sw-experimentation` and I think it's fairly clean and scalable. It would provide us with an easy way to add common code that we want to expose to test programs but not users of the library. No per-test configuration required; just link against both `oqs` and `oqs-internal`. It _should_ (I believe) also resolve a previous linking issue when building shared libraries on MacOS, where the custom `randombytes` setting was overridden without a bespoke configuration.

[...]

PR #1655, opened 2024-01-08
Update `sig_stfl.h` document for #1650

ducnguyen-sb:
This PR address the comments in #1650 

This PR rewrite the documentation in `sig_stfl.h` header for a few APIs, make simpler and (somewhat) easier to read. 


src/sig_stfl/sig_stfl.h
```diff
@@ -284,37 +283,46 @@ typedef struct OQS_SIG_STFL_SECRET_KEY {
...
-	 *            The application manages this memory.
-	 * @returns  status of the operation populated with key material none zero length.
-	 * Caller is responsible to **unallocate** the buffer `sk_buf`.
+	 * @param[out] sk Pointer to an uninitialized `OQS_SIG_STFL_SECRET_KEY` object to hold the secret key.
+	 * @param[in] sk_len The length of the secret key byte stream.
+	 * @param[in] sk_buf Pointer to the byte stream containing the serialized secret key data.
+	 * @param[in] context Pointer to application-specific data, handled externally, associated with the key.
+	 * @returns OQS_SUCCESS if the deserialization succeeds, with the `sk` object populated with the key material.
+	 *
+	 * @remark The caller must ensure that resources (e.g., memory for `sk_buf`) are properly managed.
```
baentsch, line 325: What is the meaning of "properly managed"? Would it be better to state that explicitly (who/when alloc/free need to be called; is both heap and stack allocation OK?)?

src/sig_stfl/sig_stfl.h
```diff
@@ -433,7 +447,7 @@ OQS_API OQS_STATUS OQS_SIG_STFL_sign(const OQS_SIG_STFL *sig, uint8_t *signature
 OQS_API OQS_STATUS OQS_SIG_STFL_verify(const OQS_SIG_STFL *sig, const uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *public_key);
 
 /**
- * Query number of remaining signatures
+ * Query the number of remaining signatures.
```
baentsch, line 450: "remain" undocumented?

src/sig_stfl/sig_stfl.h
```diff
@@ -469,105 +483,118 @@ OQS_API void OQS_SIG_STFL_free(OQS_SIG_STFL *sig);
...
  *
  */
 OQS_API void OQS_SIG_STFL_SECRET_KEY_SET_lock(OQS_SIG_STFL_SECRET_KEY *sk, lock_key lock);
 
 /**
- * OQS_SIG_STFL_SECRET_KEY_SET_unlock .
+ * Attach an unlock mechanism to a secret key object.
  *
- * Sets function to prevent multiple processes from using the sk at the same time.
+ * This allows for proper synchronization * in a multi-threaded or multi-process environment,
```
baentsch, line 509: superfluous "*"?

src/sig_stfl/sig_stfl.h
```diff
@@ -469,105 +483,118 @@ OQS_API void OQS_SIG_STFL_free(OQS_SIG_STFL *sig);
...
  *
  */
 OQS_API void OQS_SIG_STFL_SECRET_KEY_SET_unlock(OQS_SIG_STFL_SECRET_KEY *sk, unlock_key unlock);
 
 /**
- * OQS_SIG_STFL_SECRET_KEY_SET_mutex .
+ * Assign a mutex function to handle concurrency control over the secret key.
  *
- * Sets function to prevent multiple processes from using the sk at the same time.
+ * This is to ensure that only one process can access or modify the key at any given time.
```
baentsch, line 522: What is the implication of _not_ setting such mutex? Will the library not function, even in a single-threaded environment?

src/sig_stfl/sig_stfl.h
```diff
@@ -469,105 +483,118 @@ OQS_API void OQS_SIG_STFL_free(OQS_SIG_STFL *sig);
...
+ * @return None.
  *
  */
 OQS_API void OQS_SIG_STFL_SECRET_KEY_SET_mutex(OQS_SIG_STFL_SECRET_KEY *sk, void *mutex);
 
 /**
- * OQS_SIG_STFL_SECRET_KEY_lock .
- *
- * Locks the secret key so only one application that holds the lock can access it.
+ * Lock the secret key to ensure exclusive access in a concurrent environment.
```
baentsch, line 532: I'd assume this function will fail if no mutex has been set?! Worthwhile stating? When does this function need to be called?

src/sig_stfl/sig_stfl.h
```diff
@@ -469,105 +483,118 @@ OQS_API void OQS_SIG_STFL_free(OQS_SIG_STFL *sig);
...
+ * in order to prevent simultaneous operations that could compromise the stateful signature security.
  *
+ * @param[in] sk Pointer to the secret key to be locked.
+ * @return OQS_SUCCESS if the lock is successfully applied; OQS_ERROR otherwise.
  */
 OQS_API OQS_STATUS OQS_SIG_STFL_SECRET_KEY_lock(OQS_SIG_STFL_SECRET_KEY *sk);
 
 /**
- * OQS_SIG_STFL_SECRET_KEY_unlock .
+ * Unlock the secret key, making it accessible to other processes.
```
baentsch, line 543: Same question as above: When does (an application) need to call this function?

baentsch (commented):
Thanks for the updates addressing many of my original comments, @ducnguyen-sb ! I only re-reviewed the changes in the PR, not the merge-to-main PR again: Please let me know when I should do that (to close out/mark as resolved my comments there).

src/sig_stfl/sig_stfl.h
ducnguyen-sb: Oh right, thanks for the question, I add code to catch this behavior. 
Fixed in 
https://github.com/open-quantum-safe/liboqs/pull/1655/commits/efccbe58a339266838a0617df8562886edc068da

ducnguyen-sb: Updated the doc in https://github.com/open-quantum-safe/liboqs/pull/1655/commits/c577c2ed8d6597c1766f5d9612397ed876061aa3

src/sig_stfl/sig_stfl.h
ducnguyen-sb: Caught the case when no mutex is set in https://github.com/open-quantum-safe/liboqs/pull/1655/commits/efccbe58a339266838a0617df8562886edc068da.  
Updated in https://github.com/open-quantum-safe/liboqs/pull/1655/commits/c577c2ed8d6597c1766f5d9612397ed876061aa3

src/sig_stfl/sig_stfl.h
ducnguyen-sb: My bad, removed in https://github.com/open-quantum-safe/liboqs/commit/c577c2ed8d6597c1766f5d9612397ed876061aa3

src/sig_stfl/sig_stfl.h
ducnguyen-sb: Added the description why the remaining signature and total signature matters in https://github.com/open-quantum-safe/liboqs/commit/c577c2ed8d6597c1766f5d9612397ed876061aa3

src/sig_stfl/sig_stfl.h
ducnguyen-sb: I updated the doc, but I'm not sure it's right. 
If the `sk_buf` is stack, then user don't have to deallocate. So I say a generic one "properly managed". 
I changed to 
`* @attention The caller is responsible for ensuring that `sk_buf` is securely deallocated when it's no longer needed.` in https://github.com/open-quantum-safe/liboqs/pull/1655/commits/c577c2ed8d6597c1766f5d9612397ed876061aa3

src/sig_stfl/sig_stfl.h
ducnguyen-sb: So I added: 
```
* This operation is essential in multi-threaded or multi-process contexts
 * in order to prevent simultaneous operations that could compromise the stateful signature security.
 
 ...
* @note It's not required to use lock and unlock functions in a single-threaded environment.

```
So, it's should be call in multi-threaded environment. 
I wonder if the current documentation is good enough, let me know if it's unclear.

ducnguyen-sb:
Thanks @baentsch , I addressed your comments. 
I believe it's simpler to address a smaller changes here rather than the big PR in #1650 , once it's resolved here we can mark the comments in the main branch as resolved.

src/sig_stfl/sig_stfl.h
baentsch: > I wonder if the current documentation is good enough, let me know if it's unclear.

I can only judge from my perspective and cannot know what other's may (mis)understand, so here goes my question now: Which (`liboqs` API) calls ("operations" in your wording above) need to be guarded with this function in multi-threaded environments? And a much more basic question: If there are such library functions failing to properly operate in a multi-threaded environment, wouldn't it be better to document _them_ accordingly? And as follow-on question: Does it need such `liboqs`-based "lock/unlock" functions at all if such documentation exists? Why isn't use of any external, e.g. as provided by the application-chosen threading library, lock/unlock mechanisms sufficient?

I guess what I'd like to suggest in a more "brutal question": Should we not strive to implement a thread-safe library API OR completely get out of the way (giving the app developer full control of what they want/need to do)? Anything in between IMO has the potential for a "support nightmare" ("use API XYZ if you are using threading library ABC; if not using any threading, disregard XYZ but use API XZZ", etc....)

ducnguyen-sb: I answer each question below. 

1. What function need to be guarded with this function in a concurrent environment? 

Only the Signing operation.

Thanks to your question, I updated this note: 

```
* @note It's not necessary to use this function in either Keygen or Verifying operations.
 *       In a concurrent environment, the user is responsible for locking and unlocking the private key,
 *       to make sure that only one thread can access the private key during a Signing operation.
 ```

2. Why such lock/unlock APIs are there? Is it meant we strive to be a thread-safe library?

These APIs intention is to limit to one owner of private key at a time in a concurrent environment.
If users supply lock/unlock logic with a mutex, we acquire the lock to be the only owner of the private key, and unlock after the Signing operation complete. 
If users don't  supply the lock/unlock logic or a mutex, they are simply NOP operations. 

I think the key different here is: We don't execute stateful signature in multi-thread  at the library itself, but we provide a way to let users use multi-thread securely, in the right way. 

3. Does it need such liboqs-based "lock/unlock" functions at all if such documentation exists? 

I think to have `lock/unlock` functions help liboqs in:

- Testing. Someone has to foreseen the usage of our proposed APIs in multi-thread environment, so we need `lock/unlock` to test if it's still safe in such if our low-level implementation change. 

So we have it in test, well, why not exposed them to library user, it's convenience anyway. So there are `liboqs`-based `lock/unlock` functions.

4. Why isn't use of any external, e.g. as provided by the application-chosen threading library, lock/unlock mechanisms sufficient?

Actually we don't concretely define the `lock/unlock` mechanisms in `liboqs`, if the function pointers of `lock/unlock` are non-NULL then we will run them to perform `lock/unlock` the private key. 

So, we open to any application-chosen threading library, and it's user responsibility to create a proper `lock/unlock` functions, and it's up to them to define their `mutex`. 

5. Anything in between IMO has the potential for a "support nightmare" ("use API XYZ if you are using threading library ABC; if not using any threading, disregard XYZ but use API XZZ", etc....)

I totally agree with you, we don't want to get into "support nightmare". I want to run away from it as far as possible. :)) 

You have more experience than me in this matter, please let me know if there is something I need to change to get out of "support nightmare".

ducnguyen-sb: @baentsch  before I merge this PR, do I answer your questions?

baentsch: > You have more experience than me in this matter, please let me know if there is something I need to change to get out of "support nightmare".

I'm not sure I do -- I did not use these APIs -- yet. My "general" question regarding minimizing support effort is: Could we completely do away with those functions in the external OQS API? If we could, we won't have to support them.

> So we have it in test, well, why not exposed them to library user, it's convenience anyway. So there are liboqs-based lock/unlock functions.

That's what I'm concerned with: If it's mere "convenience" (for us) we shouldn't externally make it available. Any API externally visible creates more support effort. Please see the discussion in #1648 and decide whether mutex/lock/unlock is another candidate for this internal API.

ducnguyen-sb: I double check, these APIs just for testing purpose, so I moved them to internal APIs. 
Updated in https://github.com/open-quantum-safe/liboqs/pull/1655/commits/8b0e15e5ca0772c958a4ac368a8cf0a13d63b70a

ducnguyen-sb: The `set_lock, set_mutex, and set_unlock` should still be public APIs. 
@baentsch

baentsch: :-( I won't respond to user questions about them.

PR #1664, opened 2024-01-15
Zephyr: CMake fixes

Frauschi:
* Minor fixes for the `CMakeLists.txt` file in the `zephyr` directory propably happened during rebasing of #1641.
* Minor improvements to the Zephyr specific CMake workarounds
* RiscV distinct board references have been removed to support all RiscV boards Zephyr supports.

### DIFF
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -8,32 +8,22 @@ if(CONFIG_LIBOQS)
                 # Workaround as the generic name "arm" is not a supported architecture in liboqs.
                 # In Zephyr, however, it is exclusively used for 32-bit ARM architectures.
                 set(CMAKE_SYSTEM_PROCESSOR "armv7")
-                # We have to set that manually as CMake can't detect it properly in Zephyr
-                set(CMAKE_SIZEOF_VOID_P 4)
         elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "riscv")
-                if(BOARD STREQUAL "qemu_riscv32")
-                        set(CMAKE_SYSTEM_PROCESSOR "riscv32")
-                        # We have to set that manually as CMake can't detect it properly in Zephyr
-                        set(CMAKE_SIZEOF_VOID_P 4)
-                elseif(BOARD STREQUAL "hifive1_revb")
-                        set(CMAKE_SYSTEM_PROCESSOR "riscv32")
-                        # We have to set that manually as CMake can't detect it properly in Zephyr
-                        set(CMAKE_SIZEOF_VOID_P 4)
-                elseif(BOARD STREQUAL "qemu_riscv64")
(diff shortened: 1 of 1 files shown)


zephyr/CMakeLists.txt
```diff
@@ -161,11 +151,14 @@ if(CONFIG_LIBOQS)
...
         if(CMAKE_SYSTEM_PROCESSOR STREQUAL "armv7")
-                # Undo the workaround from above to not interfere with other modules
                 set(CMAKE_SYSTEM_PROCESSOR "arm")
         elseif(CMAKE_SYSTEM_PROCESSOR EQUAL CMAKE_HOST_SYSTEM_PROCESSOR)
-                # Undo the workaround from above to not interfere with other modules
                 set(CMAKE_SYSTEM_PROCESSOR "posix")
+        elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "riscv")
+                set(CMAKE_SYSTEM_PROCESSOR "riscv")
+        elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
+                set(CMAKE_SYSTEM_PROCESSOR "x86")
```
dstebila, line 162: Can you explain this part here about undoing?  And in particular, why is `x86_64` being undone to `x86`?

Frauschi: Those workarounds (and their cleanup at the end) are necessary and exclusive to Zephyr. Zephyrs CMake infrastructure calls the `CMakeLists.txt` file in the `zephyr` directory of each module (in non-deterministic order). All CMake build environment variables (e.g. `CMAKE_SYSTEM_PROCESSOR`) are set by the Zephyr CMake infrastructure to its own, sometimes “uncommon” values (e.g. “arm” for all 32-bit ARM targets, “arm64” for all 64-bit targets and “x86” for all 32-bit **and** 64-bit x86 targets). Within the `zephyr/CMakeLists.txt` file, the main liboqs `CMakeLists.txt` file in the root directory is called. Hence, we have to convert the “custom” Zephyr build environment variables to the generic ones used in the root `CMakeLists.txt` to not have to make custom changes to that file exclusively for the Zephyr port (e.g. changing `CMAKE_SYSTEM_PROCESSOR` to “x86_64” in case the Zephyr value is “x86” and the 64-bit configuration is enabled, indicating “x86_64”). 

However, as there may be other Zephyr modules parsed after liboqs, we have to undo our modifications of the build environment variables to not interfere with other modules (and their potential workarounds). As this cleanup is done after the parsing of the root `CMakeLists.txt` of liboqs, the temporal workarounds only apply to those build targets.

dstebila: Ah, I see now this is limited in scope to the Zephyr CMake setting, thanks.

dstebila (approved):

PR #1667, opened 2024-01-16
Make internal API available to (only) test programs

### LINKED ISSUE #1648: Better separate internal and public APIs

SWilson4:
This PR serves three purposes:
1. to separate and define the OQS "internal API"
2. to make the internal API available to our test programs, and
3. to ensure that the internal API is not made available to library users.

It does so by doing the following:
- Move the NIST DRBG to the OQS internal API.
- Build the OQS internal API code (AES, SHA2, SHA3, and the NIST DRBG) as a separate `oqs-internal` library, with visibility set to "default".
- Don't install the internal API headers in the CMake include directory.
- Remove the internal API `.h` files from `oqs.h` and `.Doxyfile`.
- Add `#include` statements as necessary to address compilation failures due to the above.
- Link test programs against both `oqs` and `oqs-internal`.

This will hopefully make testing easier and in particular will unblock PR #1560. I've added it to the 0.10.0 milestone for the latter reason.

Fixes #1648.

Yes, the API is changed, but it's not much. The NIST DRBG is no longer exposed in the public API. This means changes to two public API functions: `OQS_randombytes_switch_algorithm` will now error when passed the string "NIST-KAT" (with or without capitalization) and `OQS_randombytes_nist_kat_init_256bit` has been deleted.


SWilson4:
I also should mention that I deleted the `nistseedexpander` PQClean shim, which is no longer needed since #1585.

dstebila (approved):

baentsch (commented):
Thanks @SWilson4 for this PR: After a first review, it looks good to me -- and reduces the "OQS support interface width" as (I think is) desirable. Allow me a bit more time to "test-drive" this branch in an `oqs-provider` setup. If you feel there's other sub-projects that may benefit from such test, be my guest :)

baentsch (approved):
Tested to work fine downstream. Thanks again for the reduction in potential support effort!

PR #1671, opened 2024-01-19
Call set_available_cpu_extensions using pthread_once

zxjtan:
Dependent on #1653.

`set_available_cpu_extensions()` is synchronized with a thread-unsafe array element. This is similar to #1549, except in this case instead of causing any crashes, the thread-unsafe behaviour simply leads some early callers to believe that the relevant CPU extensions are not available, and hence use reference implementations instead before the extensions are eventually calculated. This can and has set off TSAN.

It is also possible in rare instances for the thread calling `set_available_cpu_extensions()` to set `cpu_ext_data[OQS_CPU_EXT_INIT] = 1` and then crash before setting any of the other `cpu_ext_data` fields, preventing future callers from ever calling `set_available_cpu_extensions()` again, thus causing the program to use only reference implementations for its entire lifetime. This PR would also fix this edge case.

### DIFF
--- a/src/common/common.c
+++ b/src/common/common.c
@@ -13,6 +13,10 @@
 #include <stdlib.h>
 #include <string.h>
 
+#if defined(OQS_DIST_BUILD) && defined(OQS_USE_PTHREADS)
+#include <pthread.h>
+#endif
+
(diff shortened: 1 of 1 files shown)


dstebila (approved):

baentsch (approved):

baentsch:
Thanks for the contribution!

PR #1676, opened 2024-01-23
Disable Stateful Signatures in the build by default

ashman-p:
This commit will disable XMSS and LMS from the build by default. When enabled key and signature generation is also disabled by default. They can be enabled by defining the following variables on the config command.

-DOQS_ENABLE_SIG_STFL_KEY_SIG_GEN=ON
-DOQS_ENABLE_SIG_STFL_XMSS = ON
-DOQS_ENABLE_SIG_STFL_LMS = ON


baentsch (commented):
Thanks for adding this separator (and turning it off by default)! May I ask whether it would be possible to completely replace [OQS_SIG_STFL](https://github.com/open-quantum-safe/liboqs/blob/3bf7e800f93f6f9ee362f2cdbc068e33c0fb1b7c/src/sig_stfl/sig_stfl.h#L184) with [OQS_SIG](https://github.com/open-quantum-safe/liboqs/blob/3bf7e800f93f6f9ee362f2cdbc068e33c0fb1b7c/src/sig/sig.h#L108) if "OQS_ALLOW_SFTL_KEY_AND_SIG_GEN" is not set?

.CMake/alg_support.cmake
```diff
@@ -450,6 +450,24 @@ cmake_dependent_option(OQS_ENABLE_SIG_STFL_lms_sha256_h20_w8_h10_w8 "" ON "OQS_E
...
+endif()
+
+if (${OQS_ENABLE_SIG_STFL_KEY_SIG_GEN} AND ${OQS_ENABLE_SIG_STFL_LMS})
+    set(OQS_ALLOW_LMS_KEY_AND_SIG_GEN ON)
+else()
+    set(OQS_ALLOW_LMS_KEY_AND_SIG_GEN OFF)
+endif()
+
+if(OQS_ALLOW_SFTL_KEY_AND_SIG_GEN  STREQUAL "ON")
+    message(STATUS "Experimental stateful key and signature generation is enabled. Ensure secret keys are securely stored to prevent multiple simultanous sign operations.")
```
SWilson4, line 469:
```suggestion
    message(STATUS "Experimental stateful key and signature generation is enabled. Ensure secret keys are securely stored to prevent multiple simultaneous sign operations.")
```

CONFIGURE.md
```diff
@@ -112,6 +112,26 @@ When `OQS_USE_OPENSSL` is `ON`, CMake also scans the filesystem to find the mini
...
+
+XMSS and LMS are the two supported Hash-Based Signatures schemes
+OQS_ENABLE_SIG_STFL_XMSS and OQS_ENABLE_SIG_STFL_LMS contorl these algorithms which are disabled by default.
+A thrid variable, OQS_ENABLE_SIG_STFL_KEY_SIG_GEN also controls the ability to generate keys and signatures. This is also disabled by efault.
+Each of these variables can be set to `ON` or `OFF`.
+When all three are `ON`, stateful signatures is fully functional and can generate key-pairs, sign data, and verify signatures.
+If OQS_ENABLE_SIG_STFL_KEY_SIG_GEN is `OFF` signature verification is the only functioanl operation.
+
+Standards bodies, such as NIST, recommend that key and signature generation only by done in hardware in-order to best enforce the one-time use of secret keys.
+Keys stored in a file system is extreamly susceptible to simultaneous use.
```
SWilson4, line 125:
```suggestion
## Stateful Hash Based Signatures 

XMSS and LMS are the two supported Hash-Based Signatures schemes.
`OQS_ENABLE_SIG_STFL_XMSS` and `OQS_ENABLE_SIG_STFL_LMS` control these algorithms, which are disabled by default.
A thrid variable, `OQS_ENABLE_SIG_STFL_KEY_SIG_GEN`, also controls the ability to generate keys and signatures. This is also disabled by default.
Each of these variables can be set to `ON` or `OFF`.
When all three are `ON`, stateful signatures are fully functional and can generate key pairs, sign data, and verify signatures.
If `OQS_ENABLE_SIG_STFL_KEY_SIG_GEN` is `OFF` signature verification is the only functional operation.

Standards bodies, such as NIST, recommend that key and signature generation only by done in hardware in order to best enforce the one-time use of secret keys.
Keys stored in a file system are extremely susceptible to simultaneous use.
```

tests/kat_sig_stfl.c
```diff
@@ -297,7 +298,44 @@ OQS_STATUS sig_stfl_kat(const char *method_name, const char *katfile) {
 
 	ret = OQS_SUCCESS;
 	goto cleanup;
+#else
+	/*
+	 * Signature generation is disabled so only signature verification can be tested.
```
SWilson4, line 303: I know that we don't really have any "unhappy path" tests in liboqs, but I think they would be useful here as a sanity check to make sure that calling keygen / sign does, in fact, result in an error when those operations are disabled in the config.

SWilson4 (commented):
Thanks for putting this together, Norm. Just a few typo/formatting fixes from me and one suggestion regarding test coverage.

ashman-p:
> Thanks for adding this separator (and turning it off by default)! May I ask whether it would be possible to completely replace [OQS_SIG_STFL](https://github.com/open-quantum-safe/liboqs/blob/3bf7e800f93f6f9ee362f2cdbc068e33c0fb1b7c/src/sig_stfl/sig_stfl.h#L184) with [OQS_SIG](https://github.com/open-quantum-safe/liboqs/blob/3bf7e800f93f6f9ee362f2cdbc068e33c0fb1b7c/src/sig/sig.h#L108) if "OQS_ALLOW_SFTL_KEY_AND_SIG_GEN" is not set?

Hi, could you give an example or describe more about how that would look? Thanks.

tests/kat_sig_stfl.c
ashman-p: Thanks. Will do this on another test since the KAT test compares hashes of the expected output to determine success status of the test.

baentsch:
> > Thanks for adding this separator (and turning it off by default)! May I ask whether it would be possible to completely replace [OQS_SIG_STFL](https://github.com/open-quantum-safe/liboqs/blob/3bf7e800f93f6f9ee362f2cdbc068e33c0fb1b7c/src/sig_stfl/sig_stfl.h#L184) with [OQS_SIG](https://github.com/open-quantum-safe/liboqs/blob/3bf7e800f93f6f9ee362f2cdbc068e33c0fb1b7c/src/sig/sig.h#L108) if "OQS_ALLOW_SFTL_KEY_AND_SIG_GEN" is not set?
> 
> Hi, could you give an example or describe more about how that would look? Thanks.

Sure, my question is whether it would be possible (for the rest of the code to continue working) when completely wrapping the current definition for OQS_SIG_STFL in "sig_stfl.h" in an #ifdef OQS_ALLOW_SFTL_KEY_AND_SIG_GEN and 

    #define OQS_SIG_STFL OQS_SIG

in the #else (#ifndef OQS_ALLOW_SFTL_KEY_AND_SIG_GEN) branch.

baentsch:
@ashman-p I've started fleshing out my proposal above in the branch "mb-stfl-no-sigkeygen", so see e.g. https://github.com/open-quantum-safe/liboqs/blob/14999164201ffa61c20731de63abb9a3fbbcba07/src/sig_stfl/sig_stfl.h#L181-L183 ...

Question: What code generator did you use to create the many variants? It's tedious to manually #ifdef the now apparently unneeded replicate references to "oid", "sigs_remaining", "sigs_total"... I started this in the branch to see whether compilation succeeds -- it does: Those structs are apparently only needed in the disabled (under unset OQS_ALLOW_SFTL_KEY_AND_SIG_GEN) `keygen` and `sign` routines... So now I wonder whether there's a master template that can be adapted and used to generate the many *xmss*.c files(?).

If this were successful, XMSS/LMS algs might indeed be used (for verification) just like any other OQS_SIG & no downstream changes appear to be necessary to use this code right out of the box (only if configured without sig/gen).

baentsch:
I hate commenting my own comment but I have to:

> So now I wonder whether there's a master template that can be adapted and used to generate the many xmss.c files

After understanding the code a bit better I'd now see a much simpler solution, doing away with a possible generator and dozens of files and compile jobs, namely #defines for the `...new` etc functions. What is the reason this has not been done, @ashman-p ? This would make the code much easier to maintain, adapt and faster to build, no? Example:

```
// macro to en/disable OQS_SIG_STFL-only structs used only in sig&gen case:
#ifdef OQS_ALLOW_SFTL_KEY_AND_SIG_GEN
#define XMSS_SIGGEN(designator) \
        sig->oid = OQS_SIG_STFL_alg_xmss##designator##_oid; \
        sig->sigs_remaining = OQS_SIG_STFL_alg_xmss##designator##_sigs_remaining;\
        sig->sigs_total = OQS_SIG_STFL_alg_xmss##designator##_sigs_total;
#else
#define XMSS_SIGGEN(designator)
#endif

// generator for all alg-specific functions:
#define XMSS_ALG(mt, designator) \
OQS_SIG_STFL *OQS_SIG_STFL_alg_xmss##designator##_new(void) { \
\
        OQS_SIG_STFL *sig = (OQS_SIG_STFL *)malloc(sizeof(OQS_SIG_STFL)); \
        if (sig == NULL) { \
                return NULL; \
        } \
        memset(sig, 0, sizeof(OQS_SIG_STFL)); \
\
        XMSS_SIGGEN(designator) \
        sig->method_name = OQS_SIG_STFL_alg_xmss##designator; \
        sig->alg_version = "https://datatracker.ietf.org/doc/html/rfc8391"; \
        sig->euf_cma = true; \
\
        sig->length_public_key = OQS_SIG_STFL_alg_xmss##designator##_length_pk; \
        sig->length_secret_key = OQS_SIG_STFL_alg_xmss##designator##_length_sk; \
        sig->length_signature = OQS_SIG_STFL_alg_xmss##designator##_length_signature; \
\
        sig->keypair = OQS_SIG_STFL_alg_xmss##designator##_keypair;\
        sig->sign = OQS_SIG_STFL_alg_xmss##designator##_sign;\
        sig->verify = OQS_SIG_STFL_alg_xmss##designator##_verify;\
\
        return sig;\
} \
\
OQS_SIG_STFL_SECRET_KEY *OQS_SECRET_KEY_XMSS##designator##_new(void) {\
        return OQS_SECRET_KEY_XMSS_new(OQS_SIG_STFL_alg_xmss##designator##_length_sk);\
}\
\
OQS_API OQS_STATUS OQS_SIG_STFL_alg_xmss##designator##_keypair(XMSS_UNUSED_ATT uint8_t *public_key, XMSS_UNUSED_ATT OQS_SIG_STFL_SECRET_KEY *secret_key) {\
\
        if (public_key == NULL || secret_key == NULL || secret_key->secret_key_data == NULL) {\
                return OQS_ERROR;\
        }\
\
        if (xmss##mt##_keypair(public_key, secret_key->secret_key_data, OQS_SIG_STFL_alg_xmss##designator##_oid)) {\
                return OQS_ERROR;\
        }\
\
        return OQS_SUCCESS;\
}\
\
OQS_API OQS_STATUS OQS_SIG_STFL_alg_xmss##designator##_sign(uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, OQS_SIG_STFL_SECRET_KEY *secret_key) {\
        return OQS_SIG_STFL_alg_xmss##mt##_sign(signature, signature_len, message, message_len, secret_key);\
}\
\
OQS_API OQS_STATUS OQS_SIG_STFL_alg_xmss##designator##_verify(const uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *public_key) {\
        return OQS_SIG_STFL_alg_xmss##mt##_verify(message, message_len, signature, signature_len, public_key);\
}\
\
OQS_API OQS_STATUS OQS_SIG_STFL_alg_xmss##designator##_sigs_remaining(unsigned long long *remain, const OQS_SIG_STFL_SECRET_KEY *secret_key) {\
        return OQS_SIG_STFL_alg_xmss##mt##_sigs_remaining(remain, secret_key);\
}\
\
OQS_API OQS_STATUS OQS_SIG_STFL_alg_xmss##designator##_sigs_total(unsigned long long *total, const OQS_SIG_STFL_SECRET_KEY *secret_key) {\
        return OQS_SIG_STFL_alg_xmss##mt##_sigs_total(total, secret_key);\
}

// each of these lines replaces a complete file/compile target
// possibly wrapped with #ifdef OQS_ENABLE_SIG_STFL_xmss_....
XMSS_ALG(, _sha256_h16)
XMSS_ALG(mt, mt_sha256_h60_12)
....
```

Please let me know whether you'd want to adopt this -- or whether I'm overlooking a problem with this. If the former, let me know whether I should prepare a complete PR or you'd rather do it yourself.

ashman-p:
Yes, I agree that this macro would speed things up in general.
In your proposed changes. We would also not set vector functions for 'keygen' and 'sign' since i think those signatures are different between OQS_SIG and OQS_SIG_STFL. The oid values are also not used during a verify.

I am in the process of making an update to address one of Spencer's comments and also fix a build issue.
Please let me complete that task and then revisit yours?

baentsch:
> We would also not set vector functions for 'keygen' and 'sign' since i think those signatures are different between OQS_SIG and OQS_SIG_STFL. The oid values are also not used during a verify.

OK, that could be facilitated easily by moving those two assignments to the proposed "XMSS_SIGGEN" macro.

> Please let me complete that task and then revisit yours?

Sure. Just wanted to ask whether my thinking in general is OK for you. Thanks for the feedback.

ashman-p:
@baentsch, Let me study this a little, get feedback from @ducnguyen-sb and run some tests to try it.
I believe it will work but, having two APIs like this feels a little strange to me. That is, using the 'stateless' API to verify a 'stateful' signature. I will also look at converting the files to use a macro where applicable.

tests/kat_sig_stfl.c
ashman-p: Incorporated some of this in test_sig_stfl.c.

baentsch:
> I believe it will work but, having two APIs like this feels a little strange to me. That is, using the 'stateless' API to verify a 'stateful' signature. 

Yes, that may sound "unconventional" -- but the question seems to be: Is it logically right? And technically doable, i.e., is signature state sufficiently encoded in the signatures to permit verification? If the latter, I don't really see a mismatch: OQS_SIG then (continues to) support state-free signature operations (on all algorithms supporting that) while OQS_SIG_STFL adds functionality for stateful operations (on those algs needing/supporting this).

> I will also look at converting the files to use a macro where applicable.

That would be great: I got stuck a bit trying that in the pre-existing macro machinery loading "external/xmss.c" that you probably understand much better than I do.

dstebila:
> > I believe it will work but, having two APIs like this feels a little strange to me. That is, using the 'stateless' API to verify a 'stateful' signature.
> 
> Yes, that may sound "unconventional" -- but the question seems to be: Is it logically right? And technically doable, i.e., is signature state sufficiently encoded in the signatures to permit verification? If the latter, I don't really see a mismatch: OQS_SIG then (continues to) support state-free signature operations (on all algorithms supporting that) while OQS_SIG_STFL adds functionality for stateful operations (on those algs needing/supporting this).

It is an interesting idea... are we using the prefix `OQS_SIG_` to denote the family of algorithm or a class of functionality?

In the context of key generation, both `OQS_KEM_..._keygen` and `OQS_SIG_..._keygen` have the same function signature (output a public key and secret key pair) and both are intended for the same purpose (key generation), but we prefixed them separately rather than have a generic `OQS_..._keygen`.  

Granted, `SIG` and `SIG_STFL` are "closer" conceptually than KEM and SIG.

I don't have a conclusion at this point, but you made me think.

ashman-p:
I would like to treat this proposal as a separate task. If it is OK, let me merge this branch to 'stateful-sigs' with the stated goal of disabling the feature by default and only enable stateful verification.

I will continue the investigation of using OQS_SIG interface as well as converting some macro generated functions.

baentsch:
> Granted, SIG and SIG_STFL are "closer" conceptually than KEM and SIG.

Kinda... but I'm not sure I even fully concur with that: Granted, SIG and SIG_STFL both contain one function with pretty much identical signature and functionality (`verify`), but two with different signatures (`keygen` and `sign`) and wildly different usage logic (though conceptually keygen and sig). Taking the former as a case in point: Couldn't one argue the `keygen` in OQS_KEM is closer logically to the one in OQS_SIG than the one in OQS_SIG_STFL? This argument can be seen "in action" in `oqsprovider`: The code making use of the two is _very_ similar. The code to activate `keygen` of OQS_SIG_STFL in turn however would need a completely new logic and support structure (as would SIG_STFL:sign) -- both [moot points](https://github.com/open-quantum-safe/liboqs/pull/1650#issuecomment-1895747286).

> I don't have a conclusion at this point, but you made me think.

Thanks. I in turn am also re-thinking my original proposal to activate this OQS_SIG "verify-only" API if "OQS_ALLOW_SFTL_KEY_AND_SIG_GEN" is not set: I now begin to wonder whether this should always be used: This would also make it clear to every user which of the SIG_STFL operations are "different" (not saying [dangerous](https://github.com/open-quantum-safe/liboqs/pull/1650#issuecomment-1893957110)).

ducnguyen-sb:
Hi @ashman-p , sorry for the late reply. Please let me know if I can help in the next PR.

Issue #1666, opened 2024-01-16
Update constant-time suppression files for Sphincs+ and McEliece generic configuration


SWilson4:
Running the constant-time tests with the `generic` optimization target (`cmake -GNinja -DOQS_DIST_BUILD=OFF -DOQS_OPT_TARGET=generic -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON ..`) results in failures for Sphincs+ and McEliece. We haven't been running CT tests for these algorithms in CI and until recently we weren't running CT tests for the `generic` target due to a [CI configuration bug](https://github.com/open-quantum-safe/liboqs/pull/1618).

The full constant-time tests need to be run (locally) with the generic configuration and the suppression files updated so that there are no more failures.

PR #1677, opened 2024-01-23
Update McEliece suppression files for generic config

SWilson4:
This PR updates the suppression files for the "clean" (pure C) implementation of Classic McEliece. I've run the tests both in a container built from the CI image and locally on my machine.

For now, I've labelled all of these as "issues", as I'm not knowledgeable enough about McEliece to confidently classify them as false positives or true instances of secret-dependent behaviour. I believe that this is in line with what we had done previously for the AVX2 constant-time failures.

To test:
```bash
rm -rf build && mkdir build && cd build
cmake -GNinja -DOQS_DIST_BUILD=OFF -DOQS_OPT_TARGET=generic -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON .. && ninja ; cd -
python3 -m pytest -n=auto -k "McEliece" -v tests/test_constant_time.py
```

Partially addresses #1666.

I suspect that this may also address #1540. @bhess @praveksharma is it plausible that the "env-specific" constant-time errors were simply caused by building without AVX2 optimization? There were previously no suppressions for the "clean" implementation.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -16,7 +16,7 @@ jobs:
             container: openquantumsafe/ci-ubuntu-focal-x86_64:latest
             CMAKE_ARGS: -DOQS_DIST_BUILD=OFF -DOQS_OPT_TARGET=generic -DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON
             PYTEST_ARGS: --numprocesses=auto -k 'test_constant_time'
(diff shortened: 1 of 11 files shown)


SWilson4:
I've just updated the weekly constant-time runs so that the smallest parameter set of McEliece is tested there. I don't think it will add too much overhead, and it will (hopefully) make sure that we don't let these suppression files get out of sync in the future.

Issue #1366, opened 2023-01-20
Wrong flag passed to Clang when memory sanitizer build is requested.


xvzcf:
In [compiler_opts.cmake](https://github.com/open-quantum-safe/liboqs/blob/main/.CMake/compiler_opts.cmake#L112), we have the following code:

```
elseif(USE_SANITIZER STREQUAL "Memory")
    add_compile_options(-fsanitize=address)
    set(SANITIZER_LD_FLAGS "-fsanitize=memory")
elseif(USE_SANITIZER STREQUAL "MemoryWithOrigins")
...
```
I opened this issue to ask whether we should be running a memory-sanitizer build regularly to test this option.

dstebila:
Your point is that it should be `add_compile_options(-fsanitize=memory)`?  Yes, let's fix that.  

As for running it regularly in CI, sure, if the runtime is not problematic.

SWilson4:
I tried running the tests with the memory sanitizer turned on after creating #1680. The build was intensive, and the tests failed _very_ loudly. Interestingly enough there were quite a few complaints about functions in `libcrypto`, so some of the failures might be beyond our control. (That said, the tests also failed loudly with `-DOQS_USE_OPENSSL=OFF`.)

#1680 contains the one-line fix for the compile flag, so I'm going to preemptively rename this issue to clarify that it's tracking the proposed CI configuration. However, I think it will be a significant amount of work to get the memory sanitizer working in CI.

PR #1680, opened 2024-01-26
Set the correct compile flag for the memory sanitizer build

SWilson4:
Fixes a typo in our `USE_SANITIZER` configuration.

Partially addresses #1366.

No to both.


baentsch (approved):
Thanks for catching this bummer @SWilson4 !

dstebila (approved):

PR #1683, opened 2024-02-05
Update SPHINCS+ "clean" suppression files

### LINKED ISSUE #1666: Update constant-time suppression files for Sphincs+ and McEliece generic configuration

SWilson4:
Update the Sphincs+ suppression files so that constant-time tests are green. For now I've classified all of the failures as "issues"; someone with more knowledge about SPHINCS+ might want to take a look and see whether or not they're false positives.

The "SHA2-f" variants of SPHINCS+ are pretty fast, so I've added them to the weekly constant-time runs. Hopefully this keep us honest with regards to keeping suppression files up to date. It seems likely that this change will catch most constant-time errors in the "SHAKE" and "SHA2-s" variants as well.

Resolves #1666, pending merge of #1677.


baentsch (approved):
Same comment as before, now pertaining to the "constant time" property: I also see this as a feature of "product quality" code: Why spend the compute cycles checking something again and again that [OQS doesn't promise to provide](https://openquantumsafe.org/about/#limitations)? So, I'm OK with any of this -- until a clear decision is taken as to which properties OQS shall have (my goal would be towards "product usefulness", but this is a team decision). As and when that has been decided, we need to carefully comb through all tests and carefully review and enable those deemed essential (maybe only for a subset of algorithms), so for now we're good, I'd gather: Agreed, @dstebila ?

Issue #1561, opened 2023-09-26
Falcon-1024 KATs differ from upstream


SWilson4:
From #1560:

> our implementation of Falcon-1024 differs from the upstream on a single KAT, number 82. I put together a demonstration here: https://github.com/SWilson4/falcon-diff

> This seems to be explained by the signature buffer size used for the KATs: our implementation uses the signature size of 1280 bytes while the upstream code passes a larger buffer of 1330 bytes (even though the signature size upstream is still defined to be 1280 bytes). I'm not sure yet why this is done, but redefining the signature size to be 1330 in liboqs does make all 100 KATs pass.

SWilson4:
I recently received insight into this issue from Gregor Seiler of the Falcon team. This is my paraphrase of what was written to me:

The scheme samples (and resamples) signatures until one which is short enough to fit in the given buffer is found. The "maximum size" (quotation mine) of 1280 bytes was chosen somewhat arbitrarily. It's a "nice" (presumably meaning divisible by 128) number such that signatures usually are small enough on the first try. However, the buffer size for the KAT testing was chosen such that no retrying would be required---hence the larger value of 1330 bytes.

I've confirmed that this is the cause of the mismatch by modifying the Falcon reference KAT generation program to use a 1280-byte buffer and to retry signatures if necessary. With these modifications, it agrees with the ([yet unmerged](https://github.com/open-quantum-safe/liboqs/pull/1560)) OQS test program on all 100 KATs.

I see three different approaches we could take to resolve this:
1. Use our own KAT program's output (equal to that of the modified reference KAT program) as the source of truth.
2. Modify our KAT test program to use a larger buffer size for Falcon-1024, so that it agrees with the reference KAT program.
3. Change the Falcon signature size macro to 1330 from 1280.

I don't love any of these, especially not the third option. I lean toward the first. If anyone would like to weigh in on one of these or present a different path, please do so here.

SWilson4:
I'm not sure if anyone will get a notification about this by default, so pinging @baentsch @dstebila @praveksharma @bhess for input.

praveksharma:
Thank you for looking into this @SWilson4!  

I'm not sure if testing the the KAT program against its own output would be entirely appropriate if that's what you're suggesting in 1? I'm leaning towards 2 but I don't know the changes that this would necessitate and how this would impact maintainability.

baentsch:
[...]

PR #1684, opened 2024-02-05
Rename weekly runs and skip Falcon-1024 [skip ci]

SWilson4:
As expected, Falcon-1024 fails the weekly extended KAT tests. This will be the case until #1561 is resolved. For now, the failure creates unnecessary noise. This PR
1. renames the weekly test workflow to reflect that it's no longer just a constant-time test run and
2. skips the Falcon-1024 extended KAT tests.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -1,4 +1,4 @@
-name: Weekly constant time tests
+name: Weekly extended tests
 
 on:
   schedule:
@@ -46,10 +46,12 @@ jobs:
             container: openquantumsafe/ci-ubuntu-focal-x86_64:latest
             CMAKE_ARGS: -DOQS_DIST_BUILD=OFF -DOQS_OPT_TARGET=generic
             PYTEST_ARGS: --numprocesses=auto -k 'test_kat_all'
+            SKIP_ALGS: 'Falcon-1024' # re-enable when #1561 is resolved
           - name: extensions
             container: openquantumsafe/ci-ubuntu-focal-x86_64:latest
             CMAKE_ARGS: -DOQS_DIST_BUILD=OFF -DOQS_OPT_TARGET=haswell
             PYTEST_ARGS: --numprocesses=auto -k 'test_kat_all'
+            SKIP_ALGS: 'Falcon-1024' # re-enable when #1561 is resolved
     container:
       image: ${{ matrix.container }}
     steps:


baentsch (approved):
LGTM. I wonder whether we may even want to drop weekly tests altogether: I see these as a feature supporting "production grade" code. And as this doesn't seem to be a clear(ly articulated) goal, why waste these compute cycles? Maybe add the test to release testing to not lose the work, but that should suffice, no?

dstebila:
> LGTM. I wonder whether we may even want to drop weekly tests altogether: I see these as a feature supporting "production grade" code. And as this doesn't seem to be a clear(ly articulated) goal, why waste these compute cycles? Maybe add the test to release testing to not lose the work, but that should suffice, no?

It's a goal to work towards, let's not throw away the work we've done now towards it.

SWilson4:
> LGTM. I wonder whether we may even want to drop weekly tests altogether: I see these as a feature supporting "production grade" code. And as this doesn't seem to be a clear(ly articulated) goal, why waste these compute cycles? Maybe add the test to release testing to not lose the work, but that should suffice, no?

I think that it's worth running the extended KAT tests on a weekly basis regardless of the focus of `liboqs`. Having implementations that match the full set of available KATs is (IMO) important from both experimental and production perspectives. The utility of those tests is, I think, illustrated by the fact that they caught a Falcon implementation inconsistency in PQClean that had existed for a couple of years.

The constant-time tests are perhaps more production-focused, but there is active research on side-channel attacks on standards candidates, and our tests are (again IMO) useful for this. For example, they [detected](https://github.com/open-quantum-safe/liboqs/issues/995) the non-CT behaviour that led to [a key-recovery attack on HQC](https://eprint.iacr.org/2021/1485). This attack led to the HQC Round 4 submission being revised. I don't think our tests directly contributed to that attack being discovered, but I could see them providing a starting point for side-channel analysis of (say) the signature on-ramp candidates if/when those are integrated into the library.

I suppose that running the extended KAT tests and CT tests only at release time would eventually catch these issues, but from a maintenance point of view I'd rather find out about them before we get to the point of cutting a release candidate.

baentsch:
> It's a goal to work towards, let's not throw away the work we've done now towards it.

I did not suggest throwing stuff away, but to consciously decide what to do -- and then pursue that rigorously.

> The utility of those tests is, I think, illustrated by the fact that they caught a Falcon implementation inconsistency in PQClean that had existed for a couple of years

The utility of the tests is not what I question, it's the frequency of execution. A single (properly configured :-/) run of the tests (when integrating a new code base, say) suffices to unearth such problems. In the pursuit to speed up the PR CI cycle, it clearly was my fault to suggest running tests weekly _without_ instituting a rigorous procedure around checking and resolving issues they discover. The new "automated downstream" merge trigger (when someone checks one of the "new code" boxes) could be the place where these tests could live in the future. Or we agree that a "red" weekly CI mark in the dashboard needs to be addressed before the next weekly call.

> from a maintenance point of view I'd rather find out about them before we get to the point of cutting a release candidate.

I completely second that -- and that rationale was AFAIK the reason why we instituted them. But again, I think we do not have sufficient clarity what we want to achieve by passing those tests (or which algorithms _must_ pass them). Only that decision will inform us about the rigour (or not; or maybe different for different algorithms) we need to pursue them.

My rationale for keeping these tests "green" was to increase the level of trust people can have in the library to the degree that they can deploy it productively and every carve-out hurt me (with that goal in mind). But without that goal in mind we can (should?) focus our work effort on other things, e.g., bring XMSS "over the finish line". By way of example: If "product readiness" were a (stated and pursued) goal, I'd fight integrating XMSS+LMS tooth and nail (until they pass all --to be defined-- essential tests, such as constant-time, no memory leaks, standardized KATs/interop, downstream utility/"integrate-ability", proper licensing, etc... whatever we deem (document!) as prerequisite).

dstebila (approved):

SWilson4:
Merging now, but linking #1689 so that the discussion here leaves a footprint there.

PR #1687, opened 2024-02-06
Na stateful macro

ashman-p:
The Stateful Hash-Based Signature feature, when enabled, does signature verification by default. In this mode it uses the OQS_SIG API and only populates the 'verify' vector function. In the experimental full mode (verify, key and signature generation) use of OQS_SIG_STFL is needed.
Also, this PR includes code refactored to use macros to generate repetitive functions for HBS variants.


src/sig_stfl/sig_stfl.c
```diff
@@ -920,20 +921,36 @@ OQS_API OQS_STATUS OQS_SIG_STFL_verify(const OQS_SIG_STFL *sig, const uint8_t *m
 	}
 }
 
+
 OQS_API OQS_STATUS OQS_SIG_STFL_sigs_remaining(const OQS_SIG_STFL *sig, unsigned long long *remain, const OQS_SIG_STFL_SECRET_KEY *secret_key) {
+#ifndef OQS_ALLOW_SFTL_KEY_AND_SIG_GEN
+	(void)sig;
+	(void)remain;
+	(void)secret_key;
+	return OQS_ERROR;
```
Martyrshot, line 930: Would it make sense to add an OQS_UNUSED macro in our common defines?

Likely doesn't make sense for the PR, and if this is the only spot in the codebase that has unused params it might not make sense at all...

Martyrshot: cc: @baentsch @dstebila @SWilson4 @praveksharma

Martyrshot: Sounds good to me. :)

ashman-p:
> Excellent -- Thanks very much! This reduces the LOC count markedly and makes this code much more maintainable. Only nit: Do I understand it right that only a single alg gets tested? No longer too much of an issue as all of them now share the same code base, though.

Thanks @baentsch. 
I may have missed it but, what do you mean "single algorithm gets tested"?
My expectation is that all the enabled HBS variants would be tested. 
tests/KATs/sig_stfl/kats.json

ashman-p:
The example that makes the most sense is the KATs used for verification tests.
e.g. python3 tests/test_kat.py

baentsch:
> The example that makes the most sense is the KATs used for verification tests. e.g. python3 tests/test_kat.py

Thanks for this pointer to the test code. Open question remaining: Which CI runs do not skip these tests? Which CI runs trigger only verification, which ones keygen+sign+verify?

PR #1688, opened 2024-02-07
properly document release support level [skip ci]

baentsch:
Update security status of library releases. [Release wiki](https://github.com/open-quantum-safe/liboqs/wiki/Release-process#code-change) has been amended to avoid such omissions from happening again.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,9 +4,14 @@
 
 We only support the most recent release.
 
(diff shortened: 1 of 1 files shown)


SWilson4 (approved):
Good catch, thanks Michael.

dstebila (approved):

Issue #1630, opened 2023-12-14
Improve CI for macOS on Apple Silicon


dstebila:
Due to poor CI support this is currently a Tier 3 system according to our PLATFORMS.md document, but it should be better.

https://github.com/open-quantum-safe/liboqs/pull/1605#discussion_r1414029843

SWilson4:
@dstebila @baentsch It looks it should now be feasible to use an M1 runner through GitHub actions: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source.

dstebila:
> @dstebila @baentsch It looks it should now be feasible to use an M1 runner through GitHub actions: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source.
> 
> Edit to add: The M1 runner outperforms the Intel runner in a (very) limited `oqs-provider` test run: https://github.com/open-quantum-safe/oqs-provider/actions/runs/7846813573

It would be great to start using this and move away from our ad hoc solution involving one of our own machines.

baentsch:
> It would be great to start using this and move away from our ad hoc solution involving one of our own machines.

We presently do that basically for profiling (in the process also testing). Indeed it would be nice to also do that on github (weekly run) if we decide to keep that sub project alive...

SWilson4:
Related to today's meeting: GitHub Actions does seem to have Arm runners, which are currently in private beta: https://github.blog/changelog/2023-10-30-accelerate-your-ci-cd-with-arm-based-hosted-runners-in-github-actions

A GitHub executive recently mentioned in the OQS GitHub that they had helped to get LF projects additional GitHub features in the past—perhaps it's worth following up with them to see if we can get into the private beta?

Martyrshot:
Worth mentioning: Someone from github joined the `pqca` discord and offered to talk to the admins about getting extra features. Ry Jones responded saying that he would contact them once they are back in office, so it might be worth talking to Ry about.

SWilson4:
> Worth mentioning: Someone from github joined the `pqca` discord and offered to talk to the admins about getting extra features. Ry Jones responded saying that he would contact them once they are back in office, so it might be worth talking to Ry about.

This is what I meant to convey, but I typed "GitHub" instead of "Discord" :speak_no_evil:

Martyrshot:
[...]

PR #1692, opened 2024-02-09
Update GitHub Actions workflows for stateful signatures

SWilson4:
Integrate stateful signatures into our CI workflows.

I took the approach of simply switching on LMS and XMSS for all of our GitHub Actions builds that support all algorithms and adding an additional run in select places to test the configuration in which keygen and sign are disabled.

It looks like the logic around enabling / disabling keygen/sign needs to be tweaked in the XMSS KAT testing function.

We still need to test these on MacOS, which will require updating CircleCI, but I thought it best to put this draft up here as is for now to unblock anyone who wants to work on the test failures.

### DIFF
--- a/scripts/build-android.sh
+++ b/scripts/build-android.sh
@@ -6,12 +6,13 @@ set -e
 
 show_help() {
     echo ""
-    echo " Usage: ./build-android <ndk-dir> -a [abi] -b [build-directory] -s [sdk-version]"
+    echo " Usage: ./build-android <ndk-dir> -a [abi] -b [build-directory] -s [sdk-version] -f [extra-cmake-flags]"
 
     echo "   ndk-dir: the directory of the Android NDK (required)"
     echo "   abi: the Android ABI to target for the build"
     echo "   build-directory: the directory in which to build the project"
     echo "   sdk-version: the minimum Android SDK version to target"
+    echo "   extra-cmake-flags: extra flags to use for CMake configuration"
     echo ""
     exit 0
 }
@@ -52,12 +53,13 @@ MINSDKVERSION=21
 BUILDDIR="build"
 
 OPTIND=2
-while getopts "a:s:b:" flag
+while getopts "a:s:b:f:" flag
 do
     case $flag in
         a) ABI=$OPTARG;;
         s) MINSDKVERSION=$OPTARG;;
         b) BUILDDIR=$OPTARG;;
+        f) EXTRAFLAGS="$OPTARG";;
         *) exit 1
     esac
 done
@@ -107,7 +109,10 @@ cmake .. -DOQS_USE_OPENSSL=OFF \
          -DBUILD_SHARED_LIBS=ON  \
          -DCMAKE_TOOLCHAIN_FILE="$NDK"/build/cmake/android.toolchain.cmake \
          -DANDROID_ABI="$ABI" \
-         -DANDROID_NATIVE_API_LEVEL="$MINSDKVERSION"
+         -DANDROID_NATIVE_API_LEVEL="$MINSDKVERSION" \
+         -DOQS_ENABLE_SIG_STFL_LMS=ON \
+         -DOQS_ENABLE_SIG_STFL_XMSS=ON \
+         $EXTRAFLAGS
 cmake --build ./
 
 # Provide rudimentary information following build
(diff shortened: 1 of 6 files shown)


SWilson4:
@baentsch @dstebila For your consideration: it looks to me like a lot of our CircleCI runs (most of the Linux and MacOS configs) could be ported over to GitHub Actions, and [we might get M1 Mac testing "for free" in the process](https://github.com/open-quantum-safe/liboqs/issues/1630#issuecomment-1936219108). Is there a reason to prefer keeping them on CircleCI? Since we have to revise the workflows anyway to enable stateful sigs, this seems like an opportunity to kill two birds with one stone: move as many as possible to GH and enable stateful sigs en route.

baentsch (approved):
Thanks! LGTM. What about adding one test config with only LMS, one only XMSS, one both disabled, too? I'd not suggest adding this to the matrix but maybe just to our primary build platform, a linux under `x64` to catch "interdependency" errors in that alg category?

baentsch:
> Is there a reason to prefer keeping them on CircleCI?

Not really: I have a long-running goal of [moving off CCI](https://github.com/open-quantum-safe/oqs-provider/issues/248) but there's (been?) [a catch](https://github.com/open-quantum-safe/oqs-provider/issues/248#issuecomment-1712417062).

ducnguyen-sb:
Hi @SWilson4 , I can help fix the XMSS errors. I've been away for a while, can you let me know what is the changes that lead to XMSS kat errors?

SWilson4:
> Hi @SWilson4 , I can help fix the XMSS errors. I've been away for a while, can you let me know what is the changes that lead to XMSS kat errors?

It looks to me like it's just a matter of the remaining / total signatures not printing correctly when keygen / sign are disabled: https://github.com/open-quantum-safe/liboqs/blob/0a44a2583534f8b96992c791b0bedac0d51e4d67/tests/kat_sig_stfl.c#L327

ducnguyen-sb:
Thanks @SWilson4 , I will handle it from here. Can I make a commit to this PR? (I assume yes, just ask to be sure)

baentsch:
> Can I make a commit to this PR? (I assume yes, just ask to be sure)

That would be great -- Goal would be to "turn CI green": Thanks in advance!

SWilson4:
> Thanks @SWilson4 , I will handle it from here. Can I make a commit to this PR? (I assume yes, just ask to be sure)

Yes please!

dstebila:
> @baentsch @dstebila For your consideration: at a glance it looks to me like a lot of our CircleCI runs (most of the Linux and MacOS configs) could be ported over to GitHub Actions, and [we might get M1 Mac testing "for free" in the process](https://github.com/open-quantum-safe/liboqs/issues/1630#issuecomment-1936219108). Is there a reason to prefer keeping them on CircleCI? Since we have to revise the workflows anyway to enable stateful sigs, this seems like an opportunity to kill two birds with one stone: move as many as possible to GH and enable stateful sigs en route.

Go for it!  No reason to stick with CircleCI if we can get what we need from GitHub Actions.

ducnguyen-sb:
All checks have passed. Please review this PR. @SWilson4 @baentsch

ashman-p (approved):

scripts/build-android.sh
```diff
@@ -107,7 +109,10 @@ cmake .. -DOQS_USE_OPENSSL=OFF \
          -DBUILD_SHARED_LIBS=ON  \
          -DCMAKE_TOOLCHAIN_FILE="$NDK"/build/cmake/android.toolchain.cmake \
          -DANDROID_ABI="$ABI" \
-         -DANDROID_NATIVE_API_LEVEL="$MINSDKVERSION"
+         -DANDROID_NATIVE_API_LEVEL="$MINSDKVERSION" \
+         -DOQS_ENABLE_SIG_STFL_LMS=ON \
```
baentsch, line 113: This activates XMSS/LMS by default for android builds. Sensible for CI, not so for "unsuspecting" users of the script: Suggest to add at least an output warning message that this is an dangerous and not-recommended setting.

baentsch (approved):
Thanks for adding tests also with algs=ON and keygen/sig=ON|OFF! Apart from one nit LGTM.

scripts/build-android.sh
SWilson4: Good catch. I've reworked the script so that XMSS/LMS are disabled by default.

PR #1694, opened 2024-02-12
Refactor OpenSSL Implementation of SHA3 SHAKE to use new Squeeze API

### LINKED ISSUE #1539: Update OpenSSL integration
OpenSSL is adding the feature for [SHAKE to squeeze multiple times with different output sizes](https://github.com/openssl/openssl/pull/21511) which changes the [baseline in liboqs OpenSSL integration for SHA3](https://github.com/open-quantum-safe/liboqs/blob/0dafd4f02c5a8fc000502ae51005a9962b671860/src/common/sha3/ossl_sha3.c#L147-L162).
This issue is to propose using the new OpenSSL API to help validate it.

Eddy-M-K:
Refactor OpenSSL Implementation of SHA3 SHAKE to use new `EVP_DigestSqueeze()` allowing for multiple calls to squeeze

Resolves #1539

Please correct me if I'm wrong but it looks like `n_out` of `intrn_shake*_inc_ctx`s are used solely to keep track of the byte number of the input and are unnecessary with the new Squeeze API. The `test_sha3` test ran fine when I removed all instances of `n_out` so I'm curious if this is no longer necessary. 

If `n_out` is indeed unnecessary, then perhaps we can also remove the wrapper structs `intrn_shake128_inc_ctx` and `intrn_shake256_inc_ctx` and instead point `state->ctx` directly to an instance of `EVP_MD_CTX`.


beldmit:
May I ask you to put this changes under some #ifdef? OpenSSL 3.0/3.1 are still available and these branches don't have the Squeeze API

baentsch (commented):
Thanks @Eddy-M-K for putting this PR together. Conceptually this looks OK. The [comment](https://github.com/open-quantum-safe/liboqs/pull/1694#issuecomment-1938130925) explains why [this CI test is failing](https://github.com/open-quantum-safe/liboqs/actions/runs/7867628612/job/21463873772?pr=1694) (it enables "OQS_USE_SHA3_OPENSSL" on a downlevel OpenSSL): You may simply wrap the new code in a suitable `#ifdef OPENSSL_VERSION_NUMBER >= ...` (do you happen to know the exact value that would be right, @beldmit ?)

You may also consider adding "-DOQS_USE_SHA3_OPENSSL=ON" to [this CI test](https://github.com/open-quantum-safe/liboqs/blob/3e34ed18548d504aadc0edf0fac53af5f664a7be/.github/workflows/linux.yml#L90) which is using 3.0.2 and maybe some other "OpenSSL master" test to test-drive the OpenSSL SHA3 code more regularly (as "OQS_USE_SHA3_OPENSSL" is OFF by default).

beldmit:
According to man pages `EVP_DigestSqueeze()` function was added in OpenSSL 3.3.

Eddy-M-K:
Oh right, completely forgot about backward compatibility with the current version. Thank you @beldmit and @baentsch - will fix this.

SWilson4:
> You may also consider adding "-DOQS_USE_SHA3_OPENSSL=ON" to [this CI test](https://github.com/open-quantum-safe/liboqs/blob/3e34ed18548d504aadc0edf0fac53af5f664a7be/.github/workflows/linux.yml#L90) which is using 3.0.2 and maybe some other "OpenSSL master" test to test-drive the OpenSSL SHA3 code more regularly (as "OQS_USE_SHA3_OPENSSL" is OFF by default).

@baentsch Do you happen to know which jobs (if any) use OpenSSL `master`? It looks to me based on CI logs like all of our GitHub Actions runs are using either 1.1.1 or 3.0.2. If there indeed aren't any, should one be added? Or is that better left for downstream testing with the provider?

baentsch:
> Do you happen to know which jobs (if any) use OpenSSL master?

Good question! I don't (recall us ever having had that goal), so the answer probably is "None". Until this PR I would have said "so what" as `liboqs` never used any unreleased `openssl` features. This PR apparently changes this, so maybe should also introduce `openssl` "master" tests. But then again, I'd argue against targeting "master" as that may stall `liboqs` CI for PRs based on things this project does not control. Also "master"-based testing would increase CI time substantially. I'd therefore argue for tests (particularly targeting this code) using a specific `openssl` commit in the "3.3.0-dev" range -- and replacing that with a tagged "3.3.0" build as and when that is released. This way, the `openssl` build can be cached in CI and will run fast(er than a "master" build always created from scratch/code).

.github/workflows/unix.yml
```diff
@@ -218,3 +218,50 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          path: .localopenssl330
+          key: ${{ runner.os }}-openssl330
+      - name: Checkout the OpenSSL v3.3.0 commit
+        if: steps.cache-openssl330.outputs.cache-hit != 'true'
+        uses: actions/checkout@v4
+        with:
+          repository: 'openssl/openssl'
+          ref: '536649082212e7c643ab8d7bab89f620fbcd37f0'
```
SWilson4, line 241: It looks like OpenSSL has recently (a few days ago) tagged this commit as `openssl-3.3.0-beta1`. Might as well use that tag here to improve readability.
```suggestion
          ref: 'openssl-3.3.0-beta1'
```

SWilson4 (approved):
Looks good to me, with one minor suggestion. Thanks, Eddy!

dstebila (approved):
Thanks Eddy!

Eddy-M-K:
Had to make a couple force pushes due to a failed attempt to sign off the last commit :cold_sweat:

PR #1702, opened 2024-02-21
Small fixes after adding ML-*

bhess:
Two small fixes after #1626:
- The fragment to filter "STD" signature algorithms contained an error
- The script to reproduce ML-* vectors used an incorrect directory name


dstebila (approved):

.CMake/alg_support.cmake
```diff
@@ -448,7 +448,7 @@ if(NOT ((OQS_MINIMAL_BUILD STREQUAL "") OR (OQS_MINIMAL_BUILD STREQUAL "OFF")))
...
 elseif (${OQS_ALGS_ENABLED} STREQUAL "STD")
 ##### OQS_COPY_FROM_UPSTREAM_FRAGMENT_LIST_STANDARDIZED_ALGS_START
-	filter_algs("KEM_ml_kem_512_ipd;KEM_ml_kem_768_ipd;KEM_ml_kem_1024_ipd;SIG_dilithium_;SIG_dilithium_;SIG_dilithium_;SIG_ml_dsa_44_ipd;SIG_ml_dsa_65_ipd;SIG_ml_dsa_87_ipd;SIG_falcon_512;SIG_falcon_1024;SIG_sphincs_sha2_128f_simple;SIG_sphincs_sha2_128s_simple;SIG_sphincs_sha2_192f_simple;SIG_sphincs_sha2_192s_simple;SIG_sphincs_sha2_256f_simple;SIG_sphincs_sha2_256s_simple;SIG_sphincs_shake_128f_simple;SIG_sphincs_shake_128s_simple;SIG_sphincs_shake_192f_simple;SIG_sphincs_shake_192s_simple;SIG_sphincs_shake_256f_simple;SIG_sphincs_shake_256s_simple")
+	filter_algs("KEM_ml_kem_512_ipd;KEM_ml_kem_768_ipd;KEM_ml_kem_1024_ipd;SIG_ml_dsa_44_ipd;SIG_ml_dsa_65_ipd;SIG_ml_dsa_87_ipd;SIG_falcon_512;SIG_falcon_1024;SIG_sphincs_sha2_128f_simple;SIG_sphincs_sha2_128s_simple;SIG_sphincs_sha2_192f_simple;SIG_sphincs_sha2_192s_simple;SIG_sphincs_sha2_256f_simple;SIG_sphincs_sha2_256s_simple;SIG_sphincs_shake_128f_simple;SIG_sphincs_shake_128s_simple;SIG_sphincs_shake_192f_simple;SIG_sphincs_shake_192s_simple;SIG_sphincs_shake_256f_simple;SIG_sphincs_shake_256s_simple")
```
baentsch, line 451: (Still) not quite perfect: This apparently disables "KEM_ml_kem_512". Intentional? This is the reason for `oqsprovider` CI now failing. Positive is that this exercizes (kind-of successfully) `oqsprovider` without a single KEM active in `liboqs` (when built with STD config).

What about adding a test exercizing the "alias" feature (ascertaining that the same OQS_... defines get set)?  "oqsconfig.h" otherwise looks like this:

```
#define OQS_ENABLE_KEM_ml_kem_512_ipd 1
/* #undef OQS_ENABLE_KEM_ml_kem_512 */
...
``

**Review by baentsch — CHANGES_REQUESTED** [review]
Please ascertain aliases and originals operate identically. Current problem: Header-defines (can -- and do) mismatch (see single comment).

File: .CMake/alg_support.cmake (hunk 1 above)
Comment (bhess) [reply .CMake/alg_support.cmake]: Thank you for catching this! I've now added the aliases to the fragment.
Ideally the alias would be automatically enabled in oqsconfig.h with a `cmake_dependent_option`. I'll look into this and a test next week.
But there seems another issue with `filter_algs` which appears to set everything to OFF except those explicitly provided.
For example, using the STD config I get
```
#define OQS_ENABLE_SIG_FALCON 1
#define OQS_ENABLE_SIG_falcon_512 1
/* #undef OQS_ENABLE_SIG_falcon_512_avx2 */
/* #undef OQS_ENABLE_SIG_falcon_512_aarch64 */
...
```
But I'd expect the aarch64-optimized version to be enabled on my laptop, which is set with the cmake default config.

baentsch: > I'll look into this and a test next week.
But there seems another issue with filter_algs which appears to set everything to OFF except those explicitly provided.

OK, #1703 created to track and unblock this PR.

baentsch (approved):

baentsch:
A (non-merge-blocking) comment, though: I understand you have a propensity to retain the "*-ipd" algorithms as "STD" algorithms, but I personally consider this logically incorrect and my suggestion is to _not_ include them in this algorithm selection.

Rationale: As and when a v2 ipd arrives or when ipd gets approved as final, they will need to be manually removed from this list again -- in other words, no matter what, they will need to be removed eventually; in turn, the algorithm names without "-ipd" moniker _are_ standardized and will remain the name of valid "STD" algorithms, whatever their underlying implementation.

-> If you don't want to subscribe to this logic, may I then suggest to at least add a comment to this code along the lines "TBD - manually revise as standard process moves on"? Then we could create a script scouring for "TBD" monikers and consciously decide whether or not this (and other instances containing such aide memoire) is still OK when doing a release.

bhess:
I agree that adding the non-ipd alias in the STD list would be sufficient to track the latest versions (which I think is the goal of "STD" as the algorothms there are not in final-standard phase yet but rather "selected to be standardized and in their latest draft phase"). With #1703 / #1704 addressed, the latest version (-ipd or -ipdv2...) macros would then automatically be enabled as well, and explicitly adding -ipd to the STD list will not be needed anymore.

Issue #1474, opened 2023-05-19
Multithreading tests


baentsch:
AFAIK `liboqs` only has multi-processing tests, but no multi-threading tests. Please point to such tests if I'm wrong.
Otherwise this issue is to suggest adding such tests to catch problems such as https://github.com/open-quantum-safe/liboqs/issues/1471 in CI.

sreenath-tm:
Hi @baentsch is this issue still under consideration and is there any workaround where we can run Liboqs in a multi threaded manner?

dstebila:
> is this issue still under consideration and is there any workaround where we can run Liboqs in a multi threaded manner?

We don't have any further progress on this issue, but would welcome a contribution if you are interested in working on it.

Issue #1408, opened 2023-02-26
Test all scripts


baentsch:
This issue is to suggest adding tests to (automatically) exercise all scripts within this project. 

The goals would be 
- to ensure no "local setup" problems will hit anyone (new) using any `liboqs` script
- scripts are resilient (e.g., idempotent, not changing code on a given `git` state)
- script run results are reproducible (also in CI)

Proposed list:

- [x] ./tests/test_code_conventions.py
- [x] ./tests/test_distbuild.py
- [x] ./tests/test_leaks.py
- [x] ./tests/test_binary.py
- [x] ./tests/test_speed.py
- [x] ./tests/test_constant_time.py
- [x] ./tests/test_kat.py
- [x] ./tests/helpers.py
- [x] ./tests/test_alg_info.py
- [x] ./tests/test_hash.py
- [x] ./tests/test_mem.py
- [x] ./tests/test_cmdline.py
- [ ] ./scripts/update_docs_from_yaml.py
- [ ] ./scripts/genkatdict.py
- [ ] ./scripts/doxyfy.py
- [ ] ./scripts/update_cbom.py
- [ ] ./scripts/format_docs_yaml.py
- [ ] ./scripts/noregress.py
- [ ] ./scripts/copy_from_upstream/copy_from_upstream.py
- [ ] ./scripts/copy_from_upstream/update_upstream_alg_docs.py
- [x] ./tests/run_astyle.sh
- [x] ./tests/test_spdx.sh
- [ ] ./scripts/build-android.sh
- [ ] ./scripts/git_commit.sh
- [ ] ./scripts/copy_from_xkcp/checkout.sh
- [ ] ./scripts/copy_from_xkcp/update_patches.sh
- [ ] ./scripts/copy_from_xkcp/package.sh
- [ ] ./scripts/run_doxygen.sh
- [ ] ./scripts/validate_cbom.sh
- [ ] ./scripts/genkatsha256.sh
- [ ] ./scripts/noregress.sh

Further additions welcome as are suggestions to prune scripts not deemed necessary any more. Scripts run in CI already checked off.

vikasmulaje:
Hi @baentsch is this still open for work?

baentsch:
> Hi [@baentsch](https://github.com/baentsch) is this still open for work?

It sure is and it'd be great if this improvement gets done @vikasmulaje . Tagging @xuganyu96 for his take regarding the list above: Worth while pruning / extending given how old this issue is?

[...]

Issue #1215, opened 2022-05-25
Add fuzzing testing


baentsch:
Follow the approach taken by [OpenSSL](https://github.com/openssl/openssl/blob/master/fuzz/README.md) or another one: Suggestions welcome below.

silvergasp:
I'd like to suggest and champion an effort integrating liboqs with [google/oss-fuzz](https://github.com/google/oss-fuzz). If you aren't familiar with it, Google offers a free (for open source) continuous fuzzing service called OSS-fuzz.

I've opened up a draft pull request to add a super basic fuzz-testing harness here https://github.com/open-quantum-safe/liboqs/pull/1905. It needs a little bit of tidying before it's ready to go but I thought I'd guage interest before polishing it up.

The general process would look something like this.
- Merge #1905
- Apply for integration into oss-fuzz, this is a pretty simple PR into the repo that I can do on your behalf. All I'd need is a comment with approval from someone with write access to this repo.
- Integrate the project, this includes a Dockerfile and a shell script to build liboqs in the oss-fuzz environment.
- OSS-fuzz integration would be complete and the fuzzer would run every night for a few hours on a distributed cluster.
- Integrate clusterfuzzlite, this would run a stripped down version of oss-fuzz in github actions for X minutes (usually 10mins, but this is configurable) on every PR. The motivation behind this differs slightly in that it's intended to catch shallow bugs before they are merged. In comparison OSS-fuzz will run for significantly longer each night to attempt to find those harder to reach bugs (if they exist).

Let me know what your thoughts are on this :)

baentsch:
> I'd like to suggest and champion an effort integrating liboqs with [google/oss-fuzz](https://github.com/google/oss-fuzz).

This would be _very_ welcome, @silvergasp ! Thanks a million for the suggestion and apparent commitment!! As you seem to be an Independent Contributor like me (trying to establish that notion towards the corporate/LF folks :) I shall provide any possibly needed assistance with this, e.g., helping #1905 move to merge-ability, so please be sure to tag me when needed.

> Let me know what your thoughts are on this :)

In a nutshell: LGTM :) Details to follow once this moves forward, I guess.

silvergasp:
[...]

Issue #167, opened 2017-08-14
Code coverage


tlepoint:
Add code coverage using coveralls

aidenfoxivey:
In all honesty, I don't have a tonne of experience with setting up code coverage in non-Golang environments.

I think that https://coveralls.io looks very nice and has a good interface. That said, I'm always a little bit wary to adopt SaaS when there could be a more "long lived" solution. (for example, choosing `astyle` over `clang-format` despite `astyle` existing on Sourceforge)

I was recommended `gcov` as well. I'm probably going to read the manual for `gcov` and then give it a go. If anyone has firm opinions and/or recommendations, I would be very interested to hear them.

Seems like a lot of C test coverage tools are proprietary.

aidenfoxivey:
Well, this is a bit later...

![Image](https://github.com/user-attachments/assets/9eae4d72-33dd-4974-9225-2bc8ab4bf204)

aidenfoxivey:
I'm not 100% sure this actually makes 100% sense, so please do check me if something this diagnostic is showing seems inaccurate.

It seems like coveralls has a nice interface, but it's also just using gcov underneath.

So much for a diverse ecosystem of C code coverage tools! :)

dstebila:
Thanks for revisiting this. Is this output for a single execution of test_kem for a BIKE algorithm?  Is it able to aggregate coverage across multiple runs?

aidenfoxivey:
![Image](https://github.com/user-attachments/assets/0804dbaa-da7e-4f38-9e14-edad4fbd290f)

Yeah it appears to be able to do a decent job of aggregating across multiple tests being run. I just need to figure out how to run all the tests haha.

aidenfoxivey:
![Image](https://github.com/user-attachments/assets/02dbefc9-2af4-4653-b12c-66cf5514f6b6)

also works for AES.

aidenfoxivey:
Seems to use pytest if I'm not wrong.

aidenfoxivey:
![Image](https://github.com/user-attachments/assets/61ab35ac-1722-43e1-8da2-9057084605dd)

aidenfoxivey:
Alright, looks like this is the approximate flow:

```
# build with --coverage flag
cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_FLAGS="--coverage" -DCMAKE_CXX_FLAGS="--coverage" ..

# run files
<trying to sort this bit out>

[...]

PR #1708, opened 2024-02-27
Create scorecard.yml (OpenSSF)

### LINKED ISSUE #1706: Add OpenSSF scorecard
I suggest we generate [OpenSSF Scorecards](https://securityscorecards.dev/) 

We are offering assets in the security space. In addition to assurance of those assets in algorithmic terms & testing, there are [additional criteria](https://securityscorecards.dev/#the-checks) relating to the management of the project in github, packaging, dependencies, workflows, contributor diversity, and use of various tools.

scorecards are becoming more discussed as we all worry about supply-chain security, and some organizations are using them as criteria as to which projects can be used. 

The tests can be done automatically in a [github action](https://securityscorecards.dev/#using-the-github-action) to at least generate a local report - can consider later how to share further.  

I think by doing this we add credibility (including in being more closely aligned to openssf) - even though initially we will likely fail on multiple criteria, but it gives us a best-practice list to work to. 

Note - See also https://github.com/pq-code-package/tsc/issues/14, but since we have code out there with open-quantum-safe, how about adding it here? If useful, we can consider extending it to other relevant repos - oqs-provider comes to mind, perhaps others/all. I'd be happy to propose a PR/take this on, and it should have very little risk/impact

planetf1:
Adds a github action to run the OpenSSF scorecard tool, and posts results to security->code scanning

Fixes #1706

This is a new CI action. It's only effect is
 - small amount of resource to run the workflow
 - creates items under Security->code scanning, visible to those with 'write' access in the repo

No branch protection rules are being changed to require successful completion of this action
 
 No additional testing is required
 
 Example output is shown in the referenced issue
 

* [NO] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)

As per https://github.com/ossf/scorecard it's possible some additional permissions may be needed (depending on branch rotection setup).  These can be dealt with as a second iteration

This PR also does not attempt to address any findings from the scorecard scan. It's purely activating the tool so that we can review and act on those findings subsequently.

Draft for
 - testing
 - minor changes/comments cleanup


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,72 @@
...
+          # repo_token: ${{ secrets.SCORECARD_TOKEN }}
+
+          # Public repositories:
+          #   - Publish results to OpenSSF REST API for easy access by consumers
+          #   - Allows the repository to include the Scorecard badge.
+          #   - See https://github.com/ossf/scorecard-action#publishing-results.
+          # For private repositories:
+          #   - `publish_results` will always be set to `false`, regardless
+          #     of the value entered here.
+          publish_results: true
```
baentsch, line 57: I don't think it's prudent to publish results before the team had a chance to analyze/fix.

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,72 @@
...
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: '29 9 * * 4'
```
baentsch, line 13: Hmm -- this means the scan isn't activated even in this PR, so we don't get to see results, right? Suggest to change temporarily to get such a report. Prior to merge it seems prudent to add it to the [weekly scan](https://github.com/open-quantum-safe/liboqs/blob/main/.github/workflows/weekly.yml).

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,72 @@
...
+      # format to the repository Actions tab.
+      - name: "Upload artifact"
+        uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # v3.1.0
+        with:
+          name: SARIF file
+          path: results.sarif
+          retention-days: 5
+
+      # Upload the results to GitHub's code scanning dashboard.
+      - name: "Upload to code-scanning"
```
baentsch, line 69: As discussed before, suggest to enable only after initial results in CI logs have been created and acted upon.

.github/workflows/scorecard.yml
planetf1: That makes sense. Will update to false.

.github/workflows/scorecard.yml
planetf1: By default no, not in the PR, in part because the action publishes to the code scanning results for the 'definitive' code in main.

The serif output file is also generated.

The action is not run at all for PRs. I could also trigger on PR and just generate the serif which we can just read. That could be just for PR verification, or on every PR. I'll assume the latter, we can remove before merging if required.

As an example of running the action on a merge commit see issue or https://github.com/planetf1/liboqs/actions/runs/8063103943

.github/workflows/scorecard.yml
planetf1: There are three different ways the results are persisted
a) The serif file - this is entirely local to the action run ie attached. 
b) The 'publish_results' above. This makes the results available via a REST API and I believe forms part of the scan data for many projects which have scorecards. So this is very much external to the project
c) Uploading (here) to code scanning results. This is the step that makes the results visible to maintainers (or those with 'write' access) on the 'security' tab.

As above I completely agree with b). Happy to go with what is felt to be right for c. -- it does make the results easier to see for the dev team, but can go with just a) initially (I'll comment out the other steps) if we just want to review the serif output to get a feel of the scope of what is being suggested.

planetf1:
@baentsch many thanks for the comments!

planetf1:
PR updated - only generates .sarif file, and runs on both PRs and merges to main
 - can review the sarif file and decide if we want to push to code scanning results in this repo
 - I can squash the commits when we're ready for final review if that's preferred

planetf1:
Example sarif file can be found as attachment on https://github.com/open-quantum-safe/liboqs/actions/runs/8066088271?pr=1708

planetf1:
In the scan above, we have two categories of issues reported:
* [Token Permissions](https://github.com/ossf/scorecard/blob/376f465c111c39c6a5ad7408e8896cd790cb5219/docs/checks.md#token-permissions) - recommending some refinements to permissions for the github actions

To fix these carries some risk of trial/error, but ultimately should not change results, it's just being more explicit about what we expect an action to do

* [Pinned Dependencies](https://github.com/ossf/scorecard/blob/376f465c111c39c6a5ad7408e8896cd790cb5219/docs/checks.md#pinned-dependencies) - recommending pinning the versions of actions we use by hash. 

This requires making the version of github actions we use explicit, so that a release can always be regenerated. The downside is maintenance to update even within minor versions. I see the reasoning for it, it may work better alongside an automated tool like dependabot (I know it works on major minor versions, not sure on commit hash)

I didn't notice other issues in the sarif (viewed with the sarif viewer in vscode, from ms dev labs)

.github/workflows/scorecard.yml
baentsch: > The serif output file is also generated.

With the latest commit it now is, thanks!

> we can remove before merging if required.

That depends on what this tool finds: Does it have any relationship whatsoever to the code (within a PR) or only procedural issues in the github setup? If the latter, it doesn't make any sense to run it on PRs.

.github/workflows/scorecard.yml
baentsch: > if we just want to review the serif output to get a feel of the scope of what is being suggested.

I personally would prefer that: From the below, it seems this is solely project setup matters that this tool finds: Right or wrong?

planetf1: Right

... Mostly ie obviously some of that setup is in code, and affected by commits

.github/workflows/scorecard.yml
planetf1: I can update the PR to be executed on merge only, and to just generate the serif for now. I'll leave the option to a) upload to scanning results b) post to openssf commented out.

We hopefully can then merge the PR.

I'll leave the issue open to track that followup discussion. I think the desired end goal would be to uncomment the two additional steps, though a final decision that we don't may happen too!

baentsch: > I can update the PR to be executed on merge only

I still don't quite understand why this should run during a PR/merge process: This seems to check (only) the "procedural" github setup not exactly influenced by code. Right or wrong? If right, I'd suggest running this only during the weekly tests (as well as at release). If wrong, I'd be grateful for an example of (code) problems this tool can point out and that a contributor can then fix during the PR process. Running only at merge seems wrong (either too late or too often).

> We hopefully can then merge the PR.

I'd prefer merging this only when we don't have a "red status" right away, i.e., the issues this highlights got fixed: This surely should not stall the next release planned for early next week.

planetf1: By 'merge' I mean when code is merged into main, by the process of merging the PR.
The actual workflow trigger is on a push to main
Yes, these are basically procedural checks

The intent is simply to record the state of the 'current' code.

We could run weekly - though the resource taken by this check is very low in terms of github actions. 

Scorecard is an OpenSSF initiative. It's just a checklist to provide some assurance that the project is adopting good practice. In some ways it's like having a badge or certificate for what you've learnt. You already know it, but want to demonstrate that in a standard way. 

This change is absolutely not relevant for the release  next week. I think the PR should wait until after that in any case

The results (if we go with the change here) are only published to the serif file attached to the workflow, nowhere else. But if it makes sense to wait until the underlying issues are fixed, I could look at PRs for those, and leave this in draft.

planetf1:
Rebased, to review current findings and correct.

planetf1:
Update on this PR - good things
 - added pinned SHAs for github actions - this seems to work, and pass the scorecard tests
  - added explicit permissions for github actions - this again seems to work & pass tests
  - updated the scorecard yaml itself as I noticed 2 deprecations (one was in checkout... looking at that across the repo should be a distinct action)
  
  There are however still a few errors relating to freezing of pip dependencies
   - a build failure due to some dependencies in copy_from_upstream being pinned, but not all (? may need recursive definition?)
    - a warning from scorecard where the unix build explicitly does a 'pip install'. this needs to add SHAs (I'd only done requirements.txt)
     - a warning from scorecard with requirements.txt explicit SHAs' Either a bug, or may relate to the build failure - ie I only pinned what was specified, but not the full resolved list of dependencies.....

planetf1:
Pip requirements.txt install was failing in the build since once shas are used, ALL dependencies need to be listed (resursively) and a few were missing. 

fixed by
* installing the existing dependencies into a clear virtual environment (running `python -m venv /tmp/env2; . /tmp/env2/bin/activate`; pip install -r requirements.txt`)
* freezing the dependencies to get a seed list `pip freeze > requirements.txt2`
* running the following script to get the actual shas
```
cat requirements.txt | while read line                                                                                                                                            <<<
do
hashin $line
done
```
* finally discarding the original & using the new requirements.txt2

Also tested loading the new requirements.txt into a clean pip venv which works fine.

planetf1:
@baentsch the checks are nearly clean, with the only OSSF negative points (score 8/10) caused by:
```
        run: env HOMEBREW_NO_AUTO_UPDATE=1 brew install ninja && pip3 install --break-system-packages pytest pytest-xdist pyyaml
```
We have the same in Windows. It's not detected, but should fix in the same way

Aside: ossf scorecard doesn't know/check brew packages, which of course can also change the results!

To address the ossf report, we would need to specify hashes. I did this for our import script (see above) which uses requirements.txt

I'm think this is the cleanest way to do here too - and in future may be easier to maintain if we use dependabot?

I could create the required 'requirements.txt' somewhere and refactor the above. The big question would be where in the source tree. it could go in the .github/workflows directory, but is this clutter? somewhere else? What do you think?

planetf1:
The mac builds are using Python 3.12.x whilst the Windows build is using 3.9.x -- this is why the dependencies resolve differently. These versions are the standard, supplied, python versions for the respective runners

I initially used a common file, then a slight variation from a test machine running windows but with python 3.12

Additional consideration - for consistent results, having a managed, specific version of python may be advisable.

planetf1:
Changed approach to build requirements txt to
```
pip install pip-tools
pip-compile requirements.txt --generate-hashes -o requirements_new.txt
```

This is done in a clean venv, and has been tested on macos (3.12) and windows (3.9)

planetf1:
This is now ready for review:
 * Scorecard results are clean with all reported warnings clean
 * results are only reported in sarif file attached to action
 
 Follow-on activity (after merge)
  * Open a new PR to reset the schedule (ie 'on:' to OSSF recommendations) - but get the main PR in first.
  * Add information about scorecard & tools to aid in pinning dependencies to documentation (discuss where)
  * report-back to dev call on change
  * roll out similar changes to other repositories

ryjones:
this is really cool! thanks for doing the heavy lift

scripts/copy_from_upstream/requirements.txt
```diff
@@ -1,12 +1,153 @@
...
-markdown-it-py==2.2.0
-MarkupSafe==2.1.3
-mdit-py-plugins==0.3.4
-PyYAML==6.0.1
-tabulate==0.8.10
-typing-extensions==3.7.4.3
-wget==3.2
-zipp==3.4.0
+attrs==20.3.0 \
+    --hash=sha256:31b2eced602aa8423c2aea9c76a724617ed67cf9513173fd3a4f03e3a929c7e6 \
```
baentsch, line 2: Please add documentation how to generate these hashes (and how to handle any errors they may trigger).

.github/workflows/requirements-3-12.txt
```diff
@@ -0,0 +1,97 @@
+#
+# This file is autogenerated by pip-compile with Python 3.12
```
baentsch, line 2: This documentation is good(ness). However, I'm not impressed by the apparent need to generate separate files for different python versions: Do I understand this right? If so, this seems to add an undue maintenance burden: Which requirements.txt file is master? How to generate the different files? How many are required? One for each python version? In other words, will this break once CI upgrades python? How to handle any errors the hashes may generate?

baentsch:
> This is now ready for review:
> 
>     * Scorecard results are clean with all reported warnings clean

Very good, thanks. Also good to know that no software changes are/were required to get to this "green state".

> 
>     * results are only reported in sarif file attached to action
> 
> 
> Follow-on activity (after merge)
> 
>     * Open a new PR to reset the schedule (ie 'on:' to OSSF recommendations) - but get the main PR in first.

Before doing that, I'd like to 
- get confirmation about which (downstream) project or user needs this; this would not be required if you can guarantee that this tooling will never generate alarms that may cause work to us (i.e., be 100% automated). Looking at the current changes I have doubts about this (i.e., that this has the potential to require manual interventions) but would like to hear your thoughts on this, @planetf1 
- how to ascertain that any warnings this tool may generate will not be considered blockers for use or progress, i.e.,
- how people can themselves fix any errors this tooling may highlight (aka documentation; see below)

> 
>     * Add information about scorecard & tools to aid in pinning dependencies to documentation (discuss where)

Such documentation seems prudent to have in this PR already (also see my single questions) to avoid a situation in which things fail for users and they don't know where to look for documentation to "dig themselves" out of the proverbial hole this may have plunged them into. My suggestion would be to place this in a file "PROCEDURES.md" in the "docs" directory. It already should contain three sub items:
- DCO (why needed/benefits, e.g., with reference to project charter), e.g., to reference in [PRs failing because of this](https://github.com/open-quantum-safe/liboqs/pull/1758)
- CBOM (why needed/benefits with documentation how to handle CI errors if so triggered, avoiding future instances of [for example this](https://github.com/open-quantum-safe/liboqs/pull/1745#issuecomment-2046714322)).
- Scorecard (why needed/benefits with documentation how to handle CI or local execution errors if so triggered)

> 
>     * report-back to dev call on change

Good proposal, but this feature must be understandable (by reading documentation) by anyone not attending dev calls too.

> 
>     * roll out similar changes to other repositories

Suggest to wait with this until 
a) this feature is proven to not cause any procedural complications to technical progress (quick-and-easy PRs, etc.)
b) [we have finally decided which projects to carry forward at which support/maintenance level](https://github.com/open-quantum-safe/tsc/issues/2)

baentsch (changes requested):
Please add "self-help" documentation as per discussion item(s).

dstebila:
Thanks Nigel for getting this started!  And thanks Michael for a thorough analysis.  We will definitely need some hand-holding as this lands and the first few times it pops up during a PR.

scripts/copy_from_upstream/requirements.txt
planetf1: Agreed - I'll answer in more detail below

.github/workflows/requirements-3-12.txt
planetf1: thanks for the comments, and Yes, you do....

Once ANY hashes are used in the requirements.txt, all dependencies must be qualified by hashes, including the transitive dependencies. So those lists must be complete

Currently the 'unix' build (running on mac) uses Python 3.12 ie current, whilst the Windows (2022) environment is using Python 3.9. There are some dependencies we use that have slightly different dependencies themselves when being used on python 3.9 or 3.12

So this means two different files, which are likely (not guaranteed) to be correct for minor versions, and most probably often major versions

Currently there is no concept introduced of a 'master' dependency. If we wanted this, the clearest change would probably be to build using a specific version of python, and to ensure this is installed within the relevant build script - whether it's 3.9, 3.12, or something else since they can, and will vary

Whilst the python dependence could affect our direct dependencies, currently it seems it's only our transitive dependencies, so we could provide a script which builds the hash-inclusive requirements.txt from a less precise list. However we can't include this as part of the build process without going against the spirit of the OSSF checks, since the intent is to ensure there is always manual review when any change, and that the subsequent builds are repeatable. Any script would also need access to the relevant python version. It's not dissimilar to what I did to develop the change, and I totally agree it should form part of the documentation.

If CI upgrades python, yes, there is a chance of breakage. As above it's most likely on major version changes, and we know there is such a change between 3.9 & 3.12 since we have two distinct files. We could look into what version guarantees the CI images offer, but again the way to reduce dependence (and ensure more repeatability) would be to specify the version of all build tools being used, including python.

If the dependency chain is not correct, the python will fail, probably in the pip install step.

I think the best way to
 - reduce the changes of such changes breaking us
 - remove the need for two distinct requirements files (I hope, assuming there's no platform dependence...)
 
 is to use the [setup-python](https://github.com/actions/setup-python) github action, and explicitly specify python 3.12. **I could make a change to the build scripts for this and refactor?**

planetf1:
To answer the remaining comments:

* It's not possible to guarantee 100% that there will be no work from this change for the following reasons
 - If *we* want to add an additional python module, or change the versions we are using, we have to do more than list the dependency or the version, but now also need to add the sha, and list the entire set of dependencies after transitive resolution. Mitigation: document process
 - If the python version supplied by github were to upgrade python we could find a transitive dependency changes, leading to a build break until above was done. Mitigation: use the github-python plugin to use a relatively fixed version of python (much much smaller change with minor fixes) << as in earlier comment
 
 Both of the above seem reasonable to me given the intent of the scorecard recommended approach is to avoid unknown changes to what is being produced - for example not just a new, but a modified/hacked/illicit version of a library.
 
 * We have two things
  - a) The scorecard check - this is passive - it is only giving us a score, so will not progress, though we should monitor & plan action if we regress
  - b) the change to the pip install - this is within the PR - but see previous bullet point
  
* How people can fix? - documentation  

PROCEDURES.md - I can add changes there, but would suggest within this PR it focusses on scorecard, and we raise additional issue/issues to handle DCO and SBOM (which I can also help with). Agree with the point about making sure the change is understandable from the contents of the repo/docs (I could also add a pointer/info in the requirements file itself for extra clarity)

Rolling out to repo - makes sense we allow a little time (a few weeks?) after the change is merged into liboqs (suggest a checkpoint at the TSC?) before working on the other repos -- and of course they should each go through their own pr approval). As to which repos, ultimately any that has any supported code within.

planetf1:
So in summary my plan (if reviewers agree) is to:
 - [ ] Add setup-python into each build script (using 3.12)
 - [ ] remove/rename the additional requirements file
 - [ ] Add a docs/PROCEDURES.md describing how the requirements file can be built with hashes etc, and link back from requirements file or script as makes sense
 - [ ] Raise issue to document DCO process
 - [ ] Raise issue to document CBOM process
 
 Will return to TSC/rollout once this PR is merged.

baentsch:
Please go ahead with your plan, @planetf1 . But please also keep in mind that this must be as easy-to-use as possible for developers as well as users, e.g., with any `python` version and any CI system (and/or add suitable documentation catering for all possible problems you may already have heard about).

baentsch:
> We will definitely need some hand-holding as this lands and the first few times it pops up during a PR.

This should never pop up during a PR as I understand it, but it may pop up during development (prior to PR) and in weekly tests (if some hashes are wrong), right @planetf1 ?

The latter doesn't concern me as it's just "decoration" for now. The former is what I'd like to avoid at all cost: Making developers scratch their heads why the "requirements.txt" file fails to properly load or having to install a specific `python` version just to satisfy this tool: Things like this should be running in the background raising their head only when something's _really_ wrong and otherwise get out of the way of development progress.

Yes, this "progress-over-process" priority will change eventually as and when and if OQS takes on specific product properties. But we have still not decided which components and which properties those are, let alone their priorities. For example, I'd consider a serious code review for critical `liboqs` parts or #1540 or #1474 or #1366 or #167 or #1215 _much_ more urgent than this OSSF badge (don't get me wrong, @planetf1 , it's nice to have and addresses a concern, but it's ascertaining properties for sub-components of scripts [that we don't even have tests for](https://github.com/open-quantum-safe/liboqs/issues/1408).....)

I guess this argues for closure on https://github.com/open-quantum-safe/tsc/issues/1 before all else, @dstebila.

planetf1:
We use python dependencies in several areas during a pr & merge build. Specifically in both the unix (macos/linux) and windows builds:
 1. copy_from_upstream - this script uses python to copy some implementations across from pqclean etc. 
 2. for macOS/windows only, prior to running tests (see below)
 
 So the code is used then - but I think the important point is why anything would change ie break it. As I see it this would be
  - where a developer needs additional dependencies in these places (or others). They would need to add the right sha, and may need to recalculate the dependency chain. Docs should cover this
   - where a particular dependency gets removed, perhaps replaced. I think this is unlikely as usually libraries just add new versions (which we'd only use when we choose to)
   - where we decide to review/update our dependencies. either manually, or using a tool like dependabot - but it should just be part of the process, so answered by documentation
   
For the docs, making sure the developer knows where to look is important - I suggested above adding a pointer within the actual requirements.txt. Could also add a link within the relevant script, or even print output into the build logs if that would be useful?

There are lots of steps to move to a trusted production environment - this is just a very small step, and I completely agree those other issues need reviewing too, as does the decision around prod vs research. We can't do it all at once. I hope that this change also gives more confidence to other projects looking at oqs (and other pqca projects) as a source, that we have a good handle on supply chain security (I know there's many other parts, a lot already done, and this checklist is simplistic) - reputation.

We then have the pinning of github actions

We wouldn't expect versions to be withdrawn, so this is more a maint task when we want to upgrade our build action versions.

OSSF themselves say:

_Pinned dependencies reduce several security risks:

They ensure that checking and deployment are all done with the same software, reducing deployment risks, simplifying debugging, and enabling reproducibility.

They can help mitigate compromised dependencies from undermining the security of the project (in the case where you've evaluated the pinned dependency, you are confident it's not compromised, and a later version is released that is compromised).

They are one way to [counter dependency confusion (aka substitution) attacks](https://azure.microsoft.com/en-us/resources/3-ways-to-mitigate-risk-using-private-package-feeds/), in which an application uses multiple feeds to acquire software packages (a "hybrid configuration"), and attackers fool the user into using a malicious package via a feed that was not expected for that package.

However, pinning dependencies can inhibit software updates, either because of a security vulnerability or because the pinned version is compromised. Mitigate this risk by:

using automated tools to notify applications when their dependencies are outdated;
quickly updating applications that do pin dependencies._


The last section is relevant -- A follow-on action would ideally be to look at automatic dependency updates/recommendations. For example dependabot manages this well for github actions . It should also manage python. I'll add:
- [ ] open issue on dependency management tools

 
 Finally - I note the linux build uses the openquantumsafe/ci-debian-buster-amd64 image - where does this get created?

planetf1:
Added
 - [ ] open issue on ci image
 
 Perhaps we could build within our repos? Or maybe use a standard image? (I don't know enough about it to comment)

baentsch:
> Perhaps we could build within our repos? 

We do: https://github.com/open-quantum-safe/ci-containers

> Or maybe use a standard image?

That would most likely run counter to https://github.com/open-quantum-safe/tsc/issues/5

planetf1:
Added explicit python setup using 'setup-python' github action where used in workflow.

In testing this fix, I hit the doxygen deprecation issue which has been addressed in https://github.com/open-quantum-safe/liboqs/pull/1775/files

Will redo these changes back once that PR has been merged

ryjones:
[interesting](https://scorecard.dev/viewer/?uri=github.com/open-quantum-safe/liboqs)

baentsch:
> [interesting](https://scorecard.dev/viewer/?uri=github.com/open-quantum-safe/liboqs)

Thanks for this easy-to-digest pointer! Indeed, finally a reference maybe now triggering interest/motivation to work on #1215 and #1366. What is a bit surprising is the lack of a constant (run-)time checkbox in this list, though. The remaining checkboxes seem to be fixable by re-configuring the github setup, right?

planetf1:
> [interesting](https://scorecard.dev/viewer/?uri=github.com/open-quantum-safe/liboqs)

That scorecard was NOT generated by the definitions here. For example ours is still running the v4 scorecard, whilst that is for v5 (prerelease) where they have added more checks.

I suspect (but can't be sure) this was run by the openssf team directly, perhaps as a precursor to the PQCA tac meeting. Not something I was aware of. I know they do proactively run on 'top' projects, but when I started these changes we were not included. Note that much of this PR is the mitigation not the actual scheduling of the check.

There's now a few things to think of:

* We may wish to merge the changes to date so that we are in control of the runs & any reviewers can see gradually improving metrics (including some of the fixes here). Outstanding changes (v5 additions, docs) can be added through a further PR
* We should probably move to the v5 scorecard as soon as it is GA
* There will be more potential aspects to check - there won't be quick fixes for all of these, and even scorecard itself doesn't get 10/10

planetf1:
> > [interesting](https://scorecard.dev/viewer/?uri=github.com/open-quantum-safe/liboqs)
> 
> Thanks for this easy-to-digest pointer! Indeed, finally a reference maybe now triggering interest/motivation to work on #1215 and #1366. What is a bit surprising is the lack of a constant (run-)time checkbox in this list, though. The remaining checkboxes seem to be fixable by re-configuring the github setup, right?

Some, not all. Varying amounts of work. This PR already has fixes to some (pinned dependencies, token permissions) - albeit not verified with v5.

Others may require more work -- for example CII best Practices is a significant effort in it's own right

ryjones:
yes, OpenSSF scans the top million repos automatically, which is why it's there.

planetf1:
Added PROCEDURES.md & a template to document & make it easier to maintain pinned versions

.github/workflows/scorecard.yml
planetf1: The checks appear to be running clean on the v4 scorecard (as coded).

Please refer to the sarif results generated by the action.

An upgrade to scorecard v5 can be done as a new PR.

I will end off this PR update with a change to move the scans to release, once approval is secured. Closing this conversation.

scripts/copy_from_upstream/requirements.txt
planetf1: Now added in PROCEDURES.md

.github/workflows/requirements-3-12.txt
planetf1: Scripts are changed to use python 3.12

planetf1:
Hopefully I've addressed the review issues. No findings are reported. I propose
- [ ] This PR is merged (once final review is completed). This will continue running the scorecard check within the PR and not publish. 
- [ ] A new PR will be opened to move the execution to a scheduled check as per openSSF recommendations, 
- [ ] A new PR will be opened to migrate from scorecard v4 to scorecard v5
- [ ] Further PRs will be opened if/when any fixes are needed.

Trying to maintain a PR with build changes (albeit minor) is a little tricky in merging as there is other activity in this area. I hope the changes are sufficient know to close this one out. In addition proactive scans are already being done by the openssf team, and getting our mitigations in place will improve our already public score.

planetf1:
Seeing one unresolved build error
Screenshot 2024-06-17 at 20 10 39

The referenced requirements.txt appears to correctly specify all hashes, and runs ok locally......

dstebila (approved):

ryjones:
@planetf1 could you do a squash commit or similar?

docs/PROCEDURES.md
```diff
@@ -0,0 +1,69 @@
+# Additional procedures for code maintenance 
+
+## Managing pinned dependencies
+
+The OpenSSF, via the [scorecard](https://securityscorecards.dev/) project recommends that projects pin any
+dependencies they use:
+
+* to ensure reproducibility
+* to reduce the change for rogue dependency updates to compromise software
```
baentsch, line 9: Don't understand the term "reduce the change": Do you mean "reduce the risk"?

docs/PROCEDURES.md
```diff
@@ -0,0 +1,69 @@
...
+
+It's important to note that this requires any changes to dependencies are properly reviews, and
```
baentsch, line 11: typo: reviews -> reviewed, right? What (manual steps) does this review entail (beyond usual PR-change-of-code reviews)?

docs/PROCEDURES.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Python dependencies
+
+Python dependencies used in the build process should be pinned to a specific version to ensure reproducibility.
+
+This is done by:
+ * Using the `--require-hashes` option on any `pip install` command line
+ * Adding the required hash in the `requirements.txt`
+
+ Currently this is used withou `.github/workflows` but the same principle applies elsewhere.
```
baentsch, line 22: typo: "withou" -> "within" (?) or "throughout"?

docs/PROCEDURES.md
```diff
@@ -0,0 +1,69 @@
...
+
+ To make this easier, a version of the `requirements.txt` without hashes has been saved as `requirements.in`. This is
+ to make maintenance easier, but it is not used at script execution time.
```
baentsch, line 25: So when updating a dependency version, one first has to update "requirements.in", then run the tool, right?

docs/PROCEDURES.md
```diff
@@ -0,0 +1,69 @@
...
+ The `pip-compile` tool must be installed via the [pip-tools](https://pypi.org/project/pip-tools/) package.
+ 
+ To add a new, or changed dependency:
+ ```
+ pip-compile --generate-hashes --output-file=requirements.txt requirements.in
+ ```
+
+ This will update requirements.txt with the correct hashes.
+
+ Correction function should be verified, and then both `requirements.txt` and `requirements.in` checked in.
```
baentsch, line 36: I do not understand this sentence: What is a "Correction function" and what does one (contributor and/or reviewer) have to do to verify it?

docs/PROCEDURES.md
```diff
@@ -0,0 +1,69 @@
...
+The exact hash specified after `@` is the git commit hash within the repo where the action is found.
+
+The [pin github action](https://github.com/mheap/pin-github-action) tool can be used to maintain these
+by, for example, running:
+
+```shell
+pin-github-action unix.yml
+```
+
+This will update the hashes with the latest version if no pinned value is set.
```
baentsch, line 57: For completion of the if condition, may be worth while adding "This tool does nothing if a pinned value has been set."?

docs/PROCEDURES.md
```diff
@@ -0,0 +1,69 @@
...
+For major updates, update the comment ie `pin@v4` to `pin@v5` and the tool will attempt to find the new hash.
+
+To help in explanation here's an example of a similar code fragment between tool executions:
+
+* Original entry is `uses: actions/checkout@v3`
+* run `pin-github-action unix.yml`
+* We now see `uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # pin@v3`
+* later we want to go to v4, so update the text to `uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # pin@v4`
+* Now run `pin-github-action unix.yml` to correct the sha
+* File now shows `uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # pin@v4`
```
baentsch, line 68: Ah, now I get it: The tool relies on the comments it generated in previous runs, moving the original version string in the code ("v3" in the example) to a comment, so one now has to edit the comments and then run the tool if one wants to upgrade what's actually executed, right?

scripts/copy_from_upstream/requirements.txt
baentsch: Thanks for the documentation! Our etiquette so far was that comments are resolved by the people that raised them. Please let me know if that doesn't suit you.

baentsch (commented):
Thanks for the documentation regarding version pinning.

Good to know there's different tools that update these hashes. That in my mind begs the question what manual step is required (by committers as well as reviewers) to ascertain that the tool did the right thing/obtained a "proper" hash? Without this manual step, one could obviously simply run the tool "blindly" on the (original) "requirements.in" or (version comments in the) CI job .yml files, much like an automated preprocessing step that could be integrated to CI... Or is the recommendation to just check _whether_ the hashes change in a PR?

docs/PROCEDURES.md
planetf1: I did mean that - thanks for spotting.

docs/PROCEDURES.md
planetf1: The usual with any code change - assess the impact. so if there's an update to a dependency try and test it locally, check the release notes.

docs/PROCEDURES.md
planetf1: corrected

docs/PROCEDURES.md
planetf1: yes.

docs/PROCEDURES.md
planetf1: A typo. It should say 'correct functionality'

docs/PROCEDURES.md
planetf1: If a commented value like 'pin@v4' is set, then the tool will update hash to the latest available v4. If it's set to an explicit SHA then it will leave as is. I think the extra info may confuse the explanation. I've reworded slightly, and added another reference to the doc.

docs/PROCEDURES.md
planetf1: yes, that's right.

We may start with 'actions/myaction@2.1'
Then run the tool and get `actions/myaction@a1b2c3d4 # pin@2.1' (etc)
Then run the tool again and get `actions/myaction@e5f6a1 # pin@2.1' if there's a new update that is still 2.1
But to go to 2.2, I'd update the 'pin@2.1' comment to 'pin@2.2'

This tool felt reasonable to use when I was making the changes, and reduces the effort.

planetf1:
> Thanks for the documentation regarding version pinning.
> 
> Good to know there's different tools that update these hashes. That in my mind begs the question what manual step is required (by committers as well as reviewers) to ascertain that the tool did the right thing/obtained a "proper" hash? Without this manual step, one could obviously simply run the tool "blindly" on the (original) "requirements.in" or (version comments in the) CI job .yml files, much like an automated preprocessing step that could be integrated to CI... Or is the recommendation to just check _whether_ the hashes change in a PR?

The reason for suggesting it's manual is that the author is then likely to be closely checking the action logs and output, plus be aware if any failures occur.

The exact hash could be searched for in github to match the version used.

I've added a note about this in PROCEDURES.md

docs/PROCEDURES.md
baentsch: OK -- then why not make this explicit in the documentation? Along the lines "When updating dependencies, a contributor should first update 'requirements.in', then run the pip-tools as per the below, then check for correct operation and only then commit."

docs/PROCEDURES.md
baentsch: It sure does. But again, for the unwary user it may be better to state explicitly that the comments _must_ be preserved for the tool to work again in case of an update (and that the version/reference update _must_ happen in the comment, not in the code).

What surprises me, though, is that it is possible that the hash may change upon re-running the tool: How/why can this happen? Wouldn't this be a tell-tale sign of someone tampering with the versioned/referenced action?

docs/PROCEDURES.md
baentsch: OK. But the typo still remains.

planetf1: fixed (in next push)

docs/PROCEDURES.md
planetf1: Reworded in a more prescriptive way. Hopefully this is clearer. (in next push)

docs/PROCEDURES.md
planetf1: Updated comments for first para.

On the second topic, agree, it would be something to look into. Normally a version won't be republished, but it could be. If the hash changes, it's different code. Any change would show up on a PR diff, and as part of a review it would be an appropriate question to ask.

without the hash mechanism in place we would silently consume the modified code in our build process... this way we at least go through a change review process.

planetf1:
Updated docs, rebased - @baentsch please check the updates are ok for you.

baentsch (approved):

baentsch:
@planetf1 Our etiquette is that the person who created a PR also merges it, squashing all commits into one with a reasonably small set of commit messages to let posterity know what this is _really_ about (e.g., removing all commit messages related only to PR discussions).

planetf1:
Thanks, that's helpful and what I expected. I'll sort the squash out soon (juggling tasks!) and merge

planetf1:
Squashed to one commit following successful review

@baentsch @dstebila I do not have permissions to merge. Once tests complete would you be able to do this?

ryjones:
@planetf1 do you want me to merge it?

baentsch:
Merged (shortened title a bit, moving text to contents). Thanks for the contribution @planetf1 !

PR #1709, opened 2024-02-27
Move MacOS CI tests to GitHub Actions; add M1 CI tests

### LINKED ISSUE #1630: Improve CI for macOS on Apple Silicon

SWilson4:
This PR moves our macOS CI tests from CircleCI to GitHub Actions. It also takes advantage of the [new M1 runner](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/). I believe this promotes macOS on aarch64 to a ["Tier 1" platform](https://github.com/open-quantum-safe/liboqs/blob/bb23b3f61b4c2cba00ba80980216b4584b345c9f/PLATFORMS.md#tier-1).

Fixes #1630.


.github/workflows/unix.yml
```diff
@@ -181,3 +181,33 @@ jobs:
       - name: Build
         run: ninja
         working-directory: build
+
+  macos:
+    needs: buildcheck
+    strategy:
+      fail-fast: false
+      matrix:
+        os: # macos-14 runs on aarch64; the others run on x64
```
Martyrshot, line 190: Can we also run macos-14 tests on x86? If it is possible I think we should since apple still supports MacOS 14 on x86.

Martyrshot (approved):
Nice PR! Just a quick question about macos x86. :)

baentsch:
> Nice PR! Just a quick question about macos x86. :)

Well, what about (Apple on) PowerPC? :-)

baentsch (approved):
Thanks very much! LGTM

Martyrshot:
> > Nice PR! Just a quick question about macos x86. :)
> 
> Well, what about (Apple on) PowerPC? :-)

X86_64 is still supported by all MacOS releases. PowerPC is not. :)

.github/workflows/unix.yml
Martyrshot: Fair enough! Thanks for checking. :)

Issue #1608, opened 2023-11-09
Support multiple Falcon signature formats


SWilson4:
The current (20211101) version of Falcon supports three different signature formats. Currently, we intend to support the "padded" format, which has fixed-length signatures. (As of writing, we don't actually do that and neither does PQClean: see #1561 and https://github.com/PQClean/PQClean/issues/523.) It would be nice to support both "compressed" and "padded" formats in liboqs, and possibly the "constant time" format as well.

More detail on formatting from the [Falcon reference implementation](https://falcon-sign.info/impl/falcon.h.html):

>  127  * There are three formats for signatures:
  128  *
  129  *   - COMPRESSED: this is the default format, which yields the shortest
  130  *     signatures on average. However, the size is variable (see below)
  131  *     though within a limited range.
  132  *
  133  *   - PADDED: this is the compressed format, but with extra padding bytes
  134  *     to obtain a fixed size known at compile-time. The size depends only
  135  *     on the degree; the FALCON_SIG_PADDED_SIZE macro computes it. The
  136  *     signature process enforces that size by restarting the process
  137  *     until an appropriate size is obtained (such restarts are uncommon
  138  *     enough that the computational overhead is negligible).
  139  *
  140  *   - CT: this is a fixed-size format, which furthermore allows
  141  *     constant-time processing with regard to the signature value and
  142  *     message data. This is meant for uncommon situations in which
  143  *     the signed data is secret but of low entropy, and the public key
  144  *     is not actually public. The CT format is larger than the
  145  *     COMPRESSED and PADDED formats.

baentsch:
This is still tagged for 0.10.0 milestone -- but I think we should either close (if we don't intend to add CT format, too?) or move to another milestone (if we do). I'm fine either way, so please do as you see fit, @SWilson4 .

SWilson4:
Fixed (to the extent that we intend to support Falcon formats) by #1710.

PR #1710, opened 2024-02-28
Support Falcon PADDED format

### LINKED ISSUE #1561: Falcon-1024 KATs differ from upstream

SWilson4:
Support the fixed-size PADDED format for Falcon signatures and update the existing Falcon code (which intends to implements the COMPRESSED format) to bring it in line with the spec. Reenable Falcon-1024 in weekly extended KAT tests, as it should now pass.

A couple of points to make review simpler:
- The only differences between the COMPRESSED and PADDED code should be contained in `api.h` and `pqclean.c`. It's probably worth looking at a side-by-side diff of these: for example, `pqclean_falcon-512-padded_clean/pqclean.c` vs `pqclean_falcon-512_clean/pqclean.c`.
- There are no NIST KATs available for the PADDED variant, so I generated my own from the Falcon reference implementation: see https://github.com/SWilson4/falcon-padded-KATs.
- The corresponding upstream PR is PQClean/PQClean#530.

Fixes #1561 and #1608.


dstebila (approved):

SWilson4:
@Frauschi Can you please take a look at the [Zephyr test failure](https://github.com/open-quantum-safe/liboqs/actions/runs/8086936503/job/22097805629?pr=1710)? It looks like an out-of-storage error to me, and I'm unsure of the best way to resolve it.

SWilson4:
The `oqs-provider` release tests have passed on this branch and the appropriate tracker: https://github.com/open-quantum-safe/oqs-provider/actions/runs/8101766836/job/22142722419

README.md
```diff
@@ -58,7 +58,7 @@ The list below indicates all algorithms supported by liboqs, but not all those a
 
 <!--- OQS_TEMPLATE_FRAGMENT_LIST_SIGS_START -->
 - **CRYSTALS-Dilithium**: Dilithium2, Dilithium3, Dilithium5
-- **Falcon**: Falcon-512, Falcon-1024
+- **Falcon**: Falcon-512, Falcon-1024, Falcon-512 (PADDED), Falcon-1024 (PADDED)
```
baentsch, line 61: Is this naming prudent? It makes specifying this alg a bit of a hassle (escaping space and bracket if used on a command line; always applying CAPS lock). What about "Falcon-padded-512"/1024?

docs/algorithms/sig/falcon.md
```diff
@@ -7,16 +7,18 @@
...
   - **Implementation license (SPDX-Identifier)**: MIT
 
 
 ## Parameter set summary
 
-|  Parameter set  | Parameter set alias   | Security model   |   Claimed NIST Level |   Public key size (bytes) |   Secret key size (bytes) |   Signature size (bytes) |
-|:---------------:|:----------------------|:-----------------|---------------------:|--------------------------:|--------------------------:|-------------------------:|
-|   Falcon-512    | NA                    | EUF-CMA          |                    1 |                       897 |                      1281 |                      666 |
-|   Falcon-1024   | NA                    | EUF-CMA          |                    5 |                      1793 |                      2305 |                     1280 |
+|    Parameter set     | Parameter set alias   | Security model   |   Claimed NIST Level |   Public key size (bytes) |   Secret key size (bytes) |   Signature size (bytes) |
```
baentsch, line 16: This change confuses me: The length of the artifacts created by the (currently named) "Falcon" algs change while the (new, padded) algs take on those lengths: Is this intentional? Isn't it surprising then that KATs for the ("original") Falcon algs don't change?

baentsch:
> @Frauschi Can you please take a look at the [Zephyr test failure](https://github.com/open-quantum-safe/liboqs/actions/runs/8086936503/job/22097805629?pr=1710)? It looks like an out-of-storage error to me, and I'm unsure of the best way to resolve it. The same test fails on the current `main`, so it's not specific to this PR.

Indeed, what does "No space left on device" mean in this context? Also FYI, @Frauschi we're aiming of doing a `liboqs` release containing this PR early next week so resolving this quickly would be very helpful & appreciated.

src/sig/falcon/pqclean_falcon-1024-padded_aarch64/LICENSE
```diff
@@ -0,0 +1,57 @@
+This ARMv8 NEON implementation is provided under the Apache 2.0 license:
+
+/*
+ * =============================================================================
+ * Copyright (c) 2023 by Cryptographic Engineering Research Group (CERG)
+ * ECE Department, George Mason University
+ * Fairfax, VA, U.S.A.
+ * Author: Duc Tri Nguyen
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
baentsch, line 9: Hmm -- Falcon right now is listed as "MIT" (only) if I'm not mistaken(?). Doesn't this new LICENSE text (adding Apache2) warrant a change to this (at least in top level doc .yml/.md)? Are the SPDX statements in line?

Frauschi:
> > @Frauschi Can you please take a look at the [Zephyr test failure](https://github.com/open-quantum-safe/liboqs/actions/runs/8086936503/job/22097805629?pr=1710)? It looks like an out-of-storage error to me, and I'm unsure of the best way to resolve it. The same test fails on the current `main`, so it's not specific to this PR.
> 
> Indeed, what does "No space left on device" mean in this context? Also FYI, @Frauschi we're aiming of doing a `liboqs` release containing this PR early next week so resolving this quickly would be very helpful & appreciated.

I just created PR #1714 that should fix the failing Zephyr CI tests. The problem was the Zephyr installation that has been created within the test environment. A default Zephyr installation is several GBs in size, apparently overfilling the disk space of the Github runner. Hence, the "No space left on device".

docs/algorithms/sig/falcon.md
SWilson4: The only length which actually changes is the signature length. The "original" / "currently named" Falcon algorithm produces variable-length signatures, so this length is only an upper bound. We had (incorrectly, per the spec) been using the size of the fixed-length "PADDED" format as an upper bound previously.

The behaviour of the "original" / "currently named" algs _does_ change with regard to KATs. In fact, the behaviour changes precisely so that Falcon-1024 now produces the correct KATs. Previously, our implementation had been out of sync with the KATs: see #1561.

README.md
SWilson4: Good point regarding the space... I went with all caps because that's how the formats are referred to in the reference documentation (https://falcon-sign.info/impl/falcon.h.html, starting at line 120). There are no uses of `strcmp` in `liboqs/src`, only `strcasecmp`, so I don't think the capitalization poses an issue. (I do see some `strcmp`s in the KAT test helpers, which I introduced—oops.)

How about `Falcon-512-PADDED` / `Falcon-1024-PADDED`? This is in line with `FrodoKEM-640-SHAKE` / `FrodoKEM-640-AES` / `FrodoKEM-1344-SHAKE` / etc. I'd prefer to have the format modifier at the end, as this really is fundamentally the same algorithm as Falcon-512/1024.

This space/paren/caps convention has already made its way into PQClean, but I don't imagine it will be difficult to amend.

beldmit: Hmm... If you are looking for smth case-insensitive using `strcasecmp` inside liboqs, you should beware of the TR locale. `Dilithium` with its several `i`s is especially in danger. In OpenSSL we had to introduce `OPENSSL_strcasecmp` to deal with it.

src/sig/falcon/pqclean_falcon-1024-padded_aarch64/LICENSE
SWilson4: Good catch. The existing Falcon `aarch64` code is also by @ducnguyen-sb and also uses the Apache2 licence, so it should be updated as well.

ducnguyen-sb: Hi, 

I don't mind changing the License. I remember back when I develop aarch64, it was Apache 2.0, so I set mine as Apache 2.0. 
If you want MIT, I can make it instantly.

README.md
baentsch: > I went with all caps because that's how the formats are referred to in the reference documentation (https://falcon-sign.info/impl/falcon.h.html, starting at line 120)

This argument is somewhat weak as it otherwise begs the question why we don't have "COMPRESSED" in the (original alg's) name, too. And No: I don't argue for that!

> How about Falcon-512-PADDED / Falcon-1024-PADDED? This is in line with FrodoKEM-640-SHAKE / FrodoKEM-640-AES / FrodoKEM-1344-SHAKE / etc. I'd prefer to have the format modifier at the end, as this really is fundamentally the same algorithm as Falcon-512/1024.

Yup -- I always hated the Frodo naming :-) Superfluous "KEM" and (to me) unlogical: Most Algs have a "base name", possibly a "variant modifier" and then a "strength indicator". That said, Sphincs also deviates from this, so why not another one? :-/ My argument btw does not contradict yours: falcon remains the basic algorithm, so falcon-padded is just a variant -- in two strength versions.

docs/algorithms/sig/falcon.md
baentsch: > The behaviour of the "original" / "currently named" algs does change with regard to KATs.

Aha. That then has an implication for https://github.com/open-quantum-safe/oqs-provider/pull/362... (new OID necessary for Falcon-512/1024).

src/sig/falcon/pqclean_falcon-1024-padded_aarch64/LICENSE
baentsch: I honestly don't mind. Whatever is easiest to bring LICENSE, documentation and SPDX in sync.

README.md
SWilson4: > > I went with all caps because that's how the formats are referred to in the reference documentation (https://falcon-sign.info/impl/falcon.h.html, starting at line 120)
> 
> This argument is somewhat weak as it otherwise begs the question why we don't have "COMPRESSED" in the (original alg's) name, too. And No: I don't argue for that!
>

For argument's sake, "COMPRESSED" is the default format, so it's implied.

> > How about Falcon-512-PADDED / Falcon-1024-PADDED? This is in line with FrodoKEM-640-SHAKE / FrodoKEM-640-AES / FrodoKEM-1344-SHAKE / etc. I'd prefer to have the format modifier at the end, as this really is fundamentally the same algorithm as Falcon-512/1024.
> 
> Yup -- I always hated the Frodo naming :-) Superfluous "KEM" and (to me) unlogical: Most Algs have a "base name", possibly a "variant modifier" and then a "strength indicator". That said, Sphincs also deviates from this, so why not another one? :-/ My argument btw does not contradict yours: falcon remains the basic algorithm, so falcon-padded is just a variant -- in two strength versions.

The naming (presumably) has more of an impact downstream, and that's your department, so I'll concede. :stuck_out_tongue_winking_eye: `Falcon-padded-512` and `Falcon-padded-1024` it is. At any rate, this will be a moot point in the near future (hopefully) when Falcon becomes NTRU-DSA or whatever name NIST settles on for the standardized version.

SWilson4: @baentsch Following up on this after looking into patching in an updated name:

The `Falcon-* (PADDED)` name has already landed in PQClean. (I should probably have solicited feedback on that choice, earlier, mea culpa. :grimacing:) It should be an easy and quick fix to patch it (in `liboqs`) to `Falcon-*-padded` or `Falcon-*-PADDED` or what-have-you, in alg-param-variant order. However, switching to alg-variant-param will change the PQClean algorithm namespace from `PQCLEAN_FALCON512PADDED_*` to `PQCLEAN_FALCONPADDED512_*`. That's too big a change to fix easily with a patch, so it would require a PR upstream (one with a very large, albeit inconsequential diff) to change the name.

I'm OK with doing the work, but since we're aiming to cut a release candidate tomorrow, I thought I should check to see if it's worth potentially delaying that for naming conventions.

baentsch: Sorry in turn for having brought this up so late in the game. I might have compromised on the place where "padded" is placed, but the space and brackets are no-gos in my opinion: If we'd do a release with that choice, this will come and haunt us (as a release by definition is a commitment).

So indeed, two choices:
1) create a huge patch to enable running `copy-from-upstream` on the release with the "logical" name, removing that when upstream changed.
2) First fix upstream and delay release until that happened. 

1) is superfluous work which I'd hate; so how much time are we talking about on 2?

src/sig/falcon/pqclean_falcon-1024-padded_aarch64/LICENSE
SWilson4: I've updated the documentation to clarify that the `aarch64` implementation follows the Apache 2.0 licence.

README.md
SWilson4: Update: I just added said patch and re-ran `copy_from_upstream`.

baentsch: > The space and brackets are an easy fix; that doesn't change the namespace. I can add a very small patch to change Falcon-512 (PADDED) to Falcon-512-padded in the api.h and META.yml files, and nothing else will need to be done.

Thanks for that change. But this would still mean that we'd have to keep supporting an unlogical name :-( I still wouldn't if we don't release 0.10.0 right away. And [the tracking sheet](https://github.com/open-quantum-safe/liboqs/milestone/23) doesn't let me think we will...

If the upstream change is taking too long, what about also just changing "pretty_name_full" in [copy_from_upstream.yml](https://github.com/open-quantum-safe/liboqs/blob/b952262ad473cfb464cb6f300a80e9ac10fe96eb/scripts/copy_from_upstream/copy_from_upstream.yml#L231) (leaving "pqclean_scheme" unchanged until the naming gets "logic-fied" there too)?

SWilson4: I went ahead and made the upstream PR: https://github.com/PQClean/PQClean/pull/549. It hit a lot of files but was pretty quick with some help from `sed`. I'll wait for your comments there before hitting merge.

docs/algorithms/sig/falcon.md
SWilson4: > > The behaviour of the "original" / "currently named" algs does change with regard to KATs.
> 
> Aha. That then has an implication for [open-quantum-safe/oqs-provider#362](https://github.com/open-quantum-safe/oqs-provider/pull/362)... (new OID necessary for Falcon-512/1024).

The latest updates to https://github.com/open-quantum-safe/oqs-provider/pull/362 assign new OIDs for Falcon-512 and Falcon-1024. This will of course require another update once the new names are integrated, but from an OID standpoint I think it's OK now.

README.md
baentsch: > I'll wait for your comments there before hitting merge.
 
None (negative :) from my side. Worthwhile trying to run `copy_from_upstream` on that branch, just to test it here already? Would give someone with write access in PQClean time to review more deeply. Thanks for doing this, @SWilson4 -- I honestly didn't think my comment would touch upon so many files -- but as you said, `sed` to the rescue :-)

src/sig/falcon/pqclean_falcon-1024-padded_aarch64/LICENSE
baentsch: Just remembered that we don't expect the "upstream" files to have SPDX statements in their files. The above then resolves this comment completely. Thanks also for this, @SWilson4

docs/algorithms/sig/falcon.md
baentsch: > but from an OID standpoint I think it's OK now.

I tend to agree just looking at the YML. But when checking the resultant documentation I'm a bit confused (see https://github.com/open-quantum-safe/oqs-provider/pull/362#discussion_r1513177365). Is this the consequence of getting the KATs right -- did we indeed not change the underlying code? I thought we did -- wouldn't then a different algorithm version ID be prudent?

baentsch (approved):
Thanks for all this work, @SWilson4 !

scripts/copy_from_upstream/copy_from_upstream.yml
```diff
@@ -13,8 +13,8 @@ upstreams:
   -
     name: pqclean
     git_url: https://github.com/PQClean/PQClean.git
-    git_branch: master
-    git_commit: 0657749a785db30e7f49e9435452cb042edb1852
+    git_branch: sw-falcon-padded-rename
```
baentsch, line 16: Just as reminder: This needs changing before final merge.

PR #1712, opened 2024-02-29
Add return status for XMSS lock/unlock functions.

ducnguyen-sb:


baentsch (approved):
I'm a big fan of reporting errors for upper layers to give them a chance of handling them, so thanks for this improvement, @ducnguyen-sb .

PR #1714, opened 2024-03-01
Fix for the Zephyr CI tests

Frauschi:
The Zephyr Github CI Actions used a complete Zephyr installation to build and run the test code. The resulting size of the data has been up to several GBs, overfilling the Github runner storage space.

With this patch of the Zephyr workflow, only a minimal Zephyr installation is created to run the tests, resulting in a much smaller space footprint.

Furthermore, the timeout of the Signature test has been increased, as sometimes the tests take longer than the prior 10 minutes.

### DIFF
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -17,23 +17,35 @@ jobs:
             - zephyr-ref: v3.5.0
 
     steps:
-        - name: Init west workspace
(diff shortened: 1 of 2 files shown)


baentsch (approved):
LGTM. Thanks for the quick fix, @Frauschi ! @SWilson4 please take a look, approve & re-base #1710.

PR #1717, opened 2024-03-01
Proposed build wrapper script for static libs macOS / iOS / Android platforms

andybrucenet:
We wish to integrate kyber and other algorithms in our apps via oqs-provider - and this requires a built liboqs foreach of our cross-compiled solutions in addition to a built openssl.

The script has the following features:
* CMake output for all platforms (e.g. `build/apple/iphoneos/arm64` or `build/android/arm64-v8a`).
* The script can also generate `build/export` in a versioned format; example: `build/export/android/0.10.0-dev/armeabi-v7a/lib`.
* The export version is determined automatically from the cmake-generated headers.
* Because the cmake-generated headers are platform specific via `#defines` the output includes headers for each platform. Example: `build/export/apple/0.10.0-dev/macosx/arm64/include`
* Libraries are always built static and macOS / iOS libraries include all required architectures.
* Script can be invoked with any of the included functions; by default, builds all platforms and creates export output.

The output from the script can be used as-is with the analogous script for the `oqs-provider` project - there is a separate PR for that project.

No 

No new features added. The script requires access to a versioned openssl with subfolders to each platform.
Examples:
* Android - `[dir]/android/3.2.1` with subfolders similar to `arm64-v8a/{include,lib}`
* Apple - `[dir]/apple/3.2.1` with subfolders similar to `iphoneos/{include,lib}`

* [NO] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)


baentsch (commented):
Suggest to close as per https://github.com/open-quantum-safe/oqs-provider/pull/363#discussion_r1658649316. Tagging @dstebila @andybrucenet for additional feedback.

SWilson4:
Closing, cf. https://github.com/open-quantum-safe/oqs-provider/pull/363#issuecomment-2260481617

PR #1720, opened 2024-03-06
Fix SPHINCS+ naming in CT tests [skip ci]

### LINKED ISSUE #1718: Extended generic constant-time tests failing for SPHINCS+
https://github.com/open-quantum-safe/liboqs/actions/runs/8126799220

We should update the suppression files as necessary before the 0.10.0 release.

SWilson4:
Looks like the names were overwritten by mistake during the recent ML-DSA / ML-KEM additions, leading to CI failures.

Fixes #1718.


baentsch (approved):
LGTM. Thanks for finding the reason, @SWilson4 . Looking at #1626 that apparently necessitated this PR, I wonder how this could happen given we have a merge CI checking for unintended changes (e.g., to algs not impacted by a change to `copy_from_upstream.yml`)? Something worth while improving to avoid this may happen again?

dstebila (approved):

dstebila:
Thanks for catching this!

PR #1722, opened 2024-03-08
Reformat LMS / XMSS KAT files

SWilson4:
As per comments on PR #1650.


ducnguyen-sb:
I'm okay with removing KAT name and new line.It's just a little bit inconvenience in the CLI output. But I think we can see the what file is tested from the commandline.

ducnguyen-sb (approved):

ashman-p (approved):

PR #1725, opened 2024-03-11
Always build "internal" library as static

### LINKED ISSUE #1723: 0.10.0-rc1: kat-kem: liboqs-internal.so => not found
Previously (for ALT) I was installing (for `liboqs-tests` package) some test binaries but now they require `liboqs-internal.so` which is not installed. Is this intended? If so I can easily remove installation of them (perhaps this tests package altogether).

List of binaries now requiring the lib (they are prefixed with `oqs-` to not accidentally conflict with other packages and to group them with a namespace):
```
builder@x86_64:~/tmp/liboqs-buildroot/usr/bin$ for i in *; do (set -x; ldd $i) | grep internal; done
+ ldd oqs-dump-alg-info
+ ldd oqs-example-kem
+ ldd oqs-example-sig
+ ldd oqs-kat-kem
        liboqs-internal.so => not found
+ ldd oqs-kat-sig
        liboqs-internal.so => not found
+ ldd oqs-speed-common
        liboqs-internal.so => not found
+ ldd oqs-speed-kem
+ ldd oqs-speed-sig
+ ldd oqs-test-aes
        liboqs-internal.so => not found
+ ldd oqs-test-hash
        liboqs-internal.so => not found
+ ldd oqs-test-kem
+ ldd oqs-test-kem-mem
+ ldd oqs-test-sha3
        liboqs-internal.so => not found
+ ldd oqs-test-sig
+ ldd oqs-test-sig-mem
+ ldd oqs-vectors-kem
+ ldd oqs-vectors-sig
```
 - OS: ALT Linux Sisyphus
 - liboqs version 0.10.0-rc1

SWilson4:
Since `liboqs-internal` (see #1667) is never actually installed on the user's system, it doesn't make sense to build it as a shared library and then link the test programs against it. This PR forces `liboqs-internal` to be built as a static library regardless of the `BUILD_SHARED_LIBS` variable.

Fixes #1723.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -63,7 +63,7 @@ add_library(oqs kem/kem.c
                 ${COMMON_OBJS})
 
 # Internal library to be used only by test programs
-add_library(oqs-internal ${INTERNAL_OBJS})
+add_library(oqs-internal STATIC ${INTERNAL_OBJS})
 
 if(DEFINED SANITIZER_LD_FLAGS)
     target_link_libraries(oqs PUBLIC ${SANITIZER_LD_FLAGS})
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -64,6 +64,10 @@ target_link_libraries(example_kem PRIVATE ${TEST_DEPS})
 
 add_executable(kat_kem kat_kem.c test_helpers.c)
 target_link_libraries(kat_kem PRIVATE ${TEST_DEPS})
+if(CMAKE_CROSSCOMPILING AND CMAKE_SYSTEM_NAME STREQUAL "Windows" AND BUILD_SHARED_LIBS)
+    # workaround for Windows .dll cross-compiling
+    target_link_options(kat_kem PRIVATE -Wl,--allow-multiple-definition)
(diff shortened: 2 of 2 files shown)


baentsch:
> This PR forces liboqs-internal to be built as a static library regardless of the BUILD_SHARED_LIBS variable.

Hmm -- does this also guarantee that the (test) executables then are statically linked (to the symbols in the internal lib) while they're still dynamically linked to the symbols in the public (shared) lib?

SWilson4:
> Hmm -- does this also guarantee that the (test) executables then are statically linked (to the symbols in the internal lib) while they're still dynamically linked to the symbols in the public (shared) lib?

Yes, I believe so. (Based on the test executables now running if `liboqs-internal.a` is removed but not if `liboqs.so.5` is removed.)

dstebila (approved):

Martyrshot (approved):
LGTM (But from this account)

dstebila:
Thanks @baentsch!  The discussion in the OQS status call yesterday was that we would tag an -rc2 after this landed, which I will do tonight unless I hear otherwise.

PR #1732, opened 2024-03-15
Minor update for the Zephyr port

Frauschi:
Hi all,

I know this is very last minute, but this PR improves the Zephyr port and properly makes the new algorithms ML-KEM and ML-DSA properly configurable for users. Before that, the new algorithms were always enabled and couldn't be disabled on the Zephyr port.
As these two now “supersede” Kyber and Dilithium Round 3 versions, I disabled those two by default now. However, they can always still be enabled by a user via the Kconfig interface. 

The new algorithms (also the Falcon padded version) have already been part of the Zephyr CI testing, so no functionality is changed besides the configuration interface. 

I hope it is not too late for version 0.10.


baentsch (approved):

dstebila (approved):

dstebila:
Thanks Tobias!

Since you're the expert on Zephyr, I'll take your advice on whether you think a new release candidate is needed to fully test the Zephyr-related updates, or whether you're satisfied with it based on your tests on this PR.

Frauschi:
> Since you're the expert on Zephyr, I'll take your advice on whether you think a new release candidate is needed to fully test the Zephyr-related updates, or whether you're satisfied with it based on your tests on this PR.

I don't think that we need a new release candidate only for these changes, as they are basically the same options as the existing ones only for the new algorithms. I tested the options on my side and verified them to work properly. Furthermore, the CI tests also use them, so everything should be fine.

PR #1743, opened 2024-04-02
switching to dev mode again

baentsch:
Moving version number forward. Also tests MacOS CI dependencies.


dstebila (approved):

dstebila:
In my own dev environment I've been getting annoying python3 failure messages from brew that mention pipx, but I haven't figured out what I'm actually supposed to do.  Is it as simple as using pipx instead of pip3 everywhere?

baentsch:
> In my own dev environment I've been getting annoying python3 failure messages from brew that mention pipx, but I haven't figured out what I'm actually supposed to do. Is it as simple as using pipx instead of pip3 everywhere?

No clue yet. Without local MacOS environment, I'm trying remote CI debugging right now. Turnaround sucks. If you'd have time and a local environment with the same problem, please give it a go.

baentsch:
> since it doesn't do proper package management.

Do we need to care about this in a temporary CI environment? Just giving the option a go....

dstebila:
> > since it doesn't do proper package management.
> 
> Do we need to care about this in a temporary CI environment? Just giving the option a go....

Probably it's fine, give it a try.

baentsch:
> Probably it's fine, give it a try.

Just done -- without DCO/`-s` option, though :-( That anyway makes me wonder: Why did we activate that? Wouldn't it be more reasonable that maintainers add this to commits while reviewing and not contributors? Any documentation how LF wants us to use this feature?

baentsch:
> > Probably it's fine, give it a try.
> 
> Just done -- without DCO/`-s` option, though :-( That anyway makes me wonder: Why did we activate that? Wouldn't it be more reasonable that maintainers add this to commits while reviewing and not contributors? Any documentation how LF wants us to use this feature?

Tagging @ryjones also here: Any documentation that'd help us understand what you want us to do wrt DCO?

baentsch:
For now, I try to remember adding "-s" to commit messages. Not exactly rocket science, but I'm not sure what this buys us (or any other committer). But again, suggest to add this new LF formalism "HowTo" also to https://openquantumsafe.org/about/#getting-started.

baentsch:
@SWilson4 please provide second review and approval to this "quick fix". I think this is acceptable at this time as [OQS did not agree to provide "production grade" software](https://github.com/open-quantum-safe/tsc/issues/1), so an addition like `--break-system-packages` seems "OK" for now.

praveksharma (approved):

Issue #1466, opened 2023-05-16
Integrate Kyber implementation from libjade


dstebila:
libjade is a Jasmin project involving @cryptojedi and others that has a formally verified implementation of Kyber for some platforms. We have talked about incorporating that into liboqs.

From an email to me by @cryptojedi:

> On May 6, 2023, at 8:24 AM, Peter Schwabe <peter@cryptojedi.org> wrote:
> 
> As we discussed interfacing between OQS/OpenSSL/Sandwich and Libjade in
> Lisbon, I thought I'd let you know that yesterday we got a second
> release of Libjade out, which is hopefully a good starting point:
> https://github.com/formosa-crypto/libjade/releases/latest
> 
> We mostly made sure that the Kyber code matches the code we formally
> verified, added a derandomized API, and improved documentation; see
> https://github.com/formosa-crypto/libjade/blob/release/2023.05/README.md
> https://github.com/formosa-crypto/libjade/blob/release/2023.05/doc/api.md
> 
> As a word of warning: the AVX2 implementation of Kyber in Libjade is
> currently the "fully optimized" one and not the "fully verified" one,
> i.e., the matrix generation is not proven. See
> https://eprint.iacr.org/2023/215, just before Section 5. Also, we
> haven't integrated the proofs with Libjade, yet, and consequently
> they're not covered by CI. That's one of the plans for the next release.
> 
> We are, of course, happy for any kind of feedback by mail or in the
> Formosa-Crypto Zulip: https://formosa-crypto.zulipchat.com/login/

dstebila:
Even if there isn't a full chain of formal verification for the entire library, having higher assurance in some components still seems worthwhile.

baentsch:
To state something probably painfully obvious: This integration ought to be automatically repeatable after the first time, e.g., by way of using `copy_from_upstream` or github submodule, right? And this is still intended for 0.10.0?

dstebila:
> To state something probably painfully obvious: This integration ought to be automatically repeatable after the first time, e.g., by way of using `copy_from_upstream` or github submodule, right? And this is still intended for 0.10.0?

Yes, it is intended for this to be repeatable. Unclear at this point whether it'll be for 0.10.0, this is also tied up with #1521.

baentsch:
Any reason not to close this issue @praveksharma ?

praveksharma:
Closing now @baentsch.

PR #1745, opened 2024-04-03
Integrate Kyber from libjade 

praveksharma:
This PR integrates Kyber512 and Kyber768 from libjade for `x86_64`  with and without AVX2 optimizations as described in #1466. Once Kyber1024 is made available in libjade the updated `copy_from_upstream.py` script will pull and integrate that code into liboqs as well, requiring only modifications to `copy_from_libjade.yml` and `copy_from_upstream.yml`.

This PR modifies `copy_from_upstream.py` and adds a 'libjade' mode in addition 'copy' and 'verify. This new 'libjade' functions independent of 'copy' mode and copies code for the KEM algorithms specified in `copy_from_libjade.yml` and updates relevant documentation and the library's CBOM. While this additional 'libjade' copy_from_upstream infrastructure can also pull signature algorithms it would likely require minor tweaks to the various templates used by `copy_from_upstream.py`; since libjade doesn't provide full support for any signature algorithm at the moment, those changes are best dealt with a separate PR. 

UPDATE (from comment below):

At the moment, I've turned off weekly constant time tests when building with `-DOQS_LIBJADE_BUILD=ON`. Jasmin compiler doesn't include debug information in the x86 assembly it produces so Valgrind is unable to resolve specific function names and locations in source. I've attached output from memcheck for reference: [Kyber768.txt](https://github.com/open-quantum-safe/liboqs/files/15045566/Kyber768.txt) [Kyber512.txt](https://github.com/open-quantum-safe/liboqs/files/15045567/Kyber512.txt)

There is a [closed PR](https://github.com/jasmin-lang/jasmin/pull/684) in the Jasmin lang repo which added a feature to include DWARF2 information in the generated assembly but this update [ is not included in the latest release](https://github.com/jasmin-lang/jasmin/blob/dd1fad7b94b6f577e7c6e804b3c5acd839b2660d/CHANGELOG.md?plain=1#L72). Once merged, I can create an issue to track this and update the versions of Jasmin compiler used in CI after that feature is released.   

[No] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged.)


praveksharma:
The CBOM generation script doesn't account for the BOM including multiple implementations of the same algorithm targeting the same platform. I've tweaked the script to include additional information about the implementation in the bom-ref for components pulled from libjade to avoid clashes (see `/docs/cbom.json`). Since the [Cyclone DX CBOM guide](https://cyclonedx.org/guides/OWASP_CycloneDX-Authoritative-Guide-to-CBOM-en.pdf) only list a uniqueness criterion that any component's bom-ref must satisfy I think this should be okay; I'd appreciate your thoughts on this @bhess.

bhess:
> The CBOM generation script doesn't account for the BOM including multiple implementations of the same algorithm targeting the same platform. I've tweaked the script to include additional information about the implementation in the bom-ref for components pulled from libjade to avoid clashes (see `/docs/cbom.json`). Since the [Cyclone DX CBOM guide](https://cyclonedx.org/guides/OWASP_CycloneDX-Authoritative-Guide-to-CBOM-en.pdf) only list a uniqueness criterion that any component's bom-ref must satisfy I think this should be okay; I'd appreciate your thoughts on this @bhess.

I agree @praveksharma, for this use case this should be fine, only the bom-ref needs to be unique. Thanks for updating the script. I've also created issue #1753 since the CBOM spec was just merged to the official CycloneDX 1.6 release.

Do I understand it correctly that the libjade version of Kyber would co-exist with the pq-crystals implementation, but implementing the same algorithm for the same target? Is there a use case to have both implementations in liboqs?

praveksharma:
The libjade code wasn't accessible due a error in one of the templates. Thank you for pointing that out @SWilson4! 

At the moment, I've turned off weekly constant time tests when building with `-DOQS_LIBJADE_BUILD=ON`. The Jasmin compiler doesn't include debug information in the x86 assembly it produces so Valgrind is unable to resolve specific function names and locations in source. There is a PR in the Jasmin lang repo which adds

docs/algorithms/kem/kyber.yml
```diff
@@ -117,6 +146,30 @@ parameter-sets:
...
+    - architecture: x86_64
+      operating_systems:
+      - Linux
+      - Darwin
+      required_flags:
+      - avx2
+      - bmi2
+      - popcnt
+    no-secret-dependent-branching-claimed: true
+    no-secret-dependent-branching-checked-by-valgrind: true
```
baentsch, line 171: Possibly silly question: How does ´valgrind´ check this given [this comment](https://github.com/open-quantum-safe/liboqs/pull/1745#issuecomment-2067269179)?

scripts/copy_from_upstream/patches/libjade-kyber-meta.patch
```diff
@@ -0,0 +1,120 @@
...
++  - Tancrède Lepoint
++  - Vadim Lyubashevsky
++  - John M. Schanck
++  - Gregor Seiler
++  - Damien Stehlé
+ implementations:
+-    - name: amd64/ref
+-      version: TODO
++    - name: ref
++      version: NIST Round 3 submission
```
baentsch, line 35: Is it sufficient to use this as version information? Wouldn't it be better to include reference to libjade?

baentsch (commented):
Thanks very much for this large chunk of work, @praveksharma ! Am I overlooking CI actually activating and running this code?

docs/algorithms/kem/kyber.yml
```diff
@@ -72,6 +77,30 @@ parameter-sets:
...
```
SWilson4, line 102: Should probably mark this as false if we're not planning on including these in the weekly runs.

SWilson4, line 102:
```suggestion
    no-secret-dependent-branching-checked-by-valgrind: false
```

docs/algorithms/kem/kyber.yml
SWilson4:
```suggestion
    no-secret-dependent-branching-checked-by-valgrind: false
```

.github/workflows/weekly.yml
SWilson4, line 1: Might it be worthwhile including the libjade code in the weekly 100-KAT runs?

src/kem/kyber/kem_kyber_512.c
```diff
@@ -46,7 +46,45 @@ extern int PQCLEAN_KYBER512_AARCH64_crypto_kem_enc(uint8_t *ct, uint8_t *ss, con
...
+#if defined(OQS_DIST_BUILD)
+	if (OQS_CPU_has_extension(OQS_CPU_EXT_AVX2) && OQS_CPU_has_extension(OQS_CPU_EXT_BMI2) && OQS_CPU_has_extension(OQS_CPU_EXT_POPCNT)) {
+#endif /* OQS_DIST_BUILD */
+		return (OQS_STATUS) libjade_kyber512_avx2_keypair(public_key, secret_key);
+#if defined(OQS_DIST_BUILD)
+	} else {
+		return (OQS_STATUS) libjade_kyber512_ref_keypair(public_key, secret_key);
+	}
+#endif /* OQS_DIST_BUILD */
+#elif defined(OQS_ENABLE_KEM_kyber_512aarch64)
```
SWilson4, line 74: @praveksharma looks like the fix to remove the extra underscore had an unintended side effect here. :( I think this should be `OQS_ENABLE_KEM_kyber_512_aarch64`.

src/kem/kyber/kem_kyber_512.c
```diff
@@ -70,9 +108,35 @@ OQS_API OQS_STATUS OQS_KEM_kyber_512_keypair(uint8_t *public_key, uint8_t *secre
...
+#if defined(OQS_DIST_BUILD)
+	if (OQS_CPU_has_extension(OQS_CPU_EXT_AVX2) && OQS_CPU_has_extension(OQS_CPU_EXT_BMI2) && OQS_CPU_has_extension(OQS_CPU_EXT_POPCNT)) {
+#endif /* OQS_DIST_BUILD */
+		return (OQS_STATUS) libjade_kyber512_avx2_enc(ciphertext, shared_secret, public_key);
+#if defined(OQS_DIST_BUILD)
+	} else {
+		return (OQS_STATUS) libjade_kyber512_ref_enc(ciphertext, shared_secret, public_key);
+	}
+#endif /* OQS_DIST_BUILD */
+#elif defined(OQS_ENABLE_KEM_kyber_512aarch64)
```
SWilson4, line 126: ditto

src/kem/kyber/kem_kyber_512.c
```diff
@@ -96,9 +160,35 @@ OQS_API OQS_STATUS OQS_KEM_kyber_512_encaps(uint8_t *ciphertext, uint8_t *shared
...
+#if defined(OQS_DIST_BUILD)
+	if (OQS_CPU_has_extension(OQS_CPU_EXT_AVX2) && OQS_CPU_has_extension(OQS_CPU_EXT_BMI2) && OQS_CPU_has_extension(OQS_CPU_EXT_POPCNT)) {
+#endif /* OQS_DIST_BUILD */
+		return (OQS_STATUS) libjade_kyber512_avx2_dec(shared_secret, ciphertext, secret_key);
+#if defined(OQS_DIST_BUILD)
+	} else {
+		return (OQS_STATUS) libjade_kyber512_ref_dec(shared_secret, ciphertext, secret_key);
+	}
+#endif /* OQS_DIST_BUILD */
+#elif defined(OQS_ENABLE_KEM_kyber_512aarch64)
```
SWilson4, line 178: ditto

src/kem/kyber/kem_kyber_768.c
```diff
@@ -46,7 +46,45 @@ extern int PQCLEAN_KYBER768_AARCH64_crypto_kem_enc(uint8_t *ct, uint8_t *ss, con
...
+#if defined(OQS_DIST_BUILD)
+	if (OQS_CPU_has_extension(OQS_CPU_EXT_AVX2) && OQS_CPU_has_extension(OQS_CPU_EXT_BMI2) && OQS_CPU_has_extension(OQS_CPU_EXT_POPCNT)) {
+#endif /* OQS_DIST_BUILD */
+		return (OQS_STATUS) libjade_kyber768_avx2_keypair(public_key, secret_key);
+#if defined(OQS_DIST_BUILD)
+	} else {
+		return (OQS_STATUS) libjade_kyber768_ref_keypair(public_key, secret_key);
+	}
+#endif /* OQS_DIST_BUILD */
+#elif defined(OQS_ENABLE_KEM_kyber_768aarch64)
```
SWilson4, line 74: ditto

src/kem/kyber/kem_kyber_768.c
```diff
@@ -70,9 +108,35 @@ OQS_API OQS_STATUS OQS_KEM_kyber_768_keypair(uint8_t *public_key, uint8_t *secre
...
+#if defined(OQS_DIST_BUILD)
+	if (OQS_CPU_has_extension(OQS_CPU_EXT_AVX2) && OQS_CPU_has_extension(OQS_CPU_EXT_BMI2) && OQS_CPU_has_extension(OQS_CPU_EXT_POPCNT)) {
+#endif /* OQS_DIST_BUILD */
+		return (OQS_STATUS) libjade_kyber768_avx2_enc(ciphertext, shared_secret, public_key);
+#if defined(OQS_DIST_BUILD)
+	} else {
+		return (OQS_STATUS) libjade_kyber768_ref_enc(ciphertext, shared_secret, public_key);
+	}
+#endif /* OQS_DIST_BUILD */
+#elif defined(OQS_ENABLE_KEM_kyber_768aarch64)
```
SWilson4, line 126: ditto

src/kem/kyber/kem_kyber_768.c
```diff
@@ -96,9 +160,35 @@ OQS_API OQS_STATUS OQS_KEM_kyber_768_encaps(uint8_t *ciphertext, uint8_t *shared
...
+#if defined(OQS_DIST_BUILD)
+	if (OQS_CPU_has_extension(OQS_CPU_EXT_AVX2) && OQS_CPU_has_extension(OQS_CPU_EXT_BMI2) && OQS_CPU_has_extension(OQS_CPU_EXT_POPCNT)) {
+#endif /* OQS_DIST_BUILD */
+		return (OQS_STATUS) libjade_kyber768_avx2_dec(shared_secret, ciphertext, secret_key);
+#if defined(OQS_DIST_BUILD)
+	} else {
+		return (OQS_STATUS) libjade_kyber768_ref_dec(shared_secret, ciphertext, secret_key);
+	}
+#endif /* OQS_DIST_BUILD */
+#elif defined(OQS_ENABLE_KEM_kyber_768aarch64)
```
SWilson4, line 178: ditto

src/kem/kyber/libjade_kyber512_avx2/api.c
```diff
@@ -0,0 +1,19 @@
+#include <oqs/rand.h>
+#include "api.h"
+
+int libjade_kyber512_avx2_keypair(uint8_t *public_key, uint8_t *secret_key) {
+    uint8_t keypair_coins[JADE_KEM_kyber_kyber512_amd64_avx2_KEYPAIRCOINBYTES];
+    OQS_randombytes(keypair_coins, JADE_KEM_kyber_kyber512_amd64_avx2_KEYPAIRCOINBYTES/2);
```
SWilson4, line 6: Quick sanity check for my own understanding: I'm assuming we can't get the randomness in one shot because that would screw up the KATs and their counter-based PRNG?

SWilson4 (changes requested):
Looks like there's still one templating bug to iron out. Other than that, everything looks great to me. I didn't carefully review the assembly code (it's formally verified anyway, right?) or the Python scripts.

SWilson4:
A few comments from an older review (addressed in-person with @praveksharma) slipped in. I think I've cleaned them all up; sorry for any confusion.

scripts/copy_from_upstream/patches/libjade-kyber-meta.patch
praveksharma: This patch is applied to Kyber512/768's `META.yml` files in the libjade repo pulled (and subsequently deleted) by `copy_from_upstream.py`. The information in these YAML files is used by `copy_from_upstream.py` but they don't make their way into liboqs source code in their entirety. 

`docs/algorithms/kem/kyber.yml`, which is populated by `update_upstream_alg_docs.py`, does explicitly list libjade as an upstream and links to the specific release of libjade that liboqs is currently pulling code from (`update_upstream_alg_docs.py` gets this information from `copy_from_libjade.yml` which would need to be updated to pull from newer libjade releases ensuring `docs/algorithms/kem/kyber.yml` is also updated).

`CONFIGURE.md` which documents the `OQS_LIBJADE_BUILD` option (this option must be enabled at build time to include libjade code in liboqs) also links to the libjade repo.

Are there other places you think should also include a reference to libjade?

scripts/copy_from_upstream/copy_from_upstream.py
```diff
@@ -65,15 +65,27 @@ def shell(command, expect=0):
     if ret.returncode != expect:
         raise Exception("'{}' failed with error {}. Expected {}.".format(" ".join(command), ret, expect))
 
-def generator(destination_file_path, template_filename, family, scheme_desired):
+# Generate template from specified scheme to replace old file in 'copy' mode
+# but preserves additions made to file made in priour runs of 'libjade' mode
```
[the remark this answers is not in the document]
praveksharma: Fixed.

src/kem/kyber/kem_kyber_768.c
```diff
@@ -70,9 +108,35 @@ OQS_API OQS_STATUS OQS_KEM_kyber_768_keypair(uint8_t *public_key, uint8_t *secre
 #else
 	return (OQS_STATUS) pqcrystals_kyber768_ref_keypair(public_key, secret_key);
 #endif
+#endif /* OQS_LIBJADE_BUILD */
 }
 
 OQS_API OQS_STATUS OQS_KEM_kyber_768_encaps(uint8_t *ciphertext, uint8_t *shared_secret, const uint8_t *public_key) {
+#if OQS_LIBJADE_BUILD
+#if defined(OQS_ENABLE_LIBJADE_KEM_kyber_768__avx2)
```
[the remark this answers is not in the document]
praveksharma: Fixed.

src/kem/kyber/kem_kyber_768.c
```diff
@@ -96,9 +160,35 @@ OQS_API OQS_STATUS OQS_KEM_kyber_768_encaps(uint8_t *ciphertext, uint8_t *shared
 #else
 	return (OQS_STATUS) pqcrystals_kyber768_ref_enc(ciphertext, shared_secret, public_key);
 #endif
+#endif /* OQS_LIBJADE_BUILD */
 }
 
 OQS_API OQS_STATUS OQS_KEM_kyber_768_decaps(uint8_t *shared_secret, const uint8_t *ciphertext, const uint8_t *secret_key) {
+#if OQS_LIBJADE_BUILD
+#if defined(OQS_ENABLE_LIBJADE_KEM_kyber_768__avx2)
```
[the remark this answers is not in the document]
praveksharma: Fixed.

docs/algorithms/kem/kyber.yml
praveksharma: Fixed.

praveksharma: Fixed.

.github/workflows/weekly.yml
praveksharma: Thanks for pointing this out. As pointed out by Michael, the code wasn't being tested CI after I stopped constant time testing for libjade code. This is fixed now.

src/kem/kyber/kem_kyber_512.c
praveksharma: :( indeed, but fixed, so :).

src/kem/kyber/kem_kyber_512.c
praveksharma: Fixed.

src/kem/kyber/kem_kyber_512.c
praveksharma: Fixed.

src/kem/kyber/kem_kyber_768.c
praveksharma: Fixed.

src/kem/kyber/kem_kyber_768.c
praveksharma: Fixed.

src/kem/kyber/kem_kyber_768.c
praveksharma: Fixed.

src/kem/kyber/libjade_kyber512_avx2/api.c
praveksharma: Yes, precisely for the reason you mention.

praveksharma:
@baentsch @SWilson4, thank you for your reviews! I think I've addressed all of them. 

> Am I overlooking CI actually activating and running this code?

You weren't Michael, I've fixed that. The code is being activated by the GitHub Actions workflows but not CircleCI (specifically the ubuntu-bionic-i386 image which from what I can tell isn't being used in GitHub Actions workflows). However, the code is being tested on both Linux and Darwin which are they only platforms libjade supports at the moment.

docs/algorithms/kem/kyber.yml
```diff
@@ -26,6 +26,11 @@ optimized-upstreams:
       with copy_from_upstream patches
     spdx-license-identifier: CC0-1.0 and (CC0-1.0 or Apache-2.0) and (CC0-1.0 or MIT)
       and MIT
+formally-verified-upstreams:
+  libjade:
+    source: https://github.com/formosa-crypto/lib/commit/a6845a257a0ebbab4fd93c19423a62557c838a10
+      with copy_from_upstream patches
+    spdx-license-identifier: CC0-1.0
```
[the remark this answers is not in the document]
praveksharma: I did bring this up in the last status meeting, and I think we're good as long the libjade code isn't enabled by default. Am I paraphrasing you correctly @baentsch?

That said, the libjade team does wish to switch to a Apache/CC0 dual-license. It's unclear what the timelines on that are at the moment.

scripts/copy_from_upstream/copy_from_libjade.yml
```diff
@@ -0,0 +1,29 @@
+# When adding an algorithm to copy_from_libjade.yml, the boolean
```
baentsch, line 1: What is the conceptual benefit from having a separate .YML for libjade? Why is/could this information not be integrated into our (so far one-and-only) central copy_from_upstream.yml?

scripts/copy_from_upstream/patches/libjade-kyber-meta.patch
baentsch: > but they don't make their way into liboqs source code in their entirety.

And this may be a problem: Looking at the resultant .md/.yml files I don't see libjade version information (other than the "Source" reference). Please be aware that whichever "version information finding" logic is required, this must be implemented in [oqs-provider](https://github.com/open-quantum-safe/oqs-provider/blob/main/oqs-template/generate.py) -- or any other possible future downstream caring to make algorithm version information available to users, so having this easily available would be appreciated by anyone (human users and downstream software logic alike) looking at the .yml/.md files.

> Are there other places you think should also include a reference to libjade?

No; to the contrary, there should be as few places as possible to do so, ideally just one -- but that should contain all information easily consumed, something along the lines "Kyber512 - libjade vx.y.z".

scripts/copy_from_upstream/copy_from_libjade.yml
praveksharma: While adding libjade as an upstream it was my desire that it be possible to pull code from libjade independently of liboqs' various other upstreams (more on that below). An additional YAML file seems easier to maintain than the additional scaffolding (the already rather large) copy_from_upstream.py would require to support filtering out libjade specific information interspersed through a single YAML needed for the libjade mode. I think it also makes more sense to have separate YML files for separate modes: one for copy mode, one for libjade mode. 

Why should copy and libjade be separate modes? This PR compiles libjade *.jazz files into *.S assembly files using the Jasmin compiler instead of pulling pre-compiled *.S files from libjade releases. This approach has two advantages:

1. It allows verifying libjade proofs (e.g., correctness, constant time) in CI once libjade releases this feature.
2. We can compile the assembly files with our choice of Jasmin compiler flags, such as those enabling emitting DWARF2 info.

However, the Jasmin compiler isn't widely packaged (only available in Nix and OPAM). Also, compiling the assembly files locally takes considerable time. Requiring all liboqs developers to install a not-easy-to-install package and compile libjade even to update non-libjade code would hinder development.

scripts/copy_from_upstream/patches/libjade-kyber-meta.patch
praveksharma: > Looking at the resultant .md/.yml files I don't see libjade version information (other than the "Source" reference).
>  so having this easily available would be appreciated by anyone (human users and downstream software logic alike) looking at the .yml/.md files.

To my knowledge, this is consistent with how all other upstreams for Kyber are documented at the moment. The "Source" reference includes both a link to the repo and the specific release of libjade the implementation is being pulled from. Do we want to document additional information for libjade implementations?

> "Kyber512 - libjade vx.y.z".

Am I missing where this information is documented for pqclean and pqcrystals implementations of Kyber? Or are do we want to change how libjade implementations are documented? At the moment the .md file lists the upstreams (and their sources) that liboqs pulls from, the algorithm's parameters sets, and the specific implementations liboqs provides for each parameter (each implementation clearly lists its upstream). I'm not sure why we would want to list

.CMake/alg_support.cmake
```diff
@@ -168,6 +168,15 @@ cmake_dependent_option(OQS_ENABLE_SIG_sphincs_shake_256f_simple "" ON "OQS_ENABL
 cmake_dependent_option(OQS_ENABLE_SIG_sphincs_shake_256s_simple "" ON "OQS_ENABLE_SIG_SPHINCS" OFF)
 ##### OQS_COPY_FROM_UPSTREAM_FRAGMENT_ADD_ENABLE_BY_ALG_END
 
+##### OQS_COPY_FROM_LIBJADE_FRAGMENT_ADD_ENABLE_BY_ALG_START
+if ((OQS_LIBJADE_BUILD STREQUAL ON))
```
SWilson4, line 172: I'm not sure it makes a difference or not, but it matches the conventions in the rest of the file.
```suggestion
if ((OQS_LIBJADE_BUILD STREQUAL "ON"))
```

.CMake/alg_support.cmake
```diff
@@ -497,6 +506,24 @@ endif()
 
 ##### OQS_COPY_FROM_UPSTREAM_FRAGMENT_ADD_ENABLE_BY_ALG_CONDITIONAL_END
 
+##### OQS_COPY_FROM_LIBJADE_FRAGMENT_ADD_ENABLE_BY_ALG_CONDITIONAL_START
+if ((OQS_LIBJADE_BUILD STREQUAL ON))
```
SWilson4, line 510:
```suggestion
if ((OQS_LIBJADE_BUILD STREQUAL "ON"))
```

docs/algorithms/kem/kyber.yml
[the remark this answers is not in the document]
SWilson4: Sounds good. We should probably document it in the README regardless: https://github.com/open-quantum-safe/liboqs/blob/main/README.md#license

scripts/copy_from_upstream/patches/libjade-kyber-meta.patch
SWilson4: I've always interpreted these version strings as tracking the version of the algorithm specification that each implementation implements, which in this case is the Round 3 version.

src/common/CMakeLists.txt
```diff
@@ -71,13 +71,20 @@ else() # using XKCP
     set(SHA3_IMPL sha3/xkcp_sha3.c sha3/xkcp_sha3x4.c)
 endif()
 
+if ((OQS_LIBJADE_BUILD STREQUAL ON))
```
SWilson4, line 74:
```suggestion
if ((OQS_LIBJADE_BUILD STREQUAL "ON"))
```

src/kem/kyber/kem_kyber_512.c
```diff
@@ -46,7 +46,45 @@ extern int PQCLEAN_KYBER512_AARCH64_crypto_kem_enc(uint8_t *ct, uint8_t *ss, con
...
+#if defined(OQS_DIST_BUILD)
+	} else {
+		return (OQS_STATUS) libjade_kyber512_ref_keypair(public_key, secret_key);
+	}
+#endif /* OQS_DIST_BUILD */
+#elif defined(OQS_ENABLE_KEM_kyber_512_aarch64)
+#if defined(OQS_DIST_BUILD)
+	if (OQS_CPU_has_extension(OQS_CPU_EXT_ARM_NEON)) {
+#endif /* OQS_DIST_BUILD */
+		return (OQS_STATUS) PQCLEAN_KYBER512_AARCH64_crypto_kem_keypair(public_key, secret_key);
```
SWilson4, line 78: It seems to me that if the user has specified a libjade build on an ARM platform, they probably want a formally verified implementation. Perhaps we ought to use the libjade "ref" implementation irrespective of available extensions.

.github/workflows/unix.yml
```diff
@@ -206,6 +229,16 @@ jobs:
...
           - -DBUILD_SHARED_LIBS=ON -DOQS_DIST_BUILD=OFF
+        libjade-build:
+          - -DOQS_LIBJADE_BUILD=OFF
+          # Restrict -DOQS_LIBJADE_BUILD=ON build to algs provided by
+          # libjade to minimise repeated tests
+          - -DOQS_LIBJADE_BUILD=ON -DOQS_MINIMAL_BUILD=$LIBJADE_ALG_LIST
+        exclude: 
+          # macos-14 runs on aarch64, libjade targets x86
+          # Skip testing libjade on macos-14
+          - os: macos-14
```
SWilson4, line 240: It seems like there is code that calls the libjade "ref" implementation on aarch64 platforms (see my comment on kem_kyber_512.c). Is this something that will actually work, or should we disable libjade completely on aarch64? (e.g., throw an error via CMake if the user passes OQS_LIBJADE_BUILD=ON on an ARM system)

src/kem/kyber/kem_kyber_768.c
```diff
@@ -46,7 +46,45 @@ extern int PQCLEAN_KYBER768_AARCH64_crypto_kem_enc(uint8_t *ct, uint8_t *ss, con
...
+#if defined(OQS_DIST_BUILD)
+	if (OQS_CPU_has_extension(OQS_CPU_EXT_AVX2) && OQS_CPU_has_extension(OQS_CPU_EXT_BMI2) && OQS_CPU_has_extension(OQS_CPU_EXT_POPCNT)) {
+#endif /* OQS_DIST_BUILD */
+		return (OQS_STATUS) libjade_kyber768_avx2_keypair(public_key, secret_key);
+#if defined(OQS_DIST_BUILD)
+	} else {
+		return (OQS_STATUS) libjade_kyber768_ref_keypair(public_key, secret_key);
+	}
+#endif /* OQS_DIST_BUILD */
+#elif defined(OQS_ENABLE_KEM_kyber_768_aarch64)
```
SWilson4, line 74: Same comment as on kem_kyber_512.c

SWilson4 (commented):
Just a couple of questions around config for ARM-based systems.

src/kem/kyber/kem_kyber_512.c
praveksharma: At the moment libjade only supports x86. This was a poorly thought out idea for the eventuality that libjade adds selective ARM support for algorithms and I've removed this line. This "solution" clearly wouldn't work, either the code wouldn't compile or that return statement would be unreachable.

praveksharma: I've edited `CMakelists.txt` so that CMake explicitly disallows building liboqs with libjade on on non-x86 platforms, I'd missed this earlier. This would need to be reevaluated once libjade supports other platforms.

praveksharma: See [this comment](https://github.com/open-quantum-safe/liboqs/pull/1745#discussion_r1580123937).

.github/workflows/unix.yml
praveksharma: I've edited `CMakelists.txt` so that CMake explicitly disallows building liboqs with libjade on on non-x86 platforms, I'd missed this earlier. This would need to be reevaluated once libjade supports other platforms.

src/kem/kyber/kem_kyber_768.c
praveksharma: See [this coment](https://github.com/open-quantum-safe/liboqs/pull/1745#discussion_r1580123937).

CMakeLists.txt
```diff
@@ -100,6 +96,10 @@ else()
     message(FATAL_ERROR "Unknown or unsupported processor: " ${CMAKE_SYSTEM_PROCESSOR} ". Override by setting OQS_PERMIT_UNSUPPORTED_ARCHITECTURE=ON")
 endif()
 
+if (NOT (CMAKE_SYSTEM_NAME MATCHES "Linux|Darwin") AND (OQS_LIBJADE_BUILD STREQUAL ON) AND (ARCH_X86_64 STREQUAL "ON"))
```
SWilson4, line 99: I think this condition is not quite right: for instance, it seems like `-DOQS_LIBJADE_BUILD=ON` would make it through without a warning on Windows ARM builds.

SWilson4:
```suggestion
if (NOT ((CMAKE_SYSTEM_NAME MATCHES "Linux|Darwin") AND (ARCH_X86_64 STREQUAL "ON")) AND (OQS_LIBJADE_BUILD STREQUAL "ON"))
```
I think this does the trick

SWilson4 (approved):

scripts/copy_from_upstream/copy_from_libjade.yml
baentsch: Thanks for the explanation. LGTM.

baentsch:
> @baentsch @SWilson4, thank you for your reviews! I think I've addressed all of them.
> 
> > Am I overlooking CI actually activating and running this code?
> 
> You weren't Michael, I've fixed that. The code is being activated by the GitHub Actions workflows but not CircleCI (specifically the ubuntu-bionic-i386 image which from what I can tell isn't being used in GitHub Actions workflows). However, the code is being tested on both Linux and Darwin which are they only platforms libjade supports at the moment.

Thanks for this improvement. So this means that if this is merged (and activated), Linux and MacOS will run automatically run `libjade` Kyber code and all else (Windows, Raspberry, IBM platforms, etc.) will run reference code? If so, is this already documented? Should it be? How?

bhess (commented):
Great work adding Kyber from libjade!

Two points I noticed:

- When running `copy_from_upstream.py libjade` I seem to be missing some dependencies. Should they be documented?

- If I compile with `-DOQS_LIBJADE_BUILD=ON` and run, e.g., `test_kem Kyber768`, the `OQS_LIBJADE_BUILD` flag isn't reported under "OQS build flags". I think it would be useful to know the flag is activated by reporting it in [system_info.c](https://github.com/open-quantum-safe/liboqs/blob/main/tests/system_info.c).

Now it looks like this:
```
Configuration info
==================
Target platform:  x86_64-Linux-6.1.19
Compiler:         gcc (11.4.0)
OQS version:      0.10.1-dev
Git commit:       864be05738585e975540dd3ee1095d580904517f (+ local modifications)
OpenSSL enabled:  Yes (OpenSSL 3.0.2 15 Mar 2022)
AES:              NI
SHA-2:            OpenSSL
SHA-3:            C
CPU exts active:  ADX AES AVX AVX2 AVX512 BMI1 BMI2 PCLMULQDQ VPCLMULQDQ POPCNT SSE SSE2 SSE3
```

baentsch:
As per the performance discussion from yesterday: Indeed this code is faster than what we have in the system until now, so maybe you'd like to add some marketing speak to the build option "OQS_LIBJADE_BUILD" along those lines. 

```
$ ../build-plain/tests/speed_kem Kyber768
Configuration info
==================
Target platform:  x86_64-Linux-6.5.0-26-generic
Compiler:         gcc (11.4.0)
OQS version:      0.10.1-dev
Git commit:       a5ec23cf19763d36a558b8358345823ae45d57e5
OpenSSL enabled:  Yes (OpenSSL 3.0.2 15 Mar 2022)
AES:              NI
SHA-2:            OpenSSL
SHA-3:            C
CPU exts active:  ADX AES AVX AVX2 BMI1 BMI2 PCLMULQDQ POPCNT SSE SSE2 SSE3
Speed test
==========
Started at 2024-05-08 08:00:58
Operation                            | Iterations | Total time (s) | Time (us): mean | pop. stdev | CPU cycles: mean          | pop. stdev
------------------------------------ | ----------:| --------------:| ---------------:| ----------:| -------------------------:| ----------:
Kyber768                             |            |                |                 |            |                           |           
keygen                               |     246687 |          3.000 |          12.161 |      2.489 |                     23015 |       4694
encaps                               |     207357 |          3.000 |          14.468 |      0.635 |                     27388 |        754
decaps                               |     276601 |          3.000 |          10.846 |      0.408 |                     20520 |        324
Ended at 2024-05-08 08:01:07

$ ./tests/speed_kem Kyber768
Configuration info
==================
Target platform:  x86_64-Linux-6.5.0-26-generic
Compiler:         gcc (11.4.0)
OQS version:      0.10.1-dev
Git commit:       864be05738585e975540dd3ee1095d580904517f
OpenSSL enabled:  Yes (OpenSSL 3.0.2 15 Mar 2022)
AES:              NI
SHA-2:            OpenSSL
SHA-3:            C
CPU exts active:  ADX AES AVX AVX2 BMI1 BMI2 PCLMULQDQ POPCNT SSE SSE2 SSE3
Speed test
==========
Started at 2024-05-08 08:01:16
Operation                            | Iterations | Total time (s) | Time (us): mean | pop. stdev | CPU cycles: mean          | pop. stdev
------------------------------------ | ----------:| --------------:| ---------------:| ----------:| -------------------------:| ----------:
Kyber768                             |            |                |                 |            |                           |           
keygen                               |     297927 |          3.000 |          10.070 |      2.127 |                     19048 |       4013
encaps                               |     243501 |          3.000 |          12.320 |      0.526 |                     23318 |        495
decaps                               |     303813 |          3.000 |           9.874 |      0.398 |                     18679 |        369
Ended at 2024-05-08 08:01:25
```

What I'd surely recommend before merge is adding at least "OQS_LIBJADE_BUILD" to the output of the "OQS build flags" output as otherwise there's really no indication as to what explains performance differences (in the above it's just my distinction of a "plain" build directory (without libjasmin) and a current build (on branch "ps-jasmin").... Even cooler of course would be output next to each algorithm name as to which code (level, version, etc.) it's running on the given platform, thus doing away with having to read all caveats as to what is active on a specific platform -- but that's arguably quite some effort.  Created #1786 to track separately.

src/kem/kyber/libjade_kyber768_ref/api.c
```diff
@@ -0,0 +1,19 @@
+#include <oqs/rand.h>
+#include "api.h"
+
+int libjade_kyber768_ref_keypair(uint8_t *public_key, uint8_t *secret_key) {
+    uint8_t keypair_coins[JADE_KEM_kyber_kyber768_amd64_ref_KEYPAIRCOINBYTES];
+    OQS_randombytes(keypair_coins, JADE_KEM_kyber_kyber768_amd64_ref_KEYPAIRCOINBYTES/2);
+    OQS_randombytes((uint8_t *)(keypair_coins + (JADE_KEM_kyber_kyber768_amd64_ref_KEYPAIRCOINBYTES/2)), JADE_KEM_kyber_kyber768_amd64_ref_KEYPAIRCOINBYTES/2);
```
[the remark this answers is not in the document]
SWilson4: Answering on @praveksharma's behalf: since the NIST DRBG uses a counter, the sequences of calls needs to be duplicated to not screw up the KATs.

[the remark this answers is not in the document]
bhess: Thanks, but still not sure why generating the randombytes in one go should lead to different bytes. AFAIK, the counter in the DRBG is incremented after each 16 byte block and the state is kept after each call. As long as one doesn't truncate generated bytes (which doesn't seem the case here), there shouldn't be a difference.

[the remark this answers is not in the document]
dstebila: It does seem counterintuitive, but I've heard it from a few people, most recently from @tfaoliveira, who I think understands this subtle problem.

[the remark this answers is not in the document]
SWilson4: Looking at the source code, I believe it's not because of the counter (as I had thought) but because the AES key is updated after each `randombytes` call: https://github.com/open-quantum-safe/liboqs/blob/a5ec23cf19763d36a558b8358345823ae45d57e5/src/common/rand/rand_nist.c#L118

praveksharma:
The license files have been updated and the PR is ready to merge.

CONFIGURE.md
```diff
@@ -198,3 +200,18 @@ At the moment, this is **only** considered for random number generation, as both
 **Attention**: When this option is enabled, you have to supply a custom callback for obtaining random numbers using the `OQS_randombytes_custom_algorithm()` API before accessing the cryptographic API. Otherwise, all key generation and signing operations will fail. 
 
 **Default**: `OFF`.
+
+## OQS_LIBJADE_BUILD
+Can be `ON` or `OFF`. When `ON` liboqs is built to use high assurance implementations of cryptographic algorithms from [Libjade](https://github.com/formosa-crypto/libjade). The cryptographic primitives in Libjade are written using [Jasmin](https://github.com/jasmin-lang/jasmin) and built using the Jasmin compiler. The Jasmin compiler is proven (in Coq) to preserve semantic correctness of a program, maintain secret-independence of control flow, and maintain secret independence of locations of memory access through compilation. Additionaly, the Jasmin compiler guarantees thread safety because Jasmin doesn't support global variables (with the caveat that the caller respects the function's contract).
```
baentsch, line 205: Typo "additionally". And I don't understand the caveat: Is this relevant for users of `liboqs` or for the integration _to_ `liboqs` (i.e., wouldn't need to be mentioned here)?

docs/.Doxyfile
```diff
@@ -378,7 +378,7 @@ TOC_INCLUDE_HEADINGS   = 0
 # The default value is: DOXYGEN.
 # This tag requires that the tag MARKDOWN_SUPPORT is set to YES.
 
-MARKDOWN_ID_STYLE      = DOXYGEN
+MARKDOWN_ID_STYLE      = GITHUB
```
baentsch, line 381: Why is this necessary/related to a `libjade` integration?

src/kem/kyber/kem_kyber_512.c
```diff
@@ -46,7 +46,35 @@ extern int PQCLEAN_KYBER512_AARCH64_crypto_kem_enc(uint8_t *ct, uint8_t *ss, con
...
+extern int libjade_kyber512_avx2_dec(uint8_t *ss, const uint8_t *ct, const uint8_t *sk);
+#endif
+
+
 OQS_API OQS_STATUS OQS_KEM_kyber_512_keypair(uint8_t *public_key, uint8_t *secret_key) {
+#if defined(OQS_LIBJADE_BUILD) && (defined(OQS_ENABLE_LIBJADE_KEM_kyber_512))
+#if defined(OQS_ENABLE_LIBJADE_KEM_kyber_512_avx2)
+#if defined(OQS_DIST_BUILD)
+	if (OQS_CPU_has_extension(OQS_CPU_EXT_AVX2) && OQS_CPU_has_extension(OQS_CPU_EXT_BMI2) && OQS_CPU_has_extension(OQS_CPU_EXT_POPCNT)) {
+#endif /* OQS_DIST_BUILD */
```
baentsch, line 67: Wow - one impressive set of config options... Do we have enough new CI that triggers (and not) this? I found mostly MINIMAL_BUILD configs in GH actions files. Are these sufficient?

src/oqsconfig.h.cmake
```diff
@@ -201,6 +201,17 @@
 #cmakedefine OQS_ENABLE_SIG_mayo_5_avx2 1
 ///// OQS_COPY_FROM_UPSTREAM_FRAGMENT_ADD_ALG_ENABLE_DEFINES_END
 
+///// OQS_COPY_FROM_LIBJADE_FRAGMENT_ADD_ALG_ENABLE_DEFINES_START
+
+#cmakedefine01 OQS_LIBJADE_BUILD
```
baentsch, line 206: what's cmakedefine01?

baentsch (commented):
Conceptually LGTM, but didn't have enough time to thoroughly look at it all and will be on the road from tomorrow until next Wed, so please rely on other's opinions. Thanks for the work @praveksharma !

CONFIGURE.md
praveksharma: Good catch Michael, you're right this isn't necessary. I've pushed a commit remove the unnecessary caveat.

src/kem/kyber/kem_kyber_512.c
praveksharma: I'm not sure I understand your question entirely but if you're asking "Do the AVX2 implementations from libjade get tested in CI?" then yes they do. These config options are standard for all AVX2 implementations even for those not from libjade.

src/oqsconfig.h.cmake
praveksharma: CMake 3.10 added support for both `cmakedefine` and `cmakedefine01` to `configure_file()`. Both are similar and from CMake 3.15 docs:

> Input file lines of the form #cmakedefine01 VAR will be replaced with either #define VAR 1 or #define VAR 0 similarly. 

It's syntactic sugar.

docs/.Doxyfile
praveksharma: It isn't. I likely made this change while debugging documentation generation and this option fixes some issues I was facing with GH style anchors.

SWilson4 (approved):
One comment about docs that may require updating. Otherwise LGTM!

README.md
```diff
@@ -194,6 +194,7 @@ liboqs includes some third party libraries or modules that are licensed differen
 - `src/kem/classic_mceliece/pqclean_*`: public domain
 - `src/kem/kyber/pqcrystals-*`: public domain (CC0) or Apache License v2.0
 - `src/kem/kyber/pqclean_*`: public domain (CC0), and public domain (CC0) or Apache License v2.0, and public domain (CC0) or MIT, and MIT
+- `src/kem/kyber/libjade_*` public domain (CC0)
```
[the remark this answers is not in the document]
baentsch: It surely should be.

src/kem/kyber/kem_kyber_512.c
baentsch: @praveksharma Sorry for not being clearer before. What I meant to ask is whether there is CI that tests all config options possible, thus triggering all code bases embedded/enabled by the various defines/config options that may be chosen by users, i.e., libjade ref and avx2 code, pqcrystals ref and avx2 and pqclean aarch64 code?

praveksharma: @baentsch Thank you for clarifying! I believe the aarch64 code does end up getting tested on our ARM tests but you're right our CI tests don't explicitly use the OQS_USE_CPUFEATURE_INSTRUCTIONS option. So it would seem we do not actually test all options available to users in CONFIGURE.md. I can create a separate issue to track this.

src/common/libjade_shims/libjade_randombytes.c
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: CC0-1.0
```
dstebila, line 1: Should this be MIT license?

src/kem/kyber/libjade_kyber768_ref/api.c
```diff
@@ -0,0 +1,19 @@
+#include <oqs/rand.h>
```
dstebila, line 1: Missing SPDX license header?

src/kem/kyber/libjade_kyber768_ref/api.h
```diff
@@ -0,0 +1,45 @@
+#ifndef JADE_KEM_kyber_kyber768_amd64_ref_API_H
```
dstebila, line 1: Missing SPDX license header?

src/common/libjade_shims/libjade_randombytes.c
praveksharma: Thank you for catching this. I copied a portion of this this from libjade and I've changed the license to Apache-2.0 OR CC0-1.0. Although since the code is relatively straightforward I can change this MIT if that if preferable.

src/common/libjade_shims/libjade_randombytes.h
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: CC0-1.0
```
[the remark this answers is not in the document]
praveksharma: Same as https://github.com/open-quantum-safe/liboqs/pull/1745#discussion_r1720118393.

src/kem/kyber/libjade_kyber512_avx2/api.c
```diff
@@ -0,0 +1,19 @@
+#include <oqs/rand.h>
```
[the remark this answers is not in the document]
praveksharma: Files matching `kem/.*/.*/.*` are copied from upstream and are not checked by `tests/test_spdx.sh` (presumably to avoid having to patch individual upstream files). Instead the license information for these files is provided by README.md (this PR updates it to include license information for files matching `src/kem/kyber/libjade_*`) : https://github.com/open-quantum-safe/liboqs?tab=readme-ov-file#license

src/kem/kyber/libjade_kyber512_avx2/api.h
```diff
@@ -0,0 +1,45 @@
+#ifndef JADE_KEM_kyber_kyber512_amd64_avx2_API_H
```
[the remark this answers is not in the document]
praveksharma: Same as: https://github.com/open-quantum-safe/liboqs/pull/1745#discussion_r1720123289

src/kem/kyber/libjade_kyber512_ref/api.c
```diff
@@ -0,0 +1,19 @@
+#include <oqs/rand.h>
```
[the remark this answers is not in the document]
praveksharma: Same as: https://github.com/open-quantum-safe/liboqs/pull/1745#discussion_r1720123289

src/kem/kyber/libjade_kyber512_ref/api.h
```diff
@@ -0,0 +1,45 @@
+#ifndef JADE_KEM_kyber_kyber512_amd64_ref_API_H
```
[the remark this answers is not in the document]
praveksharma: Same as: https://github.com/open-quantum-safe/liboqs/pull/1745#discussion_r1720123289

src/kem/kyber/libjade_kyber768_avx2/api.c
```diff
@@ -0,0 +1,19 @@
+#include <oqs/rand.h>
```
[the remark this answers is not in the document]
praveksharma: Same as: https://github.com/open-quantum-safe/liboqs/pull/1745#discussion_r1720123289

src/kem/kyber/libjade_kyber768_avx2/api.h
```diff
@@ -0,0 +1,45 @@
+#ifndef JADE_KEM_kyber_kyber768_amd64_avx2_API_H
```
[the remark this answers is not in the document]
praveksharma: Same as: https://github.com/open-quantum-safe/liboqs/pull/1745#discussion_r1720123289

src/kem/kyber/libjade_kyber768_ref/api.c
praveksharma: Same as: https://github.com/open-quantum-safe/liboqs/pull/1745#discussion_r1720123289

src/kem/kyber/libjade_kyber768_ref/api.h
praveksharma: Same as: https://github.com/open-quantum-safe/liboqs/pull/1745#discussion_r1720123289

src/kem/kyber/libjade_kyber512_avx2/api.c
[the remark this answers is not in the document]
dstebila: Makes sense! Thanks.

dstebila (approved):

praveksharma:
Merging despite Travis failing, please see #1888.

baentsch:
@SWilson4 @praveksharma May I ask a question: Does this mean that when a GH CI job is incorrectly formatted, it simply does not run and no UI information is given so one is tempted to (and in this case did) merge even though CI doesn't run? Second question: Is there no way one can check "at a glance" whether all configured GH jobs ran (and are configured) OK (along the lines of the Travis CI badge)? Worthwhile checking back with GH?

SWilson4:
> @SWilson4 @praveksharma May I ask a question: Does this mean that when a GH CI job is incorrectly formatted, it simply does not run and no UI information is given so one is tempted to (and in this case did) merge even though CI doesn't run? Second question: Is there no way one can check "at a glance" whether all configured GH jobs ran (and are configured) OK (along the lines of the Travis CI badge)? Worthwhile checking back with GH?

I believe it does shows up, but not as a "red" failure (simply greyed out iirc). In this case it was probably even less obvious because of the Travis failures. I think this will be improved by #1880.

SWilson4:
> I believe it does shows up, but not as a "red" failure (simply greyed out iirc). In this case it was probably even less obvious because of the Travis failures. I think this will be improved by #1880.

Never mind, it doesn't show up at all on the interface next to the merge button. :( See below for an example of a PR which inherits the same failure:

Screenshot from 2024-08-21 15-19-03

baentsch:
> Never mind, it doesn't show up at all on the interface next to the merge button. 

So I thought. This is not good: We start to regularly disregard CI failures (Travis) and don't see (look for) badly configured CI jobs, too (not exactly surprising as we already seem to expect CI to fail) -- and on top [violate the least privilege principle wrt controlling who can do what to a project](https://github.com/open-quantum-safe/tsc/pull/62#issuecomment-2295311836) -- and discuss this all in closed-and-merged PRs, i.e., after the fox has left the barn (or whatever the proverb is :).

I think we ought to discuss how to improve things here. Any chance we can take some minutes for this in the next status call, @dstebila ? If not, my proposal below:
- Do not accept PRs failing any CI but rather remove deemed irrelevant CI for good, creating an issue to activate it when possible; possibly document changes thus caused in PLATFORMS.md so this is becoming visible to the community and possibly creating motivation for the community to help solve things.
- Add testing to highlight incorrect CI scripts _and_ do not allow PRs before such errors are fixed, in effect making them a priority
- PRs changing GH access permissions should be reviewed by at least as many people as changes to the logic of `liboqs` as errors here have the potential for more damage than simple deletion of a project, namely the uncontrolled integration of code into a code base that aims to be trusted (at least in my view).

dstebila:
> I think we ought to discuss how to improve things here. Any chance we can take some minutes for this in the next status call, @dstebila ?

Done!  https://github.com/open-quantum-safe/tsc/blob/main/oqs-status-meetings/2024-08-27%20-%20OQS%20status%20meeting%20-%20agenda.md

PR #1747, opened 2024-04-04
remove "maximum" words for most length fields

wangweij:
Remove the "(maximum)" words for most length fields. All of them are actually fixed and the APIs do not support variable lengths.

Note: the only variable length is the signature length, where the `OQS_SIG_sign` function returns it.


SWilson4 (approved):
LGTM, thanks for clarifying our documentation. Please address the failing DCO check (should just be a matter of committing with the `-s` option).

praveksharma (approved):

PR #1754, opened 2024-04-13
add compile_commands.json to .gitignore

carsonRadtke:
We set CMAKE_EXPORT_COMPILE_COMMANDS=ON so the compilation database is automatically exported to the build directory. However, many language servers expect this file to be in the project root directory, so a common post-build step is to:

```shell
$ ln -s <build-directory>/compile_commands.json
```

This PR enables developers to do this without having to worry about accidentally commiting a symlink file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,4 +35,5 @@ __pycache__
 .pytest_cache
 .cache
 .CMake/a.out
+compile_commands.json
 


SWilson4 (approved):
I have no objection to this. Thanks for the contribution!

dstebila (approved):

PR #1755, opened 2024-04-16
Stateful sigs: Rename keygen / sign option, add more tests, fix memory errors

SWilson4:
- Rename `OQS_ENABLE_SIG_STFL_KEY_SIG_GEN` to `OQS_EXPERIMENTAL_ENABLE_SIG_STFL_KEY_SIG_GEN`
- Test stateful signatures on macOS
- Enable valgrind tests for stateful signatures
- Fix two memory errors detected by valgrind

One of the memory errors was related to argument parsing in `test_sig_stfl`; it was possible to trigger a segfault. The other arose when keys were written to disk in the same test program: Valgrind complained because the `aux_data` field for an LMS key was partially uninitialized. This seems like a valid bug to me, so I changed the code to zero-initialize `aux_data`.


ashman-p (approved):

src/sig_stfl/lms/sig_stfl_lms_functions.c
```diff
@@ -287,6 +287,7 @@ int oqs_sig_stfl_lms_keypair(uint8_t *pk, OQS_SIG_STFL_SECRET_KEY *sk, const uin
 		OQS_MEM_insecure_free(oqs_key_data);
 		return -1;
 	}
+	memset(aux_data, 0, len_aux_data);
```
cothan, line 290: You can always use `calloc` instead of `malloc` then `memset`.

cothan (approved):

PR #1774, opened 2024-04-29
Errors not printed out when OPENSSL_NO_STDIO is set

bencemali:
The handleErrors function should print out errors when OPENSSL_NO_STDIO is set. The ERR_clear_error function is called instead since ERR_print_errors_fp would empty the error queue.


baentsch (approved):
Looks sensible to me. Thanks for the improved guard, @bencemali !

baentsch:
Please rebase as the current CI error should be fixed as #1775 landed.

dstebila (approved):

SWilson4 (approved):

PR #1769, opened 2024-04-25
Update README.md

vsoftco:
fixed typo in link

* [ ] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [ ] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged.)

PR #1715, opened 2024-03-01
fix documentation generation

baentsch:
Fixes #1711.

* [ ] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [ ] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) and [OQS-OpenSSH](https://github.com/open-quantum-safe/openssh) will also need to be ready for review and merge by the time this is merged.)

PR #1775, opened 2024-04-29
Fix README.md to work with Doxygen release 1.10.0

praveksharma:
Undoes changes from #1769 and re-introduces changes from #1715 to satisfy Doxygen release 1.10.0. 

GitHub markdown provides a few extension in addition to the standard Markdown spec; a valid GitHub Markdown file isn't necessarily a valid Markdown file (see [this comment](https://github.com/doxygen/doxygen/issues/9412#issuecomment-1161591387) by a Doxygen contributer). Changing `#linuxmacos` to either `#linuxmacOS` or `#linux/macOS` worked for me locally; I chose `#linuxmacOS` because that seems to keep the link intact at the moment. 

CI containers currently use Doxygen 1.9.2, [openquantumsafe/ci-containers#76](https://github.com/open-quantum-safe/ci-containers/pull/76) updates them to use Doxygen 1.10.0.


Martyrshot:
Looks good to me. :) Thanks for this

SWilson4 (approved):
Thanks for taking care of this, @praveksharma!

dstebila:
If we were to rename the section title from `Linux/macOS` to `Linux and macOS` would that result in a shortcode that is compatible with both Doxygen and Github?

praveksharma:
> If we were to rename the section title from `Linux/macOS` to `Linux and macOS` would that result in a shortcode that is compatible with both Doxygen and Github?

I considered that. Since both  `#linuxmacOS` and `#linux/macOS` (I didn't state this explicitly but `#linux/macos` does not work) I think the issue is the upper case letters in macOS and not the /.

dstebila:
> > If we were to rename the section title from `Linux/macOS` to `Linux and macOS` would that result in a shortcode that is compatible with both Doxygen and Github?
> 
> I considered that. Since both `#linuxmacOS` and `#linux/macOS` (I didn't state this explicitly but `#linux/macos` does not work) I think the issue is the upper case letters in macOS and not the /.

Okay... would `Linux/Mac` or `Linux and Mac' work?

praveksharma:
> Okay... would `Linux/Mac` or `Linux and Mac` work?

Either would work, I ended up going with `Linux and Mac` since that seems to better follow GitHub Markdown conventions (I could be wrong but I don't think the [GitHub Markdown spec](https://github.github.com/gfm/) or the [GitHub markdown formatting documentation](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#section-links) explicitly address how section links handle non-alphabetic characters and uppercase alphabets).

praveksharma:
I pushed some more changes: the CI needed to be tweaked a little once the CI containers started using the updated images from [openquantumsafe/ci-containers#76](https://github.com/open-quantum-safe/ci-containers/pull/76). CI now uses `scripts/run_doxygen.sh` (which does some formatting to make GitHub markdown work with Doxygen) instead of calling Doxygen directly. I'm not why the older way earlier but doesn't anymore; for what it's worth, the new way of using  `scripts/run_doxygen.sh` follows how `ninja gen_docs` generates documentation.

Martyrshot (approved):
Seems good to me, but I would wait for @SWilson4 @dstebila, and @baentsch to take a look. :)

docs/.Doxyfile
```diff
@@ -341,6 +369,17 @@ MARKDOWN_SUPPORT       = YES
...
+# The MARKDOWN_ID_STYLE tag can be used to specify the algorithm used to
+# generate identifiers for the Markdown headings. Note: Every identifier is
+# unique.
+# Possible values are: DOXYGEN use a fixed 'autotoc_md' string followed by a
+# sequence number starting at 0 and GITHUB use the lower case version of title
+# with any whitespace replaced by '-' and punctuation characters removed.
+# The default value is: DOXYGEN.
+# This tag requires that the tag MARKDOWN_SUPPORT is set to YES.
+
+MARKDOWN_ID_STYLE      = DOXYGEN
```
dstebila, line 381: This comment seems to suggest a new feature controlling how anchors for headers are generated for Markdown files. Would this have solved our problem?

dstebila (approved):

docs/.Doxyfile
praveksharma: I think it might. I'll merge this so CI can pass for other PRs with red CI and create a new PR to address configuring Doxygen more appropriately.

PR #1794, opened 2024-05-15
Add comparsion for shared secret after encapsulation

yogappu:
The NIST provided Intermediate TestVectors testcases for ML-KEM takes encaps_k(shared secret) as an argument 4 but however does not use it for comparison after the encapsulation stage.
This PR adds this check and aborts the code flow incase the comparison fails.


yogappu:
will create a MR soon again with more changes and add @baentsch @SWilson4

Issue #1804, opened 2024-05-31
CI: macOS build failures


planetf1:
Currently macOS builds are failing 
 - [Linux and MacOS tests](https://github.com/open-quantum-safe/liboqs/actions/workflows/unix.yml)
 - Noted in PRs #1708, #1707 
 
 These are failing with gcc on macOS 13 & macOS 14
 
<img width="298" alt="Screenshot 2024-05-31 at 11 09 48" src="https://github.com/open-quantum-safe/liboqs/assets/7292002/31773dde-3821-4221-8319-6acdcc18d899">

For macOS 14 :
| Component | Working | Failed |
| --| -- | -- |
| runner | 2.316.1 | 2.316.1 |
| Xcode | 15.0.1 | 15.0.1 |
| gcc | Apple 13.2.0 | gcc 13.3.0 |
| _example log_ | https://github.com/open-quantum-safe/liboqs/actions/runs/9211032819/job/25339585288|https://github.com/open-quantum-safe/liboqs/actions/runs/9309161112/job/25624155404|

For example
 - [macos-14, -DCMAKE_C_COMPILER=gcc-13](https://github.com/open-quantum-safe/liboqs/actions/runs/9309161112/job/25624155404) - reports the compiler as gcc 13.3.0 (13 hours ago)
 - [macos-14, -DCMAKE_C_COMPILER=gcc-13](https://github.com/open-quantum-safe/liboqs/actions/runs/9211032819/job/25339585288) - reports the compiler as gcc 13.2.0
 
 The current runner software versions are documented [here](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md). 
 
 |  | Working | Failed |
 | -- | -- | -- |
 | Date | 2024-05-15 | 2024-05-29 | 
 | Image Version | 20240514.3 | 20240526.1 |
 | OS Version |13.6.6 (22G630)| 13.6.7 (22G720) |

planetf1:
Additional commit reverts the version reversion ... 

- [macOS 13 gcc](https://github.com/open-quantum-safe/liboqs/actions/runs/9317360310/job/25647584042) - FAIL
- [macOS 14 gcc](https://github.com/open-quantum-safe/liboqs/actions/runs/9317360310/job/25647584850) - FAIL

This seems to prove the cause.

Also note that some annotations are added - just due to stderr when installing direct using ruby. Cosmetic, can be cleaned up if we want to go with a workaround
```
[...]

PR #1805, opened 2024-05-31
Force gcc 13.2.0 over 13.3.0

planetf1:
Attempts to verify cause of recent build failures by patching the gcc compiler for macOS 13.3.0 -> 13.2.0

See issue #1804


planetf1:
Pushed a second commit, which comments out the change. This is to validate that it is this gcc revert that fixes the build -- in case the cause is another external problem

planetf1:
OQS status meeting 2024-06-04 agreed ok to go with workaround, but keep issue open to revert

SWilson4 (approved):

praveksharma (approved):

dstebila (approved):

PR #1816, opened 2024-06-08
CMakeLists: add ppc case to known archs

barracuda156:
This is a trivial PR: it simply adds `ppc` to a list of known archs, retaining a warning re no CI for this platform.

I have confirmed the build on Darwin ppc for `liboqs` 0.10.1 with no other alterations to the code.


SWilson4 (approved):
Thanks for the PR, @barracuda156. This mirrors the current configuration for ppc64, so it seems reasonable to me.

barracuda156:
@SWilson4 I just thought we may need to verify one thing. What is the value of `CMAKE_SYSTEM_PROCESSOR` when the build is done on 10.5 for `ppc64`?
If it uses `uname -p` under the hood, this returns `powerpc`, but the build arch should be `ppc64`.

I cannot test that immediately, since I have no build environment on 10.5, but let me check with CMake documentation.

A fallback would be to use `CMAKE_OSX_ARCHITECTURES` when on Apple, that returns a value matching the arch of the build. (Since gcc does not support fat builds, it will be a single arch always, on PowerPC.)

CMakeLists.txt
```diff
@@ -80,6 +80,13 @@ elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "(ppc64|powerpc64)")
     if(${OQS_DIST_BUILD})
         set(OQS_DIST_PPC64_BUILD ON)
     endif()
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "(ppc|powerpc)")
+    message(WARNING "There is currently no CI for: " ${CMAKE_SYSTEM_PROCESSOR})
```
baentsch, line 84: I do not quite understand this warning: Isn't there ppc CI via Travis, @bhess ? But then again, that CI link in the README is dead. Created #1827 to track.

barracuda156: For the warning, I just reproduced what other cases had. If unneeded, perhaps should be removed accordingly.

barracuda156:
@SWilson4 @baentsch Sorry for a delay, will finalize this over a coming weekend.

CMakeLists.txt
baentsch: No, I'm OK with what you did: Just checking again, I indeed don't find PPC CI and our PLATFORMS.md file does not document such. So, please disregard my comment.

baentsch (approved):

baentsch:
> @SWilson4 @baentsch Sorry for a delay, will finalize this over a coming weekend.

@barracuda156 Please let us know when you're ready. Should this be moved to Draft to avoid an unintended merge?

baentsch:
@barracuda156 This PR is now a bit old and I removed my approval for the time being: Can you please comment (by) when you may have time to move this to completion as discussed above?

barracuda156:
> @barracuda156 This PR is now a bit old and I removed my approval for the time being: Can you please comment (by) when you may have time to move this to completion as discussed above?

@baentsch Yeah, sorry, will finalize this by this weekend.

barracuda156:
Ok, so yeah, CMake will have the same value for ppc and ppc64 builds on Darwin, since it uses `uname -m`: https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html

I will rewrite a patch in a way it gonna work correctly now.

barracuda156:
@baentsch @dstebila Could you please review?

CMakeLists.txt
```diff
@@ -81,6 +81,24 @@ elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "(ppc64|powerpc64)")
...
+    # CMake uses uname to derive CMAKE_SYSTEM_PROCESSOR value, so on Darwin
+    # the value is identical for ppc and ppc64. To have the right build arch
+    # in 64-bit case, we use CMAKE_OSX_ARCHITECTURES.
+    if(APPLE AND CMAKE_OSX_ARCHITECTURES STREQUAL "ppc64")
+        set(ARCH "ppc64")
+        set(ARCH_PPC64 ON)
+        if(${OQS_DIST_BUILD})
+            set(OQS_DIST_PPC64_BUILD ON)
+        endif()
+    else()
```
baentsch, line 95: Is this `else` correct? It seems to set the following also for non-Apple platforms...

baentsch (commented):
Please see the question, @barracuda156 . I also triggered CI...

CMakeLists.txt
barracuda156: @baentsch I believe yes, because other systems using `ppc64` should be taken care of by the case above. So at this point we are left with `ppc32` – plus an odd case of macOS, which supports both ppc and ppc64 in the same OS (while on *BSD the arch of the OS itself must be either ppc or ppc64).

baentsch: Thanks for that rationale @baentsch : Makes sense.

baentsch (approved):
Thanks again for the contribution @barracuda156 !

barracuda156:
Thanks for merging!

Issue #1815, opened 2024-06-07
unknown type OQS_SIG in sig_stfl.h included in oqs.h


lingxinh-aws:
**Describe the bug**
"#include <oqs/sig_stfl.h>" was added in oqs.h in [971773a](https://github.com/open-quantum-safe/liboqs/commit/971173ad82327ede5027b6d48e81bcaff92f417c#diff-9ba35166c0f1d2190622871ceb7f323735a1e7c7078d93994600bd5d85814bf5). It caused an error of "unknown type name 'OQS_SIG' " when compiling oqs-provider/oqsprov/oqs_sig.c, which includessig.h, which includes oqs.h. Then, OQS_SIG is referred in sig_stfl.h, but it is defined after #include <oqs/oqs.h> in the body of sig.h, so compilation failed.

**To Reproduce**
1. Git clone the oqs-provider
2. Run: cd oqs-provider
3. Run:  scripts/fullbuild.sh

**Expected behavior**
OpenSSL, liboqs and oqs-provider are all successfully built.

**Screenshots**
<img width="522" alt="sig_stfl_h_error" src="https://github.com/open-quantum-safe/liboqs/assets/94720905/26334e8f-3d67-4e03-9583-4563d318a377">

**Environment (please complete the following information):**
 - OS [ Amazon Linux 2 ]
 - OpenSSL version [3.3.0-dev]
 - Compiler version used [gcc 7.3.1]
 - Build variables used ["-DCMAKE_BUILD_TYPE=Debug"]
 - liboqs version [main branch]

**Additional context**

baentsch:
Thanks for this report. Problem permeates through the complete OQS stack, see https://github.com/open-quantum-safe/oqs-provider/issues/427 and https://github.com/open-quantum-safe/oqs-demos/issues/281 . Possible work-around [discussed here](https://github.com/open-quantum-safe/oqs-provider/issues/427#issuecomment-2157249645). A better solution of course is needed, see [comment here](https://github.com/open-quantum-safe/liboqs/pull/1817#pullrequestreview-2106361120): downstream-triggering CI should have caught this; still wonder why it didn't (ideas @SWilson4 ?)

SWilson4:
[...]

PR #1820, opened 2024-06-11
Forward-declare OQS_SIG in sig_stfl.h

SWilson4:
When an application (e.g., `oqs-provider`) includes `<oqs/sig.h>`, the following happens:

- `sig.h` includes `oqs.h` _at the top of the file_.
- `oqs.h` includes `sig.h`, but this include is skipped because of the `OQS_SIG_H` include guard.
- `oqs.h` includes `sig_stfl.h`.
- `sig_stfl.h` references `OQS_SIG`, which has not yet been defined.

This breaks the application's build: see https://github.com/open-quantum-safe/liboqs/issues/1815, https://github.com/open-quantum-safe/oqs-provider/issues/427, and https://github.com/open-quantum-safe/oqs-demos/issues/281.

This PR resolves the issue by forward-declaring the OQS_SIG type in `sig_stfl.h`. This fixes the failing `oqs-provider` build: see the downstream tests [before](https://github.com/open-quantum-safe/oqs-provider/actions/runs/9468608791/job/26085334692) and [after](https://github.com/open-quantum-safe/oqs-provider/actions/runs/9468780524/job/26085916757).

I don't believe that the forward-declaration violates any C programming rules / best practices, but it would be great if someone with more experience could confirm this.

### DIFF
--- a/src/sig_stfl/sig_stfl.h
+++ b/src/sig_stfl/sig_stfl.h
@@ -179,6 +179,7 @@ OQS_API int OQS_SIG_STFL_alg_count(void);
 OQS_API int OQS_SIG_STFL_alg_is_enabled(const char *method_name);
 
 #ifndef OQS_ALLOW_STFL_KEY_AND_SIG_GEN
+typedef struct OQS_SIG OQS_SIG;
 #define OQS_SIG_STFL OQS_SIG
 #else
 /**


cothan (approved):
Great work!

praveksharma (approved):

ashman-p:
@SWilson4 Thanks for looking at this and resolving it. 
I did also find that an alternative resolution but i prefer this fix.

PR #1821, opened 2024-06-14
Fix test_alg_info.py on Windows platform

qnfm:
Fix test_alg_info.py on Windows which cause the ninja run_tests failure.

Since `gerp` is not available on Windows, making use of powershell is a good alternative way to replace grep.

### DIFF
--- a/tests/test_alg_info.py
+++ b/tests/test_alg_info.py
@@ -16,7 +16,11 @@ def test_alg_info_kem(kem_name):
     alg_info = yaml.safe_load(output)['KEMs'][kem_name]
     assert(not(alg_info['isnull']))
     # find and load the datasheet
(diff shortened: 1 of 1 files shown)


tests/test_alg_info.py
```diff
@@ -66,5 +74,6 @@ def test_alg_info_sig(sig_name):
 
 if __name__ == "__main__":
     import sys
+    print("platform:",platform.system)
```
baentsch, line 77: Is this necessary?

baentsch (approved):
LGTM. Thanks for bringing in Windows knowhow!

tests/test_alg_info.py
qnfm: Aha, forgot to remove. My fault.

dstebila (approved):

SWilson4 (approved):
Thank you very much for the contribution!

Issue #1753, opened 2024-04-10
Update CBOM to CycloneDX 1.6 specification


bhess:
The specification used for [CBOMs](https://github.com/open-quantum-safe/liboqs/blob/main/docs/cbom.json) has been upstreamed to CycloneDX 1.6:
https://github.com/CycloneDX/specification/releases/tag/1.6

This issue is to update the CBOM in liboqs, and its generation script to the CycloneDX 1.6 specification.

baentsch:
As per https://github.com/open-quantum-safe/liboqs/pull/1708#issuecomment-2069210942 also document CBOM utility, use and development/developer's implications.

planetf1:
See also https://github.com/open-quantum-safe/liboqs/issues/1831

I missed this issue when originally opening the above - apologies. I added some observations. We can close one, though I did try to add my non-expert comments. @bhess let me know if you want any help with this.

dstebila:
Closing this issue as a dupe of #1831.

PR #1834, opened 2024-07-09
Update CBOM format to upstream v1.6

### LINKED ISSUE #1831: Update CBOM to CycloneDX 1.6 format
## State of current BOM:

The current BOM at `docs/cbom.json` mostly validates against the [1.4 spec](https://github.com/IBM/CBOM/blob/main/bom-1.4-cbom-1.0.schema.json) with just 1 issue

* metadata.,*timestamp* value of `2024-03-05T11:49:42.428605` fails with `String is not a RFC3339 date-time.`

In [RFC3339](https://www.rfc-editor.org/rfc/rfc3339) a time-offset is required, ie can be `Z`.

Using `2024-03-05T11:49:42.428605Z` passes validation.

This change is needed in the tooling that builds/tests the BOM

## Improvement to include pointer to schema

Many BOMs include a pointer to the spec ie we could add
```
  "$schema": "https://raw.githubusercontent.com/IBM/CBOM/main/bom-1.4-cbom-1.0.schema.json#",
```

Aside - as an alternative, and for testing, the following can be added in .vscode settings:
```
    "json.schemas": [
        

        {
            "fileMatch": [
                "cbom.json"
            ],
            "url": "https://raw.githubusercontent.com/IBM/CBOM/main/bom-1.4-cbom-1.0.schema.json"
        }
    ]
```

## Cyclone DX 1.6

CBOM is now incorporated into CycloneDX 1.6, but here the schema as evolved a little

We can validate against the [new spec](http://cyclonedx.org/schema/bom-1.6.schema.json) - the example below includes the vscode definition:

```
    "json.schemas": [


        {
            "fileMatch": [
                "cbom.json"
            ],
            "url": "http://cyclonedx.org/schema/bom-1.6.schema.json"
        }
    ]
```

If `$schema` is included in the bom, then other tools wouldn't need to explicitly be configured, and should find the schema definition automatically

In CycloneDX 1.6 the following is noted:

* ` "bomFormat": "CBOM",` -> can only be 'CycloneDX'
* `"type": "crypto-asset",` -> 'cryptographic-asset'
* In algorithmProperties, variant & implementationLevel are not allowed. where the last segment of Variant includes a platform like `x86_64` this probably is now best matched to `implementationPlatform` , whilst implementationLevel is `executionEnvironment`
* in cryptoProperties, nistQuantumSecurityLevel is not allowed - it has moved under `algorithmProperties`
* `"primitive": "blockcipher",` -> `block-cipher`
* "dependencyType": "uses" - not allowed. there is a new property `provides` which seems to make sense where an algorithm is provided, whilst just having a dependency presumes uses ..?

@bhess @dstebila I think it would be useful to consider this update now that the 1.6 spec is out, and we're seeing more tooling support CBOM.

Happy to work on some updates if it's helpful - or perhaps you plan to @bhess ?

bhess:
Updates the CBOM generated to the format from the official CycloneDX 1.6 release:
https://cyclonedx.org/docs/1.6/json/

Closes #1831 #1753 

### DIFF
--- a/docs/cbom.json
+++ b/docs/cbom.json
@@ -1,3135 +1,3134 @@
 {
-  "bomFormat": "CBOM",
-  "specVersion": "1.4-cbom-1.0",
-  "serialNumber": "urn:uuid:b953d460-1246-4cbb-aff9-642a0308d18b",
+  "$schema": "https://raw.githubusercontent.com/CycloneDX/specification/1.6/schema/bom-1.6.schema.json",
+  "bomFormat": "CycloneDX",
+  "specVersion": "1.6",
+  "serialNumber": "urn:uuid:679a1c2f-1dd4-4692-b098-1d4dacdfc75d",
   "version": 1,
   "metadata": {
-    "timestamp": "2024-08-26T18:04:44.668645",
+    "timestamp": "2024-10-01T14:21:06.857613+00:00",
     "component": {
       "type": "library",
-      "bom-ref": "pkg:github/open-quantum-safe/liboqs@062e793edf54cbc1073b54d0689795063fd41910",
+      "bom-ref": "pkg:github/open-quantum-safe/liboqs@bf7cbdca1bf2866da22b4b75e04b68baf6707a7b",
       "name": "liboqs",
-      "version": "062e793edf54cbc1073b54d0689795063fd41910"
+      "version": "bf7cbdca1bf2866da22b4b75e04b68baf6707a7b"
     }
   },
   "components": [
     {
       "type": "library",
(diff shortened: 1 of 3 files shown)


SWilson4:
@bhess @planetf1 Any plans to move this PR forward?

bhess:
Thanks for the reminder @SWilson4, yes I'll update the PR.

bhess:
Rebased and made the PR ready for review.

dstebila (approved):

SWilson4 (approved):

Issue #1838, opened 2024-07-10
Decode FORS indices similarly to WOTS


GillonB:
Can we add coming changes to SPHINCS+ implementation to liboqs? Could be as a temporary branch as done in  https://github.com/sphincs/sphincsplus/pull/51. 

This change breaks forward/backward compatibility between old/new signatures.

SWilson4:
Closing as this is superseded by https://github.com/open-quantum-safe/liboqs/issues/1894.

PR #1839, opened 2024-07-10
Decode FORS indices similarly to WOTS

GillonB:
Update the decoding of the FORS indices as proposed by latest NIST changes. This has already been proposed in the sphincsplus repo (https://github.com/sphincs/sphincsplus/pull/51). Would be good to have the same for liboqs as this change will eventually land. https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/88tuvtb7nN4/m/DA1QCoJWBAAJ

* This changes breaks SPHINCS+ signature compatibility: Signatures generated before cannot be verified correctly with the implementation after this change. Also old implementation cannot verify correctly newly generate signatures.
* No change to available algorithms

relates to #1838


dstebila:
Because we're pulling out SPHINCS+ implementation from PQClean, we'd prefer to see this change made first in PQClean, and then our automated scripts can pull it in.  Would you be able to do that?

Issue #1847, opened 2024-07-18
Running `cmake -LA` changes library configuration


SWilson4:
**Describe the bug**
If `cmake -LA ..` is run between the library configuration (`cmake -GNinja <config options> ..`) and build (`ninja`) commands, the library configuration may be changed.

**To Reproduce**
From the `liboqs` directory:
```sh
rm -rf build && mkdir build && cd build
cmake -GNinja -DOQS_MINIMAL_BUILD="KEM_kyber_512;SIG_dilithium_2" -DCMAKE_BUILD_TYPE=Debug ..
ninja
gdb -ex "b pqcrystals_dilithium2_avx2_keypair" -ex r --args ./tests/test_sig Dilithium2
```

![avx2_enabled](https://github.com/user-attachments/assets/c0d800c7-9ff6-4a60-bda9-13f1ab376f6f)

As expected, the breakpoint is hit: the AVX2 code is enabled in the build.

Here is the exact same command sequence with `cmake -LA ..` added between config and build:
```sh
rm -rf build && mkdir build && cd build
cmake -GNinja -DOQS_MINIMAL_BUILD="KEM_kyber_512;SIG_dilithium_2" -DCMAKE_BUILD_TYPE=Debug ..
cmake -LA ..
ninja
gdb -ex "b pqcrystals_dilithium2_avx2_keypair" -ex r --args ./tests/test_sig Dilithium2
```

![avx2_disabled](https://github.com/user-attachments/assets/0ad3c11e-e6ee-4560-803f-edf72ac9df5a)

The breakpoint cannot be set: the AVX2 code is disabled. Also note that the number of targets built is greater in the first case (127) than in the second (101).

**Expected behavior**
The `cmake -LA ..` call should have no side effects.

**Screenshots**
See above

**Environment (please complete the following information):**
 - OS: Ubuntu 22.04
 - OpenSSL version: 3.0.2
 - Compiler version used: GCC 11.4.0
 - Build variables used: see above
 - liboqs version: `main`, at commit 5670edf59fa7a910654b9171229b7dc1ea7ece4c

**Additional context**
[...]

PR #1848, opened 2024-07-19
Use `cmake -LA -N` instead of `cmake -LA` in CI

SWilson4:
To ensure that printing config variables doesn't have undesired side effects.

Mitigates #1847.


SWilson4:
Note that all CI tests passed on e7071a7033202f6872ba8fd5fb5c42902b84eec0, the subsequent commit simply removes a reference to `cmake -LA` (without `-N`) from documentation.

dstebila (approved):

baentsch (approved):

Issue #1783, opened 2024-05-07
Remove unnecessary downstream CI


baentsch:
Current [CI failures](https://app.circleci.com/pipelines/github/open-quantum-safe/liboqs/3701/workflows/bc6f3080-474b-4e6b-aa63-98a1e866bd15/jobs/29065) indicate the CI pipeline triggers CI on downstream projects that are not maintained any more. This is in direct contradiction to https://github.com/open-quantum-safe/tsc/issues/5. But it probably also points to the need to finally tackle https://github.com/open-quantum-safe/tsc/issues/2 (tagging @dstebila @bhess for scheduling discussion/resolution).

PR #1849, opened 2024-07-19
Move from CircleCI to GitHub Actions

### LINKED ISSUE #1795: Move from CCI to GH CI
[...] move them over at your leisure.

_Originally posted by @ryjones in https://github.com/open-quantum-safe/liboqs/issues/1789#issuecomment-2111144389_

SWilson4:
This PR moves our remaining CircleCI workflows over to GitHub Actions. I've left it in draft for now as it's based on the branch for #1848.

I have tried to duplicate the testing coverage from CircleCI without additions or removals. This PR isn't intended to overhaul or evaluate our test coverage, it's just meant to move it to a different platform.

Three items of note:

(1) The CCI [`bionic-i386`](https://github.com/open-quantum-safe/liboqs/blob/60cd3d83c92ae2679df3460955921570a6e0f05d/.circleci/config.yml#L297) could not be ported over, as the GitHub `checkout` action requires a newer version of `node` and can't run in the older container (at least not without some finicky work). Bionic is no longer supported anyhow, so it's probably for the best that we're no longer using it for testing. However, the multiarch images that we'd been using don't have options for i386 past Bionic, so there is no easy way to continue testing on i386. We don't list Linux / x86 as a [supported platform](https://github.com/open-quantum-safe/liboqs/blob/60cd3d83c92ae2679df3460955921570a6e0f05d/PLATFORMS.md), so I've opted to simply remove CI support for it.

(2) The CCI `focal-clang-15` job had `-DOQS_OPT_TARGET=skylake` set. It didn't configure `-DOQS_DIST_BUILD=OFF`, however, so the opt target value was ignored. I have omitted the opt target variable in GitHub CI to reflect what was actually being tested.

(3) Finally, @ryjones helped me add a CircleCI API token to the liboqs environment so that GitHub can trigger CircleCI for downstream projects. These triggers are contained in the new `commit-to-main.yml` file. We may want to revise them for completeness, but for now I have left them as they were on CCI.

Fixes #1795.


.github/workflows/commit-to-main.yml
```diff
@@ -0,0 +1,87 @@
...
+    steps:
+      # TODO: missing projects?
+      - name: Trigger OQS-OpenSSL CI
+        run: |
+          curl --silent \
+               --write-out "\n%{response_code}\n" \
+               --user ${BUILD_TRIGGER_TOKEN}: \
+               --request POST \
+               --header "Content-Type: application/json" \
+               --data '{ "branch": "OQS-OpenSSL_1_1_1-stable", "parameters": { "run_downstream_tests": true } }' \
```
baentsch, line 24: Doesn't look necessary to keep this one

.github/workflows/commit-to-main.yml
```diff
@@ -0,0 +1,87 @@
...
+               https://circleci.com/api/v2/project/gh/open-quantum-safe/boringssl/pipeline | tee curl_out \
+          && grep -q "201" curl_out
+      - name: Trigger OQS-OpenSSH CI
+        run: |
+          curl --silent \
+               --write-out "\n%{response_code}\n" \
+               --user ${BUILD_TRIGGER_TOKEN}: \
+               --request POST \
+               --header "Content-Type: application/json" \
+               --data '{ "branch": "OQS-v8", "parameters": { "run_downstream_tests": true } }' \
```
baentsch, line 44: When to add/change this to v9?

.github/workflows/commit-to-main.yml
```diff
@@ -0,0 +1,87 @@
...
+          && grep -q "201" curl_out
+      - name: Trigger liboqs-java CI
+        run: |
+          curl --silent \
+               --write-out "\n%{response_code}\n" \
+               --user ${BUILD_TRIGGER_TOKEN}: \
+               --request POST \
+               --header "Content-Type: application/json" \
+               --data '{ "branch": "master" }' \
+               https://circleci.com/api/v2/project/gh/open-quantum-safe/liboqs-java/pipeline | tee curl_out \
```
baentsch, line 75: Doesn't look as necessary as triggering liboqs-rust, does it?

.github/workflows/unix.yml
```diff
@@ -85,7 +85,7 @@ jobs:
           - name: arm64
             runner: oqs-arm64
             container: openquantumsafe/ci-ubuntu-focal-arm64:latest
-            PYTEST_ARGS: --numprocesses=auto --maxprocesses=10 --ignore=tests/test_kat_all.py
+            PYTEST_ARGS: --maxprocesses=10 --ignore=tests/test_kat_all.py
```
baentsch, line 88: Why did the 'auto' (# of processes started) param get dropped?

baentsch (approved):

.github/workflows/commit-to-main.yml
dstebila: The v9 branch doesn't yet work; the [current PR from @geedo0](https://github.com/open-quantum-safe/openssh/pull/160) gets the basic test suite running but still doesn't have full PQ or hybrid support.  I think we can merge this PR with v8, and then when v9 is ready, we can come back and update the liboqs CI.

dstebila (approved):

.github/workflows/unix.yml
SWilson4: It's still in effect; I just factored it out into the `python3 -m pytest ...` command below. It was used almost universally, so I figured that there was no point in duplicating it across all jobs.

.github/workflows/commit-to-main.yml
SWilson4: Or liboqs-go... My intent with this PR was just to duplicate existing CCI functionality as closely as possible to make sure it worked on GH and make any improvements in a follow-up PR. But I could do both in one shot, if that would be preferable from a review standpoint.

.github/workflows/commit-to-main.yml
SWilson4: Same comment as below re liboqs-rust applies here.

.github/workflows/commit-to-main.yml
SWilson4: @baentsch I'll interpret the thumbs-up as an endorsement of a follow-up PR to address #1783 and merge this as is.

SWilson4:
The [downstream trigger](https://github.com/open-quantum-safe/liboqs/actions/runs/10080975646) failed after the merge. I can trigger downstream pipelines locally with the same API token that should be in CI, so I think this is a GitHub variables config error. Will follow up with @ryjones later today.

PR #1854, opened 2024-07-23
ML-KEM NIST tests, fix order of d and z

### LINKED ISSUE #1853: Incorrect Order of z and d in NIST tests
In the NIST tests for ML-KEM, the random number stream is constructed as [z || d || m] i.e. the first random byte returned by randombytes() would be z followed by d and m.

However, in the code

64 bytes random is generated together https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/ml_kem/pqcrystals-kyber-standard_ml-kem-512-ipd_ref/kem.c#L54
and then passed to derand api for keypair generation.

In the derand api, first 32 byte is used for keypair generation https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/ml_kem/pqcrystals-kyber-standard_ml-kem-512-ipd_ref/kem.c#L29
which makes the first 32 byte as 'd' as per Algo 12 of FIPS-203.

The next 32 bytes are appended to end of private key https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/ml_kem/pqcrystals-kyber-standard_ml-kem-512-ipd_ref/kem.c#L33 
which makes it 'z' as per Algo 15.

So, the random byte stream construction should be [d || z || m] so that randombytes are output in the order of d->z->m

Currently, it works for NIST Intermediate Vectors as the value of z and d are same for all KEM version. Incase, they are different the tests would fail.

A simple switch of z and d at https://github.com/open-quantum-safe/liboqs/blob/main/tests/test_kem_vectors.sh#L26 should solve this IMO.

bhess:
Thank you for catching the issue @abhinav-thales! Could you please confirm if this fixes the issue?

Fixes #1853

### DIFF
--- a/tests/test_kem_vectors.sh
+++ b/tests/test_kem_vectors.sh
@@ -23,7 +23,7 @@ keygen_sk=$(grep "keygen_dk: " "$file")
 encaps_c=$(grep "encaps_c: " "$file")
 encaps_K=$(grep "encaps_K: " "$file")
 
-output=$($build_dir/tests/vectors_kem $1 "$keygen_z$keygen_d$encaps_m" "$encaps_ek" "$encaps_k" "$decaps_dk" "$decaps_c" "$decaps_kprime")
+output=$($build_dir/tests/vectors_kem $1 "$keygen_d$keygen_z$encaps_m" "$encaps_ek" "$encaps_k" "$decaps_dk" "$decaps_c" "$decaps_kprime")
 if [ $? != 0 ]; then
     echo "$output"
     exit 1
--- a/tests/vectors_kem.c
+++ b/tests/vectors_kem.c
@@ -222,7 +222,7 @@ int main(int argc, char **argv) {
 	}
 
 	char *alg_name = argv[1];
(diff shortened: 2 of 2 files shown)


dstebila (approved):

SWilson4 (approved):

yogappu:
Looks good @bhess . 

may be a comment here could also help : https://github.com/open-quantum-safe/liboqs/blob/main/tests/vectors_kem.c#L225

abhinav-thales:
changes look good @bhess 

may be  a comment update [here ](https://github.com/open-quantum-safe/liboqs/blob/main/tests/vectors_kem.c#L225) could also help.

bhess:
> may be a comment update [here ](https://github.com/open-quantum-safe/liboqs/blob/main/tests/vectors_kem.c#L225) could also help.

Updated in [2f630e0](https://github.com/open-quantum-safe/liboqs/pull/1854/commits/2f630e019d0a900806d12d73a815779165a63466), thanks!

Issue #1867, opened 2024-07-26
Add CodeQL query to enforce OpenSSL return code handling


SWilson4:
To prevent issues similar to #1862.

@trailofbits used the following query to detect #1862:
```ql
import  cpp

from  FunctionCall  call, Function f 
where
  f = call.getTarget()  and 
  // function name starts with EVP 
  f.getName().matches(  "EVP%"  )  and 
  // and the function does not return a pointer or void 
  not  f.getType() instanceof PointerType  and 
  not  f.getType() instanceof VoidType  and 
  // and the function is not guarded by the OQS_OPENSSL_GUARD macro 
  not  exists(MacroAccess m | 
    m.getLocation().subsumes(call.getLocation())  and 
    m.getMacroName() =  "OQS_OPENSSL_GUARD" 
  )
select call
```

Vishnu2707:
Working on this, will add a CodeQL workflow and custom query file based on the Trail of Bits query provided in the issue description.

Issue #1866, opened 2024-07-26
Add CI workflow analysis tooling


SWilson4:
This will help to prevent (potentially serious) errors of a similar nature to #1863. Suggestions include [actionlint](https://github.com/rhysd/actionlint) and [poutine](https://github.com/boostsecurityio/poutine).

Reported by @trailofbits in Week 1 of their audit of `liboqs`.

SWilson4:
Fixed by https://github.com/open-quantum-safe/liboqs/pull/1916.

PR #1869, opened 2024-07-26
Quick fixes from Trail of Bits audit Week 1

### LINKED ISSUE #1862: Missing OpenSSL guards in common code
Calls to the OpenSSL EVP API are not being error-checked in the SHA2 code. We should be using the `OQS_OPENSSL_GUARD` macro here.

https://github.com/open-quantum-safe/liboqs/blob/4cc88845e8bb66dca20cd61ff88c2162f38c3230/src/common/sha2/sha2_ossl.c#L16-L25

We could also use the macro here instead of error-checking manually:

https://github.com/open-quantum-safe/liboqs/blob/4cc88845e8bb66dca20cd61ff88c2162f38c3230/src/common/rand/rand_nist.c#L59-L70

Reported by @trailofbits in Week 1 of their audit of `liboqs`.

### LINKED ISSUE #1863: Undefined arguments in CI workflows
https://github.com/open-quantum-safe/liboqs/blob/4cc88845e8bb66dca20cd61ff88c2162f38c3230/.github/workflows/weekly.yml#L41-L67

https://github.com/open-quantum-safe/liboqs/blob/4cc88845e8bb66dca20cd61ff88c2162f38c3230/.github/workflows/unix.yml#L273-L273

Reported by @trailofbits in Week 1 of their audit of `liboqs`.

### LINKED ISSUE #1865: Broken link and misplaced comment in common.c
The link in the following comment is broken:

https://github.com/open-quantum-safe/liboqs/blob/4cc88845e8bb66dca20cd61ff88c2162f38c3230/src/common/common.c#L40-L45

The comment currently on line 108 refers to the code currently at line 113 and should be moved for clarity.

https://github.com/open-quantum-safe/liboqs/blob/4cc88845e8bb66dca20cd61ff88c2162f38c3230/src/common/common.c#L107-L114

Reported by @trailofbits in Week 1 of their audit of `liboqs`.

SWilson4:
Fixes #1862
Fixes #1863 
Fixes #1865 

### DIFF
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -24,12 +24,12 @@ extern "C" {
  * Macro for terminating the program if x is
  * a null pointer.
  */
-#define OQS_EXIT_IF_NULLPTR(x, loc)    \
-    do {                          \
-        if ( (x) == (void*)0 ) {  \
+#define OQS_EXIT_IF_NULLPTR(x, loc)                                                   \
+    do {                                                                              \
+        if ( (x) == (void*)0 ) {                                                      \
             fprintf(stderr, "Unexpected NULL returned from %s API. Exiting.\n", loc); \
-            exit(EXIT_FAILURE); \
-        }  \
+            exit(EXIT_FAILURE);                                                       \
+        }                                                                             \
     } while (0)
 
 /**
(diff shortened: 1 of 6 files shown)


SWilson4:
The most recent commit is tagged with `[skip ci]`, but note that CI passed on the actual code changes on commit [15fcdbe](https://github.com/open-quantum-safe/liboqs/pull/1869/commits/15fcdbec6f4a9a38acf587cfdfeff713fe27201c).

bhess (approved):
Thanks @SWilson4

src/common/common.h
```diff
@@ -43,13 +43,24 @@ extern "C" {
...
+        if( 1 != (x) ) {                                                               \
             fprintf(stderr, "Error return value from OpenSSL API: %d. Exiting.\n", x); \
-            exit(EXIT_FAILURE); \
-        }                       \
+            OSSL_FUNC(ERR_print_errors_fp)(stderr);                                    \
+            exit(EXIT_FAILURE);                                                        \
+        }                                                                              \
     } while (0)
+#else
+#define OQS_OPENSSL_GUARD(x)                                                           \
```
baentsch, line 56: Is the following #define triggered if "OQS_USE_OPENSSL" is NOT set? Wouldn't seem quite right.

baentsch (approved):
One nit (see separate comment); otherwise LGTM. Thanks for the improvements @SWilson4 .

I'm not seeing new CI catching these types of errors in the future: Am I looking in the wrong place?

SWilson4:
> One nit (see separate comment); otherwise LGTM. Thanks for the improvements @SWilson4 .
> 
> I'm not seeing new CI catching these types of errors in the future: Am I looking in the wrong place?

Not included in this PR, but I opened #1866 and #1867 for future-proofing based on suggestions from ToB.

src/common/common.h
SWilson4: Good catch. I've refactored the directives so that the guard is defined if and only if `OQS_USE_OPENSSL` is defined.

ashman-p (approved):

PR #1890, opened 2024-08-13
#1830 update scorecard to v5 (gh action 2.4.0)

### LINKED ISSUE #1830: scorecard: update to version 5
The openSSF scorecard execution is currently running with version 4 of scorecard

Once v5 goes GA, we should update to v5 (currently at v5.0.0-rc2)

See https://github.com/ossf/scorecard/releases

planetf1:
- Updates to use scorecard v5
 - re-pinned actions in scorecard.yaml to latest levels
 - fixed report of unpinned dependencies in unix.yml
 - enabled publishing of scorecard results

Note: Results are published by the openssf team at https://scorecard.dev/viewer/?uri=github.com%2Fopen-quantum-safe%2Fliboqs
Enabling our own scan is recommended by openssf, allows us to enable badges, and permits some tweaking of rules
The current scan results are [here](https://scorecard.dev/viewer/?uri=github.com%2Fopen-quantum-safe%2Fliboqs)

Fixes #1830 

n/a

No change to crypto. This is build pipeline only.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -47,7 +47,7 @@ jobs:
           # For private repositories:
           #   - `publish_results` will always be set to `false`, regardless
           #     of the value entered here.
-          publish_results: false
+          publish_results: true
 
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
@@ -56,9 +56,9 @@ jobs:
         with:
           name: SARIF file
           path: results.sarif
-          retention-days: 5
+          retention-days: 28
       # Upload the results to GitHub's code scanning dashboard.
-      #- name: "Upload to code-scanning"
-      #  uses: github/codeql-action/upload-sarif@e949a1676c32f4c215780f7429eb9f00ff18b225 # pin@v2
-      #  with:
-      #    sarif_file: results.sarif
+      - name: "Upload to code-scanning"
+        uses: github/codeql-action/upload-sarif@2c779ab0d087cd7fe7b826087247c2c81f27bfa6 # pin@v3
+        with:
+          sarif_file: results.sarif
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -9,6 +9,7 @@ on:
   workflow_call:
   workflow_dispatch:
 
+
 jobs:
   analysis:
     name: Scorecard analysis
@@ -24,12 +25,12 @@ jobs:
 
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # pin@v4
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # pin@v4
         with:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@0864cf19026789058feabb7e87baa5f140aac736 # pin@v2.3.1
+        uses: ossf/scorecard-action@62b2cac7ed8198b15735ed49ab1e5cf35480ba46 # pin@v2.4.0
         with:
           results_file: results.sarif
           results_format: sarif
@@ -51,7 +52,7 @@ jobs:
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
       - name: "Upload artifact"
-        uses: actions/upload-artifact@1746f4ab65b179e0ea60a494b83293b640dd5bba # pin@v4
+        uses: actions/upload-artifact@834a144ee995460fba8ed112a2fc961b36a5ec5a # pin@v4
         with:
           name: SARIF file
           path: results.sarif
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -259,7 +259,7 @@ jobs:
     container: openquantumsafe/ci-ubuntu-latest:latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # pin@v4
       - name: Configure
         run: mkdir build && cd build && scan-build --status-bugs cmake -GNinja ..
       - name: Build


.github/workflows/scorecard.yml
```diff
@@ -3,16 +3,17 @@ name: Scorecard supply-chain security
 permissions: read-all
 
 on:
-  # For Branch-Protection check. Only the default branch is supported. See
-  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
-  branch_protection_rule: # To guarantee Maintained check is occasionally updated. See
+  branch_protection_rule:
+    # To guarantee Maintained check is occasionally updated. See
+    # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
```
baentsch, line 8: Err -- please tell me that this is not a cronjob aimed at creating (just) the impression of "maintained"-ness, @planetf1 : If it were, I'd violently demand this to be dropped: If no-one maintains the project this fact should become visible. Otherwise this whole scorecard thing really is just a hoax.

.github/workflows/scorecard.yml
```diff
@@ -3,16 +3,17 @@ name: Scorecard supply-chain security
...
 
-  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
   schedule:
     - cron: '29 9 * * 4'
   push:
     branches: [ "main" ]
-  pull_request:
+  # Enable for testing only 
+  #pull_request:
```
baentsch, line 15: Testing only code should not be in "main" should it be? Otherwise, please document with some more words how this would be used in testing.

.github/workflows/scorecard.yml
```diff
@@ -51,18 +52,18 @@ jobs:
           # For private repositories:
           #   - `publish_results` will always be set to `false`, regardless
           #     of the value entered here.
-          publish_results: false
+          publish_results: true
```
baentsch, line 55: Well, it seems this value is pointless given the results are published anyway, so what's the difference between true and false here?

baentsch (dismissed):
Please re-base and check the single comments.

.github/workflows/scorecard.yml
dstebila: If that is indeed what this action is doing, then agreed -- having an artificial commit to meet a metric seems like a bad idea.

planetf1: This code **isn't performing a commit**, it's just [triggering if the branch protection rules are updated](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#branch_protection_rule) so that the scorecard can be refreshed. 

The comment isn't the best placed and has likely moved a little after merges - it's a generic comment (or could be removed).

However the comment is accurate, in that the [check it refers to](https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained) i is looking at recent commits and using that to determine an overall score. There is no particular threshold. .. and yes it can be gamed by introducing the noise of a unnecessary commit.

I agree the whole area of these 'heuristics' in general is debated a lot (see the [chaoss](https://chaoss.community/) community!)

Some organizations may use these metrics to understand more about a project. Overall we score fairly highly...

liboqs is in fact being scanned by openssf directly - the results are [here](https://securityscorecards.dev/viewer/?uri=github.com/open-quantum-safe/liboqs)

The action allows a few more checks to work, enables badging, raises events (security) for tracking - and in fact is needed for smaller projects they've yet to scan.

.github/workflows/scorecard.yml
planetf1: I can clean up the comment.

By 'testing' I meant 'testing of the action' - the intent with scorecard is to only run it on code merged into main. However when we were getting started with it, and particularly the requirement to mitigate some of the issues,  the checks were done during a PR for expediency

.github/workflows/scorecard.yml
planetf1: true means the results from this scan are published, rather than a [weekly scan done by the openssf team]. This includes a few extra checks, access to a badge, and raising of any issues found in the 'security tab'

Initially they weren't doing this on any of our repos. today AFAIK it's liboqs only.

planetf1:
The notice above relates to the scorecard analysis being available as entries under the 'security' tab.

.github/workflows/scorecard.yml
```diff
@@ -24,12 +25,12 @@ jobs:
 
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # pin@v4
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # pin@v4
```
SWilson4, line 28: The hash has updated here, but the pin comment remains the same—is this still v4?

.github/workflows/scorecard.yml
```diff
@@ -46,18 +47,18 @@ jobs:
...
           #   - `publish_results` will always be set to `false`, regardless
           #     of the value entered here.
-          publish_results: false
+          publish_results: true
 
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
       - name: "Upload artifact"
-        uses: actions/upload-artifact@1746f4ab65b179e0ea60a494b83293b640dd5bba # pin@v4
+        uses: actions/upload-artifact@834a144ee995460fba8ed112a2fc961b36a5ec5a # pin@v4
```
SWilson4, line 55: Same comment as above: is this still v4?

SWilson4 (approved):
Looks good to me. I don't think it's an issue to merge this before we release 0.11.0, since it only touches CI (and even that only minimally), but it would be good to get a second opinion on that.

.github/workflows/scorecard.yml
planetf1: yes. There's some info as to how this works in [PROCEDURES.md](https://github.com/open-quantum-safe/liboqs/blob/main/docs/PROCEDURES.md#github-actions)

.github/workflows/scorecard.yml
planetf1: as above

planetf1:
@SWilson4 though of course I want to get this merged, I'd say leave until after release. It just feels with days to go we should minimize any unncessary changes so seems a good general practice

dstebila (approved):

Issue #1888, opened 2024-08-09
Re-activate Travis CI for Tier-3 IBM platforms


SWilson4:
Travis CI builds have been failing with no output for some time now. Pinging @bhess, who I believe maintains the Travis setup.

bhess:
Thanks for reporting @SWilson4 . I've created a ticket with TravisCI about the issue.

bhess:
Opened #1901 to disable TravisCI. I'd suggest to keep this issue open to allow to revert the change if the ticket at TravisCI is fixed.

SWilson4:
In https://github.com/open-quantum-safe/liboqs/pull/1909 I propose that we remove s390x and powerpc from PLATFORMS.md, since we presumably won't be able to run release tests on those platforms for 0.11.0. We will want to add them back whenever Travis is up and running again.

dstebila:
Has this been resolved by #1988?

SWilson4:
Indeed it has.

PR #1901, opened 2024-08-22
Disable erroring TravisCI build

bhess:
Since TravisCI is failing since 2 weeks (#1888) and the ticket there is still under investigation, I'd suggest to disable it temporarily to not longer interrupt the liboqs CI pipeline.


baentsch (approved):
Absolutely agree -- big surprise as this is actually implementing part of the proposal in https://github.com/open-quantum-safe/liboqs/pull/1745#issuecomment-2303989757 :). The completion on this line item would be creation of an issue to add Travis back eventually (if someone cares about those platforms).

SWilson4 (approved):

PR #1919, opened 2024-09-11
Add ML-DSA / FIPS 204 final

### LINKED ISSUE #1891: ML-DSA: integrate final standard
The final standards were published today: [FIPS203](https://csrc.nist.gov/pubs/fips/203/final) and [FIPS204](https://csrc.nist.gov/pubs/fips/204/final).

This is to track the integration of the updated algorithms.

I'd suggest the following steps:

- Pull update from pq-crystals upstream once available
- The current ML-DSA-ipd/ML-KEM-ipd implementations would be dropped
- Integrate test vectors from NIST [ACVP repository](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files)

update (Aug 22):
- pq-crystals/kyber [upstream](https://github.com/pq-crystals/kyber) has been updated
- ML-KEM / ML-DSA OIDs have been [published](https://csrc.nist.gov/projects/computer-security-objects-register/algorithm-registration)
- [ACVP repository](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files) has been updated
- #1899 integrates ML-KEM and ACVP vectors

Update (Aug 27):
ML-KEM landed in main, remaining is ML-DSA integration

Update (Sep 5):
ML-DSA is available upstream: https://github.com/pq-crystals/dilithium

bhess:
Adds ML-DSA from [FIPS204](https://csrc.nist.gov/pubs/fips/204/final)

Pulls ML-DSA from pq-crystals/dilithium upstream: https://github.com/pq-crystals/dilithium
Removes ML-DSA-ipd alias in favor of ML-DSA.

TODOs:

- [x] Awaiting internal API upstream to be able to run NIST ACVP tests: https://github.com/pq-crystals/dilithium/issues/88#issuecomment-2333831981
- [x] ~NIST ACVP vectors against the external API (that will be exposed to the user), expected by October: https://github.com/open-quantum-safe/liboqs/issues/1891#issuecomment-2320297333~ Postponed for ACVP availability
- [x] ~Pre-hash variants / Algorithm 4/5 in [FIPS 204](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf)~ Postponed for ACVP availability
- [x] oqs-provider "-tracker" branch
- [x] add API to be able to pass 'ctx' string
- [x] update constant-time tests

FIPS 204 further defines a context parameter 'ctx' passed to the sign/verify API. This is currently not supported with the OQS API. It also defines
> By default, the context is the empty string, though applications may specify the use of a non-empty context string.

This PR makes only this default case available. Asking for feedback if we should adapt/add to the OQS Sign API to accomodate such a 'ctx' string, or if we restrict support to the default case.

Closes #1891


bhess:
A progress update:
1. The "internal" API was integrated in PQ-Crystals upstream last Friday: https://github.com/pq-crystals/dilithium/commit/444cdcc84eb36b66fe27b3a2529ee48f6d8150c2. These are algorithms 7&8 in [FIPS 204](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf).
2. This PR makes the internal API available for the ACVP tests. Tests are integrated and are passing.
3. Added a new API for signing and verifying that allows to pass an optional "context" string as specified in FIPS 204 (Algorithms 2&3).

The context parameter is needed by some use cases like for [composite signatures](https://github.com/open-quantum-safe/liboqs/issues/1891#issuecomment-2447548756). 
For this purpose this PR introduces two APIs:
- https://github.com/open-quantum-safe/liboqs/blob/444bccf9debb1d1d6ceec65c49474f0e23a84a66/src/sig/sig.h#L307-L324
- https://github.com/open-quantum-safe/liboqs/blob/444bccf9debb1d1d6ceec65c49474f0e23a84a66/src/sig/sig.h#L339-L352
The API will also be needed for [FIPS 205](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.205.pdf), formerly SPHINCS+.
For algorithms that don't support context strings, the API returns an error code.
The PR also leaves the existing API without context string to (1) avoid breaking the API, (2) provide an easy way to support the empty context string, which is the default in FIPS 204, (3) support other algorithms that don't have context strings (i.e., the official API for NIST-PQC submissions doesn't have a context string).
Seeking feedback from @open-quantum-safe/liboqs-committers and @johngray-dev especially on the new API and if it fulfils our needs.

I plan to wrap up this PR this week, also adding tests for the context API. The initial plan was to also add ACVP tests for the external API and also adding pre-hash variants. Since the publication of these vectors was postponed to within [November](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/vybKwrXx53k/m/ap3Un89jAQAJ), I'd suggest to do this as a follow-up after this PR.

baentsch:
This PR represents a break with what `liboqs` aimed to be, namely a generic collection of PQ algorithms, by adding an API only of use for some algorithms (1 right now, possibly 2 if/as someone introduces & adds this also to SLH-DSA).

This is clearly visible by the huge amount of code this PR adds just to introduce structures all basically returning "OQS_ERROR".

This is a parallel to what is proposed in https://github.com/open-quantum-safe/liboqs/issues/1985 -- with the difference that in this PR, an (one!) upstream code base actually supports the feature.

In my opinion, this addition makes OQS even harder to maintain, more fragile (see https://github.com/open-quantum-safe/liboqs/pull/1919#issuecomment-2460535228) and harder to support than it already is. For example, the only reason all this empty boilerplate code stands a chance of being OK is the code generation logic built into "copy_from_upstream": This duplicates easily a lot of code, but also makes `liboqs` ever larger and less efficient. It also makes it less and less likely that code reviews are done diligently (I bet no-one will spend time in review to vet all 243 files changed are OK).

In sum, this PR
- invalidates the concept of having one API equally supported for all PQ algs
- makes the library harder to maintain and more fragile
- reduces the likelihood that code gets reviewed
- makes the library even larger/less efficient

I see 2 alternatives to counter these problems:
1) amend the existing sig APIs to also take a context string (returning OQS_ERROR if not supported)
2) make this new API compile-time selectable, eg., when building for FIPS compliance

Option 1) would retain the nature of `liboqs` as a generic PQ library, reduce the amount of boilerplate code substantially and avoid code growth but necessitate downstream code changes. Option 2) would be an alternative approach beginning to technically tackle the internal split of the library into a "productive use/standards track" and "experimental/research track" code base allowing downstream use(rs) to properly differentiate between pretty much best-effort/unmaintained research code and (hopefully, also at the upstreams) more well-maintained standards-track code.

bhess:
Thanks for your feedback @baentsch, I agree with your concerns about maintainability when adding this extra API.

> Option 1) would retain the nature of `liboqs` as a generic PQ library, reduce the amount of boilerplate code substantially and avoid code growth but necessitate downstream code changes. Option 2) would be an alternative approach beginning to technically tackle the internal split of the library into a "productive use/standards track" and "experimental/research track" code base allowing downstream use(rs) to properly differentiate between pretty much best-effort/unmaintained research code and (hopefully, also at the upstreams) more well-maintained standards-track code.

I think we could justify adapting the generic OQS sign/verify API to include the context strings. In my view context strings are not specific to the internals of the ML-KEM/SLH-DSA algorithms but were added as a good practice for signature schemes and allow domain separation between applications. I didn't do option 1) right away, indeed because of the API break and downstream code updates needed. However, at least oqs-provider needs to be updated anyways because [composite-signatures](https://github.com/open-quantum-safe/liboqs/issues/1891#issuecomment-2447548756) will use context strings. So I'd go for option 1) for this issue.

baentsch:
Thanks for going along with option 1, @bhess. The core logic update to `oqsprovider` indeed is trivial (composite notwithstanding).

That said, I'm now seeing another shortcoming of `liboqs`: AFAIK there is no simple library version define that a downstream could utilize to adapt to API changes in `liboqs`: The only define there is is "OQS_VERSION_TEXT" and I'm not sure that's making it sufficiently easy to adapt a downstream's code to `liboqs` capabilities: The goal of (and one purpose for) `oqsprovider` is to build against all versions of `liboqs`, so a simple-to-use code-level library version test facility should be there, e.g. similar to what we have in [openssl](https://github.com/openssl/openssl/blob/012353bdf21b98def920ac317b94c4a9ed501b79/include/openssl/opensslv.h.in#L29-L31). Worthwhile a separate issue/PR or simple enough to do as part of this PR? If the latter, the corresponding `oqsprovider` update could be done as a "-tracker" update as this _is_ an API-changing PR (either way).

dstebila:
@bhess In today's OQS status call we talked about a timeline for the liboqs 0.12.0 release. The main thing to be included is ML-DSA. The hope was to be able to make a release around two weeks from now. What timeline do you anticipate for landing this PR?

dstebila:
> I see 2 alternatives to counter these problems:
> 
>  1. amend the existing sig APIs to also take a context string (returning OQS_ERROR if not supported)
>  2. make this new API compile-time selectable, eg., when building for FIPS compliance
> 
> Option 1) would retain the nature of `liboqs` as a generic PQ library, reduce the amount of boilerplate code substantially and avoid code growth but necessitate downstream code changes. Option 2) would be an alternative approach beginning to technically tackle the internal split of the library into a "productive use/standards track" and "experimental/research track" code base allowing downstream use(rs) to properly differentiate between pretty much best-effort/unmaintained research code and (hopefully, also at the upstreams) more well-maintained standards-track code.

For option 1, does "amend the existing sig API" mean to have two signature APIs always available, one without the context string (i.e., the current API) and one with the context string (i.e., a new API), or does it mean having only one signature API, but it being the one with the context string?  If it means having only one signature API, being the one with the context string, then that will be a breaking change for all existing applications calling the OQS signature API, which seems problematic. Unless there's a way in C to allow optional arguments; I know there's `va_list` and `va_arg` in C, but I'm a bit rusty on how those work.

baentsch:
> does it mean having only one signature API, but it being the one with the context string

That's my understanding.

> that will be a breaking change for all existing applications calling the OQS signature API, which seems problematic

I don't see this as problematic: Yes, it requires adaptations by users, but truly trivial ones for people who don't need the new feature -- and essential ones for people who want to use the standard algs. The big advantage I see is that this change minimizes support and maintenance effort for the OQS team, minimizes code changes in OQS PRs, minimizes library size. Also, it's a change supported by more mature versioning support.

SWilson4:
Taking a [more critical/in-depth look](https://github.com/open-quantum-safe/liboqs/pull/1919#issuecomment-2460535228) at the API updates after today's meeting, here are my thoughts:

Based on the fact that both SLH-DSA and ML-DSA share the same context string--supporting API, it seems likely (to me) that future NIST standards (e.g., for Falcon and potential selections from the signature on-ramp process) will follow this API. Does that seem like a correct assessment? If so, then I think it's desirable to have our main signature API (`OQS_SIG_sign`) conform to the FIPS documents. In order to accomplish that, breaking backwards compatibility will be required at some point. There will never be a good time to do this, but if it must be done, then it would be best if it were done quickly (i.e., now). Users have a strong incentive to upgrade to version 0.12.0 (and hence adopt the new API). It also seems to me that many (maybe most?) users of liboqs don't actually use the API directly but instead access it via oqs-provider, in which case they will see no change at all. (Please correct me if I'm wrong on this point, @baentsch.)

Other notes:
- I think it would be good to get input from @Martyrshot here as a downstream consumer of the "un-providered" liboqs API.
- I imagine we would have to update all of the language wrappers (at least the ones we wish to support). For some of them (e.g., Java, C++, Python) we might be able to take advantage of overloading and optional/default arguments.
- Are you aware of a plan for OpenSSL to handle this @baentsch? It's not immediately clear to me how one would support the context string using the existing `EVP_DigestSign` API.

bhess:
The API has now been updated to include a single sign/verify interface that incorporates the context string:

```
OQS_API OQS_STATUS OQS_SIG_sign(const OQS_SIG *sig, uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *ctx_str, size_t ctx_str_len, const uint8_t *secret_key);
```

```
OQS_API OQS_STATUS OQS_SIG_verify(const OQS_SIG *sig, const uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *ctx_str, size_t ctx_str_len, const uint8_t *public_key);
```

Note that the SIG_STFL API also needed to be updated, as the API signatures for sign/verify must be identical to the ones of SIG.

> @bhess In today's OQS status call we talked about a timeline for the liboqs 0.12.0 release. The main thing to be included is ML-DSA. The hope was to be able to make a release around two weeks from now. What timeline do you anticipate for landing this PR?

I believe the primary open issue is reaching consensus on the API. So far, the discussion seems to favor a unified sign/verify API, which has now been added. Beyond that, there are two smaller tasks to address: creating a tracker branch in the OQS provider and updating the constant-time tests. I should be able to handle these this week.

dstebila:
> > that will be a breaking change for all existing applications calling the OQS signature API, which seems problematic
> 
> I don't see this as problematic: Yes, it requires adaptations by users, but truly trivial ones for people who don't need the new feature -- and essential ones for people who want to use the standard algs. The big advantage I see is that this change minimizes support and maintenance effort for the OQS team, minimizes code changes in OQS PRs, minimizes library size. Also, it's a change supported by more mature versioning support.

I don't think it's quite that simple.  We have two levels of public APIs -- the "object-oriented" API in the `OQS_SIG` struct ([`OQS_SIG->sign`](https://github.com/open-quantum-safe/liboqs/blob/2ee908df244245c466e8e57f7aad5b07d3da7bb6/src/sig/sig.h#L222)) and then per-algorithm public API (e.g., [`OQS_SIG_ml_dsa_44_ipd_sign`](https://github.com/open-quantum-safe/liboqs/blob/2ee908df244245c466e8e57f7aad5b07d3da7bb6/src/sig/ml_dsa/sig_ml_dsa_44_ipd.c#L87C20-L87C46)). We would need to make changes for both of these APIs across all algorithms. Now admittedly all of this is done via code generation, but it's still non-trivial; and the code generation will have to know whether the underlying implementation it's calling takes a context argument or not, unless the plan is to also change the function signatures of all the underlying implementations, which I think is a non-starter because then we're touching dozens of implementations, many of them in upstreams.

I am also worried that it is a substantial breaking change.  I think it's especially problematic to change the behaviour of an existing function name.  I'm a bit rusty on linking, but what happens when a binary that was compiled against the old sign API is dynamically linked against a new build of liboqs with the new sign API?  Will linking fail?  Or will it silently go through at link time and then fail at runtime?  Or will it silently go through at link and runtime, but just use the wrong arguments somehow?  If the latter, this seems very bad, as the 5th argument in the old sign API is the secret key, and the 5th argument in the new sign API is the context string, which suggests a risk that the secret key will now be used in an insecure way.  My understanding of linking may be horribly wrong -- apologies -- but I would like to make sure these factors are considered.  If the proposal was to rename the function name, that would at least be less risky, because things would clearly fail at any stage (compilation-time, link-time, runtime).

Given that all of the public APIs are in portions of code created by code generation, I don't see it as particularly hard to generate the second API for the algorithms that support it.  (Note that the code generation metadata **will** need to be updated to say which algorithms support it regardless of which approach we use, unless we change the function signature of all the underlying implementations as I mentioned as the last option two paragraphs above.)

baentsch:
> it's desirable to have our main signature API (OQS_SIG_sign) conform to the FIPS documents. In order to accomplish that, breaking backwards compatibility will be required at some point. There will never be a good time to do this, but if it must be done, then it would be best if it were done quickly (i.e., now). 

I whole-heartedly agree with this line of thought, @SWilson4.

>  in which case they will see no change at all. (Please correct me if I'm wrong on this point, @baentsch.)

Agreed again.

> Are you aware of a plan for OpenSSL to handle this @baentsch? It's not immediately clear to me how one would support the context string using the existing EVP_DigestSign API.

Sure -- or rather, actually, there's no particular plan required for this; one can simply use the [OSSL_PARAM](https://docs.openssl.org/3.0/man3/OSSL_PARAM_int/#description) mechanism for this (via the existing "_ex" function variants); the [SLH-DSA PR](https://github.com/openssl/openssl/pull/25882) proposes "OSSL_SIGNATURE_PARAM_CONTEXT_STRING", for example.

baentsch:
> but it's still non-trivial

Well, @bhess just did it before you wrote your comment :)

>  Or will it silently go through at link and runtime, but just use the wrong arguments somehow?

Shouldn't an SOVERSION mismatch make this impossible in the case of shared libs (for static builds there's obviously no problem at all)?

dstebila:
> > Or will it silently go through at link and runtime, but just use the wrong arguments somehow?
> 
> Shouldn't an SOVERSION mismatch make this impossible in the case of shared libs (for static builds there's obviously no problem at all)?

I don't know enough about how linking works to know if SOVERSION mismatch will catch this -- if you are certain of this, then that would be good to know.

Philosophically there's also the question of how much API we ought to be providing.  BoringSSL famously promises very little API stability, and my sense is that this makes it hard for people outside Google to rely on it.  OpenSSL seems to offer very good API stability.  That two successful projects take very different approaches means that there isn't a clear answer here for us.

Regardless though I feel this isn't the way to make breaking API changes. In the status call yesterday I talked about my preference for the Kyber deprecation -- that we advertise in the release notes of one version we plan to remove it, and then remove it in the next version, as we have done in the past for previous algorithm deprecations.  (We did make immediate removals for broken algorithms, which I think is defensible.)  Since we don't have a formal policy around API stability, my off-the-cuff approach would be to take a similar gradual approach to breaking API changes.  My current thinking is that one way to do a major API change would be:

- Make an issue/PR solely about the new API (with a new name) to solicit feedback.
- In release 0.x, include both APIs: the old `OQS_SIG->sign` API and the new `OQS_SIG->sign_with_context` API. Release notes of 0.x indicate that this will be the last release with the old API.
- In release 0.x+1, remove the old API; only the new API exists.

This approach gives a warning to users, and a transition period in which they can gently migrate from the old API to the new API within the same codebase.  Using separate names reduces any risk of expected behaviour from weird linking or confusion.   The cost of this is that it increases the workload on us -- although I think it is manageable since most of this is going to be in generated code.  So it becomes a question of what we prioritize.

To some extent this approach would also help us.  We are also users of these APIs in the rest of our subprojects -- oqs-provider, the language wrappers, and perhaps other places.  All of these will need to be updated to the new APIs.  If we do a hard, breaking change like currently being proposed, then we really should update all of those at the same time before we make a 0.12 release to be confident that they are fine.

If we have a period where both APIs are present, there is the question of whether ML-DSA is accessible via both during that time, or only via the new API.  I don't have strong opinions here either way.

Martyrshot:
> * I think it would be good to get input from @Martyrshot here as a downstream consumer of the "un-providered" liboqs API.

From my point of view I would suggest making this change sooner than later. I think it's a needed change though, and if it's done along with the roll out of official FIPS versions the breaking change is justified. If the change is made later then people will complain because "what changed??". We can at least say it's a requirement for the FIPS versions of the algorithms.

> I am also worried that it is a substantial breaking change. I think it's especially problematic to change the behaviour of an existing function name. I'm a bit rusty on linking, but what happens when a binary that was compiled against the old sign API is dynamically linked against a new build of liboqs with the new sign API? Will linking fail?

Since the function prototype changes, it should fail to link. The important thing to remember is that the header file will change as part of this breaking change. So the compiler will complain about "insufficient arguments".

dstebila:
> > I am also worried that it is a substantial breaking change. I think it's especially problematic to change the behaviour of an existing function name. I'm a bit rusty on linking, but what happens when a binary that was compiled against the old sign API is dynamically linked against a new build of liboqs with the new sign API? Will linking fail?
> 
> Since the function prototype changes, it should fail to link. The important thing to remember is that the header file will change as part of this breaking change. So the compiler will complain about "insufficient arguments".

I think you're confusing compiling and linking here. I agree the compiler will complain at compile time about insufficient arguments.  But that doesn't mean that the linker will complain at link time / runtime about linking together an old calling application with a new version of the library that has different arguments.  That's the key question I have still.

dstebila:
I wrote up a very simple collection of C files that I think exercise the concern I have.  In the attached ZIP, there's a single library function `myfunction`, which in `fv1.[ch]` has 3 arguments, and in `fv2.[ch]` has 4 arguments. In the build script, I compile a shared library containing the old version of the function, compile `main.c` against that old header file and link against the old shared library, then I delete the old shared library and compile a new shared library containing the new version of the function, and then try to run main. You'll see that main runs a second time without complaint, and that it uses the new version of the function, with an extra argument. Moreover, the value of the extra argument is nonsense (whatever was next on the stack, presumably).  So I think my concerns are justified.  The one caveat in my experiment is that I did not test whether using SOVERSION will stop this, as I don't know how to set the SOVERSION for a simple example like this, the only time we've used SOVERSION is in the context of CMake.

[badlinking.zip](https://github.com/user-attachments/files/17741699/badlinking.zip)

Martyrshot:
> > > I am also worried that it is a substantial breaking change. I think it's especially problematic to change the behaviour of an existing function name. I'm a bit rusty on linking, but what happens when a binary that was compiled against the old sign API is dynamically linked against a new build of liboqs with the new sign API? Will linking fail?
> > 
> > 
> > Since the function prototype changes, it should fail to link. The important thing to remember is that the header file will change as part of this breaking change. So the compiler will complain about "insufficient arguments".
> 
> I think you're confusing compiling and linking here. I agree the compiler will complain at compile time about insufficient arguments. But that doesn't mean that the linker will complain at link time / runtime about linking together an old calling application with a new version of the library that has different arguments. That's the key question I have still.

Ah yes I was thinking of static linking, not dynamic linking. I just did a sanity check. If a user uses a newer version of the shared object or dylib without recompiling with the new header files, the dynamic linker doesn't care and will act as if it received enough arguments. (aka, the new extra arguments will be garbage)

baentsch:
Thanks @dstebila @SWilson4 for taking the time to get to the bottom of this. While I don't think we can always protect people from shooting themselves into the foot (if I get it right, the problem occurs only if one does not use the `liboqs` build environment, omits SOVERSION in their own build environment of the library, have built a shared executable against it, and finally, exchanges the library executable against a new version thoughtlessly), we should try to keep away the gun to do it.

As I argue also in other places for caution in that `liboqs` ought to be presented "defensively" to the external world, in light of the above I change my perspective on the topic and propose the following:
- Introduction of a new API (generic "_ex" as is done by OpenSSL or specific "_ctx" to resolve this problem only; I'd personally think the latter is sufficient)
- Announcement at the next release that the current (then old) API will be removed at the subsequent release along the same lines as algorithm deprecation (unless there's a truly intuitive and automated way to map the old to the new behaviour such as to avoid additional support effort).

This will make this PR even more cumbersome to create (sorry, @bhess) and review, but this is probably the best long-term maintainable (and "protective" for the unwary) way forward.

> here is the question of whether ML-DSA is accessible via both during that time, or only via the new API. I don't have strong opinions here either way

I do have a strong opinion: This is a library promising generic access to all PQC algs in the same way.  To deliver on this promise, all public APIs must work for all algorithms.

bhess:
Thanks everyone for weighing in on the API aspects. It seems like a consensus is emerging: adding a new API to accommodate the context string, while retaining the existing API for compatibility. This aligns with what I previously implemented in [this PR](https://github.com/open-quantum-safe/liboqs/pull/1919#issuecomment-2457287369). At the same time, we would announce deprecation of the current sign/verify API, with plans to fully transition to the new API in the release following this one.

The good news is that both APIs (the one for this release and the next) are already implemented in the git history of this PR. Thus, updating is mostly a matter of reverting a few commits. 😊

Regarding the question:

> here is the question of whether ML-DSA is accessible via both during that time, or only via the new API. I don't have strong opinions here either way

My take: The implementation discussed in [the PR](https://github.com/open-quantum-safe/liboqs/pull/1919#issuecomment-2457287369) makes ML-DSA accessible via both APIs. This seems appropriate since FIPS 204 specifies the default context string as empty (handled by the old API), while the new API also enables non-empty context strings.

I believe this approach maintains consistency with our promise to provide uniform access to all PQC algorithms across the public APIs. Let me know/thumbs-up/down if you have concerns or agree with this plan. If there are no objections, I'll proceed with updating the PR later today.

bhess:
A related question to this:

> Here is the question of whether ML-DSA is accessible via both during that time, or only via the new API. I don't have strong opinions here either way

Should algorithms that don't support context strings also be accessible via the new sign_with_context API? For the sake of uniform API access, I would argue yes. This would ensure consistent behavior in the release following this one when the old API is deprecated.

For algorithms that don't support context strings, the behavior could be one of two options: either ignore the provided context string or return an error code if a non-empty context string is passed. Personally, I would prefer the latter, as it clearly indicates to users that the algorithm does not support context strings, preventing any incorrect assumptions.

dstebila:
> A related question to this:
> 
> > Here is the question of whether ML-DSA is accessible via both during that time, or only via the new API. I don't have strong opinions here either way
> 
> Should algorithms that don't support context strings also be accessible via the new sign_with_context API? For the sake of uniform API access, I would argue yes. This would ensure consistent behavior in the release following this one when the old API is removed.

Yes, I think this is a full transition of all signing algorithms to the new `sign_with_context` API (and also verify).

> For algorithms that don't support context strings, the behavior could be one of two options: either ignore the provided context string or return an error code if a non-empty context string is passed. Personally, I would prefer the latter, as it clearly indicates to users that the algorithm does not support context strings, preventing any incorrect assumptions.

I agree, it should return an error code if a non-empty context string is passed to an algorithm that doesn't support context strings, to avoid any potential confusion.

tests/vectors_sig.c
```diff
@@ -165,109 +212,340 @@ OQS_STATUS sig_vector(const char *method_name,
...
+
+	signature = malloc(sigLen);
+
+	if (signature == NULL) {
+		fprintf(stderr, "[test_acvp_sig] %s ERROR: malloc failed!\n", method_name);
+		goto err;
+	}
+
+	if ((randomized && prng_output_stream == NULL) || (sigGen_sk == NULL) || (sigGen_msg == NULL) || (sigGen_sig == NULL)) {
+		fprintf(stderr, "[test_acvp_sig] %s ERROR: inputs NULL!\n", method_name);
```
SWilson4, line 317:
```suggestion
		fprintf(stderr, "[vectors_sig] %s ERROR: inputs NULL!\n", method_name);
```

tests/vectors_sig.c
```diff
@@ -165,109 +212,340 @@ OQS_STATUS sig_vector(const char *method_name,
...
+		fprintf(stderr, "[test_acvp_sig] %s ERROR: mlca_sig_sign_internal failed!\n", method_name);
+		goto err;
+	}
+	fprintBstr(fh, "signature: ", signature, sig->length_public_key);
+
+	if (!memcmp(signature, sigGen_sig, sigLen)) {
+		ret = EXIT_SUCCESS;
+	} else {
+		ret = EXIT_FAILURE;
+		fprintf(stderr, "[test_acvp_sig] %s ERROR: public key or private key doesn't match!\n", method_name);
```
SWilson4, line 347:
```suggestion
		fprintf(stderr, "[vectors_sig] %s ERROR: public key or private key doesn't match!\n", method_name);
```

tests/vectors_sig.c
```diff
@@ -165,109 +212,340 @@ OQS_STATUS sig_vector(const char *method_name,
...
-	if (argc != 8) {
-		fprintf(stderr, "Usage: vectors_sig algname prng_output_stream sig_msg sig_sk verif_sig verif_pk verif_msg\n");
-		fprintf(stderr, "  algname: ");
-		for (size_t i = 0; i < OQS_SIG_algs_length; i++) {
-			if (i > 0) {
-				fprintf(stderr, ", ");
-			}
-			fprintf(stderr, "%s", OQS_SIG_alg_identifier(i));
-		}
+	if (argc == 222) {
```
SWilson4, line 370: Where does 222 come from?

tests/vectors_sig.c
```diff
@@ -165,109 +212,340 @@ OQS_STATUS sig_vector(const char *method_name,
...
+		fprintf(stderr, "Usage: test_acvp_sig algname testname [testargs]\n");
```
SWilson4, line 371:
```suggestion
		fprintf(stderr, "Usage: vectors_sig algname testname [testargs]\n");
```

SWilson4 (commented):
Thanks for all the work, @bhess, especially given the uncertainty around API changes.

Most of my comments are fairly trivial questions/suggestions in the `vectors_sig` file (cleaning up error messages etc). I do have one more substantial suggestion for `test_sig.c`.

src/sig/sig.c
```diff
@@ -771,6 +726,14 @@ OQS_API OQS_STATUS OQS_SIG_sign(const OQS_SIG *sig, uint8_t *signature, size_t *
 	}
 }
 
+OQS_API OQS_STATUS OQS_SIG_sign_with_ctx_str(const OQS_SIG *sig, uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *ctx_str, size_t ctx_str_len, const uint8_t *secret_key) {
+	if (sig == NULL || sig->sign_with_ctx_str == NULL || sig->sign_with_ctx_str(signature, signature_len, message, message_len, ctx_str, ctx_str_len, secret_key) != OQS_SUCCESS) {
```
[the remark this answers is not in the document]
bhess: I agree, the check was a leftover from an earlier version where `sign_with_ctx_str` could potentially be NULL. This is no longer applicable. Updated accordingly.

src/sig/sig.h
```diff
@@ -185,6 +179,9 @@ typedef struct OQS_SIG {
 	/** Whether the signature offers EUF-CMA security (TRUE) or not (FALSE). */
 	bool euf_cma;
 
+	/** Whether the signature supports signing with a context string (TRUE) or not (FALSE). */
```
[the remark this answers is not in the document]
bhess: I agree. I've added a note to the API definitions in ed194b7dbad65ed5bc795eb515833901b1bfe304 for both `sign_with_ctx_str` and `verify_with_ctx_str`. Do you think this covers it, or would it also be helpful to include a similar note for the `sig_with_ctx_support` boolean for clarity?

tests/test_sig.c
```diff
@@ -118,6 +119,57 @@ static OQS_STATUS sig_test_correctness(const char *method_name) {
...
+			goto err;
+		}
+	} else {
+		rc = OQS_SIG_sign_with_ctx_str(sig, signature, &signature_len, message, message_len, ctx, 1, secret_key);
+		if (rc != OQS_ERROR) {
+			fprintf(stderr, "ERROR: OQS_SIG_sign_with_ctx_str should fail without support for context strings\n");
+			goto err;
+		}
+
+		rc = OQS_SIG_sign_with_ctx_str(sig, signature, &signature_len, message, message_len, NULL, 0, secret_key);
```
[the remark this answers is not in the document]
bhess: Thanks a lot for the code suggestion! Committed them now.

tests/vectors_sig.c
```diff
@@ -1,6 +1,7 @@
-// SPDX-License-Identifier: MIT
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
bhess: Thanks for the catch, this wasn't intentional. Corrected it to MIT in ed194b7dbad65ed5bc795eb515833901b1bfe304

tests/vectors_sig.c
```diff
@@ -114,45 +165,41 @@ OQS_STATUS sig_vector(const char *method_name,
...
 	}
 
 	randombytes_init(entropy_input, NULL);
 
 	fh = stdout;
 
-	public_key = OQS_MEM_malloc(sig->length_public_key);
-	secret_key = OQS_MEM_malloc(sig->length_secret_key);
-	signature = OQS_MEM_malloc(sig->length_signature);
+	public_key = malloc(sig->length_public_key);
```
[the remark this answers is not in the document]
bhess: Updated all occurrences in ed194b7dbad65ed5bc795eb515833901b1bfe304.

tests/vectors_sig.c
bhess: Was a placeholder for a better check :) Thanks for the catch! Updated in ed194b7dbad65ed5bc795eb515833901b1bfe304

src/sig/ml_dsa/sig_ml_dsa_44.c
```diff
@@ -0,0 +1,122 @@
...
+
+#if defined(OQS_ENABLE_SIG_ml_dsa_44)
+OQS_SIG *OQS_SIG_ml_dsa_44_new(void) {
+
+	OQS_SIG *sig = OQS_MEM_malloc(sizeof(OQS_SIG));
+	if (sig == NULL) {
+		return NULL;
+	}
+	sig->method_name = OQS_SIG_alg_ml_dsa_44;
+	sig->alg_version = "https://github.com/pq-crystals/dilithium/tree/master";
```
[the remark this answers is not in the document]
bhess: In most algorithms, `alg_version` seems to refer to the version of the upstream code, whereas the algorithm files (e.g., ml_dsa.yml) contain details about the specification version. However, using the spec name "FIPS 204" could reduce ambiguity, especially reading the member definition:
> Implementations with the same method_name and same alg_version will be interoperable.

dstebila:
> The remaining tasks have been implemented, and the oqs-provider downstream tests are passing. The new APIs are so far named `OQS_SIG_sign_with_ctx_str` and `OQS_SIG_verify_with_ctx_str`. An alternative could be `OQS_SIG_sign_with_context` or `OQS_SIG_sign_with_context_string`. Is there a preference?
> 
> I avoided `OQS_SIG_sign_with_context` because the term "context" alone is somewhat overloaded and might be misinterpreted as referring to a state structure. While `OQS_SIG_sign_with_context_string` is more verbose, it seems to align better with the typical naming in liboqs.

Makes sense to explicitly say context string rather than just context.  I don't have a preference between `ctx_str` and `context_string`.

src/sig/ml_dsa/sig_ml_dsa_44.c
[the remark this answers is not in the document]
dstebila: Having `alg_version` point to a repository is a carry-on from when NIST submission teams would update code without updating specifications and so the code was the definition of an algorithm version. Other times we have used phrases like "Round 1" or "version 2.1" as the `sig->alg_version`.  I'd be okay with saying FIPS 204 here.

tests/test_sig.c
```diff
@@ -118,6 +119,76 @@ static OQS_STATUS sig_test_correctness(const char *method_name) {
...
+			OQS_TEST_CT_DECLASSIFY(signature, signature_len);
+			rc = OQS_SIG_verify_with_ctx_str(sig, message, message_len, signature, signature_len, ctx, i, public_key);
+			OQS_TEST_CT_DECLASSIFY(&rc, sizeof rc);
+			if (rc != OQS_SUCCESS) {
+				fprintf(stderr, "ERROR: OQS_SIG_verify_with_ctx_str failed\n");
+				goto err;
+			}
+
+			/* modify the signature to invalidate it */
+			OQS_randombytes(signature, signature_len);
```
baentsch, line 143: Pretty drastic sig change for negative test. What about only "++"ing a single random byte?

src/sig/ml_dsa/sig_ml_dsa_44.c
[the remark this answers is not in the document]
bhess: Updated to "FIPS204" in [3e6302a](https://github.com/open-quantum-safe/liboqs/pull/1919/commits/3e6302a8d05db87e768e3f05e5fe209685648263)

tests/test_sig.c
bhess: Thank you for the suggestion! The tests were originally implemented this way. Considering the security claims of the signature schemes in liboqs, all of them assert "EUF-CMA". Would this test align with that security claim? Tagging @dstebila for input and to confirm whether the test was intentionally designed this way originally.

I’ve run the suggested test by @baentsch and noticed that the CROSS signature scheme seems to fail in some cases, passing the verification of modified signatures.

For ML-DSA (FIPS 204) at least, which claims the stronger SUF-CMA property, this test should be applicable. Section 3.1 of [FIPS 204](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf) states:

> ML-DSA is designed to be strongly existentially unforgeable under chosen message attack (SUF-CMA).
That is, it is expected that even if an adversary can get the honest party to sign arbitrary messages, the
adversary cannot create any additional valid signatures based on the signer’s public key, including on
messages for which the signer has already provided a signature.

Side-note: I've updated the security claim of ML-DSA to SUF-CMA in the ml_dsa.yml.

dstebila: I guess technically for an EUF-CMA scheme it would not be a violation of security if you found a different signature for the same message, so for those schemes a better test would be changing bytes of the message, rather than bytes of the signature.

For SUF-CMA schemes, it would be a violation of security if a different signature works for the same message, so for those schemes changing bytes of the signature is a good test. I do think changing as few bytes as possible is better. 

Very curious that CROSS passes verification of modified signatures -- perhaps some of those bytes are padding or formatting?

SWilson4: I'll defer to @dstebila's expertise, but I think both versions "test" SUF-CMA security. Given a valid message-signature pair (_m_, _σ_), it does not break EUF-CMA security to produce a second valid signature _σ'_ on the same message—whether that second signature differs in a single bit or every single bit. To break EUF-CMA security, the attacker has to produce a forgery for a message _m'_ for which it does _not_ already have a valid signature. On the other hand, SUF-CMA security is broken even if the attacker produces a second valid signature for the same message.

In order to "test" EUF-CMA security, we would have to modify the message. Modifying the signature would be optional.

I have always viewed this test as not testing "security" as much as "correctness" (in quotations because the two notions are obviously connected)—it's a sanity check that the verification algorithm doesn't trivially accept. Guessing a signature to see if it verifies is pretty much the most naïve "attack" possible, and will only catch trivially broken schemes (likely not breaking the scheme itself but rather finding an implementation error).

In principle, I do agree with @baentsch that a smaller modification is better as it's more likely to capture subtle correctness errors. However, if a scheme does not claim SUF-CMA security, we may run into false positives. I'm not sure if any of the schemes in liboqs are known to _not_ be SUF-CMA secure.

tests/vectors_sig.c
```diff
@@ -165,109 +212,361 @@ OQS_STATUS sig_vector(const char *method_name,
...
+
+		hexStringToByteArray(prng_output_stream, prng_output_stream_bytes);
+		hexStringToByteArray(kg_pk, kg_pk_bytes);
+		hexStringToByteArray(kg_sk, kg_sk_bytes);
+
+
+		rc = sig_kg_vector(alg_name, prng_output_stream_bytes, kg_pk_bytes, kg_sk_bytes);
+
+	} else if (!strcmp(test_name, "sigGen_det") || !strcmp(test_name, "sigGen_rnd")) {
+		int randomized = !strcmp(test_name, "sigGen_rnd");
```
SWilson4, line 450: I think this is a bitwise NOT, which is OK for the truthiness comparison in the `else if` clause but will cause problems if we treat it as an arithmetic value. Probably should do something like this instead:
```suggestion
		int randomized = (strcmp(test_name, "sigGen_rnd") != 0);
```

tests/test_sig.c
SWilson4: Gah, beat me to it @dstebila!

src/sig/ml_dsa/sig_ml_dsa_44.c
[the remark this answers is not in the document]
SWilson4: Created https://github.com/open-quantum-safe/liboqs/issues/1996 to propose making the same change to ML-KEM for consistency.

tests/vectors_sig.c
bhess: Wouldn't `~` be the bitwise NOT operator, while `!` is the logical NOT operator, which always returns either 0 or 1?

src/sig/sig.h
```diff
@@ -233,6 +248,21 @@ typedef struct OQS_SIG {
...
 
+	/**
+	 * Signature verification algorithm, with custom context string.
+	 *
+	 * @param[in] message The message represented as a byte string.
+	 * @param[in] message_len The length of the message.
+	 * @param[in] signature The context string used for the signature. This value can be set to NULL if a context string is not needed (i.e., for algorithms that do not support context strings or if an empty context string is used).
+	 * @param[in] signature_len The context string used for the signature. This value can be set to 0 if a context string is not needed (i.e., for algorithms that do not support context strings or if an empty context string is used).
+	 * @param[in] ctx_str The context string for the signature.
+	 * @param[in] ctx_str_len The length of the context string.
```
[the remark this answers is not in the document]
bhess: Oh yeah, thanks for noticing and fixing this!

tests/vectors_sig.c
SWilson4: Of course, you are correct, disregard.

tests/test_sig.c
bhess: Thanks to both @dstebila and @SWilson4 for the reasoning, and to @baentsch for opening the related issue on CROSS: [#1995](https://github.com/open-quantum-safe/liboqs/issues/1995).

We could do the following testing strategy:

- For EUF-CMA-secure schemes: A negative test would involve modifying the message by flipping a random bit.
- For SUF-CMA-secure schemes: There would be two negative tests:
The first is the same as the EUF-CMA test (modifying the message).
The second involves modifying the signature by flipping a random bit.

To implement this, we would need an additional variable in the OQS_SIG struct, `suf_cma`, analogous to the existing `euf_cma` variable. The only scheme with documented SUF-CMA claims so far in liboqs appears to be ML-DSA.

Does this approach sound reasonable, or might it be overkill?

baentsch: Unrelated to the contents of this discussion, can I suggest moving this conversation to a new issue such as to close out this PR?

baentsch: > I'm not sure if any of the schemes in liboqs are known to not be SUF-CMA secure.

Particularly this statement by @SWilson4 below (that for some reason I cannot respond to) makes me nervous in the sense that this discussion may become "long winded" (reminds me of "unknown unknowns" :-) For now, apologies for having opened this can of worms. For now, let's keep it closed and disregard my comment and keep tests as they've "always been" but feel free to open a separate issue.

SWilson4: I'm hesitant to make this test dependent on the security model because
1. it's not a particularly good test of the security of the scheme
2. even for schemes which only claim EUF-CMA, a "weak" forgery may be unexpected and worth investigation (as we just saw with #1995).

SWilson4: Created issue #1997 to contain follow-up discussion.

SWilson4 (approved):
Everything LGTM now. Thanks to @bhess for the work and to all for the effort in discussing the API changes!

bhess:
Thanks @SWilson4 for going through all the changes and for the review!

dstebila (approved):

dstebila:
Is this ready to merge?

bhess:
Yes, this is ready to merge from my side. I will proceed with the merge in about an hour unless there are any objections.

PR #1922, opened 2024-09-12
Patch Kyber to fix ASAN error on ARM64

praveksharma:
Patch Kyber to fix address sanitiser issue on ARM64 as described in #1914. Patch copied from [PQClean](https://github.com/PQClean/PQClean/blob/0c5bb143456333cd6ac85435d344b08b351b27a8/crypto_kem/kyber512/aarch64/neon_symmetric-shake.c#L59-L60), in line with fix submitted in #1914. 

### DIFF
--- a/scripts/copy_from_upstream/copy_from_upstream.yml
+++ b/scripts/copy_from_upstream/copy_from_upstream.yml
@@ -8,7 +8,8 @@ upstreams:
     sig_meta_path: 'crypto_sign/{pqclean_scheme}/META.yml'
     kem_scheme_path: 'crypto_kem/{pqclean_scheme}'
     sig_scheme_path: 'crypto_sign/{pqclean_scheme}'
-    patches: [pqclean-dilithium-arm-randomized-signing.patch, pqclean-kyber-armneon-shake-fixes.patch, pqclean-kyber-armneon-768-1024-fixes.patch, pqclean-kyber-armneon-variable-timing-fix.patch]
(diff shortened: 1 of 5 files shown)


praveksharma:
Thank you for your work on #1914 @mingtaoy! I've copied the changes from PQClean as you pointed and added them as a patch. I've also added the authors from #1914 as co-authors. Would you please have a look at the changes?

baentsch:
@praveksharma Thanks for pulling this over from PQClean. To make reviewing easier would you mind pointing to the actual PQClean commit you pulled over? This PR doesn't seem to be the identical logic as in #1914.

SWilson4 (approved):

praveksharma:
Sorry for the inadequate documentation @baentsch. 

The changes were first introduced [this commit](https://github.com/PQClean/PQClean/commit/fb003a2a625c49f3090eec546b2383dcfa2c75d8). The logic is slightly different where instead of defining a function to round up integers this is done manually -- `KYBER_SYMBYTES` is 32 for Kyber 512, 768, and 1024 so `KYBER_SYMBYTES + 2 + 6` is divisible by 8 while `KYBER_SYMBYTES + 2` is not.

baentsch:
> Sorry for the inadequate documentation @baentsch.

No need to apologize; I've been too lazy to try to understand the logic (differences). So this LGTM then. So when this lands #1914 gets closed without merge: OK also for you @mingtaoy?

baentsch (approved):

mingtaoy:
@baentsch - Yes! When I worked on #1914, I was unaware at the time that this was already fixed upstream.

PR #1923, opened 2024-09-13
Install test binaries

ajlennon:
I am creating a Yocto Embedded Linux layer to automate building liboqs into Yocto images so people can test.

ref: https://github.com/DynamicDevices/meta-quantum-safe

As I've been working on this I see the test examples aren't setup to be installed within the CMakeFiles.txt. Yocto takes the installed output of CMake and then installs those files via packages into embedded target images.

I feel it would be really helpful to have the option to install these onto target systems so I've  created this PR which does this. Perhaps we could have a chat about whether you can accept this?

Testing: This installs the tests rather than adding a feature that needs testing

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -41,15 +41,19 @@ if(NOT WIN32)
 
     add_executable(test_aes test_aes.c)
     target_link_libraries(test_aes PRIVATE ${TEST_DEPS})
+    install (TARGETS test_aes RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
 
     add_executable(test_hash test_hash.c)
     target_link_libraries(test_hash PRIVATE ${TEST_DEPS})
+    install (TARGETS test_hash RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
 
     add_executable(test_sha3 test_sha3.c)
     target_link_libraries(test_sha3 PRIVATE ${TEST_DEPS})
+    install (TARGETS test_sha3 RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
 
     add_executable(speed_common speed_common.c)
     target_link_libraries(speed_common PRIVATE ${TEST_DEPS})
+    install (TARGETS speed_common RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
 
     set(UNIX_TESTS test_aes test_hash test_sha3 speed_common)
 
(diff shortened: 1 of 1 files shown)


ajlennon:
I'm not sure what's best to be honest. My thinking here is I'd like to try to add some wrappers to build and test and profile liboqs on embedded Linux systems with Yocto. 

My layer is coming along now and I've managed to integrate your Ninja test command into Yocto "package test" which will run on emulated targets or real hardware 

https://github.com/DynamicDevices/meta-quantum-safe

Maybe what I should do for now is to try to get those tests up and running and then come back and ask you how this should look to be most useful to you and people working in the embedded space. 

I'm not at all precious just want to provide something useful. Please let me know if there's a more sensible way of going than what I'm thinking ?

Thanks! 

Alex

dstebila:
My gut feeling is that I wouldn't want to see the test binaries installed with a default installation -- they really are just for testing, they are not standalone command-line programs, and I feel like for the vast majority of users we'd just be polluting their /usr/bin directories.  So if this is desired functionality, then I'd be inclined to have an additional build target that covers this functionality.

ajlennon:
I am just working through the test scripts one by one starting with `test_binaries.py`. I can see that they expect to be using the built library within the host build which I think might be what you were raising as a potential issue as I need the cross-compiled code on the target. So I'm separating out some Yocto patches to deal with this. Will revert when I have made a bit more progress with running these tests on-device. Cheers!

ajlennon:
OK I've made some pretty good progress here. I have the tests running in a Yocto ptest runner which can optionally be built and installed. I've patched those tests to use the on-device installed library rather than the local copy.

Just finishing up some testing and will document then assuming all goes well will close this PR as unneeded.

Then we should be able to to build loboqs within Yocto run run on any qemu simulated target or real devices for testing!

ajlennon:
ok I think I have the layer up and running with GitHub CI building QEMU x86_64 and arm64 images and running your tests as package tests. It's still a little rough around the edges but working.

https://github.com/DynamicDevices/meta-quantum-safe/actions/runs/10939380640

As you've suggested I agree this PR is the wrong way to go about things. I only make one patch which is to use the library installed on the target not the one in the build tree.

Will close this now. Really keen for any collaboration from anybody who would like to help on the embedded end of things if you can suggest anywhere you think it helpful for me to reach out.

Will close this PR now. Cheers!

dstebila:
> I wonder whether it'd be worth while adding this thread to a list of known OQS users/deplyments such as for others to more easily touch base? Thoughts where to put this @SWilson4 @dstebila ?

I had been tracking all the adoptions I knew about on the website at https://openquantumsafe.org/applications/external.html.  But that may not be the form you're suggesting... what kind of data/medium were you imagining, Michael?

baentsch:
> what kind of data/medium were you imagining

A list of people (GH IDs) and/or projects that have active experience building OQS for direct consumption in specific setups (to be detailed by the maintainers of those projects). Samples would be Fedora, homebrew or now Yocto. Goal would be 
- for users with an interest in a specific setup to directly be able to reach out to those projects with questions (and/or simply get the binaries/packages)
- for OQS to spare the team all obligations coming with maintaining binaries (at least until it has grown/matured to cater for this).

ajlennon:
fwiw from me... I don't understand all this magic stuff but I do understand embedded systems (I would hope!)

I'm very focussed on the security of embedded devices, particularly with the new and upcoming legislation, and I'm interested to put some time and energy into where we go moving forward.

My guess is that you want to focus on your core mission and not be distracted by issues downstream with, say, embedded devices. But I also imagine you'd be really interested to have performance metrics on different platforms fed back and any kinds of problems that we found "in the real world" (as it were!)

That's the kind of thing I like to do and I'd be more than happy to be some kind of Tier 2 interface working with people to get things running up and tested on embedded boards.

Just thoughts! :)

dstebila:
> A list of people (GH IDs) and/or projects that have active experience building OQS for direct consumption in specific setups (to be detailed by the maintainers of those projects). 

Sounds good. Would the Github wiki be a suitable medium for this? It would let people add themselves to the list, without needing to go through the openquantumsafe.org website update rigamarole.

Issue #1823, opened 2024-06-25
Replace malloc/free with OPENSSL_malloc/OpenSSL_free


songlingatpan:
In OpenSSL code, we can customize malloc/free functions with CRYPTO_set_mem_functions.
in oqs provider, it also utilized OPENSSL_malloc/OPENSSL_free. so CRYPTO_set_mem_functions can be used to customize allocator.
We should replace malloc/free with OPENSSL_malloc/OpenSSL_free in liboqs as well.

SWilson4:
We certainly could do this, but I don't think it's very high on our priority list given the amount of manual effort it would require and the minimal (to my knowledge) improvement to the library. However, if you would like to take a stab at it, please feel free to open a PR for us to review.

songlingatpan:
@baentsch @SWilson4
I can help on this. However, I have a few questions.
1. For the malloc and free from upstream code, if we change them to OPENSSL_malloc and OPENSSL_free, how do we maintain the code? We can write a script to replace malloc and free. Would you like to run the script during merge the code from upstream, or during build the code?
2. Would you like to use OPENSSL_malloc and OPENSSL_free only when OQS_USE_OPENSSL = 1? or we can use it by default?
3. I found quite some error handling issues, and memory leak after failure in both oqs and upstream code. Again, how do you maintain the code change to fix upstream issues?
Thanks

baentsch:
:> Would you like to run the script during merge the code from upstream, or during build the code?

My preference would be to do this during `copy_from_upstream`such as for everyone to read what code gets built.

> Would you like to use OPENSSL_malloc and OPENSSL_free only when OQS_USE_OPENSSL = 1? or we can use it by default?

Not by default. If OQS_USE_OPENSSL=0, the current, openssl-free build must still succeed. Therefore, I'd suggest you'd add 2 macros in step 1 and define those suitably depending on the setting of OQS_USE_OPENSSL.

> I found quite some error handling issues, and memory leak after failure in both oqs and upstream code. Again, how do you maintain the code change to fix upstream issues?

That's a scary question but a good one. Our usual recommendation is to contribute any fixes you develop to the upstreams so they flow back into `liboqs` when we run `copy_from_upstream`. If this does not succeed, we can create local patches -- but those would need to be maintained (as and when the upstreams change) and someone would need to take responsibility for this maintenance: Would you be willing to take that on @songlingatpan ?

baentsch:
[...]

Issue #1868, opened 2024-07-26
Add CodeQL query to check for memset calls


SWilson4:
This would prevent new instances of #1866 from being introduced into the codebase.

Suggested by @trailofbits in Week 1 of their audit of `liboqs`.

Sukuna0007Abhi:
Hi @SWilson4 can I work on this issue

dstebila:
> Hi [@SWilson4](https://github.com/SWilson4) can I work on this issue

Great!

Issue #1456, opened 2023-05-12
Add telltale error handling in void functions


baentsch:
Many OQS functions return void "for historical reasons". Nevertheless they can fail. Right now two (pretty user unfriendly) macros are used (sporadically) to react to internal failures: https://github.com/open-quantum-safe/liboqs/blob/9f912c957bfe7f4b894aa9661168a310e8dd1a58/src/common/common.h#L22-L49

This issue is to recommend improving them, e.g., along the lines suggested in https://github.com/open-quantum-safe/liboqs/pull/1454#discussion_r1192414400 and then use them in all cases.

PR #1926, opened 2024-09-17
[#1823] replace malloc/calloc/strdup/free with openssl allocator

songlingatpan:
Replaced malloc, calloc, strdup, and free with the OpenSSL memory allocator to enable the caller to customize memory allocator, addressing issue #1823. This PR does not change the existing behavior or algorithms.


songlingatpan:
I will commit another change for copy_from_upstream.

baentsch (changes requested):
Conceptually LGTM, but CI needs to pass; please see https://github.com/open-quantum-safe/liboqs/blob/main/CONTRIBUTING.md#coding-style

songlingatpan:
@baentsch Question: 'bike', 'frodokem', 'ntruprime' is not from upstream. 
To replace malloc with OQS_MEM_malloc, where should i modify the code? Thanks
`liboqs/src/kem/frodokem/kem_frodokem976shake.c:11:      OQS_KEM *kem = malloc(sizeof(OQS_KEM));
liboqs/src/kem/frodokem/kem_frodokem640aes.c:11:        OQS_KEM *kem = malloc(sizeof(OQS_KEM));
liboqs/src/kem/frodokem/kem_frodokem1344shake.c:11:     OQS_KEM *kem = malloc(sizeof(OQS_KEM));
liboqs/src/kem/frodokem/kem_frodokem976aes.c:11:        OQS_KEM *kem = malloc(sizeof(OQS_KEM));
liboqs/src/kem/frodokem/kem_frodokem1344aes.c:11:       OQS_KEM *kem = malloc(sizeof(OQS_KEM));
liboqs/src/kem/frodokem/kem_frodokem640shake.c:11:      OQS_KEM *kem = malloc(sizeof(OQS_KEM));
liboqs/src/kem/kem.h:264:OQS_API void OQS_KEM_free(OQS_KEM *kem);
liboqs/src/kem/ntruprime/kem_ntruprime_sntrup761.c:11:  OQS_KEM *kem = malloc(sizeof(OQS_KEM));
liboqs/src/kem/kem.c:493:OQS_API void OQS_KEM_free(OQS_KEM *kem) {
liboqs/src/kem/bike/kem_bike.c:9:       OQS_KEM *kem = malloc(sizeof(OQS_KEM));
liboqs/src/kem/bike/kem_bike.c:34:      OQS_KEM *kem = malloc(sizeof(OQS_KEM));
liboqs/src/kem/bike/kem_bike.c:59:      OQS_KEM *kem = malloc(sizeof(OQS_KEM));`

baentsch:
> Question: 'bike', 'frodokem', 'ntruprime' is not from upstream.

Good point. @dstebila : There's surely upstreams for these algorithms, but they're not captured by OQS automation/copy_from_upstream. Question: Are these projects still maintained (and where) or is it OK to change code straight in `liboqs` for these algs? If the former, what would it take to bring them in via "copy_from_upstream"? If the latter, should we remove them completely from OQS?

dstebila:
> > Question: 'bike', 'frodokem', 'ntruprime' is not from upstream.
> 
> Good point. @dstebila : There's surely upstreams for these algorithms, but they're not captured by OQS automation/copy_from_upstream. Question: Are these projects still maintained (and where) or is it OK to change code straight in `liboqs` for these algs? If the former, what would it take to bring them in via "copy_from_upstream"? If the latter, should we remove them completely from OQS?

FrodoKEM is maintained at https://github.com/microsoft/PQCrypto-LWEKE/; I exported the code from there any manually added it to liboqs, as we didn't have as robust a copy_from_upstream at the time.  FrodoKEM also needs to do an update from upstream, as there have been new variants introduced in the last year, but I don't have a plan for this update.  So to avoid blocking on that, I would say it's fine to make the FrodoKEM changes directly here in this repository.

NTRUPrime had been coming from PQClean, but they have stopped supporting it.  We only are keeping one variant of NTRUPrime because of its use in OpenSSH.  I think we consider a timeline for sunsetting it.  But in the interim, I think changes to NTRUPrime can be done directly here in this repository.

BIKE was contributed directly by the team at AWS.  Our main contact for that has been @dkostic, but I'm not sure if he's still the right contact. Pinging @brian-jarvis-aws for some input.

songlingatpan:
@dstebila @baentsch @bhess @jschanck 
Code change for openssl allocator has been done. 
Please review the code.

In addition, there are quite some error handling and potential memory leak.
Would you like to fix them in the current PR? or create a separate PR for review? 

Thanks

songlingatpan:
> **baentsch ** requested changes

done

dstebila:
> In addition, there are quite some error handling and potential memory leak. Would you like to fix them in the current PR? or create a separate PR for review?

I'd prefer to see those as a second PR, since those changes may be less mechanical and might require a closer look.

dstebila (approved):

songlingatpan:
@dstebila @baentsch @bhess @jschanck 
This PR is ready for review and merge.

src/kem/ntruprime/pqclean_sntrup761_avx2/params.h
```diff
@@ -32,7 +32,7 @@
 #define ppad 769
 #define endingmask _mm256_set_epi8(1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,0,0,0,0,0,0)
 #define crypto_core_weight PQCLEAN_SNTRUP761_AVX2_crypto_core_weightsntrup761
-#define p 761
+#define p_param 761
```
SWilson4, line 35: I'm not immediately seeing the reason for this name change—could you explain why it's necessary?

SWilson4:
Thanks for this work, @songlingatpan!

How would you feel about dropping the `OQS_MEM_free` function and instead using `OQS_MEM_insecure_free` (presumably modifying it to use `OPENSSL_free`)? I think it's a good idea to have a clear separation between "insecure" and "secure" frees, and we lose that if we introduce a not-explicitly-insecure free function with the same signature as `OQS_MEM_insecure_free`.

songlingatpan:
> Thanks for this work, @songlingatpan!
> 
> How would you feel about dropping the `OQS_MEM_free` function and instead using `OQS_MEM_insecure_free` (presumably modifying it to use `OPENSSL_free`)? I think it's a good idea to have a clear separation between "insecure" and "secure" frees, and we lose that if we introduce a not-explicitly-insecure free function with the same signature as `OQS_MEM_insecure_free`.

Addressed the comment in the latest PR. Please review it.

songlingatpan:
> After this lands, we may want to add some sort of static analysis to guard against raw `malloc` or `free` calls. Seems like something related to #1868.

Correct. We should phase out direct calls to malloc, free, calloc, strdup, and realloc.

src/kem/ntruprime/pqclean_sntrup761_avx2/params.h
songlingatpan: There is some conflict with openssl.
`In file included from /Volumes/GIT/git/oqs/liboqs/src/kem/ntruprime/pqclean_sntrup761_clean/kem.c:5:
In file included from /Volumes/GIT/git/oqs/liboqs/src/common/pqclean_shims/randombytes.h:6:
In file included from /Volumes/GIT/git/oqs/liboqs/build/include/oqs/rand.h:15:
In file included from /Volumes/GIT/git/oqs/liboqs/build/include/oqs/common.h:30:
In file included from /usr/local/opt/openssl@3/include/openssl/crypto.h:34:
In file included from /usr/local/opt/openssl@3/include/openssl/safestack.h:24:
/usr/local/opt/openssl@3/include/openssl/stack.h:45:60: error: expected ')'
void *OPENSSL_sk_delete_ptr(OPENSSL_STACK *st, const void *p);
                                                           ^
/Volumes/GIT/git/oqs/liboqs/src/kem/ntruprime/pqclean_sntrup761_clean/params.h:31:11: note: expanded from macro 'p'
#define p 761
          ^
/usr/local/opt/openssl@3/include/openssl/stack.h:45:28: note: to match this '('
void *OPENSSL_sk_delete_ptr(OPENSSL_STACK *st, const void *p);
                           ^
1 error generated.
[121/1636] Building C object src/kem/frodokem/CMakeFiles/frodokem.dir/external/frodo1344aes.c.o
ninja: build stopped: subcommand failed.`

dstebila:
> > After this lands, we may want to add some sort of static analysis to guard against raw `malloc` or `free` calls. Seems like something related to #1868.
> 
> Correct. We should phase out direct calls to malloc, free, calloc, strdup, and realloc.

`test_free` in `tests/test_code_conventions.py` does this for free calls using an ad hoc script. It may be better to do this with CodeQL in the long run.

songlingatpan:
> test_code_conventions.py

Addressed the comment by adding code in test_code_conventions.py.

src/kem/ntruprime/pqclean_sntrup761_avx2/params.h
SWilson4: Hmmmm, I see. Is there a clean way to handle this with #include guards or something similar? Renaming seems like a brittle workaround. Suppose for argument's sake that a future version of OpenSSL changes the function parameter to `const void *q`—then everything would break here again.

SWilson4: What I wouldn't give for C++ namespaces... :stuck_out_tongue_winking_eye:

tests/test_code_conventions.py
```diff
@@ -48,26 +48,38 @@ def test_spdx():
...
-                if not re.match(r'^.*[^_]free\(.*$', line):
+            content = f.read()
+            lines = content.splitlines()
+            for no, line in enumerate(lines, 1):
+                # Skip comments
+                if line.strip().startswith('//') or line.strip().startswith('/*'):
                     continue
-                if 'IGNORE free-check' in line:
+                # Check if we're inside a multi-line comment
+                if '/*' in content[:content.find(line)] and '*/' not in content[:content.find(line)]:
```
SWilson4, line 68: Just a note, this will produce a false positive in the following scenario:
```c
/* comment */
/*
malloc(sizeof(int));
*/
OQS_MEM_malloc(sizeof(int));
```
I think you'd have to search backwards for the closest `/*`, but that seems like overkill.

songlingatpan: @SWilson4 You are right. We can improve it later.

src/common/common.c
```diff
@@ -266,20 +270,19 @@ OQS_API void OQS_MEM_cleanse(void *ptr, size_t len) {
...
 #endif
 }
-
 void *OQS_MEM_checked_malloc(size_t len) {
-	void *ptr = malloc(len);
+	void *ptr = OQS_MEM_malloc(len);
 	if (ptr == NULL) {
 		fprintf(stderr, "Memory allocation failed\n");
-		abort();
+		return NULL; //abort();
```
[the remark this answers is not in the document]
songlingatpan: True. 
Many embedded systems have strict memory constraints, and encountering a malloc failure is not uncommon in such environments.

src/common/common.c
```diff
@@ -266,20 +270,19 @@ OQS_API void OQS_MEM_cleanse(void *ptr, size_t len) {
 	explicit_memset(ptr, 0, len);
 #elif defined(__STDC_LIB_EXT1__) || defined(OQS_HAVE_MEMSET_S)
 	if (0U < len && memset_s(ptr, (rsize_t)len, 0, (rsize_t)len) != 0) {
-		abort();
+		return; //abort();
```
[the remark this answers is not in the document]
songlingatpan: I will change it back to use abort for this case.

src/common/common.c
```diff
@@ -289,7 +292,7 @@ void *OQS_MEM_checked_aligned_alloc(size_t alignment, size_t size) {
 	void *ptr = OQS_MEM_aligned_alloc(alignment, size);
 	if (ptr == NULL) {
 		fprintf(stderr, "Memory allocation failed\n");
-		abort();
+		return NULL; //abort();
```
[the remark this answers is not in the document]
songlingatpan: yes. Please let me know if you agree to remove the _checked_ functions. If so, I will go ahead and remove them.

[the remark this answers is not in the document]
SWilson4: We could—but then we'll be left with a bunch of common-code functions that rely on the "checked" functions aborting if allocation fails. If this code is left unchanged, won't it lead to more crashes and/or memory corruption? If it needs to be changed to handle unchecked functions, is it not simpler to keep the "checked" functions rather than factor out the check everywhere?

songlingatpan:
@dstebila @baentsch @SWilson4 
Is there anything else I need to address before we proceed with merging this PR?

Thanks!

SWilson4:
> @dstebila @baentsch @SWilson4 Is there anything else I need to address before we proceed with merging this PR?
> 
> Thanks!

Please take a look at my comments [here](https://github.com/open-quantum-safe/liboqs/pull/1926#discussion_r1781360300) and [here](https://github.com/open-quantum-safe/liboqs/pull/1926#discussion_r1775256865) and let me know what you think.

src/common/common.c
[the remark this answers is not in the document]
songlingatpan: The caller must perform a NULL check before dereferencing the pointer. Additionally, in the case of a NULL check failure, the caller is responsible for releasing any allocated resources to prevent memory leaks.

songlingatpan:
> > @dstebila @baentsch @SWilson4 Is there anything else I need to address before we proceed with merging this PR?
> > Thanks!
> 
> Please take a look at my comments [here](https://github.com/open-quantum-safe/liboqs/pull/1926#discussion_r1781360300) and [here](https://github.com/open-quantum-safe/liboqs/pull/1926#discussion_r1775256865) and let me know what you think.

It seems the conflicts issue only occurs for certain openssl version/platform. Didn't see conflicts any more after revert.

songlingatpan:
@SWilson4 Addressed your 2 comments. 
Please let me know if I need to run copy_from_upstream.py or any other scripts to regenerate code before merge.
If so, what is the detail steps.

src/common/common.c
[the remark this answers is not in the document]
SWilson4: > The caller must perform a NULL check before dereferencing the pointer. Additionally, in the case of a NULL check failure, the caller is responsible for releasing any allocated resources to prevent memory leaks.

Sure—but those NULL checks are not currently present in the codebase, because the callers were relying on the `checked_malloc` functions. If we replace the `checked_malloc` functions with the unchecked versions in this PR, the NULL checks must also be added in this PR. For example [this code](https://github.com/open-quantum-safe/liboqs/pull/1926/files#diff-bb6ff52a8c0e7a4ee4104e6bd5bf779dad957ea1a751ff4b1d9bc5b0eb9f41f7R519-R522) will dereference NULL on malloc failure.

songlingatpan:
@dstebila @baentsch @SWilson4 
I have addressed existing comments. Please let me know if we can merge the PR.
Thanks

tests/test_code_conventions.py
baentsch: > We can improve it later.

@songlingatpan Such statements tend to get lost after a merge: Can I suggest creating an issue documenting this as an area for improvement, then? I'd really hate having to deal with problems that we saw coming but didn't fix beforehand.

baentsch:
> if I need to run copy_from_upstream.py or any other scripts to regenerate code before merge.

It's always a good idea to run those to preclude CI finding issues. See e.g. [this CI workflow](https://github.com/open-quantum-safe/liboqs/blob/329869f5592e555923eedd100249d2a10f546f3e/.github/workflows/basic.yml#L49-L53). @SWilson4 : Maybe a sensible addition to CONTRIBUTING.md?

baentsch:
> Please let me know if we can merge the PR.

@songlingatpan please see a single comment I just added (creating an issue to track a problem pointed out by @SWilson4 )

@SWilson4 would you please mark all your comments resolved that you consider such? Makes reviewing reviews easier (at least for me).

src/kem/ntruprime/pqclean_sntrup761_avx2/params.h
SWilson4: Resolved in https://github.com/open-quantum-safe/liboqs/pull/1926/commits/320c6e11d61f4c5c481b687f20f3999df411d3bb.

songlingatpan:
> > Please let me know if we can merge the PR.
> 
> @songlingatpan please see a single comment I just added (creating an issue to track a problem pointed out by @SWilson4 )
> 
> @SWilson4 would you please mark all your comments resolved that you consider such? Makes reviewing reviews easier (at least for me).

No need additional code changes for upstream:
python3 copy_from_upstream.py copy &&           ! git status | grep -i modified
Updating /home/shan/oqs_github/liboqs/docs/algorithms/kem/bike.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/kem/classic_mceliece.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/kem/frodokem.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/kem/hqc.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/kem/kyber.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/kem/ml_kem.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/kem/ntruprime.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/sig/cross.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/sig/dilithium.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/sig/falcon.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/sig/mayo.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/sig/ml_dsa.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/sig/sphincs.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/sig_stfl/lms.md
Updating /home/shan/oqs_github/liboqs/docs/algorithms/sig_stfl/xmss.md
Updating README.md

songlingatpan:
@SWilson4 
Please let me know if anything else I need to address in this PR.

tests/test_code_conventions.py
songlingatpan: @baentsch I will add some code to handle it shortly.

SWilson4:
> @SWilson4 Please let me know if anything else I need to address in this PR.

As I commented [here](https://github.com/open-quantum-safe/liboqs/pull/1926#discussion_r1786493334), there are still a number of checks that need to be added after `OQS_MEM_malloc` (e.g., in the SHA2 code).

songlingatpan:
> OQS_MEM_malloc

@SWilson4 Thanks. I have another PR which does the check and memory leak in https://github.com/open-quantum-safe/liboqs/pull/1929/files#
Would you like to merge the two PRs or keep them separately? The current one is for customize memory allocator.
Please let me know.

SWilson4:
> > OQS_MEM_malloc
> 
> @SWilson4 Thanks. I have another PR which does the check and memory leak in https://github.com/open-quantum-safe/liboqs/pull/1929/files# Would you like to merge the two PRs or keep them separately? The current one is for customize memory allocator. Please let me know.

I would prefer to address the issues _introduced_ by this PR in the same PR—i.e., add all checks necessitated by the removal of the `checked_malloc` functions in this PR. The additional checks are fine to defer to the next PR, but I would rather this PR not introduce new issues to fix.

src/common/sha3/ossl_sha3x4.c
```diff
@@ -81,7 +81,10 @@ static void SHA3_shake128_x4_inc_squeeze(uint8_t *out0, uint8_t *out1, uint8_t *
 		OSSL_FUNC(EVP_MD_CTX_copy_ex)(clone, s->mdctx3);
 		OSSL_FUNC(EVP_DigestFinalXOF)(clone, out3, outlen);
 	} else {
-		uint8_t *tmp = OQS_MEM_checked_malloc(s->n_out + outlen);
+		uint8_t *tmp = OQS_MEM_malloc(s->n_out + outlen);
+		if (!tmp) {
+			return;
```
SWilson4, line 86: This and similar checks should (for now, at least) call `abort` or `exit`—the SHA3 init functions are not checked in the code that uses them (maybe they should be, but that's out of scope here IMO).

SWilson4 (commented):
Hi @songlingatpan—just one more quick point regarding the replacement of the checked functions. Once that and the CI failures are fixed this is good to go from my point of view.

src/common/sha3/ossl_sha3x4.c
songlingatpan: @SWilson4 For the long term, we should return error codes and allow the caller to determine the appropriate behavior. Can we, for now, use return instead of invoking abort() or exit()?

We shouldn't assume that the caller process needs to terminate if SHA3_shake128_x4_inc_init fails. Otherwise, we won’t be able to integrate liboqs into our system. For example, in some cases, SHA3_shake128_x4_inc_init is not a critical function, and the caller may choose to continue the process even if it fails.

Please let me know your thoughts.

songlingatpan:
> Hi @songlingatpan—just one more quick point regarding the replacement of the checked functions. Once that and the CI failures are fixed this is good to go from my point of view.
@SWilson4 Need to discuss with you in email. Thanks

songlingatpan:
@SWilson4 @baentsch @baentsch 

Because eventually we need to remove all abort()/exit() from the library, currently I think it is fine to fully get rid of abort()/exit() even if subsequence is dereference NULL pointer.

This is a trivial thing I think we shouldn't waste so much time on it. We have so many critical issues to make liboqs stable, and would like to complete them more efficiently. 

Please let me know if we really need to keep abort()/exit() for now. 

Thanks

songlingatpan:
@SWilson4 @baentsch @baentsch 

Because eventually we need to remove all abort()/exit() from the library, currently I think it is fine to fully get rid of abort()/exit() even if subsequence is dereference NULL pointer.

This is a trivial thing I think we shouldn't waste so much time on it. We have so many critical issues to make liboqs stable, and would like to complete them more efficiently. 

Please let me know if we really need to keep abort()/exit() for now. 

Thanks

baentsch:
> I think it is fine to fully get rid of abort()/exit() even if subsequence is dereference NULL pointer.

If this is meant to state that the code should rather dump core than abort()s in a controlled manner (?), I'm not sure this is a good idea. I'd rather suggest tackling that as part of resolving https://github.com/open-quantum-safe/liboqs/issues/1456 (controlled error-exits throughout the library).

songlingatpan:
> > I think it is fine to fully get rid of abort()/exit() even if subsequence is dereference NULL pointer.
> 
> If this is meant to state that the code should rather dump core than abort()s in a controlled manner (?), I'm not sure this is a good idea. I'd rather suggest tackling that as part of resolving #1456 (controlled error-exits throughout the library).

This is between a rock and a hard place.
In our in-house version, after add all proper NULL checks and memory releases, and remove all abort/exit, it can work without crash at least. Basically it is not usable to have abort/exit in a library. There is no single abort/exit in openssl library code except test code. Anyway, I will add abort() back.

songlingatpan:
@SWilson4 @baentsch 
Addressed the latest comments about abort() in [1e4201b](https://github.com/open-quantum-safe/liboqs/pull/1926/commits/1e4201bc280c454a5f6501eef682554356c3c21f)
Please let me know if we are good to merge the PR.
Thanks

SWilson4 (approved):
Thanks very much for this contribution, @songlingatpan! There have been significant changes since the reviews by @baentsch and @dstebila. I'm going to dismiss those two reviews as stale and re-request reviews from them (or anybody else from @open-quantum-safe/liboqs-committers).

songlingatpan:
@SWilson4 Thanks for your time and all the comments.

songlingatpan:
@baentsch @dstebila 
Is there anything you would like me to address in this PR? 
Thanks

baentsch:
> @baentsch @dstebila Is there anything you would like me to address in this PR? Thanks

Apologies @songlingatpan -- this PR is just too large for me to take a "quick look" to approve and I have a large list of other things right now demanding urgent attention. Tagging @open-quantum-safe/liboqs-committers to help out reviewing and approving this.

src/common/common.c
```diff
@@ -369,18 +378,23 @@ void *OQS_MEM_aligned_alloc(size_t alignment, size_t size) {
...
 void OQS_MEM_aligned_free(void *ptr) {
-#if defined(OQS_HAVE_ALIGNED_ALLOC) || defined(OQS_HAVE_POSIX_MEMALIGN) || defined(OQS_HAVE_MEMALIGN)
-	free(ptr); // IGNORE free-check
+	if (ptr == NULL) {
+		return;
+	}
+#if defined(OQS_USE_OPENSSL)
+	// Use OpenSSL's free function
+	uint8_t *u8ptr = ptr;
+	OPENSSL_free(u8ptr - u8ptr[-1]);
```
baentsch, line 387: Why not simply OPENSSL_free(ptr)?

baentsch: Forget that -- that's because of "OQS_MEM_aligned_alloc".

src/common/sha2/sha2_c.c
```diff
@@ -542,8 +548,10 @@ void oqs_sha2_sha384_inc_init_c(sha384ctx *state) {
 }
 
 void oqs_sha2_sha512_inc_init_c(sha512ctx *state) {
-	state->ctx = OQS_MEM_checked_malloc(PQC_SHA512CTX_BYTES);
-
+	state->ctx = OQS_MEM_malloc(PQC_SHA512CTX_BYTES);
```
baentsch, line 551: Why this change? It seems to just silence a (imo somewhat helpful) warning output.

src/common/sha2/sha2_c.c
```diff
@@ -555,55 +563,63 @@ void oqs_sha2_sha512_inc_init_c(sha512ctx *state) {
 }
 
 void oqs_sha2_sha224_inc_ctx_clone_c(sha224ctx *stateout, const sha224ctx *statein) {
-	stateout->ctx = OQS_MEM_checked_malloc(PQC_SHA256CTX_BYTES);
-
+	stateout->ctx = OQS_MEM_malloc(PQC_SHA256CTX_BYTES);
```
baentsch, line 566: same as above

src/common/sha2/sha2_c.c
```diff
@@ -555,55 +563,63 @@ void oqs_sha2_sha512_inc_init_c(sha512ctx *state) {
...
 /* Destroy the hash state. */
 void oqs_sha2_sha224_inc_ctx_release_c(sha224ctx *state) {
-	free(state->ctx); // IGNORE free-check
+	OQS_MEM_insecure_free(state->ctx);
 }
 
 /* Destroy the hash state. */
 void oqs_sha2_sha256_inc_ctx_release_c(sha256ctx *state) {
-	free(state->ctx); // IGNORE free-check
+	OQS_MEM_insecure_free(state->ctx);
```
baentsch, line 612: This is a question to @dstebila (as you added this 6 years ago): What's the rationale for having a function explicitly not checking for ptr==NULL, even labelling it "insecure"?

src/common/sha2/sha2_c.c
```diff
@@ -691,8 +709,10 @@ void oqs_sha2_sha256_inc_finalize_c(uint8_t *out, sha256ctx *state, const uint8_
 	if (new_inlen == inlen) {
 		new_in = in;
 	} else { //Combine incremental data with final input
-		tmp_in = OQS_MEM_checked_malloc(tmp_len);
-
+		tmp_in = OQS_MEM_malloc(tmp_len);
```
baentsch, line 712: lots of instances with this pattern: Why "externalize" the ==NULL check dropping the warning output in OQS_MEM_checked_malloc? This does not provide more code safety but only less user information, no?

src/common/sha2/sha2_c.c
baentsch: And Yes, I know "free(NULL)" will not crash -- but I'd really like to understand the idea behind this function: Was it meant to contain some memory sanity check logic? Do we have to retain it? If not, this PR may be simpler....

baentsch (changes requested):
@songlingatpan most of this LGTM and improves the software (malloc->OQS_MEM_malloc & additional NULL checks). Some things I don't understand (see single comments). What I'm really unsure about is the apparent removal of (many) user-visible allocation error warnings for silent abort()s. Isn't that a step in the wrong direction? Yes, the abort() is a stupid pattern to begin with, but isn't it reasonable to let users at least know why the lib/app suddenly stopped?

songlingatpan:
> @songlingatpan most of this LGTM and improves the software (malloc->OQS_MEM_malloc & additional NULL checks). Some things I don't understand (see single comments). What I'm really unsure about is the apparent removal of (many) user-visible allocation error warnings for silent abort()s. Isn't that a step in the wrong direction? Yes, the abort() is a stupid pattern to begin with, but isn't it reasonable to let users at least know why the lib/app suddenly stopped?

@baentsch 
I understand your point.

There was some discussion with @SWilson4 about removing the checked malloc logic. If you review the OpenSSL code, it doesn’t use the checked malloc/abort() pattern. In general, a well-designed library should avoid including abort() or exit() calls, as these decisions should be left to the caller. Ideally, malloc failures should be handled consistently across the codebase by returning error codes, rather than aborting or exiting.

Since there is no fprintf logging for non-checked malloc failures, I didn’t add stderr output for abort(). I recommend fully removing abort() in the 0.12.0 release.

Our in-house version is much more stable than 0.11.0 and target for production readiness compared to the official liboqs repository. However, I’m facing challenges in integrating our in-house code changes back into the official liboqs due to abort(), and so on.

Anyway I can add fprintf back for abort.

baentsch:
> Anyway I can add fprintf back for abort.

That'd be nice. And for the avoidance of doubt: My ask (to achieve this) is basically to reduce the amount of new code (by retaining "OQS_MEM_checked_malloc" calls), not to add more new code (fprintf statements, I guess)

songlingatpan:
> > Anyway I can add fprintf back for abort.
> 
> That'd be nice. And for the avoidance of doubt: My ask (to achieve this) is basically to reduce the amount of new code (by retaining "OQS_MEM_checked_malloc" calls), not to add more new code (fprintf statements, I guess)

@baentsch We can remove the checked malloc for now. 
In the next step, we can remove abort and add return proper error and fix potential memory leaks.
If you target to the production, eventually we need to remove the checked malloc and do proper error handling and release resources, etc.

songlingatpan:
> > > Anyway I can add fprintf back for abort.
> > 
> > 
> > That'd be nice. And for the avoidance of doubt: My ask (to achieve this) is basically to reduce the amount of new code (by retaining "OQS_MEM_checked_malloc" calls), not to add more new code (fprintf statements, I guess)
> 
> @baentsch We can remove the checked malloc for now. In the next step, we can remove abort and add return proper error and fix potential memory leaks. If you target to the production, eventually we need to remove the checked malloc and do proper error handling and release resources, etc.

Please let me know if you still want to keep the checked malloc. Still, if you check openssl code, there is no checked malloc. We need to remove it eventually.

songlingatpan:
@baentsch Please review the PR.

Thanks

src/common/sha2/sha2_c.c
baentsch: fixed by OQS_EXIT_IF_NULLPTR

src/common/common.c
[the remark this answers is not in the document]
baentsch: so now done -- agree, @SWilson4 (and mark resolved)?

src/common/sha3/xkcp_sha3.c
```diff
@@ -310,7 +313,11 @@ static void SHA3_shake128(uint8_t *output, size_t outlen, const uint8_t *input,
 /* SHAKE128 incremental */
 
 static void SHA3_shake128_inc_init(OQS_SHA3_shake128_inc_ctx *state) {
-	state->ctx = OQS_MEM_checked_aligned_alloc(KECCAK_CTX_ALIGNMENT, KECCAK_CTX_BYTES);
+	if (state == NULL) {
```
dstebila, line 316: I see that this SHA3 init function got an `if (state == NULL)` guard, but most of the others didn't.  Is there a rationale for this?

src/common/common.h
```diff
@@ -32,6 +85,18 @@ extern "C" {
         }                                                                             \
     } while (0)
 
+/**
+ * Macro for terminating the program if a given pointer is NULL.
+ * @param ptr The pointer to check.
+ * @param msg The error message to display if the pointer is NULL.
+ */
+#define LIBOQS_die(ptr, msg) do { \
```
dstebila, line 93: As far as I can tell, this macros is never used anywhere in the code.

baentsch: This can be removed as it's replaced by OQS_EXIT_IF_NULLPTR. Thanks for the catch @dstebila

src/common/sha3/xkcp_sha3.c
SWilson4: I'd advocate to remove the guard for consistency/clarity. All it does is defer an eventual segfault to another SHA3 function (e.g., `absorb`). We should make sure the segfault doesn't happen, period, but that's for another PR.

src/common/common.h
songlingatpan: > As far as I can tell, this macros is never used anywhere in the code.
Sure. I will remove it shortly today.

src/common/sha3/xkcp_sha3.c
songlingatpan: I will remove the check here. thanks

dstebila (approved):

baentsch (approved):

PR #1934, opened 2024-09-25
Adds sha3 to arm compile options which is needed for ios compilation.

### LINKED ISSUE #1933: Compilation error: <instantiation>:16:5: error: instruction requires: sha3
**Describe the bug**

Compilation for ios fails:

```
13:23:30  liboqs/src/kem/kyber/oldpqclean_kyber512_aarch64/feat.S:147:5: note: while in macro instantiation
13:23:30      round
13:23:30      ^
13:23:30  <instantiation>:16:5: error: instruction requires: sha3
13:23:30      rax1.2d v29, v27, v29 ; d[3] = rotl(p[4], 1) ^ p[2]
```

I've found this [stackoverflow question](https://stackoverflow.com/questions/72100202/gcc-11-3-0-on-macos-apple-silicon-and-sha-3-instructions) that suggests adding `sha3` to the `-march` option. Doing that seems to solve the issue.

Additional question:
We want to support macos and ios arm/intel variants. How do we add some CI pipeline that tests compilation for those platforms?

fwh-dc:
Fixes #1933

### DIFF
--- a/.CMake/compiler_opts.cmake
+++ b/.CMake/compiler_opts.cmake
@@ -55,7 +55,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang|GNU")
         else()
             # Assume sensible default like -march=x86-64, -march=armv8-a, etc.
 	    if(ARCH_ARM64v8)
-                set(OQS_OPT_FLAG "-march=armv8-a+crypto")
+                set(OQS_OPT_FLAG "-march=armv8-a+crypto+sha3")
     	    else()
 		set(OQS_OPT_FLAG "")
 	    endif()


Martyrshot:
Thanks for the contribution! How would this behave on arm platforms that do not have the sha3 extensions?

fwh-dc:
> Thanks for the contribution! How would this behave on arm platforms that do not have the sha3 extensions?

TBH I don't know and I've trouble finding documentation on these flags. Perhabs you have some suggestions on where to look?

.CMake/compiler_opts.cmake
```diff
@@ -55,7 +55,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang|GNU")
         else()
             # Assume sensible default like -march=x86-64, -march=armv8-a, etc.
 	    if(ARCH_ARM64v8)
-                set(OQS_OPT_FLAG "-march=armv8-a+crypto")
+                set(OQS_OPT_FLAG "-march=armv8-a+crypto+sha3")
```
baentsch, line 58: Why is this not guarded/if'd by reference to ios ("APPLE"?) if it's indeed a problem only there?

fwh-dc: I didn't know if it is Apple specific. As I stated in my previous comment I wasn't able to find any documentation on the flags. Do you know where to find the documentation? Maybe this affects other vendors than Apple?

fwh-dc: I'll take this one back to draft and have a look at your suggestion in the related issue. Maybe I can get it to work with the toolchain file of this project.

dstebila:
Tagging @hanno-becker to see if he has any insights on ARM compilation flags.

hanno-becker:
I would recommend only using the `sha3`-flag if one is certain that the target platform supports it. Even if the SHA3-instruction-based Keccak implementation is not actually called, there is risk that a clever compiler will leverage general-purpose instructions such as `eor3` in (auto-)vectorized code, leading to an invalid instruction abort at runtime.

How does one detect/convey more details about the target in CMake?

In addition, `feat.S` should unconditionally guard the code by `__ARM_FEATURE_SHA3` (even on Apple).

.CMake/compiler_opts.cmake
cothan: Hi @fwh-dc , this PR makes compilation on Cortex-A53 (as you can see in CI) failed. It does not have SHA3 extension.

vincentvbh:
> I would recommend only using the `sha3`-flag if one is certain that the target platform supports it. Even if the SHA3-instruction-based Keccak implementation is not actually called, there is risk that a clever compiler will leverage general-purpose instructions such as `eor3` in (auto-)vectorized code, leading to an invalid instruction abort at runtime.
> 
> How does one detect/convey more details about the target in CMake?
> 
> In addition, `feat.S` should unconditionally guard the code by `__ARM_FEATURE_SHA3` (even on Apple).

On my Apple M1, __ARM_FEATURE_SHA3 is not defined by default (it will be defined if one compiles with -march=armv8-a+sha3). So I agree with Hanno that currently by default we should compile with -march=armv8-a. A feature detection is recommended to pass +sha3 conditionally.

vincentvbh:
> > Thanks for the contribution! How would this behave on arm platforms that do not have the sha3 extensions?
> 
> TBH I don't know and I've trouble finding documentation on these flags. Perhabs you have some suggestions on where to look?

The testing macros can be found here: https://developer.arm.com/documentation/101028/0012/5--Feature-test-macros. As hardware varies, you might want to test your target platforms yourself.

fwh-dc:
Nice thanks. I'll see when I can find the time to have a look at it.

fwh-dc:
Thanks for the heads up. It is not needed any longer.

PR #1938, opened 2024-09-26
Remove hardcoded build patch from test script

### LINKED ISSUE #1937: New tests in test_acvp_vectors.py fail when using custom build path
**Describe the bug**
When passing to `cmake` a custom build path (e.g. `cmake -G Ninja -B build -S liboqs-0.11.0-rc1`), the tests in new script `test_acvp_vectors.py` fail due to `FileNotFoundError` exceptions. This is because `build` is harcoded in the calls to `helpers.run_subprocess()`.

**To Reproduce**
1. Build with custom build path
2. Run `ninja -C <custom_build_path> run_tests`

**Expected behavior**
All tests should pass

 - OS: Arch Linux
 - OpenSSL version: 3.3.2
 - Compiler version used: GCC 14.2.1
 - Build variables used: Like [here](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=liboqs#n30) plus the new variables to enable stateful hash-based signatures.
 - liboqs version: 0.11.0-rc1

iyanmv:
If a user has passed a custom build path to cmake, and then calls for example `ninja -C <custom_build_path> run_tests`, the script test_acvp_vectors.py fails due to having "build" harcoded in the calls.

Instead, let's use `helpers.get_current_build_dir_name()` to get the build path and use that instead. This is already done in other scripts (e.g., test_binary.py)

Fixes #1937

### DIFF
--- a/tests/test_acvp_vectors.py
+++ b/tests/test_acvp_vectors.py
@@ -36,9 +36,10 @@ def test_acvp_vec_kem_keygen(kem_name):
                     z = testCase["z"]
                     pk = testCase["ek"]
                     sk = testCase["dk"]
-                    
+
+                    build_dir = helpers.get_current_build_dir_name()
                     helpers.run_subprocess(
-                        ['build/tests/vectors_kem', kem_name, "keyGen", d+z, pk, sk]
+                        [f'{build_dir}/tests/vectors_kem', kem_name, "keyGen", d+z, pk, sk]
                     )
 
         assert(variantFound == True)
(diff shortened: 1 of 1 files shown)


dstebila (approved):

dstebila:
Thanks for the fix!

We're about to make a release of liboqs 0.11.0 from the main branch, so let's hold off on merging this until then.

SWilson4:
LGTM as well, but since my approval will enable merge I will hold off on approving until after the release. Thanks very much for the contribution, @iyanmv.

SWilson4 (approved):

PR #1944, opened 2024-10-02
Add minimal c++ example and check

### LINKED ISSUE #1770: Add C++ test to ci
There was an issue compiling the `stateful-sigs` branch for `C++`, but CI was passing. We should probably add a simple c++ test to CI to prevent issues like this in the future.

aidenfoxivey:
Adding a simple C++ test to CI.

Fixes #1770.

Doesn't change any cryptographic algorithms.

Doesn't change any cryptographic algorithms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ The following instructions assume we are in `build`.
 	- `kat_sig_stfl`: Program for checking results against submitted KAT values using `tests/test_kat.py`
 	- `speed_kem`: Benchmarking program for key encapsulation mechanisms; see `./speed_kem --help` for usage instructions
 	- `speed_sig`: Benchmarking program for signature mechanisms; see `./speed_sig --help` for usage instructions
+	- `speed_sig_stfl`: Benchmarking program for stateful signature mechanisms; see `./speed_sig_stfl --help` for usage instructions
 	- `example_kem`: Minimal runnable example showing the usage of the KEM API
 	- `example_sig`: Minimal runnable example showing the usage of the signature API
 	- `example_sig_stfl`: Minimal runnable example showing the usage of the stateful signature API
--- a/scripts/noregress.sh
+++ b/scripts/noregress.sh
@@ -15,7 +15,18 @@ fi
 
 # Approach: Check out $1 into tmp folder, build, run speed_kem|sig and compare results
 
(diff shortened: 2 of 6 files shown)


vsoftco:
@aidenfoxivey We have a C++ wrapper that provides all the required RAII , https://github.com/open-quantum-safe/liboqs-cpp

aidenfoxivey:
@vsoftco Would it be fair to rework this with the C++ bindings and then continue working on the PR?

aidenfoxivey:
Absolutely fine to close the PR too, but I feel like it should probably be addressed in the issue first, just to make the procedure clear.

Martyrshot:
@vsoftco The purpose of this is to catch syntax errors in C++. For example when the stateful-sigs PR was in progress we almost merged something that compiled for C but failed to link against a C++ target.

Since we have #ifdef CPLUSPLUS blocks in our headers I think this check needs to be done

aidenfoxivey:
@Martyrshot Do we want to compile and run the resulting example file? (and/or use valgrind to check it for leaks?) in my existing implementation I'm just compiling and linking it.

Martyrshot:
I think valgrind for this is overkill, but running it is a good idea. It could catch weird runtime linking errors. Sorry I should have thought about this sooner. Would it be much work for you to add this test to the ninja `run_tests` rule? Then it would always be run rather then having a specific ci job.

If it's too much work, I can take a look into doing this later in the week.

aidenfoxivey:
Oh no, don't worry about it! I can absolutely sort it out later tonight or tomorrow.

aidenfoxivey:
@Martyrshot I thought about it a bit and, if I understand you correctly, this requires modifying the CMakeLists.txt to add C++ as a LANGUAGE. Are you cool with that?

Martyrshot:
Would we be able to isolate that to _just_ this test?

aidenfoxivey:
The `ninja` run-tests is generated based on CMakeLists.txt right? I think in that case, you pretty much have to change the CMake infrastructure to allow it to build whenever you call `ninja run-tests`. The reason I didn't have to earlier was because I was just hardcoding a call to `g++` in the Github action.

Martyrshot:
That's fair. Then in that case I think we should avoid modifying `CMakeLists.txt`. Let's relocate the `.cpp` file you created into the `.github` folder somewhere so that it's not mistaken as a core liboqs test. Then have the c++ action run the test and enforce that it doesn't error.

aidenfoxivey:
Should have this sorted by end of next week. Apologies about the delay - midterms :)

baentsch:
> midterms :)

Good luck!

Martyrshot:
> Should have this sorted by end of next week. Apologies about the delay - midterms :)

No worries! Good luck!

PR #1947, opened 2024-10-08
Test Improvements for ML-KEM

abhinav-thales:
This PR does the following:
	- enable ML-KEM ACVP tests on Windows
	Since ML-KEM final specification is released, it would be good to have the tests enabled on Windows
	- testcases for ML-KEM rejection key
	Add negative testcase as per Algorithm 18 of ML-KEM final specification when the calculated encrypted text does not match the passed ciphertext
       - harden checks in vector_kem for RNG lengths and correct error message in decaps


SWilson4:
Not sure if you're working on Windows @abhinav-thales, but [this script](https://github.com/open-quantum-safe/liboqs/blob/main/CONTRIBUTING.md#coding-style) might be helpful to get past the formatting check.

SWilson4:
@abhinav-thales I triggered CI on the latest commit; please see the minor build failures.

tests/test_kem.c
```diff
@@ -127,6 +132,37 @@ static OQS_STATUS kem_test_correctness(const char *method_name) {
...
+		uint8_t *buff_z_c = NULL;
+		size_t length_z_c = (size_t)32 + kem->length_ciphertext;
+		buff_z_c = OQS_MEM_malloc(length_z_c) ;
+		if (NULL == buff_z_c) {
+			fprintf(stderr, "ERROR: OQS_MEM_malloc failed\n");
+			goto err;
+		}
+		// test rejection key by corrupting the secret key
+		secret_key[0] += 1;
+		uint8_t shared_secret_r[32]; // expected output
```
bhess, line 146: Could the size refer to a macro? E.g. `OQS_KEM_ml_kem_512_length_shared_secret`, to avoid hardcoding constants.

tests/test_kem.c
```diff
@@ -127,6 +132,37 @@ static OQS_STATUS kem_test_correctness(const char *method_name) {
...
+		memcpy(&buff_z_c[32], ciphertext, kem->length_ciphertext);
+		// calculate expected secret in case of corrupted cipher : shake256(z || c)
+		OQS_SHA3_shake256(shared_secret_r, 32, buff_z_c, length_z_c);
+		OQS_MEM_secure_free(buff_z_c, length_z_c);
+		rc = OQS_KEM_decaps(kem, shared_secret_d, ciphertext, secret_key);
+		if (rc != OQS_SUCCESS) {
+			fprintf(stderr, "ERROR: OQS_KEM_decaps failed for rejection testcase\n");
+			goto err;
+		}
+		rv = memcmp(shared_secret_d, shared_secret_r, kem->length_shared_secret);
```
bhess, line 157: Did you check a constant-time test run with the code? I suspect the `shared_secret_r` variable should be declassified as the memcmp would lead to a false-positive. See the `OQS_TEST_CT_DECLASSIFY` statements earlier in the function.

bhess (changes requested):
Thank you for this additional test @abhinav-thales ! 
See the to comments in the code.
The new test is specific to ML-KEM while the existing tests apply to all KEMs. I think it would make sense to structure the code a bit differently, for example moving the new test to separate function.

tests/test_kem.c
abhinav-thales: done. macro added instead of 32

tests/test_kem.c
abhinav-thales: return value for decaps, shared_secret_r & shared_secret_d are declasified before usage

abhinav-thales:
> Thank you for this additional test @abhinav-thales ! See the to comments in the code. The new test is specific to ML-KEM while the existing tests apply to all KEMs. I think it would make sense to structure the code a bit differently, for example moving the new test to separate function.

So, i move the entire "if" condition to a new function which returns may be returns "true" for tests for ml-kem if passed(false otherwise) and for all other kem algos returns true always ?

tests/test_kem.c
bhess: Thanks!

abhinav-thales:
> Thanks for the updates! One more general comment about the test: The main use case of the rejection key in ML-KEM seems to be if active attackers modify the ciphertext, where the decapsulation should output the rejection key. A test for this would modify the ciphertext and check if the rejection keys match, while your test modifies the secret key. Could you also add a test case that modifies the ciphertext?

added the testcase now.

bhess (approved):
Thanks @abhinav-thales for adding the additional test! Looks good to me.

tests/test_kem.c
```diff
@@ -31,6 +38,89 @@ static void OQS_print_hex_string(const char *label, const uint8_t *str, size_t l
 	printf("\n");
 }
 
+/* mlkem rejection key testcase */
```
baentsch, line 41: OK -- now I get it. But then this should all be #ifdef'd on that algorithm family (being present/enabled), no?

abhinav-thales: Testcase under mlkem conditional check now

tests/test_kem.c
```diff
@@ -20,6 +25,8 @@
 #define OQS_TEST_CT_DECLASSIFY(addr, len)
 #endif
 
+#define MLKEM_SECRET_LEN      32
```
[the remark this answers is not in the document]
abhinav-thales: Yes, correctly pointed as below, the test is specific to ml-kem.

abhinav-thales:
> Thanks for the contribution, @abhinav-thales . Please see the single comments and mark/#ifdef explicitly the algorithm-specific code, e.g., using `OQS_ENABLE_KEM_ML_KEM`, to allow the system to keep its (admittedly only internal) differentiation in generalized and algorithm-specific code.

Hi @baentsch , thanks for the review. The ML-KEM tests are now under conditional check of ML-KEM configuration.

abhinav-thales:
> @abhinav-thales I triggered CI on the latest commit; please see the minor build failures.

Hi @SWilson4 , can you trigger the workflow on the latest commit as well please ?

tests/test_kem.c
```diff
@@ -31,6 +40,89 @@ static void OQS_print_hex_string(const char *label, const uint8_t *str, size_t l
...
+	}
+	ciphertext[0] -= 1; // Restore ciphertext
+	retval = true;
+cleanup:
+	if (buff_z_c) {
+		OQS_MEM_secure_free(buff_z_c, length_z_c);
+	}
+	return retval;
+}
+#endif
```
baentsch, line 124: Nit: Would be good to put name of define opening up this #ifdef into a comment to better understand the code.

tests/test_kem.c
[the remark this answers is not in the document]
abhinav-thales: Although the secret length is 32 bytes across all ML-KEM versions, the shared secret macro in the header is tied to specific ML-KEM versions. This can lead to the macro being undefined if those versions are not enabled. Therefore, the macro has been defined independently of the ML-KEM version.

Was also discussed here : https://github.com/open-quantum-safe/liboqs/pull/1947#discussion_r1808667708

tests/test_kem.c
```diff
@@ -31,6 +40,89 @@ static void OQS_print_hex_string(const char *label, const uint8_t *str, size_t l
...
+#ifdef OQS_ENABLE_KEM_ML_KEM
+/* mlkem rejection key testcase */
+static bool mlkem_rej_testcase(OQS_KEM *kem, uint8_t *ciphertext, uint8_t *secret_key) {
+	// sanity checks
+	if ((kem == NULL) || (ciphertext == NULL) || (secret_key == NULL)) {
+		fprintf(stderr, "ERROR: inputs NULL!\n");
+		return false;
+	}
+	// Only run tests for ML-KEM
+	if (!(strcasecmp(kem->method_name, "ML-KEM-512") == 0 ||
```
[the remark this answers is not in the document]
abhinav-thales: shared secret macro addressed above.

will change the hardcoding to "ML-KEM-**" to OQS macros

tests/vectors_kem.c
```diff
@@ -362,7 +362,7 @@ int main(int argc, char **argv) {
 		kg_pk = argv[4];
 		kg_sk = argv[5];
 
-		if (strlen(prng_output_stream) % 2 != 0 ||
+		if (strlen(prng_output_stream) != 128 ||
```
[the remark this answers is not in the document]
abhinav-thales: will provide brief documentation.

tests/test_kem.c
abhinav-thales: yes, will do

abhinav-thales:
Hi @SWilson4 @baentsch the review comments are addressed now. Could you please help with triggering CI pipeline if the fixes are fine. Thanks.

tests/test_kem.c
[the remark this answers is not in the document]
baentsch: With the newly added guard enabling this only when ML-KEM in general is enabled (and the recognition that ML-KEM is now standardized, so an upstream change breaking this for a single algorithm is _highly_ unlikely), I'm fine with this as-is.

SWilson4:
Looks good to me. Thanks for making the requested changes!

baentsch (approved):

baentsch:
Thanks for re-requesting review @abhinav-thales : I was under the mistaken assumption that by marking all comments resolved things turn green, sorry. Now explicitly marked so. Thanks on behalf of the FOSS community for your contributions to improve the library for everyone!

PR #1952, opened 2024-10-15
Add benchmarking for stateful hash based schemes: speed_sig_stfl

cr-marcstevens:
This pull requests adds an executable `tests/speed_sig_stfl` to be able to benchmark the stateful hash based schemes in liboqs.

The source code `tests/speed_sig_stfl.c` is directly copied from `tests/speed_sig.c`,
but making the relevant changes to use the SIG_STFL functions and types instead.

Here are some issues encountered specific to SIG_STFL code:
- A secure store needs to be set for the secret key, the context must not be NULL
- A stfl sig has a maximum number of signatures, potentially as low as 2^5=32. One needs to create a new secret key once this number of signatures has been reached.
- key generation and signing can be extremely costly for some stfl sig algorithms. Automatic testing should focus only on small parameters.
- You cannot create a new secret key over an old (used up) one. You need to free the old secret key, and create a new one.
- Bugfix new code: be careful to actually return the new secret key pointer between functions. Otherwise there is a use-after-free bug.
- for stfl sigs it is possible to disable key generation and signing, but enable verification. Benchmarking doesn't work in such a case, and it simply needs to abort and return success.

### DIFF
--- a/tests/ds_benchmark.h
+++ b/tests/ds_benchmark.h
@@ -289,4 +289,17 @@ static void _bench_init_perfcounters(void) {
         PRINT_TIMER_AVG(op_name)                                  \
     }
 
+#define TIME_OPERATION_SECONDS_MAXIT(op, op_name, secs, maxit)                                                \
+    {                                                                                                         \
+        DEFINE_TIMER_VARIABLES                                                                                \
+        INITIALIZE_TIMER                                                                                      \
+        uint64_t _bench_time_goal_usecs = 1000000 * secs;                                                     \
+        for (unsigned long long i = 0; i < (maxit) && _bench_time_cumulative < _bench_time_goal_usecs; i++) { \
+            START_TIMER { op; }                                                                               \
+            STOP_TIMER                                                                                        \
+        }                                                                                                     \
+        FINALIZE_TIMER                                                                                        \
+        PRINT_TIMER_AVG(op_name)                                                                              \
+    }
+
 #endif
--- a/tests/ds_benchmark.h
+++ b/tests/ds_benchmark.h
@@ -289,17 +289,20 @@ static void _bench_init_perfcounters(void) {
         PRINT_TIMER_AVG(op_name)                                  \
     }
 
-#define TIME_OPERATION_SECONDS_MAXIT(op, op_name, secs, maxit)                                                \
-    {                                                                                                         \
-        DEFINE_TIMER_VARIABLES                                                                                \
-        INITIALIZE_TIMER                                                                                      \
-        uint64_t _bench_time_goal_usecs = 1000000 * secs;                                                     \
-        for (unsigned long long i = 0; i < (maxit) && _bench_time_cumulative < _bench_time_goal_usecs; i++) { \
-            START_TIMER { op; }                                                                               \
-            STOP_TIMER                                                                                        \
-        }                                                                                                     \
-        FINALIZE_TIMER                                                                                        \
-        PRINT_TIMER_AVG(op_name)                                                                              \
+#define TIME_OPERATION_SECONDS_MAXIT(op, op_name, secs, maxit, refresh)                                           \
+    {                                                                                                             \
+        DEFINE_TIMER_VARIABLES                                                                                    \
+        INITIALIZE_TIMER                                                                                          \
+        uint64_t _bench_time_goal_usecs = 1000000 * secs;                                                         \
+        while (_bench_time_cumulative < _bench_time_goal_usecs) {                                                 \
+            for (unsigned long long i = 0; i < (maxit) && _bench_time_cumulative < _bench_time_goal_usecs; i++) { \
+                START_TIMER { op; }                                                                               \
+                STOP_TIMER                                                                                        \
+            }                                                                                                     \
+            if (_bench_time_cumulative < _bench_time_goal_usecs) { refresh; }                                     \
+        }                                                                                                         \
+        FINALIZE_TIMER                                                                                            \
+        PRINT_TIMER_AVG(op_name)                                                                                  \
     }
 
 #endif
--- a/tests/speed_sig_stfl.c
+++ b/tests/speed_sig_stfl.c
@@ -18,7 +18,18 @@
 #include "system_info.c"
 
 OQS_STATUS dummy_secure_storage(uint8_t *sk_buf, size_t sk_buf_len, void *context) {
+	// suppress unused parameter warning
+	(void)(sk_buf);
+	(void)(sk_buf_len);
+	(void)(context);
 	return OQS_SUCCESS;
+}
+
+// reset secret key: some schemes fail to create a new secret key over a previous secret key
+void reset_secret_key(OQS_SIG_STFL *sig, OQS_SIG_STFL_SECRET_KEY *secret_key) {
+	OQS_SIG_STFL_SECRET_KEY_free(secret_key);
+	secret_key = OQS_SIG_STFL_SECRET_KEY_new(sig->method_name);
+	OQS_SIG_STFL_SECRET_KEY_SET_store_cb(secret_key, &dummy_secure_storage, secret_key);
 }
 
 static void fullcycle(OQS_SIG_STFL *sig, uint8_t *public_key, OQS_SIG_STFL_SECRET_KEY *secret_key, uint8_t *signature, size_t signature_len, uint8_t *message, size_t message_len) {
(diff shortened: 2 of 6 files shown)


SWilson4:
Thanks for the contribution! Would you please add the new executable to the list in the README? It would also be nice to have it run in `tests/test_speed.py`, similarly to `speed_kem` and `speed_sig`.

baentsch (commented):
Thanks very much for the contribution @cr-marcstevens ! The code LGTM. One question, though: Would it be OK for you to add it to the test harness, e.g., at "scripts/noregress.sh" (or any other place where it could run regularly such as to help improve test coverage)?

cr-marcstevens:
I'll look into the extra changes needed throughout the repo with the added executable.
But I'm not fully familiar with how all that infrastructure is meant to be used.

For now I found out that `speed_sig_stfl.c` needed 2 fixes:
- signing didn't actually happen unless a secure keystore is set to the private key, so I provided a dummy one.
- timing sig ops didn't take into account max signatures. So for 2^10 max sigs, it could run out and invalidate the benchmark.

ashman-p (approved):

tests/speed_sig_stfl.c
```diff
@@ -0,0 +1,246 @@
...
+
+OQS_STATUS dummy_secure_storage(uint8_t *sk_buf, size_t sk_buf_len, void *context) {
+	// suppress unused parameter warning
+	(void)(sk_buf);
+	(void)(sk_buf_len);
+	(void)(context);
+	return OQS_SUCCESS;
+}
+
+// reset secret key: some schemes fail to create a new secret key over a previous secret key
```
SWilson4, line 28: I'm curious about this comment: I would have expected all the schemes to behave similarly. Do you have an idea as to why this occurs (or at least which schemes behave differently)?

SWilson4: Thanks for the detailed response! Looping in @ashman-p and @cothan as the primary authors of the LMS / XMSS code to see if this is something we ought to fix.

cr-marcstevens: Note: I actually disagree with not accepting context=NULL.
It might be a valid choice to use context to pass the index of the relevant secret key, which could start at 0.
Or have a secure store for exactly 1 key, and leave context NULL.

tests/ds_benchmark.h
```diff
@@ -289,4 +289,20 @@ static void _bench_init_perfcounters(void) {
         PRINT_TIMER_AVG(op_name)                                  \
     }
 
+#define TIME_OPERATION_SECONDS_MAXIT(op, op_name, secs, maxit, refresh)                                           \
```
cothan, line 292: just minor typos here, I think you mean Matrix, is it right?

cr-marcstevens: No, it is really maxit for max iterations.
A stateful hash has a maximum number of signatures, which can be as low as 2^5=32, so after that many iterations it doesn't properly sign anymore.
To ensure this effect does not result in invalid benchmarks, we have to generate a fresh secret key once in a while.

To that affect I've added a macro TIME_OPERATION_SECONDS_MAXIT that uses 2 nested loops instead and has 2 additional parameters: max innerloop iterations & outerloop code.

tests/speed_sig_stfl.c
cothan: Hi @cr-marcstevens  , thanks for the review of the code. I like your propose. 
I think we can reiterate the code better. Since we merge the LMS and XMSS, there are not many issues opened to improve the interface (perhaps no one use it yet?), so this can be one. 
We can merge this PR in first and solve this problem in separately. What do you think?

cr-marcstevens: That seems like a good choice to first merge this PR, otherwise it will likely go stale, while it is of direct use for benchmarking (at least for us) and comparing schemes.

However it seems it currently doesn't pass CI on a couple of tests.
Some of these simply timeout after 60 minutes, so perhaps sig stfl tests should be quite limited as these can take really long.

Other tests I'm not sure what the problem is.

There is one issue that I'm also aware of:
There is a memory leak when freeing the secret key, because the OQS_SIG_STFL_SECRET_KEY_free functions only clears and frees up the internal structure, but not the struct that secret key itself point to.
See:
https://github.com/open-quantum-safe/liboqs/blob/81b4452b9a653f5e203b4bd48ff6fa87a65d0147/src/sig_stfl/xmss/sig_stfl_xmss_secret_key_functions.c#L137

Since speed_sig_stfl needs to free and new secret keys often, this might be  a particular issue in running these benchmarks.

ashman-p: @cr-marcstevens, echoing gratitude for this contribution and feedback. As you pointed out there is need to protect the secret key. Protect it from reuse and loss. As a result all that weight was passed on to the caller of the API.
Perhaps better documentation would be helpful. 

Similarly, AFAIR, i think caller is also tasked with freeing memory foe the actual secret key (did not want the library to blindly free a void pointer).  
This feature was perhaps one of the more contested ones hence the hoops needed to enable key and signature generation.

cr-marcstevens: In this case the user needs to create a secret pointer through SECRET_KEY_new.
This functions first mallocs a general OQS_SIG_STFL_SECRET_KEY object (to which a pointer is returned in the end), and then mallocs a scheme dependent internal secret_key_data.
See:
https://github.com/open-quantum-safe/liboqs/blob/81b4452b9a653f5e203b4bd48ff6fa87a65d0147/src/sig_stfl/xmss/sig_stfl_xmss_secret_key_functions.c#L14

But the corresponding SECRET_KEY_free only frees the internal secret_key_data, but not the general object.
It needs another free after that to really clean things up.

I would argue against this, and have the general object be directly freed as well.
Because the naming suggests this paradigm anyway, but also because the partially freed object is of no further use: it cannot be reused, because any call to SECRET_KEY_NEW would malloc a new general object anyway, not reuse the current one.

ashman-p: @cr-marcstevens , see the API call to free secret keys. 
[https://github.com/open-quantum-safe/liboqs/blob/81b4452b9a653f5e203b4bd48ff6fa87a65d0147/src/sig_stfl/sig_stfl.c#L1512](OQS_SIG_STFL_SECRET_KEY_free)

cr-marcstevens: @ashman-p Thanks for the pointer! I noticed this weird omission in the scheme specific code and thought there was a memory-leak, so I wanted to point that out.
Thanks for pointing out this is actually handled in the generic code, my mistake.

cothan (approved):

tests/speed_sig_stfl.c
dstebila: Is this comment resolved? If so I can merge the PR.

baentsch: Thanks for pointing this out @SWilson4 ! @open-quantum-safe/liboqs-committers if you subscribe to the thought that [liboqs at some point in the future should be of "productive utility"](https://github.com/open-quantum-safe/tsc/issues/1) than please heed good practices contributing to eventually achieve this status: Existing tests must pass for PRs && new features must come with relevant new tests && PR discussions must be (marked) resolved (by the people raising the question). If you don't subscribe to this goal, well, then of course substantially less rigor can be applied, saving all of us time and effort and keep moving the library forward functionally as a research vehicle.

cothan: @cr-marcstevens , oh right, I forget to mention that the Keygen operation should be skipped, since the Key generation can take from minutes to 3 hours to complete, (or only benchmark the keygen of the smallest parameters), so I think the benchmark probably should focus on Sign and Verify operations.

cr-marcstevens:
I've noticed two issues and resolved them:
- Bugfix new code: be careful to actually return the new secret key pointer between functions. Otherwise there is a use-after-free bug.
- for stfl sigs it is possible to disable key generation and signing, but enable verification. Benchmarking doesn't work in such a case, and it simply needs to abort and return success.

cothan:
@cr-marcstevens All tests have passed. Feel free to merge it whenever you like. 
Great work.

cr-marcstevens:
Indeed I think all issues are resolved now, so it is ready for merging.

tests/speed_sig_stfl.c
baentsch: So before merging can I ask what's the conclusion now? Skip keygen? Make this optional (such that it can be omitted from CI but can be manually executed)? Again, my recommendation is that the originator of the comment marks it resolved if he's content such as to assure someone like me to judge "closure" and merge.

cr-marcstevens: My current solution is to only test algorithms with 2^10 signatures.
That is all XMSS_*_10_* and LMS*_H10_*.
So these do get tested in CI, but the computational cost is quite low.

To that end `speed_sig_stfl` has an added option `--limit10`, which instead of all algs tests only those algs.

cr-marcstevens: BTW1 the current code requires keygen to generate secret keys, to then do signing and verification with. Avoiding keygen by eg serialization/deserialization of an offline generated secret, is I think it too much complication just for a benchmarking program.
BTW2 for LMS the secret key seems to be just the seed, and signing essentially requires recomputing the entire tree again anyway.

baentsch: Thanks for the explanations @cr-marcstevens . I now also saw the command line options to that effect . I'd now be OK merging this if @SWilson4 marks the conversation resolved.

PR #1953, opened 2024-10-17
Fix for Zephyr CI

### LINKED ISSUE #1948: Zephyr builds: revert to latest ci image once fix is available
Zephyr builds in CI fail since about [a week](https://github.com/open-quantum-safe/liboqs/actions/runs/11147629806/job/30988529872), while the liboqs changes in the builds seem unrelated to zephyr.

The timing coincides with a zephyr container update https://github.com/zephyrproject-rtos/docker-image/pkgs/container/ci/282662529?tag=main-amd64.

I haven't found the root cause of the build failure, but https://github.com/zephyrproject-rtos/docker-image/issues/205 suggests there might be an issue with the SDK used in the container image and the recommendation is to downgrade to v0.26.14. Do you have any insight here @Frauschi ?

Frauschi:
* Revert to latest Zephyr SDK container (fixes #1948)
* Fix for Zephyr build (missing include of `stddef.h` for `ptrdiff_t`)
* Zephyr CI tests are now using Zephyr V3.4 (minimal supported version) and Zephyr V3.7 (current LTS release)


SWilson4 (approved):
Thanks for the fix!

bhess (approved):

PR #1955, opened 2024-10-22
Adapt existing sig fuzz harness including more algorithms

nathaniel-brough:
Adapts existing fuzz-harness to support more algorithms.

This is a continuation of the effort described in #1215 

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -95,9 +95,9 @@ add_executable(example_sig example_sig.c)
 target_link_libraries(example_sig PRIVATE ${TEST_DEPS})
 
 if(OQS_BUILD_FUZZ_TESTS AND '${CMAKE_C_COMPILER_ID}' STREQUAL 'Clang')
-    add_executable(fuzz_test_dilithium2 fuzz_test_dilithium2.c)
-    target_link_libraries(fuzz_test_dilithium2 PRIVATE ${TEST_DEPS})
-    set_target_properties(fuzz_test_dilithium2 PROPERTIES
+    add_executable(fuzz_test_sig fuzz_test_sig.c)
+    target_link_libraries(fuzz_test_sig PRIVATE ${TEST_DEPS})
+    set_target_properties(fuzz_test_sig PROPERTIES
         COMPILE_FLAGS "${FUZZING_COMPILE_FLAGS}"
         LINK_FLAGS "${FUZZING_LINK_FLAGS}"
     )
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -113,9 +113,9 @@ jobs:
           cmake -LA -N .. && \
           ! (grep -i "uninitialized variable" config.log)
       - name: Build code
-        run: ninja
+        run: ninja fuzz_test_sig
         working-directory: build
 
       - name: Short fuzz check (30s)
-        run: ./tests/fuzz_test_dilithium2 -max_total_time=30
+        run: ./tests/fuzz_test_sig -max_total_time=30
         working-directory: build
--- a/docs/FUZZING.md
+++ b/docs/FUZZING.md
@@ -15,14 +15,11 @@ errors, helping developers identify and fix bugs and security loopholes.
   - [ ] ml_kem
   - [ ] ntruprime
 - [ ] sig
-  - [ ] dilithium
-    - [x] dilithium2
-    - [ ] dilithium3
-    - [ ] dilithium5
-  - [ ] falcon
-  - [ ] mayo
-  - [ ] ml_dsa
-  - [ ] sphincs
+  - [x] dilithium
+  - [x] falcon
+  - [x] mayo
+  - [x] ml_dsa
+  - [x] sphincs
 - [ ] sig_stfl
   - [ ] lms
   - [ ] sig_stfl
--- a/tests/fuzz_test_sig.c
+++ b/tests/fuzz_test_sig.c
@@ -6,6 +6,7 @@
  * SPDX-License-Identifier: MIT
  */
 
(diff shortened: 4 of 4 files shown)


nathaniel-brough:
The failing check doesn't appear to be related to this PR. I'm unsure but it kind of looks like a flaky performance check.

cothan:
It's seem like you found a rare case that the verify operation failed to verify the signature. 
I hope the bug is within the implementation, not the algorithm itself. I don't know about CROSS signature but this one is concerning. 

https://github.com/open-quantum-safe/liboqs/blob/1d92135e808fb2f9f847e080ec8ff3124ee91ac7/tests/speed_sig.c#L29

nathaniel-brough:
Hmm ok. It doesn't look like the public/private keypair or the message is saved on failure here, so I guess it'll be a little difficult to reproduce.

While this seems unrelated to the PR, I'll see if I can modify my fuzzer in such a way to trigger the bug in a way that's reproducible.

nathaniel-brough:
Ok I've updated this fuzz harness (locally) such that it might be able to find this bug in a deterministic way. Is there a preference on stacking changes or just drip feeding in changes so that each PR is atomic i.e. doing one thing?

SWilson4:
Looping in @alexrow and @rtjk as the maintainers of CROSS.

rtjk:
Thank you @SWilson4, we're taking a look at this verification failure and will respond soon.

SWilson4:
> I managed to reproduce this very rare verification error in CROSS-RDSDP-128-small. I'm still looking for the root cause but I can confirm that it seems totaly unrelated to the new fuzzing harness.

Thanks for taking a look! I'll open a separate issue to track the error, then.

tests/CMakeLists.txt
```diff
@@ -94,13 +94,65 @@ set(KEM_TESTS example_kem kat_kem test_kem test_kem_mem speed_kem vectors_kem)
 add_executable(example_sig example_sig.c)
 target_link_libraries(example_sig PRIVATE ${TEST_DEPS})
 
+set(FUZZ_SIG_ALGORITHMS 
+    dilithium_2
```
baentsch, line 98: This list seems to be manually curated, right? This means it has to be equally manually updated whenever an upstream change occurs -- this is not good given the breadth of algorithms supported. I'd therefore strongly advocate the use of the `copy_from_upstream` logic to automatically build this list to avoid future breakage and continued, error-prone manual work.

nathaniel-brough: I found the logic around copy_from_upstream difficult to follow. I've instead just used the fuzzing data to select the algorithm to use at runtime via [OQS_SIG_alg_identifier](https://github.com/open-quantum-safe/liboqs/pull/1955/files#diff-5256ad59f90fd8a1fdf808aecf506ae6c9349ec5e67d87f6259bc4610b1f11e4R44). This should acheive approximately the same thing and reduce the number of fuzz harness executables to one.

baentsch: Yes, iterating through `OQS_SIG_alg_identifier`  looks like an OK approach to me. But is it then still necessary to list all algs manually here? This would mandate manual changes down the road even though `OQS_SIG_alg_identifier` delivers the current list, right?

nathaniel-brough: As far as I understand there would be no manual effort required to update this when algorithms are added/removed. If I'm reading [the docstring for OQS_SIG_alg_identifier](https://github.com/open-quantum-safe/liboqs/blob/3c8bde1b08ba068ebbde738165bf6f8d2ba385ce/src/sig/sig.h#L146) correctly it takes an index (from 0 to OQS_SIG_alg_identifier) and returns the string-type algorithm name which I've then passed into `OQS_SIG_new` [here](https://github.com/open-quantum-safe/liboqs/pull/1955/files#diff-5256ad59f90fd8a1fdf808aecf506ae6c9349ec5e67d87f6259bc4610b1f11e4R46). Under the hood it looks like it's [leveraging the existing code-gen support](https://github.com/open-quantum-safe/liboqs/blob/3c8bde1b08ba068ebbde738165bf6f8d2ba385ce/src/sig/sig.c#L17). So there is no need to manually update anything. There is also [OQS_SIG_algs_length](https://github.com/open-quantum-safe/liboqs/blob/3c8bde1b08ba068ebbde738165bf6f8d2ba385ce/src/sig/sig.h#L134) that is already generated as well which specifies the range of acceptable inputs to `OQS_SIG_alg_identifier`. In short there seems to be enough information around that is already generated, that I can leverage rather than adding additional logic to the code-generation system.

Also just as a note, I'm not really iterating over `OQS_SIG_alg_identifier` rather I'm taking a portion of the fuzzed data inputs [writing that into an integer](https://github.com/open-quantum-safe/liboqs/pull/1955/files#diff-5256ad59f90fd8a1fdf808aecf506ae6c9349ec5e67d87f6259bc4610b1f11e4R35-R39) then [capping the max-value of the integer at OQS_SIG_algs_length](https://github.com/open-quantum-safe/liboqs/pull/1955/files#diff-5256ad59f90fd8a1fdf808aecf506ae6c9349ec5e67d87f6259bc4610b1f11e4R39). This effectively allows the fuzzer to "choose" which algorithm to fuzz. This ensures that there is a large amount of independant "fuzzed data" available when fuzzing each algorithm. I'm happy to elaborate further, as to why this is better than iterating over all algorithms if you are interested.

baentsch: Thanks for the explanation @nathaniel-brough . The list that started this discussion is now gone from the PR, so is my concern.

SWilson4 (approved):

SWilson4:
Thanks for the contribution @nathaniel-brough!

dstebila (approved):

PR #1957, opened 2024-10-22
Add cpp linking test

aidenfoxivey:
Accidentally deleted #1944 and as a result I've remade the PR. I'll make sure to clean up the commits prior to merging if there's any desire to change what I've written here.

I've applied the advice from earlier to move the C++ example to `.github` and added a workflow to `basic.yml`.

PR #1963, opened 2024-10-25
add cpp linking test

ryjones:
This adds a linking to cpp test.

This modifies the build action to build the test.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -91,7 +91,6 @@ jobs:
     runs-on: ubuntu-latest
     container: openquantumsafe/ci-ubuntu-latest:latest
     env:
-      KEM_NAME: ml_kem_768
       SIG_NAME: ml_dsa_65
     steps:
       - name: Checkout code
@@ -102,7 +101,7 @@ jobs:
           cd build && \
           cmake -GNinja -DOQS_STRICT_WARNINGS=ON \
             -GNinja \
-            -DOQS_MINIMAL_BUILD="KEM_$KEM_NAME;SIG_$SIG_NAME" \
+            -DOQS_MINIMAL_BUILD="SIG_$SIG_NAME" \
             --warn-uninitialized .. > config.log 2>&1 && \
           cat config.log && \
           cmake -LA -N .. && \
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -86,7 +86,37 @@ jobs:
         run: ninja gen_docs
         working-directory: build
 
-        
+  cppcheck:
+    name: Check C++ linking with example program
+    runs-on: ubuntu-latest
+    container: openquantumsafe/ci-ubuntu-latest:latest
+    env:
+      KEM_NAME: ml_kem_768
+      SIG_NAME: ml_dsa_65
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # pin@v4
+      - name: Configure
+        run: |
+          mkdir build && \
+          cd build && \
+          cmake -GNinja -DOQS_STRICT_WARNINGS=ON \
+            -GNinja \
+            -DOQS_MINIMAL_BUILD="KEM_$KEM_NAME;SIG_$SIG_NAME" \
+            --warn-uninitialized .. > config.log 2>&1 && \
+          cat config.log && \
+          cmake -LA -N .. && \
+          ! (grep -i "uninitialized variable" config.log)
+      - name: Build liboqs
+        run: ninja
+        working-directory: build
+      - name: Link with C++ program
+        run: |
+          g++ ../cpp/cpp_linking_example.cpp -g \
+          -I./include -L./lib -loqs -lcrypto -std=c++11 -o example_sig && \
+          ./example_sig
+        working-directory: build
+
   fuzzbuildcheck:
     name: Check that code passes a basic fuzzing build
     needs: [ workflowcheck, stylecheck, upstreamcheck ]
--- a/cpp/cpp_linking_example.cpp
+++ b/cpp/cpp_linking_example.cpp
@@ -1,0 +1,182 @@
+/*
+ * example_sig.cpp
+ *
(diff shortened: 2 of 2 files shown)


ryjones:
Hi @aidenfoxivey I moved the code to this PR

aidenfoxivey:
I believe this is a good contender for a "how many PRs does it take an engineering student to complete a 2 file change" joke lol. @ryjones Thank you so much for your help.

.github/workflows/basic.yml
```diff
@@ -86,7 +86,37 @@ jobs:
         run: ninja gen_docs
         working-directory: build
 
-        
+  cppcheck:
+    name: Check C++ linking with example program
+    runs-on: ubuntu-latest
+    container: openquantumsafe/ci-ubuntu-latest:latest
+    env:
+      KEM_NAME: ml_kem_768
```
Martyrshot, line 94: Do we need this environment variable? I think we only need the SIG_NAME for this test.

.github/workflows/basic.yml
```diff
@@ -86,7 +86,37 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # pin@v4
+      - name: Configure
+        run: |
+          mkdir build && \
+          cd build && \
+          cmake -GNinja -DOQS_STRICT_WARNINGS=ON \
+            -GNinja \
+            -DOQS_MINIMAL_BUILD="KEM_$KEM_NAME;SIG_$SIG_NAME" \
```
Martyrshot, line 105:
```suggestion
            -DOQS_MINIMAL_BUILD="SIG_$SIG_NAME" \
```

Martyrshot (dismissed):
This looks good to me! Just a small nit. Thanks for working on this. 😄

.github/workflows/basic.yml
aidenfoxivey: Yep, this makes sense.

.github/workflows/basic.yml
aidenfoxivey: No, don't believe we do - I had left it in to be cautious, but it shouldn't be necessary.

aidenfoxivey:
One thing I had considered, was redirecting output from running the binary, but I think it's better to keep the trace on stderr for debugging.

dstebila (approved):

cpp/cpp_linking_example.cpp
```diff
@@ -0,0 +1,182 @@
...
+ * are only defined if the algorithm dilithium_2 was enabled at compile-time
+ * which must be checked using the OQS_ENABLE_SIG_dilithium_2 macro.
+ *
+ * <oqs/oqsconfig.h>, which is included in <oqs/oqs.h>, contains macros
+ * indicating which algorithms were enabled when this instance of liboqs
+ * was compiled.
+ */
+static OQS_STATUS example_stack(void) {
+
+#ifdef OQS_ENABLE_SIG_dilithium_2
```
baentsch, line 66: Hmm -- I'm not sure I understand this: CI seems to _disable_ Dilithium2 and this test basically only runs when Dilithium2 is _enabled_, so what is actually tested in CI @aidenfoxivey @ryjones ?

ryjones: @aidenfoxivey propose a change in a review comment

baentsch: > I don't have access to the repository to make a change right now.

Why can't you do a proper PR from your own repo @aidenfoxivey ? Proceeding with this PR would make @ryjones the author of record -- which simply is wrong on several counts.

If the problem is that [no reviewer/committer presses the "execute CI" button](https://github.com/open-quantum-safe/liboqs/pull/1957#issuecomment-2436069221) on a PR, well, then this project has some serious problem of getting enough attention... 

@aidenfoxivey if in turn you have problems with the [contribution process](https://github.com/open-quantum-safe/liboqs/blob/main/CONTRIBUTING.md) (where actually the way how to run CI also locally is described) please let us know as part of your PR: We want to improve our documentation such that new contributors can easily, well, contribute.

ryjones: @baentsch the issue was that the PR from his repo cannot execute all of the CI due to no access to secrets.

ryjones: @baentsch It isn't obvious, but I did do non-zero work on this PR.

As far as token availability, this is a limitation (and one I agree with) of the GitHub platform. The way to "work around it" is to set up guardrails around jobs (like triggering CircleCI) that require the job to be within the repo, not a forked repo. I planned to do this later in a follow-on PR for multiple jobs in the org, but that is an early next year thing, not a next week thing.

aidenfoxivey: Regarding next steps, I'm fine to make another pull request and move my changes there. In addition, I'm absolutely fine to just have this merged as is. I'll defer to others on the best approach for this.

baentsch: > It isn't obvious, but I did do non-zero work on this PR.

Thanks for letting me know, @ryjones . I indeed didn't see this. So who would you see as the person that should be primarily attributed with this code (and how to properly attribute the other)?

My proposal would be to do it "as usual", i.e., ask @aidenfoxivey to do a PR from his/her repo and you propose your contributions during the PR: That way, both authors are recorded and we see CI run where possible (and trigger the rest). @aidenfoxivey, it would be nice if you'd [locally execute the new test via CI](https://github.com/open-quantum-safe/liboqs/blob/main/CONTRIBUTING.md#running-ci-locally) (also checking that it actually runs :) before the PR. If you encounter problems in following the documented guidance on that, please let us know.

ryjones: @baentsch : @aidenfoxivey did most of the work, I did some of the work. I'm fine with having my attribution removed as Aiden would have made it there eventually.

I agree with having @aidenfoxivey create a new PR. As it's been approved twice, I think it should be merged as-is and allow Aiden to iterate in follow-up PRs.

aidenfoxivey:
Moved my changes over to #1971

PR #1970, opened 2024-10-31
Add Nix flake

aidenfoxivey:
@praveksharma @SWilson4 As I mentioned earlier, I wrote a little nix flake that should be able to run the environment without anything other than Nix installed on the machine - in addition to being declarative.

I'm happy to add or remove packages to the flake - I mostly followed the requirements for the QuickStart, but potentially missed something.


baentsch:
Thanks for the proposal @aidenfoxivey ! Quick initial question (not knowing anything about nix): Does this have to reside in the top level project directory?

aidenfoxivey:
Hi @baentsch!

In every project I've worked on, flake.nix and flake.lock do reside in the top level. That said, I was curious about the question, so I looked it up.

According to this reference (https://nixos.wiki/wiki/Flakes), flake.nix does have to be in the top-level.

My main goal with this is to have a declarative way to build liboqs to aid debugging and avoid having to use containers.

SWilson4:
Thanks for the submission @aidenfoxivey!

I personally don't know anything about Nix, but I think two non–Nix-related issues would need to be resolved before this lands:

1. The flake should be tested in CI somehow.
2. There should be some sort of maintenance commitment to ensure the flake stays up to date—would you be willing to take this on?

flake.nix
```diff
@@ -0,0 +1,88 @@
...
+    # we need this to access nix packages
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    # we need this to build for each system
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+  outputs = { self, nixpkgs, flake-utils }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        name = "liboqs";
+        version = "0.8.0";
```
SWilson4, line 22: Why 0.8.0?

aidenfoxivey: Good point! I'm not sure I actually recall why I chose that. Do you have a good recommendation for the version?

aidenfoxivey:
> 1. The flake should be tested in CI somehow.
Sounds good! I'll write up a little testing setup sometime this week.

> 2. There should be some sort of maintenance commitment to ensure the flake stays up to date—would you be willing to take this on?
Absolutely! I'm happy to be the designated flake maintainer.

praveksharma (commented):
The flake itself looks good to me, thank you @aidenfoxivey! 

Could you also update the quickstart section of [README.md](https://github.com/open-quantum-safe/liboqs?tab=readme-ov-file#quickstart) to include instruction on how to use the flake?

Do you know if there's some way of parameterising the flake to spin up different dev envs, say with clang instead of gcc?

aidenfoxivey:
> Could you also update the quickstart section of [README.md](https://github.com/open-quantum-safe/liboqs?tab=readme-ov-file#quickstart) to include instruction on how to use the flake?
I'll remove the instructions from the flake and put them in the quickstart along with other information.

> Do you know if there's some way of parameterising the flake to spin up different dev envs, say with clang instead of gcc?
I'll take a look!

aidenfoxivey:
I think it's maybe better to version the flake starting from `0.0.1`, but alternative perspectives are certainly appreciated.

aidenfoxivey:
Not sure what I did to boggle the code formatting lol. Maybe something in how I did the README was an issue?

flake.nix
baentsch: Currently this is "0.11.0" -- but it seems weird to have to hard-code this (and then have to manually change going forward): Any chance to automatically take this from the `cmake` logic?

README.md
```diff
@@ -15,6 +15,7 @@ liboqs is an open source C library for quantum-safe cryptographic algorithms.
 		- [Linux and Mac](#linux-and-mac)
 		- [Windows](#windows)
 		- [Cross compilation](#cross-compilation)
+		- [Nix Flake][#nix-flake]
```
SWilson4, line 18:
```suggestion
		- [Nix Flake](#nix-flake)
```
I suspect that this is what's triggering the documentation generation error.

aidenfoxivey:
@baentsch I'll look into generating it by default

flake.nix
```diff
@@ -0,0 +1,88 @@
...
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+  outputs = {
+    self,
+    nixpkgs,
+    flake-utils,
+  }:
+    flake-utils.lib.eachDefaultSystem (system: let
+      name = "liboqs";
+      version = "0.0.1";
```
cothan, line 13: I think maintaining two versions of `liboqs` and `flake.nix` are confusing. If this version of `liboqs` introduce `flake.nix`, let's start at the same version with `liboqs`. Thought? @baentsch  @SWilson4 .

result
cothan, line 1: This one shouldn't be committed.

flake.nix
```diff
@@ -0,0 +1,88 @@
+{
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
```
cothan, line 3: Can you pin to the stable version? In the `flake.nix`, I see we use regular features, so it does not need to be unstable, right?

README.md
```diff
@@ -167,6 +168,44 @@ If you want to create Visual Studio build files, e.g., if not using `ninja`, be
 
 You can cross compile liboqs for various platforms. Detailed information is available [in the Wiki](https://github.com/open-quantum-safe/liboqs/wiki/Platform-specific-notes-for-building-liboqs#cross-compiling).
 
+### Nix Flake
+
+Nix is a tool that makes builds and developer environments deterministic. With just Nix installed, you can install `liboqs` using one command.
+
+Need to install Nix? Consider using the [Lix installer](https://git.lix.systems/lix-project/lix-installer), a modern, community-focused implementation of the Nix tool, available for Linux and MacOS.
```
cothan, line 175: Why don't we use the one from Determinate system? This one is not well known.

cothan (changes requested):

README.md
aidenfoxivey: I'm not overly opposed to using the Determinate Systems one. My logic is that lix is implemented in a memory safe language and has knowledgeable developers that I know.

I personally use the Lix implementation, and since I'm planning on maintaining the flake, it seemed reasonable to me to choose what I use myself. That said, it shouldn't really be any different from the Determinate Systems installer.

flake.nix
aidenfoxivey: Sounds good - I'll see if I can build it without having a version or have the version inherit from CMake.

flake.nix
aidenfoxivey: Sure, we can use nixpkgs 24.05. I usually opt to use unstable because the lockfile locks the packages anyway (if I'm not wrong), so it's easier to have one thing to update rather than two.

result
aidenfoxivey: Whoops! Good catch.

aidenfoxivey:
Turns out the version tag is completely unnecessary, since it can be referred to by its git hash anyways. I'll wait for additional comments and then clean up the git history once I've sorted out any requests.

flake.nix
aidenfoxivey: This seems like the right call actually, I'll change it.

aidenfoxivey: Done!

README.md
cothan: > I'm not overly opposed to using the Determinate Systems one. My logic is that lix is implemented in a memory safe language and has knowledgeable developers that I know.
Hmm, Determinate system also uses memory safe language. What make Lix different than Determinate? 

I still insist on using the Determinate System installer. I don't want to apply number of stars, forks metric here, I guess it will not be just me wonder why `liboqs` made this decision, other users/maintainers as well.

flake.nix
baentsch: "Resolved" as version tag is gone.

README.md
praveksharma: For what it's worth I use the Determinate system installer too although I don't know if it's the community standard or not. Both Lix and Determinate Systems come across as rather opinionated and I don't think we want liboqs taking a stance either way. Since the flake ought to with any nix installation that supports flakes, how about we leave nix installation instructions out of the README?

cothan (approved):

cothan (approved):
LTGM! Thanks for the contribution. Please make sure the all the tests are passed before you merge.

praveksharma (approved):
This looks good, thank for you working on this @aidenfoxivey!

baentsch:
> LTGM! Thanks for the contribution. Please make sure the all the tests are passed before you merge.

Hmm -- this statement made me curious: What tests are there (to be) passed? I didn't find any (just searching for changes to .github/workflows). Wouldn't that/having such be prudent, @cothan @praveksharma ?

cothan:
Hi @baentsch , I approve the PR while there are 6 tests (unrelated to the PR) are still running, since this PR doesn't add anything to Github flow, thus I expect them all to pass, and they did (I didn't aware at the time that each PR needs 2 approval). So I comment to make sure this PR will be merged in a green state.

praveksharma:
I agree that having testing for the flake would be desirable @baentsch but I'm also uncertain what is the best way to go about it. I believe the current build targets are limited in that they don't allow the user to pass CMake options via the nix CLI; is that correct @aidenfoxivey? 

Since the project doesn't plan on moving away from CMake, at this stage it might make more sense to focus on developer quality of life improvement via `nix develop`. This would be easy to test via a weekly CI job -- on a ubuntu:latest image with just nix that must pass some cmake build tests.

Alternatively, if there is some way of transparently exposing CMake options to nix that would also be suitable. Instead of a maintainer having to ensure compatibility with CONFIGURE.md that would be handled natively by nix.

baentsch (changes requested):
> uncertain what is the best way to go about it

OK, understood -- if we are in agreement, though, that tests should be added for a feature announced in the toplevel README (?) then there must be a way for it to happen -- and just doing it in a weekly test sounds wrong to me for a feature "prominently" announced/documented.

So I see two options:
1) Remove reference to Nix from the README (basically aligning its untested status with the contents of PLATFORMS.md, i.e., where it's not mentioned as a supported environment)
2) Add testing and a suitable (level) reference to PLATFORMS.md

aidenfoxivey:
Hmm. These are good points.

My feeling about `PLATFORMS.md` is that it's a tiny bit complex given that Nix works on 

- aarch64-darwin
- aarch64-linux
- i686-linux
- x86_64-darwin
- x86_64-linux

but we also don't want to require Nix to build anything. I view it less as a platform and more as a way to develop your own tooling AND build liboqs deterministically.

I'd prefer to mention it in the README just because it seems a shame to not mention it given that it can let you develop liboqs without having to worry about installing all the dependencies in your package manager.

baentsch:
> it seems a shame to not mention it given that it can let you develop liboqs without having to worry about installing all the dependencies in your package manager

Agreed. What about adding an entry to Nix to https://github.com/open-quantum-safe/liboqs/wiki/Platform-specific-notes-for-building-liboqs instead of the explicit README mention? But then again, I'm anyway not particularly happy about having this information in the Wiki and not the core repo (so that it makes it into releases). Upon re-reading the README I've got to say the "Quickstart" is not exactly "quick", surely not short, so this PR _is_ making life easier for folks using Nix (a link to Nix in the documentation --wherever that winds up-- would by good in any case).

What about this thought @SWilson4 : Make the Linux/Mac quickstart section as short as the Windows one (basically just list `apt get ... && cmake`), pointing to a more complete documentation (current detail level) elsewhere? Add a one-liner for Nix there and more complete Nix options (as in this PR) at where the complete text for Linux/Mac winds up? This way, the Quickstart gets truly quick again (for people that know the tooling -- and the ones that don't have to read below).

aidenfoxivey:
This sounds reasonable to me!

aidenfoxivey:
@praveksharma Just to be clear, you want changes on the README, right?

praveksharma:
> @praveksharma Just to be clear, you want changes on the README, right?

Yes, listing the relevant nix alternatives to the apt commands under Quick Start > Linux and Mac > 1. Install dependencies is what I had in mind.

baentsch:
As part of my year-end review of stalled things, can I ask what the idea is with this PR:
- Add CI testing as per @SWilson4 's suggestion?
- Add a small(er) README section as per @praveksharma 's suggestion (also addressing my concerns)
- Drop PR without merge

Which of these options would you like to pursue, @aidenfoxivey ? One, all, none, something else?

aidenfoxivey:
> As part of my year-end review of stalled things, can I ask what the idea is with this PR:
> 
> * Add CI testing as per @SWilson4 's suggestion?
> * Add a small(er) README section as per @praveksharma 's suggestion (also addressing my concerns)
> * Drop PR without merge
> 
> Which of these options would you like to pursue, @aidenfoxivey ? One, all, none, something else?

Ah whoops! I've got to admit liboqs slipped by mind a little bit during the latter half of the term. I'm going to sort out the README section tonight.

Thinking about testing, I think I wrote a solution but never pushed it.

I'll look to wrap this up in the next few days.

aidenfoxivey:
So I think there are a few things to think about:
- is the Nix flake formatted? (this doesn't seem *that* important to me)
- is it valid nix?
- does the build it produces work?

Are *all* of these in scope for what tests should do?

baentsch:
> > So I think there are a few things to think about:
> > ```
> > * is the Nix flake formatted? (this doesn't seem _that_ important to me)
> > 
> > * is it valid nix?
> > 
> > * does the build it produces work?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Are _all_ of these in scope for what tests should do?
> 
> @baentsch Is this what you had in mind when it came to testing?

It's a start - with item 3 being the most relevant imo.

.github/workflows/nix-check.yml
```diff
@@ -0,0 +1,14 @@
+name: "Nix Flake Check"
```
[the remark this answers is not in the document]
aidenfoxivey: Hmm, this is a fair point - let me address this.

dstebila:
Hi @aidenfoxivey, do you think you will be able to address the point above from security scanner? Is there anything else to be done before merging this?

aidenfoxivey:
> Hi @aidenfoxivey, do you think you will be able to address the point above from security scanner? Is there anything else to be done before merging this?

I've just pinned them to the latest version I could find. I'm running it locally on my Mac to give it a check.

CleanShot 2025-02-25 at 18 19 59@2x

aidenfoxivey:
Woohoo!

```
[Nix Flake Check/tests]   ✅  Success - Main ${GITHUB_ACTION_PATH}/install-nix.sh
[Nix Flake Check/tests]   ⚙  ::set-env:: TMPDIR=/tmp
[Nix Flake Check/tests]   ⚙  ::add-path:: /nix/var/nix/profiles/default/bin
[Nix Flake Check/tests]   ⚙  ::add-path:: /root/.nix-profile/bin
[Nix Flake Check/tests]   ✅  Success - Main cachix/install-nix-action@08dcb3a5e62fa31e2da3d490afc4176ef55ecd72
[Nix Flake Check/tests]   ⚙  ::set-env:: TMPDIR=/tmp
[Nix Flake Check/tests]   ⚙  ::add-path:: /nix/var/nix/profiles/default/bin
[Nix Flake Check/tests]   ⚙  ::add-path:: /root/.nix-profile/bin
[Nix Flake Check/tests] ⭐ Run Main nix build
[Nix Flake Check/tests]   🐳  docker exec cmd=[bash -e /var/run/act/workflow/2] user= workdir=
warning: Git tree '/Users/aidenfoxivey/src/liboqs' is dirty
[Nix Flake Check/tests]   ✅  Success - Main nix build
[Nix Flake Check/tests] ⭐ Run Main nix flake check
[Nix Flake Check/tests]   🐳  docker exec cmd=[bash -e /var/run/act/workflow/3] user= workdir=
warning: Git tree '/Users/aidenfoxivey/src/liboqs' is dirty
warning: The check omitted these incompatible systems: aarch64-darwin, x86_64-darwin, x86_64-linux
| Use '--all-systems' to check all.
[Nix Flake Check/tests]   ✅  Success - Main nix flake check
[Nix Flake Check/tests] ⭐ Run Post cachix/install-nix-action@08dcb3a5e62fa31e2da3d490afc4176ef55ecd72
[Nix Flake Check/tests]   🐳  docker cp src=/Users/aidenfoxivey/.cache/act/cachix-install-nix-action@08dcb3a5e62fa31e2da3d490afc4176ef55ecd72/ dst=/var/run/act/actions/cachix-install-nix-action@08dcb3a5e62fa31e2da3d490afc4176ef55ecd72/
[Nix Flake Check/tests]   ✅  Success - Post cachix/install-nix-action@08dcb3a5e62fa31e2da3d490afc4176ef55ecd72
[Nix Flake Check/tests] ⭐ Run Complete job
[Nix Flake Check/tests] Cleaning up container for job tests
[Nix Flake Check/tests]   ✅  Success - Complete job
[Nix Flake Check/tests] 🏁  Job succeeded
```

.github/workflows/nix-check.yml
SWilson4, line 1: Could this be moved into the basic-checks workflow?

aidenfoxivey:
Should be moved into `basic.yml` now. That's the correct location right?

aidenfoxivey:
CleanShot 2025-02-26 at 13 21 57@2x
I suspect that basic.yml isn't running for me due to not having committer access to open-quantum-safe. This is from my local fork running it. I'm thinking that perhaps the container I've chosen doesn't have `sudo`. This could prove an issue.

@baentsch I'm happy to make changes in this PR or a follow up to make this more robust. I apologize for the delays in the middle of this.

(I'll be back at this after 5pm EST today to review comments and/or fix up the CI issues.)

SWilson4:
> I suspect that basic.yml isn't running for me due to not having committer access to open-quantum-safe. This is from my local fork running it. I'm thinking that perhaps the container I've chosen doesn't have `sudo`. This could prove an issue. I just used the default Ubuntu image beforehand - would that be fine to use instead?

I don't believe we install `sudo` in our containers—everything in CI runs as root anyhow. If the ubuntu-latest runner has the necessary dependencies to build the flake, that's fine (even preferable—it saves the time of downloading our image).

aidenfoxivey:
Seems the Nix tests are working - happy to squash and force push when it's all passed.

PR #1974, opened 2024-11-03
Make random/functions deterministic during fuzzing

nathaniel-brough:

### DIFF
--- a/tests/fuzz_test_sig.c
+++ b/tests/fuzz_test_sig.c
@@ -11,12 +11,51 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <stdlib.h>
 
 #include <oqs/oqs.h>
 
-void cleanup_heap(uint8_t *public_key, uint8_t *secret_key,
-                  uint8_t *signature,
+
+typedef struct {
+	uint32_t random_seed;
+	uint32_t algorithm_index;
+} fuzz_init_ctx_t;
+
+typedef struct {
+	fuzz_init_ctx_t init;
+	const uint8_t *data;
+	size_t data_len;
+} fuzz_ctx_t;
+
+fuzz_ctx_t init_fuzz_context(const uint8_t *data, size_t data_len);
+void fuzz_rand(uint8_t *random_array, size_t bytes_to_read);
+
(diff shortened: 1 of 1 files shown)


tests/fuzz_test_sig.c
```diff
@@ -11,12 +11,51 @@
...
+	const uint8_t *data;
+	size_t data_len;
+} fuzz_ctx_t;
+
+fuzz_ctx_t init_fuzz_context(const uint8_t *data, size_t data_len);
+void fuzz_rand(uint8_t *random_array, size_t bytes_to_read);
+
+void fuzz_rand(uint8_t *random_array, size_t bytes_to_read) {
+	for (size_t i = 0; i < bytes_to_read; i++) {
+		random_array[i] = (uint8_t)rand();
```
baentsch, line 34: Wouldn't it be even better to make the contents of random_array be dependent on the fuzz_data input? OK, it's both random, but that way one could reproduce failures triggered by specific input...

baentsch (commented):
Basically LGTM with a single comment/question.

tests/fuzz_test_sig.c
nathaniel-brough: I have set the seed for rand [here](https://github.com/open-quantum-safe/liboqs/pull/1974/files#diff-5256ad59f90fd8a1fdf808aecf506ae6c9349ec5e67d87f6259bc4610b1f11e4R52) using fuzzed data. So as far as I understand, it should be deterministic for a given fuzz input. That being said it is an interesting idea to drive the random number generator directly using fuzzed data. I did briefly experiment with this, however I didn't observe much improvement in fuzzing performance. It also get's a little messy as I have to copy the fuzzed data into a global variable and then copy it back out again in the "random function" keeping track of how much fuzzed data was consumed. It works, but it was not particularly easy code to read, even when I was the one writing it haha.

I may revisit this idea at some point, now that we are fuzzing more that just dilithium it might improve performance a bit for other algorithms. But for now I don't want to over-complicate things.

baentsch: ACK. I overlooked the `srand` call. Thanks for the feedback @nathaniel-brough

baentsch (approved):

dstebila (approved):

PR #1982, opened 2024-11-07
Avoid OpenSSL functions are unconditionally called at OQS_destroy

ueno:
When OQS_DLOPEN_OPENSSL is designated and low-level primitives are overridden with OQS_*_set_callbacks, OQS_destroy still indirectly calls EVP_*_free from OpenSSL. This adds a extra NULL check to avoid that.

### DIFF
--- a/src/common/ossl_helpers.c
+++ b/src/common/ossl_helpers.c
@@ -45,31 +45,41 @@ static void fetch_ossl_objects(void) {
 	}
 }
 
+static inline void cleanup_evp_md(EVP_MD **mdp) {
+	/* Always check argument is non-NULL before calling EVP_MD_free
+	 * to avoid OpenSSL functions being used when they are
+	 * overridden with OQS_*_set_callbacks.
+	 */
+	if (*mdp) {
+		OSSL_FUNC(EVP_MD_free)(*mdp);
+		*mdp = NULL;
+	}
+}
+
+static inline void cleanup_evp_cipher(EVP_CIPHER **cipherp) {
+	/* Always check argument is non-NULL before calling EVP_CIPHER_free
+	 * to avoid OpenSSL functions being used when they are
+	 * overridden with OQS_*_set_callbacks.
+	 */
+	if (*cipherp) {
+		OSSL_FUNC(EVP_CIPHER_free)(*cipherp);
+		*cipherp = NULL;
+	}
+}
+
 static void free_ossl_objects(void) {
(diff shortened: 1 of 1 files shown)


baentsch (commented):
This PR begs 1-2 questions in my mind: If the purpose is

> to avoid OpenSSL functions being used when they are overridden with OQS_*_set_callbacks

then a) why did `free_ossl_objects` get called in the first place and/or b) why is EVP_MD_free not also replaced with an "alternative TLS lib" variant?

ueno:
> then a) why did `free_ossl_objects` get called in the first place and/or b) why is EVP_MD_free not also replaced with an "alternative TLS lib" variant?

Let me answer (b) first: the objects being freed here (i.e., fetched algorithm implementations) are exclusively used by the OpenSSL implementation, not used nor needed for alternative TLS lib variants.

For (a), to avoid calling `free_ossl_objects` entirely, liboqs needs to know which objects are used and which are not, as the current mechanism allows partial override (OpenSSL for SHA3 but GnuTLS for SHA2, ...).

baentsch:
> the current mechanism allows partial override

That's the explanation I've been looking for, Thanks @ueno .

src/common/ossl_helpers.c
```diff
@@ -46,30 +46,58 @@ static void fetch_ossl_objects(void) {
...
-	aes128_ctr_ptr = NULL;
-	OSSL_FUNC(EVP_CIPHER_free)(aes256_ecb_ptr);
-	aes256_ecb_ptr = NULL;
-	OSSL_FUNC(EVP_CIPHER_free)(aes256_ctr_ptr);
-	aes256_ctr_ptr = NULL;
+	/* Always check argument is non-NULL before calling EVP_*_free
+	 * to avoid OpenSSL functions being used when they are
+	 * overridden with OQS_*_set_callbacks.
+	 */
+	if (sha256_ptr) {
```
baentsch, line 53: Wouldn't it be better to transform this into a macro (or add an IF_NONNULL variant of "OSSL_FUNC") to make this more easily readable?

ueno: Instead of a macro, I added inline functions that work in a similar fashion as GCC's cleanup functions.

baentsch: Looks cleaner and easier to maintain this way. Thanks!

SWilson4 (approved):
LGTM! Thanks @ueno.

PR #1983, opened 2024-11-08
imported fix from CROSS upstream: endianness-aware csprng

### LINKED ISSUE #1961: CROSS KAT failures / potential endianness issue
**Describe the bug**
KATs for CROSS fail on s390x after reenabling the Travis build for this platform in #1960.

S390x is the only big-endian platform in the CI pipeline, so I suspect an endianness issue in CROSS.

Tagging @rtjk 

**To Reproduce**
See https://app.travis-ci.com/github/open-quantum-safe/liboqs/jobs/627436707

rtjk:
Fixes #1961.

The rejection sampling functions in CROSS rely on pointer casting, which caused the CSPRNG to generate different numbers on big-endian versus little-endian architectures, resulting in different KATs. This fix uses the compiler flag `__BYTE_ORDER__` to detect endianness and swap the bytes when necessary. Thank you @bhess for pointing us in the right direction!


SWilson4:
Thanks for making the fix, @rtjk! I suggest that we hold merge for https://github.com/open-quantum-safe/liboqs/pull/1960 first so that the s390x tests run on this PR. Perhaps you could rebase on @bhess's branch for that PR for now in case there are CI errors that need to be resolved (e.g., due to a gcc or clang version).

rtjk:
@SWilson4 I rebased on branch `bhe-reenable-travis` and turned s390x back on for CROSS. It looks like Travis is running.

SWilson4:
> Ouch! It still failed ... time to debug some more.

FYI, looks like the s390x build uses gcc 9—does this older version support all the features you're using?

rtjk:
It turns out there were a few more instances where type punning was used, causing an endianness issue. I've fixed these, so Travis CI should now pass. I also added two lines to `copy_from_upstream.py` to create template files if they aren’t already in place. Is this okay?

baentsch (dismissed):
> I also added two lines to copy_from_upstream.py to create template files if they aren’t already in place. Is this okay?

Thanks for highlighting this @rtjk ! I don't think this is a good approach: Missing files in code generation imo are indicative of an incorrect code base and should lead to an error condition triggering a proper fix (manually adding the file(s) in question): Reason for my caution is that the code generation is so central to the proper operation of `liboqs` that I'm very wary of too much automation in this area.

baentsch (approved):
Thanks for the fix @rtjk. I can't claim to have read all changes but defer to the tests passing as this code being OK: As with all core crypto code in `liboqs` the ultimate responsibility for correctness is with the upstream.  Allow me to ask the question when/whether this code is going to flow upstream again (such as to get rid of the various "fix" statements in the version)?

bhess (approved):
Thanks for the update @rtjk!

rtjk:
> Allow me to ask the question when/whether this code is going to flow upstream again (such as to get rid of the various "fix" statements in the version)?

The plan is to release a new version of CROSS before moving on to NIST's second round (deadline January 17, 2025). This will consolidate all the changes we made since CROSS 1.2.

SWilson4 (approved):
Thanks for the fix @rtjk!

Issue #1760, opened 2024-04-22
Document DCO utility and HOWTO


baentsch:
As per https://github.com/open-quantum-safe/liboqs/pull/1708#issuecomment-2069210942

SWilson4:
Seconding this as a priority; I've recently had to ask outside contributors to fix DCO in [two](https://github.com/open-quantum-safe/liboqs/pull/1747) [different](https://github.com/open-quantum-safe/liboqs/pull/1758) PRs that were otherwise ready to merge.

dstebila:
Perhaps this would be a useful document to link to or build on to explain how to work with DCO: https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md

SWilson4:
DCO is pretty well-documented with specific instructions for fixing errors via GitHub CI, and I haven't seen a contributor lately who had difficulty resolving a DCO error. I think we can close this.

PR #1994, opened 2024-11-13
Use seed as private key format for ML-KEM

### LINKED ISSUE #1985: Support 64-byte seed form of ML-KEM private keys
**Describe the bug**
There is increasing community consensus that ML-KEM private keys should be stored in the form of 64-byte seeds (i.e., the inputs to `ML-KEM.KeyGen_internal()`), not in the expanded form described in FIPS 203.  See, for example:

https://words.filippo.io/dispatches/ml-kem-seeds/
https://datatracker.ietf.org/meeting/121/materials/slides-121-pquip-fips-issues-with-deploying-ml-kem-and-ml-dsa-04

It is impossible to implement protocols that are defined in terms of seeds.  The `OQS_KEM_ml_kem_XXX_keypair()` method produces an expanded private key, and `OQS_KEM_ml_kem_XXX_decaps()` consumes an expanded private key.

The minimal possible fix here would be to split out a method that creates an expanded private key from a seed.  That way an app could mostly just use the current API, except that when generating a completely fresh key, they would have to fill the seed with random data.  To even that out, you could make a parallel seed-only API, either by refactoring the existing interface or creating a parallel one.  Anything short of removing the current API of course, will leave the risk that a caller will use expanded keys, with the associated risk of corruption.

If we can agree on an approach here, I would be happy to send a PR.

**To Reproduce**
N/A

**Expected behavior**
N/A

**Screenshots**
N/A

 - liboqs version: main branch

**Additional context**
N/A

bifurcation:
This PR is a WIP demonstrating an approach to storing ML-KEM decapsulation keys in the form of a 64-byte seed, as opposed to the expanded form currently.  Currently, the proposed scheme is implemented (a) for ML-KEM-512 only, and (b) in-place, as opposed to defining a new scheme.  I am posting it in this state to get feedback on whether the structure of the change looks good, because the remainder of the PR will basically just be copy/pasting to cover the other parameter sets and maybe make a new KEM scheme.

If maintainers could answer the following questions, I can implement the rest of the PR:

* Does the approach shown here look correct from an OQS structure / coding standards point of view?
* Should this be implemented in the current ML-KEM provider, or as a new provider?
    * If the latter, it would be nice to key them both off of the same compile-time switches
    * For example, this could avoid having to build two copies of the PQCrystals library, which could cause symbol conflicts
    * ... since you could have one provider own the back-end library, and the other just link against it.

Fixes #1985 

Tests are currently failing because they are providing / looking for the expanded private key.  The ACVP test vectors do provide the seed values (`d` and `z`), so it should be possible to fix the tests simply by using those values instead of the expanded values (`dk`).

    * This PR changes the format of the decapsulation keys  used with ML-KEM
    * If this is undesirable, we could add a parallel KEM implementation, as discussed in #1985
* [X] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged.)
    * Not currently, since the changes are made in-place.
    * If we go with the backward-compatible approach, it will define a new algorithm.

### DIFF
--- a/src/kem/ml_kem/kem_ml_kem.h
+++ b/src/kem/ml_kem/kem_ml_kem.h
@@ -7,7 +7,7 @@
 
 #if defined(OQS_ENABLE_KEM_ml_kem_512)
 #define OQS_KEM_ml_kem_512_length_public_key 800
-#define OQS_KEM_ml_kem_512_length_secret_key 1632
+#define OQS_KEM_ml_kem_512_length_secret_key 64
 #define OQS_KEM_ml_kem_512_length_ciphertext 768
 #define OQS_KEM_ml_kem_512_length_shared_secret 32
 OQS_KEM *OQS_KEM_ml_kem_512_new(void);
@@ -18,7 +18,7 @@ OQS_API OQS_STATUS OQS_KEM_ml_kem_512_decaps(uint8_t *shared_secret, const uint8
 
 #if defined(OQS_ENABLE_KEM_ml_kem_768)
 #define OQS_KEM_ml_kem_768_length_public_key 1184
-#define OQS_KEM_ml_kem_768_length_secret_key 2400
+#define OQS_KEM_ml_kem_768_length_secret_key 64
 #define OQS_KEM_ml_kem_768_length_ciphertext 1088
 #define OQS_KEM_ml_kem_768_length_shared_secret 32
 OQS_KEM *OQS_KEM_ml_kem_768_new(void);
@@ -29,7 +29,7 @@ OQS_API OQS_STATUS OQS_KEM_ml_kem_768_decaps(uint8_t *shared_secret, const uint8
 
 #if defined(OQS_ENABLE_KEM_ml_kem_1024)
 #define OQS_KEM_ml_kem_1024_length_public_key 1568
-#define OQS_KEM_ml_kem_1024_length_secret_key 3168
+#define OQS_KEM_ml_kem_1024_length_secret_key 64
 #define OQS_KEM_ml_kem_1024_length_ciphertext 1568
 #define OQS_KEM_ml_kem_1024_length_shared_secret 32
 OQS_KEM *OQS_KEM_ml_kem_1024_new(void);
--- a/src/kem/ml_kem/kem_ml_kem_512.c
+++ b/src/kem/ml_kem/kem_ml_kem_512.c
@@ -3,6 +3,7 @@
 #include <stdlib.h>
 
 #include <oqs/kem_ml_kem.h>
+#include <randombytes.h>
 
 #if defined(OQS_ENABLE_KEM_ml_kem_512)
 
@@ -30,29 +31,36 @@ OQS_KEM *OQS_KEM_ml_kem_512_new(void) {
 	return kem;
 }
 
-extern int pqcrystals_ml_kem_512_ref_keypair(uint8_t *pk, uint8_t *sk);
+#define KYBER512_KEYPAIRCOINBYTES 64
+#define KYBER512_SECRETKEYBYTES 1632
+#define KYBER512_PUBLICKEYBYTES 800
+
+extern int pqcrystals_ml_kem_512_ref_keypair_derand(uint8_t *pk, uint8_t *sk, const uint8_t* coins);
 extern int pqcrystals_ml_kem_512_ref_enc(uint8_t *ct, uint8_t *ss, const uint8_t *pk);
 extern int pqcrystals_ml_kem_512_ref_dec(uint8_t *ss, const uint8_t *ct, const uint8_t *sk);
 
 #if defined(OQS_ENABLE_KEM_ml_kem_512_avx2)
-extern int pqcrystals_ml_kem_512_avx2_keypair(uint8_t *pk, uint8_t *sk);
+extern int pqcrystals_ml_kem_512_avx2_keypair_derand(uint8_t *pk, uint8_t *sk, const uint8_t* coins);
 extern int pqcrystals_ml_kem_512_avx2_enc(uint8_t *ct, uint8_t *ss, const uint8_t *pk);
 extern int pqcrystals_ml_kem_512_avx2_dec(uint8_t *ss, const uint8_t *ct, const uint8_t *sk);
 #endif
 
 OQS_API OQS_STATUS OQS_KEM_ml_kem_512_keypair(uint8_t *public_key, uint8_t *secret_key) {
+  uint8_t expanded_secret_key[KYBER512_SECRETKEYBYTES];
+  randombytes(secret_key, KYBER512_KEYPAIRCOINBYTES);
+
 #if defined(OQS_ENABLE_KEM_ml_kem_512_avx2)
 #if defined(OQS_DIST_BUILD)
 	if (OQS_CPU_has_extension(OQS_CPU_EXT_AVX2) && OQS_CPU_has_extension(OQS_CPU_EXT_BMI2) && OQS_CPU_has_extension(OQS_CPU_EXT_POPCNT)) {
 #endif /* OQS_DIST_BUILD */
-		return (OQS_STATUS) pqcrystals_ml_kem_512_avx2_keypair(public_key, secret_key);
+		return (OQS_STATUS) pqcrystals_ml_kem_512_avx2_keypair_derand(public_key, expanded_secret_key, secret_key);
 #if defined(OQS_DIST_BUILD)
 	} else {
-		return (OQS_STATUS) pqcrystals_ml_kem_512_ref_keypair(public_key, secret_key);
+		return (OQS_STATUS) pqcrystals_ml_kem_512_ref_keypair_derand(public_key, expanded_secret_key, secret_key);
 	}
 #endif /* OQS_DIST_BUILD */
 #else
-	return (OQS_STATUS) pqcrystals_ml_kem_512_ref_keypair(public_key, secret_key);
+	return (OQS_STATUS) pqcrystals_ml_kem_512_ref_keypair_derand(public_key, expanded_secret_key, secret_key);
 #endif
 }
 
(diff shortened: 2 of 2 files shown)


kvanals:
I'm tracking this closely, as we are heavily invested in ML-KEM at SoftIron.  I'm not really sure what we could do here that would be backwards compatible with existing expanded keys.

As a side note, I'm suspecting changes will be needed in oqs-provider as well?

`# openssl genpkey -algorithm mlkem1024 -outform pem -out mlkem.key
malloc(): unaligned tcache chunk detected
Aborted (core dumped)`

bifurcation:
@kvanals - The idea of doing it with backward compatibility would be to define a separate scheme using the same upstream library.  And yes, there would probably be changes needed in oqs-provider as well, to enable callers to reference the additional scheme.  You would have something like `openssl genpkey -algorithm mlkem1024seed ...`.

baentsch:
> Does the approach shown here look correct from an OQS structure / coding standards point of view?

Yes. A full implementation of course would have to use the code generation concepts of "copy_from_upstream" and pass all NIST tests.

> Should this be implemented in the current ML-KEM provider, or as a new provider? 

The overarching goal should be to create as little new original code (as opposed to generated code) as possible / be orthogonal to the current logic.

> I'm tracking this closely, as we are heavily invested in ML-KEM at SoftIron.

In that case, allow me to invite "SoftIron" to contribute to the work in this project. Right now this project's mission is to support research and prototyping (only) as we don't have the resources to be "everything for everyone", e.g., maintain backwards compatibility (or [other "product level" qualities](https://github.com/open-quantum-safe/liboqs?tab=readme-ov-file#limitations-and-security)). The proposal @bifurcation kindly works on here is the result of the discussion at #1985 with the goal to work with these limitations. Either way, currently, there is absolutely no guarantee that OQS will keep supporting both types of keys: For example, if the upstream ceases to support both modes, so will OQS. If the general consensus is to only use the seed-based approach, support for expanded keys most likely will vanish.

> As a side note, I'm suspecting changes will be needed in oqs-provider as well?

This seems to indicate you want to use this code via OpenSSL. If so, you may want to follow the discussion (or --again-- even better, contribute to the project) at https://github.com/openssl/openssl/issues/25885 where we're using an entirely different code base and aim to _only_ implement the relevant standards for exactly the same reason: Provide the best possible code with the smallest amount of development and maintenance resources possible.

kvanals:
> > I'm tracking this closely, as we are heavily invested in ML-KEM at SoftIron.
> 
> In that case, allow me to invite "SoftIron" to contribute to the work in this project. Right now this project's mission is to support research and prototyping (only) as we don't have the resources to be "everything for everyone", e.g., maintain backwards compatibility (or [other "product level" qualities](https://github.com/open-quantum-safe/liboqs?tab=readme-ov-file#limitations-and-security)). The proposal @bifurcation kindly works on here is the result of the discussion at #1985 with the goal to work with these limitations. Either way, currently, there is absolutely no guarantee that OQS will keep supporting both types of keys: For example, if the upstream ceases to support both modes, so will OQS. If the general consensus is to only use the seed-based approach, support for expanded keys most likely will vanish.

Cheers @baentsch, happy to upstream anything where applicable.  Presently we're using oqs-provider as-is and only maintain one out-of-tree patch to liboqs which I don't believe would meet your copy_from_upstream requirements to build for some of our ARM build targets that are lacking some of the requisite ARM NEON intrinsics.  Let me know if there is anything you'd like us to sign or review in this regard.

> 
> > As a side note, I'm suspecting changes will be needed in oqs-provider as well?
> 
> This seems to indicate you want to use this code via OpenSSL. If so, you may want to follow the discussion (or --again-- even better, contribute to the project) at [openssl/openssl#25885](https://github.com/openssl/openssl/issues/25885) where we're using an entirely different code base and aim to _only_ implement the relevant standards for exactly the same reason: Provide the best possible code with the smallest amount of development and maintenance resources possible.

I absolutely am tracking that.  And correct, we do very little work today with liboqs directly, but rather abstract its use behind high level OpenSSL EVP functions.

baentsch:
> Let me know if there is anything you'd like us to sign or review in this regard.

Regarding OQS contributions, all there is is (acceptance of) the [license](https://github.com/open-quantum-safe/liboqs/blob/main/LICENSE.txt) (by way of the DCO "signoff"; #1760). If you'd like to see your code/contributions also become used in OpenSSL, see [here](https://github.com/openssl/project/issues/431#issuecomment-2324554154).

> rather abstract its use behind high level OpenSSL EVP functions.

Good approach. Will allow use of the most well-maintained PQC code base -- whichever that is. FWIW, I like the approach of supporting both key formats [via OSSL_PARAM](https://github.com/openssl/openssl/pull/25938). Which one(s) gets an encoder/decoder will probably be a matter of which format ultimately gets an official/standardized OID assigned. If both do, the approach taken in this PR (creating a "twin" alg set that each could get its own OIDs) is probably the most simple to integrate&use.

bifurcation:
> A full implementation of course would have to use the code generation concepts of "copy_from_upstream" and pass all NIST tests.
> The overarching goal should be to create as little new original code (as opposed to generated code) as possible / be orthogonal to the current logic.

@baentsch - Happy to work along these lines, but I'm not seeing documentation of how to do this.  I see the `copy_from_upstream` script, but not what the concepts are or how to use it.  Likewise, unclear to me what in this code base is generated vs. original code.  Any pointers you could provide would be helpful.  I'm also in the libOQS Discord if that would be a more convenient place to chat.

kvanals:
It looks like @baentsch has been working on getting ML-KEM into OpenSSL from BoringSSL's implementation in https://github.com/openssl/openssl/tree/feature/ml-kem

If I'm reading this right, this supports using "OSSL_PKEY_PARAM_MLKEM_SEED" in the manner described above via an OSSL_PARAM.

baentsch:
> It looks like @baentsch has been working on getting ML-KEM into OpenSSL from BoringSSL's implementation in https://github.com/openssl/openssl/tree/feature/ml-kem
> 
> If I'm reading this right, this supports using "OSSL_PKEY_PARAM_MLKEM_SEED" in the manner described above via an OSSL_PARAM.

This is a community effort, but yes, that's what we're doing there. As always, feedback welcome to the approach taken (documentation proposal up at https://github.com/openssl/openssl/pull/26037).

kvanals:
@bifurcation FWIW, I ended up basing at least one implementation on the work in this PR as a bit of a one-off.  If useful for anyone else, I fixed a few issues with the work done in this PR and applied the logic to all three ML-KEM variants and attached an updated patch to this comment.

For my use this works, since the ASN.1 data will have the ML-KEM seed (as private key) and the ML-KEM public key as well, albeit concatenated in the private key since liboqs/oqs-provider do not yet support key derivation from the private key.  If/when this becomes more standardized, I can always rewrite the ASN.1 structure to match with the raw data needed and I avoid the issue of losing the seeds when persistent keys needs to exist.

I appreciate the initial work on this, but I also realize this isn't anything that can actually be adopted upstream.
[liboqs-mlkem-coins-as-secret.patch](https://github.com/user-attachments/files/17987143/liboqs-mlkem-coins-as-secret.patch)

baentsch:
In a year-end "clean up" effort, is this PR still active/intended to be completed @bifurcation ?

dstebila:
We discussed this issue in today's OQS status meeting. @SWilson4 noticed that the [HPKE spec for ML-KEM](https://www.ietf.org/archive/id/draft-connolly-cfrg-hpke-mlkem-04.html#name-decapsulation) says that the way to achieve decapsulation using the seed as input is just to call the derive keypair function and then decapsulate with the expanded key. So we are thinking that that would be the way to go -- keeps the API simpler. We expect to have the derive keypair functionality available once #2070 lands.

bifurcation:
Agreed, this PR is moot given that #2070 now exists.  We should get that landed instead.

baentsch:
> @SWilson4 noticed that the [HPKE spec for ML-KEM](https://www.ietf.org/archive/id/draft-connolly-cfrg-hpke-mlkem-04.html#name-decapsulation) says that the way to achieve decapsulation using the seed as input is just to call the derive keypair function and then decapsulate with the expanded key. So we are thinking that that would be the way to go

FWIW, this is the same thought as discussed in https://github.com/open-quantum-safe/liboqs/issues/2032#issuecomment-2682611883. The implication of not having a "seed" private key based algorithm representation in `liboqs` is that that format cannot be made available via the `oqs-provider` (without complete departure from its design concept of just wrapping `liboqs` key formats). Closing that issue, too then.

dstebila:
> FWIW, this is the same thought as discussed in [#2032 (comment)](https://github.com/open-quantum-safe/liboqs/issues/2032#issuecomment-2682611883). The implication of not having a "seed" private key based algorithm representation in `liboqs` is that that format cannot be made available via the `oqs-provider` (without complete departure from its design concept of just wrapping `liboqs` key formats). Closing that issue, too then.

Well, these seem separable. If the current private key format being considered in LAMPS goes through, then there will be users that want to encode their private keys in that format (which might be just a seed) and expect that an application be able to wrap/unwrap keys into that format.  Wouldn't that fall to OQS Provider?

baentsch:
> Wouldn't that fall to OQS Provider?

As usual, anything can be implemented in software; `oqsprovider` could use other crypto libs than `liboqs`, too :-)

More seriously: `oqsprovider` has been designed to be exactly that, a generic provider for all algorithms supported by the equally generic, algorithm-agnostic (lib)OQS (APIs), not a haphazard collection of algorithm-specific adaptations. Due to things like different hybrid and composite schemes `oqsprovider` already has turned to be pretty much a nightmare to maintain and IMO for that reason alone, not exactly secure. Adding more "special" cases to the code would make things worse: I don't want that to happen:

My priority goal for `oqsprovider` is first and foremost for it to remain an available and useful tool for the researchers and users that want to check out operation of new PQC algorithms in the real world.

I understand PQCA wants something different -- and everyone (there and beyond) is cordially invited to contribute development time and resources to create these capabilities: All issues are openly documented in the `oqsprovider` GH repo and I remain available to review PRs. But caveat emptor, the way things are done in `openssl` are way more generic than anything we've been discussing so far: That provider is diligently following Postel's law which for `oqsprovider` adaptations would demand even more "special purpose"/algorithm-specific code to bring the two in line for the FIPS 20x algs -- which also, users would rightfully demand and PQCA marketing will surely promise.

Again, all doable -- given an increased investment in development (time and effort). This (renewed) discussion to me feels like the opposite, though: Are we trying to minimize development efforts in `liboqs` at the detriment of its consumers (`oqsprovider` and any similar software relying on clean and generic, algorithm-independent APIs)?

> If the current private key format being considered in LAMPS goes through, then there will be users that want to encode their private keys in that format (which might be just a seed) and expect that an application be able to wrap/unwrap keys into that format.

This decision --as and when taken-- may be the opportunity for `liboqs` to create APIs suitable for that (call it the "LAMPS" API to differentiate it from the "NIST" API). It may also be the perfect time to designate those components of `liboqs` that are considered "production ready" (LAMPS?) and those that are not (NIST?) -- [a decision delayed now for more than a year](https://github.com/open-quantum-safe/tsc/issues/1), and again, arguably also due to a lack of actual development contributions.

PR #1997, opened 2024-11-19
Set ML-KEM alg_version to "FIPS203"

### LINKED ISSUE #1996: Update ML-KEM alg_version to "FIPS203"
I assume we will want to do this for consistency with ML-DSA as per discussion here: https://github.com/open-quantum-safe/liboqs/pull/1919#discussion_r1844299975

SWilson4:
To bring ML-KEM in line with the upcoming ML-DSA update.

Fixes #1996 

### DIFF
--- a/scripts/copy_from_upstream/patches/pqcrystals-ml_kem.patch
+++ b/scripts/copy_from_upstream/patches/pqcrystals-ml_kem.patch
@@ -1,174 +1,328 @@
-diff --git a/Kyber1024_META.yml b/ML-KEM-1024_META.yml
-similarity index 55%
-rename from Kyber1024_META.yml
-rename to ML-KEM-1024_META.yml
-index baa5ca3..fdfc298 100644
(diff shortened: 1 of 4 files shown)


baentsch (commented):
Pretty massive changes for 3 lines of code... Ideas why this is so?

SWilson4:
> Pretty massive changes for 3 lines of code... Ideas why this is so?

I did wonder about this also. I generated the patch using `git diff-tree` [as documented here](https://github.com/open-quantum-safe/liboqs/wiki/Patching-upstream-sources). It looks to me as if the original patch was generated with `git diff`—when I update the patch using `git diff` the diff (of the diff) is much smaller, as I would have expected.

I'm fine to update the patch with `git diff` instead to make the changes easier to view.

dstebila (approved):

dstebila:
Looking for one more review @open-quantum-safe/liboqs-committers

Martyrshot (approved):
LGTM!

PR #2000, opened 2024-11-19
Add defines for OQS version components

### LINKED ISSUE #1993: Create version #defines
Currently, AFAIK there is no simple library version define that a downstream could utilize to adapt to API changes in `liboqs`: The only define there is is "OQS_VERSION_TEXT" and I'm not sure that's making it sufficiently easy to adapt a downstream's code to `liboqs` capabilities: The goal of (and one purpose for) `oqsprovider` is to build against all versions of `liboqs`, so a simple-to-use code-level library version test facility should be there, e.g. similar to what we have in [openssl](https://github.com/openssl/openssl/blob/012353bdf21b98def920ac317b94c4a9ed501b79/include/openssl/opensslv.h.in#L29-L31). Worthwhile a separate issue/PR or simple enough to do as part of this PR? If the latter, the corresponding `oqsprovider` update could be done as a "-tracker" update as this _is_ an API-changing PR (either way).

_Originally posted by @baentsch in https://github.com/open-quantum-safe/liboqs/issues/1919#issuecomment-2469779175_

dstebila:
Fixes #1993.

Note that I had to use `#define` rather than `#cmakedefine` in oqsconfig.h.cmake to avoid it not setting the macros when the values are equal to a false-y value like `0` or `""`.


CMakeLists.txt
```diff
@@ -41,6 +41,10 @@ set(CMAKE_C_STANDARD_REQUIRED ON)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_C_VISIBILITY_PRESET hidden)
 set(OQS_VERSION_TEXT "0.11.1-dev")
+set(OQS_VERSION_MAJOR 0)
+set(OQS_VERSION_MINOR 11)
+set(OQS_VERSION_PATCH 1)
+set(OQS_VERSION_PRE_RELEASE "-dev")
```
SWilson4, line 47: What about generating `OQS_VERSION_TEXT` automatically based on the other variables? This reduces the risk of a manual error that leads to the version variables not matching.
```suggestion
set(OQS_VERSION_MAJOR 0)
set(OQS_VERSION_MINOR 11)
set(OQS_VERSION_PATCH 1)
set(OQS_VERSION_PRE_RELEASE "-dev")
set(OQS_VERSION_TEXT "${OQS_VERSION_MAJOR}.${OQS_VERSION_MINOR}.${OQS_VERSION_PATCH}${OQS_VERSION_PRE_RELEASE}")
```

src/oqsconfig.h.cmake
```diff
@@ -1,6 +1,10 @@
 // SPDX-License-Identifier: MIT
 
 #cmakedefine OQS_VERSION_TEXT "@OQS_VERSION_TEXT@"
+#define OQS_VERSION_MAJOR @OQS_VERSION_MAJOR@
+#define OQS_VERSION_MINOR @OQS_VERSION_MINOR@
+#define OQS_VERSION_PATCH @OQS_VERSION_PATCH@
+#define OQS_VERSION_PRE_RELEASE "@OQS_VERSION_PRE_RELEASE@"
```
SWilson4, line 7:
```suggestion
#cmakedefine OQS_VERSION_PRE_RELEASE "@OQS_VERSION_PRE_RELEASE@"
```
If we use cmakedefine for this variable, we can do an ifdef check when printing the version string: see suggestion on `tests/system_info.c`

SWilson4 (approved):

tests/system_info.c
```diff
@@ -219,7 +219,11 @@ static void print_oqs_configuration(void) {
 	 * USE_SANITIZER: -fsanitize= option present in compile options
 	 * OQS_ENABLE_TEST_CONSTANT_TIME: only shown below
 	 */
-	printf("OQS version:      %s\n", OQS_VERSION_TEXT);
+#if defined(OQS_VERSION_PRE_RELEASE)
```
baentsch, line 222: Is the idea to completely drop this define when doing a release then? Or if just setting an empty pre-release string is that equal to`!defined`? Maybe worth while adding proposed wording for the changes to the [Release wiki](https://github.com/open-quantum-safe/liboqs/wiki/Release-process#code-change) in the comment?

baentsch (changes requested):
I'd think it'd be prudent document (progression and change logic of) these defines somewhere such as for users to know what they can rely on (e.g., is "OQS_VERSION_PRE_RELEASE" undefined an indication of a release?). Unsure where to best place it, though: Maybe in "CONFIGURE.md" or as comment atop "oqs.h"?

dstebila:
> I'd think it'd be prudent document (progression and change logic of) these defines somewhere such as for users to know what they can rely on (e.g., is "OQS_VERSION_PRE_RELEASE" undefined an indication of a release?). Unsure where to best place it, though: Maybe in "CONFIGURE.md" or as comment atop "oqs.h"?

I've now added documentation comments in `oqsconfig.h.cmake`, which then get copied into the generated `oqsconfig.h`.

tests/system_info.c
dstebila: I've updated the release wiki to explain what's going on here. I don't have an opinion one way or the other whether it is better for `OQS_VERSION_PRE_RELEASE` to be defined-but-empty or undefined for releases.

baentsch: I'd think it'd be easier to a consumer of `liboqs` could do `ifndef OQS_VERSION_PRE_RELEASE` to detect/distinguish (using) a release.

dstebila: Okay -- which is what's currently implemented in the code. So can I mark this as resolved?

src/oqsconfig.h.cmake
```diff
@@ -1,6 +1,21 @@
...
+ */
 #cmakedefine OQS_VERSION_TEXT "@OQS_VERSION_TEXT@"
+/** 
+ * Version levels of liboqs as integers.
+ */
+#define OQS_VERSION_MAJOR @OQS_VERSION_MAJOR@
+#define OQS_VERSION_MINOR @OQS_VERSION_MINOR@
+#define OQS_VERSION_PATCH @OQS_VERSION_PATCH@
+/** 
+ * OQS_VERSION_PRE_RELEASE is defined if this is a pre-release version of liboqs, otherwise it is empty.
```
dstebila, line 14:
```suggestion
 * OQS_VERSION_PRE_RELEASE is defined if this is a pre-release version of liboqs, otherwise it is undefined.
```

baentsch (approved):

Issue #1951, opened 2024-10-15
ML-KEM doesn't perform encapsulation key check


ueno:
**Describe the bug**
When encapsulating, FIPS 204 7.1 [suggests](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf#page=36) checking input key so any integers in the encoded public key is in the range [0, q - 1]. However, the current liboqs code seems to omit this check.

**To Reproduce**
1. Compile the attached [test-kem.c](https://github.com/user-attachments/files/17372939/test-kem.c.txt) with:  `gcc -o test-kem test-kem.c $(pkg-config liboqs --cflags --libs)`
2. Run `test-kem`
3. You will see: `test-kem: test-kem.c:48: main: Assertion 'rc != OQS_SUCCESS' failed.`

**Expected behavior**
The program should exit normally.

**Environment (please complete the following information):**
 - OS: Fedora Linux 40
 - OpenSSL version 3.2.2
 - Compiler version used gcc 14.2.1
 - Build variables used `-DBUILD_SHARED_LIBS=ON -DOQS_USE_AES_OPENSSL=ON -DOQS_USE_AES_INSTRUCTIONS=OFF -DOQS_DIST_BUILD=ON -DOQS_ALGS_ENABLED=STD_IANA -DOQS_USE_SHA3_OPENSSL=ON -DOQS_DLOPEN_OPENSSL=ON -DCMAKE_BUILD_TYPE=Debug -LAH`
 - liboqs version 0.11.0

**Additional context**
This was found when running a tlsfuzzer test [script](https://github.com/tlsfuzzer/tlsfuzzer/blob/ml-kem/scripts/test-tls13-mlkem.py#L258) currently in development, against gnutls-serv (in [MR](https://gitlab.com/gnutls/gnutls/-/merge_requests/1883)).

SWilson4:
Thanks for the report, @ueno. We pull our ML-KEM implementation from the [pq-crystals/kyber](https://github.com/pq-crystals/kyber) repo. It seems to me that it would be better to add the optional checks upstream and then integrate the updated code into liboqs. Would you agree with that assessment?

As an aside, I don't believe we support the `STD_IANA` value for the `OQS_ALGS_ENABLED` variable—I assume this is something from a fork?

SWilson4:
Also tagging @bhess as he is the one most familiar with the CRYSTALS algorithms.

nidhidamodaran:
[...]

Issue #1894, opened 2024-08-15
SLH-DSA: integrate final standard


SWilson4:
In line with [FIPS 205](https://csrc.nist.gov/pubs/fips/205/final).

We have support for Round 3 SPHINCS+, but nothing more recent. Our current upstream source is https://github.com/sphincs/sphincsplus via PQClean, so we should find out their plans for updates.

Tagging @bwesterb as a significant upstream contributor: are there plans to bring the implementation in sync with the final standard?

bwesterb:
> Tagging @bwesterb as a significant upstream contributor: are there plans to bring the implementation in sync with the final standard?

Yes, we will. Personally I'm prioritising Kyber -> ML-KEM as we have that deployed quite widely. I ping the rest of the team.

psheer-spirent:
Hello,

According to Commit  36f399438892af6e6ce4dc1fdf00c8cc2f6e936b ,  liboqs is compliant with SPHINCS+ 3.1, June 10, 2022.

Now the FIPS 205 pdf reads that Sphinx+ 3.1 is a superset of SLH-DSA.

So it looks like, except for changing the cipher user-friendly names, liboqs is already compliant with FIPS 205.

Can anyone help here? Does liboqs now fully implement a binary compatible FIPS 205?

Thank you and kind regards!

Paul

SWilson4:
Hi, @psheer-spirent. I do agree that [Appendix A of FIPS 205](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.205.pdf#appendix.A) makes it sound like SLH-DSA and SPHINCS+ 3.1 should be the same on the parameter sets we support. However, I suspect that in order to implement FIPS 205, we would need to add the "external" API, which appears to add domain separation and a context string. We need to make similar changes to support the FIPS 204 final version of ML-DSA.

psheer-spirent:
> Hi, @psheer-spirent. I do agree that ....

Thanks @SWilson4 for that clarification.

Do you know if oqs-provider+liboqs intends full compliance with FIPS-203, FIPS-204, and FIPS-205?
I had assumed that 203 and 204 were already supported since the README.md for oqs-provider states under version 0.6.0, "First availability of standardized PQ algorithms FIPS.203, FIPS.204".

Is anyone right now working toward full compliance?

Kind regards,

Paul

baentsch:
[...]

PR #2009, opened 2024-12-02
Add checks for ML-KEM keys

abhinav-thales:
This PR introduces the sanity checks for encapsulation and decapsulation keys introduced in the final specification of FIPS-203 in section 7.

The changes include:
	- checking generated encaps and decaps key in the keypair step
	- checking encaps key before the encapsulation step
	- checking decaps key before the decapsulation step

Ideally this check should be in the ML-KEM source code but I am not sure if we are allowed to make changes to the upstream code. Kindly let me know otherwise and will move the code accordingly.


baentsch:
Thanks for this improvement @abhinav-thales . It seems the issue tracking this is https://github.com/open-quantum-safe/liboqs/issues/1951:

> if we allowed to make changes to the upstream code

We are allowed to do anything :-) The question is whether the upstream takes such changes. More seriously: `liboqs` has a patch mechanism to insert code not yet made available by the upstreams but we're not over-eager to use that facility as it's introducing more work in a place where it conceptually doesn't belong.

Anyway, please see the discussion in the issue above. This PR could serve as a test stopgap measure and as such imo is mergeable as-is. What's your take @bhess: Also do a patch or wait for the upstream?

bhess:
I agree that including key validation tests makes sense. As you pointed out, the pq-crystals upstream doesn't include these validations, so we’d need to rely on our patch mechanism, which would increase maintenance overhead on the liboqs side.

One alternative to consider is using this as an opportunity to start adopting the https://github.com/pq-code-package/mlkem-native implementation from PQCP. It recently had its first (alpha) release and is actively maintained. This implementation already includes the required key validations [1][2], and the advantage is that it would eliminate the need for us to maintain custom patches.

[1] https://github.com/pq-code-package/mlkem-native/blob/112dbd3d3c42aa6558a448bb80affe5f8c158ece/mlkem/kem.c#L36
[2] https://github.com/pq-code-package/mlkem-native/blob/112dbd3d3c42aa6558a448bb80affe5f8c158ece/mlkem/kem.c#L64

baentsch:
> One alternative to consider is using this as an opportunity to start adopting the https://github.com/pq-code-package/mlkem-native implementation from PQCP

Why not -- not a new issue, though -- my thoughts stated half a year ago still stand/may be worth while addressing then: https://github.com/pq-code-package/tsc/issues/15#issuecomment-2044882432 : Technically, this could be simple, no? PQCP adds "copy-from-upstream" YML files and OQS adds it as another upstream (assuming some form of agreement on APIs, of course :) My strong preference would be to then drop other upstreams to indeed make life easier and not more complicated with this addition.

baentsch:
Additional thought: As and when there are different algorithm code bases becoming available and if OQS would still want to extend its utility, wouldn't this be the perfect opportunity to finally introduce the notion of "upstream maintained" to the YML file as a way to signal code (bases) that someone may rely on for use beyond research & prototyping? Given the various professionally maintained code bases for standardized algorithms already available or with a clear plan to do so, I'm personally less and less convinced this goal is a sensible one for OQS to pursue, though -- but wouldn't want to rule it out and not lay foundations for that -- like with such distinction of actively supported upstream code bases that would make it into a "reliable build" (vs. the current "amalgam/best effort" one).

abhinav-thales:
Hi @baentsch @bhess  :
the final decision is not clear to me :(
can we do the patches(in crypto source, also reduces code complexity by using inbuilt 'poly' functions) OR 
include these changes in the test file currently and wait for the upstream switch ?

baentsch:
> the final decision is not clear to me :(

There is no final decision yet (as to which upstreams OQS keeps supporting). As far as I'm concerned, there's a lot speaking for switching for ML-KEM from PQClean/reference code to PQCP code -- if OQS decides to keep supporting ML-KEM at all (see related discussion [here](https://github.com/open-quantum-safe/liboqs/issues/1894#issuecomment-2496096033)).

> include these changes in the test file currently and wait for the upstream switch ?

That would be my preference for now. Sub-optimal as it will not be available to users of the library but anything else has a high potential to be wasted effort.

abhinav-thales:
Thanks @baentsch . we could proceed with the PR as it is then.

@bhess can I request a review from you as well please ? I don't have permission to add 'request reviews' .

tests/vectors_kem.c
```diff
@@ -11,15 +11,34 @@
...
-
+#include <oqs/sha3.h>
 #include "system_info.c"
 
+#ifdef OQS_ENABLE_KEM_ML_KEM
+/* macros for sanity checks for encaps and decaps key */
+#define ML_KEM_BLOCKSIZE        384
+#define ML_KEM_K_MAX            4
+#define ML_KEM_N                256
+#define ML_KEM_1024_PK_SIZE     1568
```
SWilson4, line 22:
```suggestion
```

This can be replaced with the existing `OQS_KEM_ml_kem_1024_length_public_key` macro.

tests/vectors_kem.c
```diff
@@ -11,15 +11,34 @@
...
+#define ML_KEM_Q                3329
+#define SHA256_OP_LEN           32
+/* since x is 12 bits, max value could be 4095. the below macro uses this to implement a simple time constant mod 3329 */
+#define MOD_Q(x)             ((x) - ((x >= ML_KEM_Q) * ML_KEM_Q))
```
SWilson4, line 26: I don't believe we can count on comparison operators being constant time. I suggest doing Barrett reduction here instead, similar to the [reference implementation](https://github.com/open-quantum-safe/liboqs/blob/dd1706bc9bcd13dc4c909dcae057227449e9a14a/src/kem/ml_kem/pqcrystals-kyber-standard_ml-kem-768_ref/reduce.c#L35). (That code computes a centred representation; we'd just need an additional addition.)

I realize that this is overkill for testing code, but there's a possibility we use this file as a guide for patching the algorithm source later.

tests/vectors_kem.c
```diff
@@ -58,13 +77,75 @@ static void hexStringToByteArray(const char *hexString, uint8_t *byteArray) {
...
+	/* sanity checks */
+	if ((NULL == sk) || (NULL == method_name) || (false == is_ml_kem(method_name))) {
+		fprintf(stderr, "[vectors_kem] %s ERROR: inputs NULL or invalid method !\n", method_name);
+		return false;
+	}
+	/* buffer to hold public key hash */
+	uint8_t pkdig[SHA256_OP_LEN] = {0};
+	/* fetch the value of k according to the ML-KEM algorithm as per FIPS-203
+	K = 2 for ML-KEM-512, K = 3 for ML-KEM-768 & K = 4 for ML-KEM-1024 */
+	uint8_t K = (0 == strcmp(method_name, OQS_KEM_alg_ml_kem_512)) ? 2 : (0 == strcmp(method_name, OQS_KEM_alg_ml_kem_768)) ? 3 : 4;
```
SWilson4, line 91: I love the ternary operator, but this logic would be much easier to parse if it used `if/else`.

tests/vectors_kem.c
```diff
@@ -58,13 +77,75 @@ static void hexStringToByteArray(const char *hexString, uint8_t *byteArray) {
...
+static inline bool sanityCheckPK(const uint8_t *pk, size_t pkLen, const char *method_name) {
+	/* sanity checks */
+	if ((NULL == pk) || (0 == pkLen) ||  (NULL == method_name) || (false == is_ml_kem(method_name))) {
+		fprintf(stderr, "[vectors_kem] %s ERROR: inputs NULL or zero or invalid method !\n", method_name);
+		return false;
+	}
+	unsigned int i, j;
+	/* fetch the value of k according to the ML-KEM algorithm as per FIPS-203
+	K = 2 for ML-KEM-512, K = 3 for ML-KEM-768 & K = 4 for ML-KEM-1024 */
+	uint8_t K = (0 == strcmp(method_name, OQS_KEM_alg_ml_kem_512)) ? 2 : (0 == strcmp(method_name, OQS_KEM_alg_ml_kem_768)) ? 3 : 4;
```
SWilson4, line 110: Same comment as above.

SWilson4 (commented):
Thanks very much for these improvements.

tests/vectors_kem.c
```diff
@@ -58,13 +77,75 @@ static void hexStringToByteArray(const char *hexString, uint8_t *byteArray) {
...
+	/* buffer to hold encoded value */
+	uint8_t buffe[ML_KEM_1024_PK_SIZE] = {0};
+	uint16_t *buff_dec;
+	/* perform byte decoding as per Algo 6 of FIPS 203 */
+	for (i = 0; i < K; i++) {
+		buff_dec = &buffd[i * ML_KEM_N];
+		const uint8_t *curr_pk = &pk[i * ML_KEM_BLOCKSIZE];
+		for (j = 0; j < ML_KEM_N / 2; j++) {
+			buff_dec[2 * j] = ((curr_pk[3 * j + 0] >> 0) | ((uint16_t)curr_pk[3 * j + 1] << 8)) & 0xFFF;
+			buff_dec[2 * j] = MOD_Q(buff_dec[2 * j]);
```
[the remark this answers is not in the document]
abhinav-thales: added '+0' as recommended.

tests/vectors_kem.c
abhinav-thales: moved it to a separate function now.

tests/vectors_kem.c
abhinav-thales: moved it to a separate function now.

tests/vectors_kem.c
abhinav-thales: correct, added a simple mod function as this was a test file. will change it to proper mod function after few tests at my end.

tests/vectors_kem.c
abhinav-thales: The macro 'OQS_KEM_ml_kem_1024_length_public_key' is only generated if 'ML-KEM-1024' is enabled. incase of builds(minimal builds?) with 'ML-KEM-1024' not enabled will result in the macro not being generated and compilation error henceforth. so, decide to hardcode here.

SWilson4: Good point!

nidhidamodaran:
@abhinav-thales  wouldn't a check like below on public be more simpler ?

+int poly_frombytes(poly *r, const uint8_t a[KYBER_POLYBYTES])
 {
   unsigned int i;
   for(i=0;i<KYBER_N/2;i++) {
     r->coeffs[2*i]   = ((a[3*i+0] >> 0) | ((uint16_t)a[3*i+1] << 8)) & 0xFFF;
     r->coeffs[2*i+1] = ((a[3*i+1] >> 4) | ((uint16_t)a[3*i+2] << 4)) & 0xFFF;
+    if((r->coeffs[2*i] >= KYBER_Q) || (r->coeffs[2*i+1]  >= KYBER_Q))
+       return -1;
   }
+  return 0;
 }

tests/vectors_kem.c
abhinav-thales: Hi @SWilson4 , thanks for the update.
But any particular reason, why non-constant time is ok ? having it time constant would be the ideal scenario IMO.
in the meantime, I have tested modQ using "Barrett reduction" and another approach using shift operators as well.

SWilson4: > Hi @SWilson4 , thanks for the update. But any particular reason, why non-constant time is ok ? having it time constant would be the ideal scenario IMO. in the meantime, I have tested modQ using "Barrett reduction" and another approach using shift operators as well.

I brought it up with the OQS team, and we decided that we're OK with a non–constant time function here because the code is limited to a test file, in the interest of not holding up the PR. If you prefer to submit a constant-time implementation, that's fine too :)

SWilson4:
> @abhinav-thales wouldn't a check like below on public be more simpler ?
> 
> ```
> +int poly_frombytes(poly *r, const uint8_t a[KYBER_POLYBYTES])
>  {
>    unsigned int i;
>    for(i=0;i<KYBER_N/2;i++) {
>      r->coeffs[2*i]   = ((a[3*i+0] >> 0) | ((uint16_t)a[3*i+1] << 8)) & 0xFFF;
>      r->coeffs[2*i+1] = ((a[3*i+1] >> 4) | ((uint16_t)a[3*i+2] << 4)) & 0xFFF;
> +    if((r->coeffs[2*i] >= KYBER_Q) || (r->coeffs[2*i+1]  >= KYBER_Q))
> +       return -1;
>    }
> +  return 0;
>  }
> ```

See discussion [here](https://github.com/open-quantum-safe/liboqs/issues/1951#issuecomment-2551316213) for why this patch is not a good idea.

nidhidamodaran:
understood. thanks for the explanation. I was just trying to understand whether we really need to encode the key data to reject keys.

dstebila:
@bhess You are planning to change the ML-KEM implementation in liboqs to [PQCP's mlkem-native](https://github.com/pq-code-package/mlkem-native). Will that include the desired checks or will it still require separate code?

bhess:
> Will that include the desired checks or will it still require separate code?

Yes, the checks are by default in mlkem-native

tests/vectors_kem.c
abhinav-thales: Got it. i will try to submit the code this week

abhinav-thales: Hi @SWilson4 , the code changes are pushed now. As suggested earlier in this conversation, MOD(Q) has been done using barrett-reduction. With the changes, the tests look fine.

Could you please review the changes ?

abhinav-thales:
> LGTM. Tagging @bhess for a second opinion as our ML-KEM expert.

Hello @bhess, could you please help to review the PR when you have a moment ?
Thanks

bhess:
Thanks @abhinav-thales , I'll take a closer look this week.

baentsch:
Can I suggest checking this PR against the code in #2041 (that should include the same logic if I'm not mistaken)?

abhinav-thales:
> Can I suggest checking this PR against the code in #2041 (that should include the same logic if I'm not mistaken)?

Thanks @baentsch for the suggestion. I tested the changes locally by fetching the branch from the mentioned PR and updating this PR changes over it. With these changes, the tests were run and they look fine.

abhinav-thales:
NOTE: The failing test is unrelated to the PR changes.

It is due to https://github.com/open-quantum-safe/liboqs/issues/2056

dstebila (approved):

tests/vectors_kem.c
```diff
@@ -58,12 +75,106 @@ static void hexStringToByteArray(const char *hexString, uint8_t *byteArray) {
...
+	uint8_t K = get_ml_kem_k(method_name);
+	if (0 == K) {
+		fprintf(stderr, "K value can be fetched only for ML-KEM !\n");
+		return false;
+	}
+	/* buffer to hold decoded value. max value used, so same buffer could be used for ML-KEM versions
+	encaps key is of length 384K bytes(384K*8 bits). Grouped into 12-bit values, the buffer requires (384*K*8)/12 = 256*K entries of 12 bits */
+	uint16_t buffd[ML_KEM_N * ML_KEM_K_MAX] = {0};
+	/* buffer to hold encoded value */
+	uint8_t buffe[ML_KEM_1024_PK_SIZE] = {0};
```
bhess, line 145: Isn't the max value `ML_KEM_1024_PK_SIZE - 32` (384*K), since the rho value isn't used in the check?

tests/vectors_kem.c
```diff
@@ -11,15 +11,32 @@
 #include <sys/stat.h>
 
 #include <oqs/oqs.h>
-
+#include <oqs/sha3.h>
 #include "system_info.c"
 
+#ifdef OQS_ENABLE_KEM_ML_KEM
+/* macros for sanity checks for encaps and decaps key */
+#define ML_KEM_BLOCKSIZE        384
```
bhess, line 19: nit: It might be better to name this `ML_KEM_POLYBYTES`, as that’s the naming convention most implementations I’m familiar with use.

tests/vectors_kem.c
```diff
@@ -58,12 +75,106 @@ static void hexStringToByteArray(const char *hexString, uint8_t *byteArray) {
...
+}
+/* sanity check for private/decaps key */
+static inline bool sanityCheckSK(const uint8_t *sk, const char *method_name) {
+	/* sanity checks */
+	if ((NULL == sk) || (NULL == method_name) || (false == is_ml_kem(method_name))) {
+		fprintf(stderr, "[vectors_kem] %s ERROR: inputs NULL or invalid method !\n", method_name);
+		return false;
+	}
+	/* buffer to hold public key hash */
+	uint8_t pkdig[SHA256_OP_LEN] = {0};
```
bhess, line 110: Better name this `SHA3_256_OP_LEN` to be clear that this is SHA3 and not SHA256.

tests/vectors_kem.c
```diff
@@ -58,12 +75,106 @@ static void hexStringToByteArray(const char *hexString, uint8_t *byteArray) {
...
+		return 2;
+	} else if (0 == strcmp(method, OQS_KEM_alg_ml_kem_768)) {
+		return 3;
+	} else if (0 == strcmp(method, OQS_KEM_alg_ml_kem_1024)) {
+		return 4;
+	} else {
+		return 0;  // Default/error case
+	}
+}
+/* sanity check for private/decaps key */
```
bhess, line 102: I'd add a reference to the spec to be clear which check is performed here (e.g. "hash check" in FIPS 203, 7.3).

tests/vectors_kem.c
```diff
@@ -58,12 +75,106 @@ static void hexStringToByteArray(const char *hexString, uint8_t *byteArray) {
...
+	}
+	/* calcualte hash of the public key(len = 384k+32) stored in private key at offset of 384k */
+	OQS_SHA3_sha3_256(pkdig, sk + (ML_KEM_BLOCKSIZE * K), (ML_KEM_BLOCKSIZE * K) + 32);
+	/* compare it with public key hash stored at 768k+32 offset */
+	if (0 != memcmp(pkdig, sk + (ML_KEM_BLOCKSIZE * K * 2) + 32, SHA256_OP_LEN)) {
+		return false;
+	}
+	return true;
+}
+/* sanity check for public/encaps key */
```
bhess, line 126: I'd add a reference to the spec to be clear which check is performed here (e.g. "modulus check" in FIPS 203, 7.2).

bhess (commented):
Thanks @abhinav-thales for adding these tests! See the few comments/nits.

mkannwischer:
> > Will that include the desired checks or will it still require separate code?
> 
> Yes, the checks are by default in mlkem-native. Test code as done in this PR may still be useful IMO.

@bhess @dstebila 
Some clarification: mlkem-native does indeed have the pk and sk validation. What we do not do by default is the pair-wise consistency check (PCT) that checks at the end of keygen that encaps+decaps works as expected. 
We added this [just now](https://github.com/pq-code-package/mlkem-native/pull/769), but it is disabled by default because it is so expensive - you can enable it by defining `MLK_KEYGEN_PCT`.
This check is _optional_ according to [FIPS203](https://csrc.nist.gov/pubs/fips/203/final), but it is _mandatory_  according to [FIPS 140-3 IG](https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf). 
Due to the cost, one likely only want to enable this in case FIPS 140-3 validation is needed. 

I have yet to find a cryptographer who thinks the PCT is useful .... but that's another discussion.

baentsch:
> you can enable it by defining MLK_KEYGEN_PCT

Thanks for letting us know -- this seems like a very sensible approach -- and also one that could be used to make me overcome a basic "unhappiness" with this PR: It's adding code of value only to a single algorithm and under specific circumstances (`liboqs` was meant to be algorithm-agnostic such as to ease the support effort for the library for the few people actively contributing).

So what about this proposal: We create a new config option, say OQS_STRICT_FIPS that would guard/enable both, these new checks as well as `MLK_KEYGEN_PCT`? It should be OFF by default so users have to select it in full acceptance of the impact. OQS supports/tests it in CI for as long as there's a committer in `liboqs` actively supporting this algorithm & feature (@mkannwischer would you be willing to take this on in case @bhess couldn't?).

tests/vectors_kem.c
abhinav-thales: added

tests/vectors_kem.c
abhinav-thales: added

tests/vectors_kem.c
abhinav-thales: modified

tests/vectors_kem.c
abhinav-thales: changed to POLY_BYTES

tests/vectors_kem.c
abhinav-thales: nice catch...changed to -32 now

abhinav-thales:
> Thanks @abhinav-thales for adding these tests! See the few comments/nits.

Hi @bhess , thanks for the review. have addressed the comments now. Tests look good.
Please review at your convenience

bhess (approved):
Thanks @abhinav-thales for the update!

@mkannwischer @baentsch I've opened https://github.com/open-quantum-safe/liboqs/issues/2088 to track the proposal for a build option for the additional checks.

baentsch:
Merged. Thanks for the contribution!

PR #2019, opened 2024-12-12
Remove hardcoded build paths & modify basic workflow to build in random path

### LINKED ISSUE #2018: New release broke (again) tests when not using default build directory
**Describe the bug**
Exactly the same issue as described in https://github.com/open-quantum-safe/liboqs/issues/1937, that was fixed with https://github.com/open-quantum-safe/liboqs/pull/1938

This time the new issue affects all the new tests added to `test_acvp_vectors.py` by https://github.com/open-quantum-safe/liboqs/pull/1919

I will provide a new patch to fix this issue on the new tests soon.

Would it be possible to modify the CI pipeline to build on a custom path to detect this kind of error before merging? What do you think?

iyanmv:
This fixes #2018 using `helpers.get_current_build_dir_name()`

In addition, [the basic workflow](https://github.com/open-quantum-safe/liboqs/blob/main/.github/workflows/basic.yml) is modified to use a random build path. Hopefully, this small change will catch future changes that introduce hardcoded paths.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -67,25 +67,27 @@ jobs:
       KEM_NAME: ml_kem_768
       SIG_NAME: ml_dsa_65
     steps:
+      - name: Create random build folder
+        run: tmp_build=$(mktemp -d) && echo "RANDOM_BUILD_DIR=$tmp_build" >> $GITHUB_ENV
       - name: Checkout code
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # pin@v4
       - name: Configure
         run: |
-          mkdir build && \
-          cd build && \
-          cmake -GNinja -DOQS_STRICT_WARNINGS=ON \
+          cmake \
+            -B ${{ env.RANDOM_BUILD_DIR }} \
             -GNinja \
+            -DOQS_STRICT_WARNINGS=ON \
             -DOQS_MINIMAL_BUILD="KEM_$KEM_NAME;SIG_$SIG_NAME" \
-            --warn-uninitialized .. > config.log 2>&1 && \
+            --warn-uninitialized . > config.log 2>&1 && \
           cat config.log && \
-          cmake -LA -N .. && \
+          cmake -LA -N . && \
           ! (grep -i "uninitialized variable" config.log)
       - name: Build code
         run: ninja
-        working-directory: build
+        working-directory: ${{ env.RANDOM_BUILD_DIR }}
       - name: Build documentation
         run: ninja gen_docs
-        working-directory: build
+        working-directory: ${{ env.RANDOM_BUILD_DIR }}
 
   cppcheck:
     name: Check C++ linking with example program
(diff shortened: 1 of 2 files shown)


SWilson4 (approved):
Thanks for the fix!

iyanmv:
I pushed too quick before catching the train 😅 I will fix the env name in a bit, and test locally the pipeline before pushing again.

.github/workflows/basic.yml
```diff
@@ -94,28 +96,30 @@ jobs:
...
+          cmake \
+            -B ${{ env.RANDOM_BUILD_DIR }} \
             -GNinja \
+            -DOQS_STRICT_WARNINGS=ON \
             -DOQS_MINIMAL_BUILD="SIG_$SIG_NAME" \
-            --warn-uninitialized .. > config.log 2>&1 && \
+            --warn-uninitialized > config.log 2>&1 && \
           cat config.log && \
-          cmake -LA -N .. && \
+          cmake -LA -N && \
```
bhess, line 112: same here

.github/workflows/basic.yml
```diff
@@ -129,23 +133,26 @@ jobs:
...
+            -B ${{ env.RANDOM_BUILD_DIR }} \
+            -GNinja \
+            -DOQS_STRICT_WARNINGS=ON \
             -DOQS_BUILD_FUZZ_TESTS=ON \
             -DOQS_MINIMAL_BUILD="SIG_$SIG_NAME" \
-            --warn-uninitialized .. > config.log 2>&1 && \
+            --warn-uninitialized > config.log 2>&1 && \
           cat config.log && \
-          cmake -LA -N .. && \
+          cmake -LA -N && \
```
bhess, line 150: same here

.github/workflows/basic.yml
```diff
@@ -67,25 +67,27 @@ jobs:
...
+          cmake \
+            -B ${{ env.RANDOM_BUILD_DIR }} \
             -GNinja \
+            -DOQS_STRICT_WARNINGS=ON \
             -DOQS_MINIMAL_BUILD="KEM_$KEM_NAME;SIG_$SIG_NAME" \
-            --warn-uninitialized .. > config.log 2>&1 && \
+            --warn-uninitialized > config.log 2>&1 && \
           cat config.log && \
-          cmake -LA -N .. && \
+          cmake -LA -N && \
```
bhess, line 83: Right now this seems to give the following warning: 
```
CMake Warning:
-- Cache values
  No source or binary directory provided.  Both will be assumed to be the
  same as the current working directory, but note that this warning will
  become a fatal error in future CMake releases.
```

Should you specify the path to `${{ env.RANDOM_BUILD_DIR }}` here as well to avoid this warning?

bhess (commented):
Thanks for this fix @iyanmv. See the comments above, otherwise LGTM.

.github/workflows/basic.yml
SWilson4: I think this will fix the warning:
```suggestion
          cmake -LA -N . && \
```

.github/workflows/basic.yml
SWilson4:
```suggestion
          cmake -LA -N . && \
```

.github/workflows/basic.yml
SWilson4:
```suggestion
          cmake -LA -N . && \
```

iyanmv:
I put back the `.` (replacing the `..` instead of nothing) at the end of each `cmake` command.

bhess (approved):

PR #2021, opened 2024-12-16
Trigger liboqs-java and liboqs-rust downstream CI

SWilson4:
This PR adds triggers for liboqs-java and liboqs-rust to the `trigger-downstream-basic` workflow.

Marked as draft pending merge of downstream PRs https://github.com/open-quantum-safe/liboqs-rust/pull/271 and https://github.com/open-quantum-safe/liboqs-java/pull/30.


dstebila (approved):

baentsch (approved):
LGTM. But why is this in Draft only?

SWilson4:
> LGTM. But why is this in Draft only?

Until https://github.com/open-quantum-safe/liboqs-rust/pull/271 lands, the `repository_dispatch` API call will fail, which would break liboqs CI on main if it were merged.

PR #2022, opened 2024-12-17
Add faces of contributors to README

dstebila:
I recently learned about [this tool](https://contrib.rocks/preview?repo=open-quantum-safe%2Fliboqs) to add pictures of contributors to the project README. Is it worth adding?

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ liboqs is an open source C library for quantum-safe cryptographic algorithms.
 	- [Contributing](#contributing)
 	- [License](#license)
 	- [Acknowledgements](#acknowledgements)
+	- [Contributors](#contributors)
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


SWilson4 (approved):
Sure, why not?

baentsch:
> Sure, why not?

Here's a probably "weirdo" European answer: Privacy.

Seeing pictures IMO arguably is OK when one is logged in to GH, where viewers "reciprocate" making available their own avatar. Exporting that to everyone viewing the README (possibly a _much_ larger audience outside of GH) may not be to everyone's liking (I'm still undecided whether I'd object personally). Sure, the technical/legal "problem" lies with GH making these "pictures" available to external extraction and the tool to "amalgamate" them, but it'd be our decision to use these possibilities.

baentsch:
On second thought -- I don't like this: There's lots of "AI" crawling around stitching together an ever more complete picture of the world (if you pardon the pun) using publicly available user images. "Fine" if users gave their consent for this (say in return for cheap search results) but IMO not OK if one provided one's picture to (only) reciprocate to fellow developers also "opening up" within a walled garden (GH in this case).

Thus -unless hearing good arguments to the opposite- no approval from my side.

baentsch (changes requested):
At least add reference to legal t's and c's explaining why this is legit.

dstebila:
Even for website visitors not logged in to Github, the images of the first 14 contributions are visible on https://github.com/open-quantum-safe/liboqs and the images of all contributors are visible on https://github.com/open-quantum-safe/liboqs/graphs/contributors.

That being said, I do not feel strongly enough to push for this; it was just a cute thing I saw on another project and wanted to suggest.

baentsch:
> Even for website visitors not logged in to Github, the images of the first 14 contributions are visible on https://github.com/open-quantum-safe/liboqs and the images of all contributors are visible on https://github.com/open-quantum-safe/liboqs/graphs/contributors.

Thanks for the pointer. Just logged out of GH to check (for the first time in many years, it seems :-). I'm not sure I like what I (still) see then. So your point is well taken, @dstebila . Withdrawing my "reservations" then. Do as you please.

dstebila:
Closing without merging due to lack of second approving review.

PR #2026, opened 2024-12-19
Update PQClean commit and delete patch for HQC

SWilson4:
PQClean adopted our patch in https://github.com/PQClean/PQClean/pull/578, so it is no longer necessary to apply it as part of the `copy_from_upstream` process. Unfortunately a lot of whitespace changes (presumably due to astyle upstream?) landed as well... I recommend using `git diff --ignore-blank-lines main` to review.


baentsch (approved):
Thanks -- also for improving the upstream!

dstebila (approved):

PR #2027, opened 2024-12-29
Exendend api and modification for allow generate keypair from fixed seed for some sig algos.

mraksoll4:
Exendend api and modification for allow generate keypair from fixed seed for some sig algos.

suporeted now .

all variants of Dilithium
all variants of Falcon

Todo add testvectors. ( tested over python calls  by hands for now.)
Todo add dirrect generation public key from private key. (done)

`OQS_SIG_keypair_from_fseed(const OQS_SIG *sig, uint8_t *public_key, uint8_t *secret_key, const uint8_t *seed);`

expect seed at args. ( 48 byte string for Falcon and 32 byte string for Dilithium)

need deep review . But seems it works as expected.

### DIFF
--- a/src/sig/dilithium/oldpqclean_dilithium2_aarch64/api.h
+++ b/src/sig/dilithium/oldpqclean_dilithium2_aarch64/api.h
@@ -21,6 +21,8 @@ int PQCLEAN_DILITHIUM2_AARCH64_crypto_sign_keypair(uint8_t *pk, uint8_t *sk);
 
 int PQCLEAN_DILITHIUM2_AARCH64_crypto_sign_keypair_from_fseed(uint8_t *pk, uint8_t *sk, const uint8_t *seed);
 
+int PQCLEAN_DILITHIUM2_AARCH64_crypto_sign_pubkey_from_privkey(uint8_t *pk, const uint8_t *sk);
+
 int PQCLEAN_DILITHIUM2_AARCH64_crypto_sign_signature(
     uint8_t *sig, size_t *siglen,
     const uint8_t *m, size_t mlen, const uint8_t *sk);
--- a/src/sig/dilithium/oldpqclean_dilithium2_aarch64/api.h
+++ b/src/sig/dilithium/oldpqclean_dilithium2_aarch64/api.h
@@ -19,6 +19,8 @@
 
 int PQCLEAN_DILITHIUM2_AARCH64_crypto_sign_keypair(uint8_t *pk, uint8_t *sk);
 
+int PQCLEAN_DILITHIUM2_AARCH64_crypto_sign_keypair_from_fseed(uint8_t *pk, uint8_t *sk, const uint8_t *seed);
+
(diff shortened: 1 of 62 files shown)


baentsch:
Thanks for sharing your activities, @mraksoll4 -- please do check our [contribution guidance](https://github.com/open-quantum-safe/liboqs/blob/main/CONTRIBUTING.md), particularly heeding [guidance on upstream code](https://github.com/open-quantum-safe/liboqs/wiki/Patching-upstream-sources) as you are modifying here, though. 

Finally, please mark this PR Ready for Review only when indeed everything is ready for review (incl. patches and tests). If you'd like to discuss the approach taken, it may be better to open an issue or discussion item to discuss the problem before doing a PR. On the topic, please be sure to also take a look at pre-existing issues like https://github.com/open-quantum-safe/liboqs/issues/1206...

mraksoll4:
> Thanks for sharing your activities, @mraksoll4 -- please do check our [contribution guidance](https://github.com/open-quantum-safe/liboqs/blob/main/CONTRIBUTING.md), particularly heeding [guidance on upstream code](https://github.com/open-quantum-safe/liboqs/wiki/Patching-upstream-sources) as you are modifying here, though.
> 
> Finally, please mark this PR Ready for Review only when indeed everything is ready for review (incl. patches and tests). If you'd like to discuss the approach taken, it may be better to open an issue or discussion item to discuss the problem before doing a PR. On the topic, please be sure to also take a look at pre-existing issues like #1206...

well, it's still far from a full-fledged solution (although these functions work as expected and are mostly built from existing ones), later I'll fix everything in the right style, although the commits are not signed at need style so what i need do ? recrate PR at rigt style or i can some how modify this exist PR,  I was created an issue alrady , but it was closed because the necessary functionality did not exist so i created it.

The most difficult thing is still planned - implementing something similar to the logic of hmac, I don't know yet if it is possible at all, we have no problems with private keys, it is quite easy to organize a key tree from them, but the ability to get a new public key from the master public key  with no decrease seccurity is questionable

baentsch:
> although the commits are not signed at need style so what i need do ?

@ryjones could help with that.

> created an issue alrady , but it was closed because the necessary functionality did not exist so i created it.

Can you please point to that? Maybe re-open or rather, create a new one with a design proposal so you can get feedback on that before doing a PR?

mraksoll4:
> > although the commits are not signed at need style so what i need do ?
> 
> @ryjones could help with that.
> 
> > created an issue alrady , but it was closed because the necessary functionality did not exist so i created it.
> 
> Can you please point to that? Maybe re-open or rather, create a new one with a design proposal so you can get feedback on that before doing a PR?

yes i think i simple create new PR for do it more clear + make separate patches for every algo, how i understand  patches need do only for algo changes, and every variant of algo as exanple falcon512 and falcon1024 need separate pacth.

mraksoll4:
PR is recreated with patches etc..

https://github.com/open-quantum-safe/liboqs/pull/2031

Issue #2030, opened 2025-01-02
Signature schemes  generate keypair from seed , and recontruct public key from private key.


mraksoll4:
Signature schemes  generate keypair from seed , and recontruct public key from private key.

SWilson4:
Hi @mraksoll4, would you consider this issue to be a duplicate of https://github.com/open-quantum-safe/liboqs/issues/1401 or https://github.com/open-quantum-safe/liboqs/issues/1802 (or both)?

mraksoll4:
> Hi @mraksoll4, would you consider this issue to be a duplicate of #1401 or #1802 (or both)?

It’s similar, but I don’t think so, since that issues there relate to encapsulations, and mine are related to Signature schemes.

SWilson4:
> > Hi @mraksoll4, would you consider this issue to be a duplicate of #1401 or #1802 (or both)?
> 
> It’s similar, but I don’t think so, since that issues there relate to encapsulations, and mine are related to Signature schemes.

That's fair. I've renamed the issue to make its categorization a little more clear.

SWilson4:
Here's a use case for deriving signature keypairs from seeds: https://github.com/open-quantum-safe/liboqs/pull/2070#issuecomment-2731791343

humuhimi:
Hi! @SWilson4  I'd like to generate ML-DSA private keys from seed.

FIPS 204 Algorithm 6 defines how to do this (ML-DSA.KeyGen_internal), but liboqs doesn't have an API for it yet. It would be great if we can use it!

Any plans to add this?

ashman-p:
I am taking a look at this issue and wanted to propose an approach.
Update signature API to add 'keypair_frm_seed(uint8_t *public_key, uint8_t *secret_key, uint8_t *seed, uint8_t seed_len
);

I will roll in algorithms one by one to keep the PRs small.
Please give feedback and other things to consider.

PR #2031, opened 2025-01-02
extended api for allow use fixed seed for generate key pair and generate public key from private key in Signature schemes

mraksoll4:
extended api for allow use fixed seed for generate key pair and generate public key from private key in Signature schemes

in this case we pass the seed as a parame.

`OQS_SIG_keypair_from_fseed(const OQS_SIG *sig, uint8_t *public_key, uint8_t *secret_key, const uint8_t *seed);`

in this case we pass the private key as a param.

`OQS_SIG_pubkey_from_privkey(const OQS_SIG *sig, uint8_t *public_key, const uint8_t *secret_key);`

falcon and dilithium at this moment - all versions. But can be done for other algos to.

"issue #2030"

old pull request  https://github.com/open-quantum-safe/liboqs/pull/2027 , recreated for with correct style.

it's extended api , but not affect native support old functions.

### DIFF
--- a/docs/algorithms/sig/falcon.md
+++ b/docs/algorithms/sig/falcon.md
@@ -7,9 +7,9 @@
 - **Authors' website**: https://falcon-sign.info
 - **Specification version**: 20211101.
 - **Primary Source**<a name="primary-source"></a>:
-  - **Source**: https://github.com/PQClean/PQClean/commit/1eacfdafc15ddc5d5759d0b85b4cef26627df181
+  - **Source**: https://github.com/PQClean/PQClean/commit/1eacfdafc15ddc5d5759d0b85b4cef26627df181 with copy_from_upstream patches
   - **Implementation license (SPDX-Identifier)**: MIT
-- **Optimized Implementation sources**: https://github.com/PQClean/PQClean/commit/1eacfdafc15ddc5d5759d0b85b4cef26627df181
+- **Optimized Implementation sources**: https://github.com/PQClean/PQClean/commit/1eacfdafc15ddc5d5759d0b85b4cef26627df181 with copy_from_upstream patches
   - **pqclean-aarch64**:<a name="pqclean-aarch64"></a>
       - **Source**: https://github.com/PQClean/PQClean/commit/7707d1bcc8ae7f9ffd296dd13b1d76d2767d14f8
       - **Implementation license (SPDX-Identifier)**: Apache-2.0
--- a/docs/algorithms/sig/falcon.yml
+++ b/docs/algorithms/sig/falcon.yml
@@ -19,6 +19,7 @@ nist-round: 3
 spec-version: 20211101
 primary-upstream:
   source: https://github.com/PQClean/PQClean/commit/1eacfdafc15ddc5d5759d0b85b4cef26627df181
+    with copy_from_upstream patches
   spdx-license-identifier: MIT
   upstream-ancestors:
   - https://www.falcon-sign.info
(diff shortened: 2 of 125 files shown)


SWilson4:
I know this is a useful feature for a number of applications, but I am hesitant to add support for it to the liboqs API. Unlike the `keygen`, `sign`, and `verify` functions, the derivation functions don't have an interfact mandated by NIST, and this will mean lots of manual patching work on our part. I don't expect that this will be a significant issue for standardized / standards-track algorithms, but I can see it slowing down additions of new schemes (e.g., from the NIST signature on-ramp).

baentsch:
> I know this is a useful feature for a number of applications

What apps are we talking about here, @SWilson4 @mraksoll4 ? fwiw, `oqsprovider` just removed Dilithium, so it's a "don't care" there by now.
 
> this will mean lots of manual patching work on our part

A very true concern, and already visible in this PR, consisting of a gazillion patches that I am not not exactly eager to review and definitely unwilling to maintain, let alone replicate across all algorithms.

> I can see it slowing down additions of new schemes (e.g., from the NIST signature on-ramp).

Completely agree on this too: Algorithm-specific patching should only be a method of "last resort", say if there's a security issue that an upstream cannot or does not fix in a timely manner. Using it to alter the externally visible behaviour of many, let alone all, algorithm(s) with many separate, bespoke patches --even if doing basically the same thing-- does not seem like a long-term viable way forward: Such changes should come from the upstream and be driven by a standards body (such as NIST).

SWilson4:
> > I know this is a useful feature for a number of applications
> 
> What apps are we talking about here, @SWilson4 @mraksoll4 ? fwiw, `oqsprovider` just removed Dilithium, so it's a "don't care" there by now.

I meant "application" in the sense of an application of cryptography, as opposed to an app. I was thinking mostly of "fancy crypto" (e.g., BIP32).

mraksoll4:
In fact, we can make a stub for the API so that it works only in supported algorithms and does not litter others, I'm a little confused there. As for patches, the function is built on the basis of the basic key pair generator, we just accept the Seed parameter from the outside, just like in the case of opening a public key from a private one we use part if it for recontrctrun pubkey - for falcon example  from f and g.

It seems that I succeeded in this, but the problem is with Falcon and other algorithms where there is no meta data from the supported functionality, although you can go the other way and use a template through checking the algorithm "name"

mraksoll4:
> > > I know this is a useful feature for a number of applications
> > 
> > 
> > What apps are we talking about here, @SWilson4 @mraksoll4 ? fwiw, `oqsprovider` just removed Dilithium, so it's a "don't care" there by now.
> 
> I meant "application" in the sense of an application of cryptography, as opposed to an app. I was thinking mostly of "fancy crypto" (e.g., BIP32).

For this to work, we also need to figure out how to securely "derive" public keys from public keys. In addition to the private key tree, there is often also a public key tree, which allows us to create a new public key and associate it with a private key without revealing the private key.

baentsch:
This is dormant now for 2 months -- can I ping @mraksoll4 @SWilson4 on their intention how to proceed with this? I'm personally inclined to just close without seeing a clear use case and equally clear maintenance path/proposal.

dstebila:
@mraksoll4 Can you say a little more about what applications/users are expected to use this functionality?

mraksoll4:
I think we can probably close this issue, given that the library is unlikely to be used by those who need it (eg cryptocurrencies), most likely the implementation will be custom for those who need it. And the idea of ​​extending this into a hardener key tree like bip32 etc... runs into NIST's prohibition on manipulating the seed before submitting it for generation to have certification.

baentsch:
Thanks for the additional feedback @mraksoll4 . Closing as proposed then.

Issue #1807, opened 2024-06-05
Extend constant-time testing


bhess:
This issue is to suggest extending the constant-time testing to include more configurations.

Background: modern compilers might add regressions by 'optimizing' code meant to be constant-time and adding conditional branches. An example is a recent regression in the [Kyber/ML-KEM](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/hqbtIGFKIpU/m/cnE3pbueBgAJ?utm_medium=email&utm_source=footer&pli=1) reference code, when using clang >= v15 on x86_64 and compiler options -Os, -O1, -O2 -fno-vectorize or -O3 -fno-vectorize.

The liboqs constant-time tooling would be able to detect the above issue, but CI runs the test only on a single platform, with one compiler (gcc) and with one configuration (cmake DEBUG).

To cover such cases, the suggestion would be to run the tests with different variables (e.g., in weekly tests). For example:
- Architectures: x86_64 and ARM (as suggested by @SWilson4 in the last developers call)
- Generic and assembly-optimized code (as already done in CI for x86_64)
- Compilers: gcc and clang using different versions (e.g., system default version and latest available version)
- Optimization levels: -Os, -O0, -O1, -O2, -O3 (or instead, using the cmake default options: Debug, MINSIZEREL, RELWITHDEBINFO, RELEASE)
- Additional compiler flags (e.g., -fno-vectorize)

I think it's unrealistic to be really exhaustive, but it should be document which configurations were tested (and what the limitations of the tests are).

meldegwi:
Hi @bhess,

I'm really interested in contributing to this project as part of the lfx program. I saw on the lfx portal that this one is part of a custom term starting this August, since August already started is it still open?

I've already applied and am currently waiting to get my school permission to participate, which I expect to receive on Wednesday. In the meantime, is there anything I can start working on or any particular issues you’d recommend I look into? I’ve already reviewed some of the good first issues and had the plan to work on one of them.

PR #2033, opened 2025-01-03
Add threat model

dstebila:
As discussed in the 2024-12-05 TSC meeting, we want to update our SECURITY.md file to mention our threat model. I have based our threat model on the one in the [OpenSSL security policy](https://openssl-library.org/policies/general/security-policy/index.html).

Feedback and discussion welcome.


SECURITY.md
```diff
@@ -12,4 +12,23 @@ Using any code prior to 0.10.1 is strongly discouraged due to a [known security
...
 ## Reporting a Vulnerability
+
 Please follow [this information to report a vulnerability](https://openquantumsafe.org/liboqs/security.html#reporting-security-bugs).
+
+## Threat Model
+
+For certain algorithms and platforms, liboqs aims for "constant-time code" in the sense of no branching or memory accesses based on secret data.
+
+- Algorithms: The [algorithm datasheets](https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/) indicate whether each algorithm and implementation aims for constant-time behaviour. These should be read in conjunction with the [issues/passes files for each algorithm's constant-time tests](https://github.com/open-quantum-safe/liboqs/tree/main/tests/constant_time).
+- Platforms: We aim for constant-time behaviour on our [Tier 1 platforms](https://github.com/open-quantum-safe/liboqs/blob/main/PLATFORMS.md).
```
SWilson4, line 23: I'm a little hesitant to say that we aim for constant-time behaviour on all Tier 1 platforms, given that we only run constant-time tests on Linux / x86_64. There's a lot of aarch64 code that isn't constant-time tested, and we certainly don't run the CT tests on macOS. (Does Valgrind even work on macOS?) Maybe it would be best to say that we will respond to reports of CT issues on Tier 1 platforms.

I did actually try out the [CT-time tests on the Arm runner](https://github.com/open-quantum-safe/liboqs/tree/sw-extended-arm64-tests) late last year, and the results were OK: just a handful of reports on Falcon and McEliece that I haven't had time to pick through. Maybe this is a signal to take up that work again.

Martyrshot: If valgrind works on macOS I've never gotten it to work. 😓

dstebila: Good point. I was hoping to tie the threat model back to our supported platforms in some way, but I guess I didn't get it right. I guess yes, we could say that we will respond to reports of CT issues on Tier 1 platforms, and for all other tiers of platforms it is nice-to-have but not actively supported (or some other wording?).

SWilson4: The PLATFORMS.md file has a † next to platforms on which constant-time tests are run—maybe we can refer to this in the threat model and make an effort to expand the list past x86_64 / Linux.

Martyrshot: This does raise an interesting conversation. Does it make sense that some tier 1 platforms be supported in different ways (in this case constant time checks) than others? Maybe it would be worth having a set of tier 0 platforms? Forgive me for not being super active so there might be other cases where we support some tier 1 platforms in different ways than that other tier 1 platforms, but having a tier 0 "we aim to support everything on this platform" might be a nice to have distinction.

SWilson4: We actually discussed this in the PR for the PLATFORMS.md file: see https://github.com/open-quantum-safe/liboqs/pull/1605#discussion_r1414042767 and https://github.com/open-quantum-safe/liboqs/pull/1605#issuecomment-1840557646.

Michael pointed out that you can go to an even higher tier than CT-tested (formally verified), which was one reason we didn't make an almighty Tier 0 for CT-tested platforms.

baentsch: Interesting discussion. But isn't this secondary (i.e., needs to be informed by) what is meant by "liboqs aims for"? Does OQS warrant it (if so, how?) or is it a "good intention" only (everyone "aims" for being a good person, say) without concrete things someone can rely on (and that OQS has to pay for in terms of effort)?

Martyrshot: That's a good point Michael. For a security policy I think "We aim for" means we have actively deployed something that we think achieves the thing we are aiming for, but it covers us from issues where there may be a testing/implementation/whatever bug.

baentsch: Well, "covering" OQS is all well and good -- so I understand why the formulation is a bit "wobbly". So let me try again: Wouldn't it be good from the perspective of a user to know what this assertion _exactly_ means? Is OQS indeed warranting (to the best of its knowledge) that algorithms marked "constant time" are such? Or is it really that OQS only and exactly asserts "no branching or memory accesses based on secret data" exists (but no instruction-level execution time independence based on secret data, say)? I guess what I'm trying to say is: If OQS "aims for" some property, should that be defined --together with the methods it asserts it-- to some more detail?

baentsch: Does "constant-time behaviour" include instruction-level execution time independence? If so, is this checked? If so, how, particularly wrt compiler optimizations?

bhess: > Does "constant-time behaviour" include instruction-level execution time independence? If so, is this checked? If so, how, particularly wrt compiler optimizations?

I believe the current constant-time analysis still needs to be extended to automatically detect issues like Clangover (where compilers optimize away constant-time code) or Kyberslash (use of non-constant-time instructions, such as div). There's an open issue (https://github.com/open-quantum-safe/liboqs/issues/1807) to extend the constant-time testing with more coverage for compilers and configurations. With these extended tests, we could have greater confidence in the constant-time claim.

baentsch: Thanks @bhess -- those are exactly (some) of my concerns (and areas of work required before I'd feel comfortable making such broad claim).

SWilson4: It seems to me that there are two important things to communicate:
1. We consider timing side-channel attacks to be within scope of the threat model. We will make an effort to address vulnerability reports involving non--constant-time behaviour, and we document known instances of non--constant-time behaviour.
2. We have automated constant-time testing that aims to detect one class of vulnerability (secret-dependent branching or memory access) on specific platforms, and all failures are documented.

With that in mind, how about the following text?

> Timing-based side-channel attacks are within the scope of our threat model. OQS tests for secret-dependent branches and memory accesses on [select Tier 1 platforms](https://github.com/open-quantum-safe/liboqs/blob/main/PLATFORMS.md#tier-1). All test failures are documented as either "passes," which we have assessed to be false positives, or "issues," which may constitute non–constant-time behaviour. The [algorithm datasheets](https://github.com/open-quantum-safe/liboqs/tree/main/docs/algorithms) indicate whether or not an implementation passes our constant-time tests. Details about passes and issues are available in the [tests/constant_time directory](https://github.com/open-quantum-safe/liboqs/tree/main/tests/constant_time). 
These tests do not encompass all classes of non–constant-time behaviour; for example, they do not detect possible variable-time instructions, such as DIV. OQS will make a best effort to respond to reports of non–constant-time behaviour that fall outside the scope of these tests.

What do you think of this verbiage @dstebila @baentsch?

baentsch: Thanks for the proposal @SWilson4. The concrete actions taken and listed are OK and correct (though still not precise enough in my eyes: which platforms are covered by the term "select", for example?).

The leading statement I think is still too broad a promise, though:

> Timing-based side-channel attacks are within the scope of our threat model.

as the trailing caveats document. Therefore, the leading statement should be "qualified" at least along the lines of 

> Some timing-based side-channel attacks are within the scope of our threat model.

And re-reading the caveat again

> OQS will make a best effort to respond to reports of non–constant-time behaviour that fall outside the scope of these tests.

I wonder what this "best effort" entails? Will we change code that OQS did not author? Will OQS only revert to upstreams offering such maintenance? What if no-one is available to take a look?

This collection of "some", "select", "best effort" makes me wonder whether this section is not too weak to retain it at all: Would users read any reliable commitment out of this? Is it thus worth while retaining in the first place?

dstebila: Thanks for iterating on the more detailed wording. I tried to incorporate observations from the discussion in the revised text below, changing the first sentence (adding the word "Some") and the last sentence.

> Some timing-based side-channel attacks are within the scope of our threat model. OQS tests for secret-dependent branches and memory accesses on [select Tier 1 platforms](https://github.com/open-quantum-safe/liboqs/blob/main/PLATFORMS.md#tier-1). All test failures are documented as either "passes," which we have assessed to be false positives, or "issues," which may constitute non–constant-time behaviour. The [algorithm datasheets](https://github.com/open-quantum-safe/liboqs/tree/main/docs/algorithms) indicate whether or not an implementation passes our constant-time tests. Details about passes and issues are available in the [tests/constant_time directory](https://github.com/open-quantum-safe/liboqs/tree/main/tests/constant_time).
These tests do not encompass all classes of non–constant-time behaviour; for example, they do not detect possible variable-time instructions, such as DIV. Reports of non–constant-time behaviour that fall outside this scope will be considered on a case-by-case basis.

Regarding Michael's comment

> This collection of "some", "select", "best effort" makes me wonder whether this section is not too weak to retain it at all: Would users read any reliable commitment out of this? Is it thus worth while retaining in the first place?

I think the goal right now is to clearly document our current practice so that users can make their own decision. I think the text you've been iterating on is pretty close to reflecting our current practice. As to whether users will find it a satisfyingly reliable commitment -- well, that's up to them. If they want a stronger commitment, then hopefully that will drive them to participate.

baentsch: > If they want a stronger commitment, then hopefully that will drive them to participate.

To move this beyond "hope", what about making this explicit, then? Maybe with an introduction to the whole document reading like this:

> This project is not commercially supported. All guidelines and goals stated below are reflections of current practices, executed by a community of voluntary contributors on a best effort basis and may change at any time. 

> Any entity seeking more reliable commitments is strongly encouraged to join our community and thus enhance the code and support that the community can provide.

baentsch: > This feels like a blanket statement that applies more broadly than the threat model. Is there a better place for it?

If we agree that it is correct statement, then of course, it might sensibly be listed in all places where we think people will try to learn about the state of the code -- and where we may be able to find new contributors.

IMO it fits perfectly at the top of SECURITY.md as that file is what GH and other docs point to where to learn about the security properties of the software -- and where OQS with this PR now documents its commitments in this regard and that the proposed text refers to.

It then of course also could be added to the main README.md, CONTRIBUTING.md and the www site for people using those to gauge the level of support that the community can provide. But those files do not contain any commitment statements such as this document, so that would arguably need different verbiage. My proposal indeed was meant to target SECURITY.md only at this time -- as that's what this PR is about.

I'd be glad to provide suitably adapted versions via PR(s) for the other places if we agree that the basic tenet is right: Please let me know if I should do.

dstebila: I think belongs in places where we're cautioning people about limitations / risks: so that would be here in SECURITY.md, in the limitations section of README.md, and the [limitations section of the website](https://openquantumsafe.org/about/).

dstebila: > > Some timing-based side-channel attacks are within the scope of our threat model. OQS tests for secret-dependent branches and memory accesses on [select Tier 1 platforms](https://github.com/open-quantum-safe/liboqs/blob/main/PLATFORMS.md#tier-1). All test failures are documented as either "passes," which we have assessed to be false positives, or "issues," which may constitute non–constant-time behaviour. The [algorithm datasheets](https://github.com/open-quantum-safe/liboqs/tree/main/docs/algorithms) indicate whether or not an implementation passes our constant-time tests. Details about passes and issues are available in the [tests/constant_time directory](https://github.com/open-quantum-safe/liboqs/tree/main/tests/constant_time).
> > These tests do not encompass all classes of non–constant-time behaviour; for example, they do not detect possible variable-time instructions, such as DIV. Reports of non–constant-time behaviour that fall outside this scope will be considered on a case-by-case basis.

I've updated the PR with (a slightly edited version of) this language.

SWilson4 (approved):

SECURITY.md
SWilson4: > If we agree that it is correct statement, then of course, it might sensibly be listed in all places where we think people will try to learn about the state of the code -- and where we may be able to find new contributors.

I agree with the overall sentiment, though I hesitate to describe the OQS community as "voluntary contributors." Certainly, a significant portion of contributions come from volunteers (one volunteer in particular :slightly_smiling_face:), but (as far as I know) most of the Waterloo contributions over the years have been from people who are compensated for their work in some form. For me, this is explicit: my contract with UW states that I am employed to work on OQS.

> I'd be glad to provide suitably adapted versions via PR(s) for the other places if we agree that the basic tenet is right: Please let me know if I should do.

@baentsch In the interest of moving this PR forward, how about creating a new issue/PR to add such a statement to all relevant places in one shot?

SWilson4: > @SWilson4 Sure: #2065. And apologies for phrasing things in a way that you felt excluded: I was aware that you're employed (as --nearly-- everyone else) but wasn't aware that (caring for) OQS is your sole task: As with everyone else I thought OQS is a topic worked on when there are no other obligations (which probably was the key aspect I meant when writing "voluntary contributors" above -- a better formulation welcome).

Thanks, and all good.

> As far as "moving this PR forward", I do not see myself holding it up: This conversation was started by yourself (and as such is for you to mark/consider resolved): If there's something I'm not happy with, I might speak up.

Marking as resolved now, I just didn't want to bury discussion that had developed subsequent to my intial CT-related concerns.

SWilson4:
Now that conversation(s) are resolved and https://github.com/open-quantum-safe/tsc/pull/124 is ready for review, this PR is also ready for review in my view.

praveksharma (approved):

SECURITY.md
```diff
@@ -12,4 +12,20 @@ Using any code prior to 0.10.1 is strongly discouraged due to a [known security
 | < 0.12  | :x:                |
 
 ## Reporting a Vulnerability
+
 Please follow [this information to report a vulnerability](https://openquantumsafe.org/liboqs/security.html#reporting-security-bugs).
+
+## Threat Model
```
bhess, line 18: > The [algorithm datasheets](https://github.com/open-quantum-safe/liboqs/tree/main/docs/algorithms) indicate whether or not an implementation passes our constant-time tests.

The datasheets also mention 'No branching-on-secrets claimed.' Including this in the text could help highlight that some implementations may not be constant-time.

baentsch: > some implementations may not be constant-time.

I personally consider (adding) this statement as "too aggressive by inference". Let me explain:

Doesn't the above imply to the casual reader that OQS considers some (the "other") implementations as CT?

In reality, though, those implementations passing the (limited) OQS testing only have one logical property (no branching on secrets) ascertained, nothing more: "Real" CT imo requires quite a bit more (code that compilers cannot wreak havoc with regarding CT; tests on all platforms, etc.). And that "full CT" property I do not think a single code base in OQS can claim to possess, nor does OQS check, let alone warrant it, right?

I'd thus suggest keeping the current text as-is, possibly adding explicitly that OQS cannot confirm presence of the constant time property, but just its absence. If that's your proposal, @bhess (?) I'm all for it then suggesting this re-phrasing:

> The [algorithm datasheets](https://github.com/open-quantum-safe/liboqs/tree/main/docs/algorithms) indicate whether or not an implementation claims the non-branching on secret data property and passes the OQS tests for this property. This property is required but not sufficient for ascertaining the constant-time property of an algorithm.

bhess: I agree @baentsch, adding such a note as you suggest might be useful to clarify that liboqs doesn't prove the constant-time property, but performs (currently a limited number of) tests to try to reject the property.

Apologies that my previous comment wasn't that clear. What I meant was that there are multiple reasons why non-constant-time behavior (secret-dependent branching) might arise. Specifically:

1. The algorithm/implementation doesn't claim to be constant-time, as indicated in the datasheet.
2. liboqs tests have identified non-constant-time behavior, and this is reflected in the datasheet.
3. Behavior that current liboqs tests currently don't detect (e.g., DIV instructions).

I believe the current text addresses points 2 and 3, but it doesn’t fully capture point 1.

SECURITY.md
```diff
@@ -12,4 +12,20 @@ Using any code prior to 0.10.1 is strongly discouraged due to a [known security
...
+
+Some timing-based side-channel attacks are within the scope of our threat model. OQS tests for secret-dependent branches and memory accesses on Linux on x86\_64. All test failures are documented as either "passes," which we have assessed to be false positives, or "issues," which may constitute non–constant-time behaviour. The [algorithm datasheets](https://github.com/open-quantum-safe/liboqs/tree/main/docs/algorithms) indicate whether or not an implementation passes our constant-time tests. Details about passes and issues are available in the [tests/constant_time directory](https://github.com/open-quantum-safe/liboqs/tree/main/tests/constant_time). These tests do not encompass all classes of non–constant-time behaviour; for example, they do not detect possible variable-time instructions, such as `DIV`. Reports of non–constant-time behaviour that fall outside this scope will be considered on a case-by-case basis, with a priority on [Tier 1 platforms](https://github.com/open-quantum-safe/liboqs/blob/main/PLATFORMS.md#tier-1).
```
SWilson4, line 20:
```suggestion
Some timing-based side-channel attacks are within the scope of our threat model. OQS tests for secret-dependent branches and memory accesses on Linux on x86\_64. All test failures are documented as either "passes," which we have assessed to be false positives, or "issues," which may constitute non–constant-time behaviour. The [algorithm datasheets](https://github.com/open-quantum-safe/liboqs/tree/main/docs/algorithms) indicate whether or not an implementation passes our constant-time tests, as well as whether or not it is expected to pass. Details about passes and issues are available in the [tests/constant_time directory](https://github.com/open-quantum-safe/liboqs/tree/main/tests/constant_time). These tests do not encompass all classes of non–constant-time behaviour; for example, they do not detect possible variable-time instructions, such as `DIV`. Reports of non–constant-time behaviour that fall outside this scope will be considered on a case-by-case basis, with a priority on [Tier 1 platforms](https://github.com/open-quantum-safe/liboqs/blob/main/PLATFORMS.md#tier-1).
```

SWilson4: @bhess How about the above addition of "... as well as whether or not it is expected to pass"?

ashman-p (approved):

SECURITY.md
bhess: Sounds good to me, thanks @SWilson4 !

PR #2041, opened 2025-01-13
Import ML-KEM from mlkem-native/PQ code package

bhess:
This PR tracks the integration of ML-KEM from the mlkem-native upstream repository.
It replaces the current ML-KEM implementation in liboqs, which was previously imported from pq-crystals, with the [mlkem-native implementation from PQCP](https://github.com/pq-code-package/mlkem-native).

Some features of mlkem-native:

- Portable C implementation (C90 compliant)
- Optimized implementation for x86_64
- Optimized implementation for ARM64
- Formal verification

The upstream code recently had a [v1.0.0-alpha release](https://github.com/pq-code-package/mlkem-native/releases/tag/v1.0.0-alpha) and is actively maintained. The goal is to synchronize the PR with an upcoming tagged release of mlkem-native.

Additionally, the upstream code includes enhanced key validation as defined by FIPS 203 by default, which resolves issue #1951.

Closes #1951.

TODOs:

- [x] Sync with the upcoming release version of mlkem-native
- [x] Update constant-time tests
- [x] Update documentation

* [ ] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [ ] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged.)

PR #2044, opened 2025-01-15
NVIDIA: Adding cuPQC as a backend for ML-KEM.

stevenireeves:
This PR adds the support for the NVIDIA library _cuPQC_ to be used as the backend for ML-KEM algorithms. 

cuPQC requires the use of an NVIDIA GPU to perform the PQC algorithms.


stevenireeves:
@praveksharma looks like a number of github actions are failing, can you help with this, or point to what needs to be changed?

praveksharma:
@stevenireeves I've pushed 3 commits to https://github.com/open-quantum-safe/liboqs/tree/libOQSxcuPQC, which builds off of your branch, to get it to pass CI.

praveksharma:
Thank you for making these changes @stevenireeves! Since I reviewed the private PR initially I'll let this one be reviewed by reviewers other than myself.

dstebila:
I talked to @praveksharma just now to understand some of the approach here, and I understand why the cupqc metadata is patched into the PQ Crystals meta.yml file (since there isn't another meta.yml to patch it into); so that makes sense. But I think a few other pieces could be done differently. I don't think the *.cu files need to be added via the patch file; couldn't they just be added directly without also being added through the patch?  And if so then could we give those directories a name that doesn't include pqcrystals -- e.g., just `src/kem/ml_kem/cupqc-ml-kem-768`?

stevenireeves:
> I talked to @praveksharma just now to understand some of the approach here, and I understand why the cupqc metadata is patched into the PQ Crystals meta.yml file (since there isn't another meta.yml to patch it into); so that makes sense. But I think a few other pieces could be done differently. I don't think the *.cu files need to be added via the patch file; couldn't they just be added directly without also being added through the patch? And if so then could we give those directories a name that doesn't include pqcrystals -- e.g., just `src/kem/ml_kem/cupqc-ml-kem-768`?

@dstebila with this organizational change will we need to mess with the copy_from_upstream stuff as well?

praveksharma:
> @dstebila with this organizational change will we need to mess with the copy_from_upstream stuff as well?

@stevenireeves I shall work on this and attemp to push the changes directly to your fork. I don't believe copy_from_upstream.py would need to be changed significantly.

stevenireeves:
> > @dstebila with this organizational change will we need to mess with the copy_from_upstream stuff as well?
> 
> @stevenireeves I shall work on this and attemp to push the changes directly to your fork. I don't believe copy_from_upstream.py would need to be changed significantly.

I have the allow edits from maintainers option selected. So _should_ be able to make those changes.

praveksharma:
> I don't think the *.cu files need to be added via the patch file; couldn't they just be added directly without also being added through the patch? And if so then could we give those directories a name that doesn't include pqcrystals -- e.g., just `src/kem/ml_kem/cupqc-ml-kem-768`?

@dstebila handling the naming shouldn't be an issue. The *.cu file must be sourced with every run of copy_from_upstream.py under delete mode, since there is no upstream the *.cu must be sourced from within the liboqs repo - a patch is the most straightforward way of doing this. Would a separate patch file for the *.cu file be adequate?

baentsch:
> I understand why the cupqc metadata is patched into the PQ Crystals meta.yml file (since there isn't another meta.yml to patch it into); so that makes sense

Hmm -- it doesn't to me: We have https://github.com/open-quantum-safe/liboqs/pull/2041 in the pipeline with the declared goal of removing the PQCrystals files. So this PR thus would need to be re-done after that landed, right? That doesn't seem sensible.

Wouldn't it be much more sensible to have this code be contained in an upstream of its own to pull it from? As far as I can see, this _is_ effectively a different implementation with different license terms, characteristics etc. Such split also would make responsibilities clear: NVIDIA is to support the interfacing to its library (that may very well change over time unbeknownst to OQS) and OQS is responsible for the proper operation within/integration into the OQS APIs (that may change unbeknownst to NVIDIA). Or is the intention by NVIDIA to become committed, well committers and/or maintainers to OQS @stevenireeves ?

stevenireeves:
> Hmm -- it doesn't to me: We have #2041 in the pipeline with the declared goal of removing the PQCrystals files. So this PR thus would need to be re-done after that landed, right? That doesn't seem sensible.
> 
> Wouldn't it be much more sensible to have this code be contained in an upstream of its own to pull it from? As far as I can see, this _is_ effectively a different implementation with different license terms, characteristics etc. Such split also would make responsibilities clear: NVIDIA is to support the interfacing to its library (that may very well change over time unbeknownst to OQS) and OQS is responsible for the proper operation within/integration into the OQS APIs (that may change unbeknownst to NVIDIA). Or is the intention by NVIDIA to become committed, well committers and/or maintainers to OQS @stevenireeves ?

Nvidia maintaining it's own fork of liboqs is not what we want to do. The intention of NVIDIA is to support libOQS on portions of liboqs that utilize cuPQC as it's backend. So if there are significant changes that require alteration of the source files related to cupqc in liboqs, we will help there (so long as we have the internal support). I believe that @praveksharma is in the process of  removing the cupqc metadata from PQ Crystals meta.yaml, correct me if I'm wrong.

dstebila:
> > I don't think the *.cu files need to be added via the patch file; couldn't they just be added directly without also being added through the patch? And if so then could we give those directories a name that doesn't include pqcrystals -- e.g., just `src/kem/ml_kem/cupqc-ml-kem-768`?
> 
> @dstebila handling the naming shouldn't be an issue. The *.cu file must be sourced with every run of copy_from_upstream.py under delete mode, since there is no upstream the *.cu must be sourced from within the liboqs repo - a patch is the most straightforward way of doing this. Would a separate patch file for the *.cu file be adequate?

Oh, I see now. I knew that we have some local implementations that `--delete` doesn't wipe out (e.g., FrodoKEM), but these algorithms aren't handled at all by copy\_from\_upstream.  But I understand now that for algorithms that are handled by copy\_from\_upstream, there's no mixing of local (not deleted) and upstream (deleted) implementations.

So I see what the patch file is adding these.

dstebila:
> > Wouldn't it be much more sensible to have this code be contained in an upstream of its own to pull it from? As far as I can see, this _is_ effectively a different implementation with different license terms, characteristics etc. Such split also would make responsibilities clear: NVIDIA is to support the interfacing to its library (that may very well change over time unbeknownst to OQS) and OQS is responsible for the proper operation within/integration into the OQS APIs (that may change unbeknownst to NVIDIA). Or is the intention by NVIDIA to become committed, well committers and/or maintainers to OQS @stevenireeves ?
> 
> Nvidia maintaining it's own fork of liboqs is not what we want to do. The intention of NVIDIA is to support libOQS on portions of liboqs that utilize cuPQC as it's backend. So if there are significant changes that require alteration of the source files related to cupqc in liboqs, we will help there (so long as we have the internal support). I believe that @praveksharma is in the process of removing the cupqc metadata from PQ Crystals meta.yaml, correct me if I'm wrong.

I think these are talking about two slightly different things. Michael isn't suggesting that NVIDIA maintain a separate fork of liboqs. The way our code has been structured is that we have scripts to pull in source code from (self-contained) implementations of algorithms in other repositories, and add them to liboqs using code generation and patches.

This could be done that way, but since it is also only one file that is being added, I don't think it's worth the effort of setting up a separate upstream for that.

baentsch:
> Nvidia maintaining it's own fork of liboqs is not what we want to do.

This is not my ask: I only (meant to :) ask whether you'd want to maintain the wrapper code around the cuPQC code in a separate project -- complete with META.yml such as for `copy_from_upstream` to pull it in -- thus obviating the need for patching other implementations' META.yml files.

SWilson4:
I added the `--delete` functionality for copy_from_upstream quite recently. I can add an exception for the cuda files so that they don't get cleaned out; it would be quite straightforward and IMO would make maintenance of those files a lot easier. Editing patch files instead of source files is a pain. How does this sound @praveksharma?

praveksharma:
> I added the `--delete` functionality for copy_from_upstream quite recently. I can add an exception for the cuda files so that they don't get cleaned out; it would be quite straightforward and IMO would make maintenance of those files a lot easier. Editing patch files instead of source files is a pain. How does this sound @praveksharma?

Thank you for offering @SWilson4 but I think it is more prudent to create a separate git repository to store cupqc_ml-kem metadata since the pqcrystals upstream is going to be deprecated soon in any case. If things work out okay @stevenireeves or someone else from Nvidia can ownership of the repo.

stevenireeves:
> Thank you for offering @SWilson4 but I think it is more prudent to create a separate git repository to store cupqc_ml-kem metadata since the pqcrystals upstream is going to be deprecated soon in any case. If things work out okay @stevenireeves or someone else from Nvidia can ownership of the repo.

@praveksharma I am not what metadata you want us to store.

praveksharma:
> I think it is more prudent to create a separate git repository to store cupqc_ml-kem metadata since the pqcrystals upstream is going to be deprecated soon in any case

I've updated the import mechanism to use https://github.com/praveksharma/cupqc-mlkem as an upstream. Ideally I would like to modify copy_from_upstream to not require ad hoc upstreams for situations such as this.

Could you please review the updated PR @SWilson4 @dstebila @baentsch ?

stevenireeves:
> I've updated the import mechanism to use https://github.com/praveksharma/cupqc-mlkem as an upstream. Ideally I would like to modify copy_from_upstream to not require ad hoc upstreams for situations such as this.

Thanks @praveksharma

CONFIGURE.md
```diff
@@ -124,6 +125,13 @@ Dynamically load OpenSSL through `dlopen`. When using liboqs from other cryptogr
 
 Only has an effect if the system supports `dlopen` and ELF binary format, such as Linux or BSD family.
 
+### OQS_USE_CUPQC
+
+Can be `ON` or `OFF`.  When `ON`, use NVIDIA's cuPQC library where able (currently just ML-KEM).  When this option is enabled, liboqs may not run correctly on machines that lack supported GPUs. To download cuPQC follow the instructions at (https://developer.nvidia.com/cupqc-download/). Detailed descriptions of the API, requirments, and installation guide are in the cuPQC documentation (https://docs.nvidia.com/cuda/cupqc/index.html).
```
baentsch, line 130: nit: typo "requirements"

docs/algorithms/kem/ml_kem.md
```diff
@@ -9,6 +9,10 @@
 - **Primary Source**<a name="primary-source"></a>:
   - **Source**: https://github.com/pq-crystals/kyber/commit/10b478fc3cc4ff6215eb0b6a11bd758bf0929cbd with copy_from_upstream patches
   - **Implementation license (SPDX-Identifier)**: CC0-1.0 or Apache-2.0
+- **Optimized Implementation sources**: https://github.com/pq-crystals/kyber/commit/10b478fc3cc4ff6215eb0b6a11bd758bf0929cbd with copy_from_upstream patches
+  - **cupqc-cuda**:<a name="cupqc-cuda"></a>
+      - **Source**: https://github.com/praveksharma/cupqc-mlkem/commit/adb8454e56979628c07b67eb7d90f9337be6dc30
```
baentsch, line 14: Is this going to move to a non-personal GH repo?

scripts/copy_from_upstream/copy_from_upstream.yml
```diff
@@ -38,6 +38,14 @@ upstreams:
     kem_meta_path: '{pretty_name_full}_META.yml'
     kem_scheme_path: '.'
     patches: [pqcrystals-ml_kem.patch]
+  - 
+    name: cupqc
+    git_url: https://github.com/praveksharma/cupqc-mlkem.git
```
baentsch, line 43: Please eventually move to long-term maintained GH repo

baentsch (commented):
Basically LGTM, @stevenireeves @praveksharma ; just some nits (see single comments). Also the question whether it would be worth while adding build documentation and a "tier 3" entry to PLATFORMS.md (as I guess we'll never be able to test this in CI, right?)

docs/algorithms/kem/ml_kem.yml
```diff
@@ -20,6 +20,10 @@ primary-upstream:
   source: https://github.com/pq-crystals/kyber/commit/10b478fc3cc4ff6215eb0b6a11bd758bf0929cbd
     with copy_from_upstream patches
   spdx-license-identifier: CC0-1.0 or Apache-2.0
+optimized-upstreams:
+  cupqc-cuda:
+    source: https://github.com/praveksharma/cupqc-mlkem/commit/adb8454e56979628c07b67eb7d90f9337be6dc30
+    spdx-license-identifier: https://docs.nvidia.com/cuda/cupqc/license.html
```
SWilson4, line 26: Based on the licensing info in the `.cu` files, should this not be Apache 2.0? I feel like this tag refers to the source files that we include in liboqs, rather than the SDK, which we don't actually ship.

SWilson4 (commented):
I know we'd need to get additional hardware to run CI tests on the CUDA code, but it is possible to add a build-only job?

docs/algorithms/kem/ml_kem.yml
praveksharma: I see, I thought this was to document the licensing requirements liboqs users ought to take into account were they to use builds which include the implementation in question. In that case, I shall move the SDK licensing information to CONFIGURE.md for visibitility.

praveksharma:
> I know we'd need to get additional hardware to run CI tests on the CUDA code, but it is possible to add a build-only job?

I'm trying to do this on a maching without a Nvidia GPU and am running into trouble. Do you know if this is possible @stevenireeves?



> Also the question whether it would be worth while adding build documentation and a "tier 3" entry to PLATFORMS.md (as I guess we'll never be able to test this in CI, right?)

I think we should do this. But I'm not familiar enough GPU architectures to know which specific platform to list. @stevenireeves, given cuPQC's requirements:
```
    CUDA Toolkit 12.4 or newer
    Supported CUDA compiler
    x86_64 CPU
    A NVIDIA GPU with one the following architectures: 70, 75, 80, 86, 89, 90
```
is listing "NVIDIA GPU architectures 70, 75, 80, 86, 89, and 90 with a x86_64 CPU" sufficient information for those in the know? For reference here are other Tier 3 platforms from PLATFORMS.md:
```
    x86 for Windows (Visual Studio Toolchain)
    ppc641e for Ubuntu (Focal)
    s390x for Ubuntu (Focal)
    loongarch64 for Debian Linux (trixie)
```

stevenireeves:
> I'm trying to do this on a maching without a Nvidia GPU and am running into trouble. Do you know if this is possible @stevenireeves?

I **think** if you have the NVIDIA toolkit 12.4 (even though the machine does not have a GPU) you should be able to build. 

> I think we should do this. But I'm not familiar enough GPU architectures to know which specific platform to list. @stevenireeves, given cuPQC's requirements:
> 
> ```
>     CUDA Toolkit 12.4 or newer
>     Supported CUDA compiler
>     x86_64 CPU
>     A NVIDIA GPU with one the following architectures: 70, 75, 80, 86, 89, 90
> ```
> 
> is listing "NVIDIA GPU architectures 70, 75, 80, 86, 89, and 90 with a x86_64 CPU" sufficient information for those in the know? For reference here are other Tier 3 platforms from PLATFORMS.md:

cuPQC only support Linux with x86_64 currently. We only _officially_ support archs 70-90 but I don't see why any GPU made post 70 wouldn't. I think we could say Linux NVIDIA GPU architectures 70, 75, 80, 86, 89, and 90 with a x86_64 CPU in that list and be good. 

In the future we can update if we support more OS/CPUs and GPUs.

praveksharma:
> I know we'd need to get additional hardware to run CI tests on the CUDA code, but it is possible to add a build-only job?

> I **think** if you have the NVIDIA toolkit 12.4 (even though the machine does not have a GPU) you should be able to build.

This would require the CI images to updated with NVIDIA toolkit so I'll get started on that.

baentsch:
> This would require the CI images to updated with NVIDIA toolkit so I'll get started on that.

If there's no concrete desire to list this feature in a higher support tier (who wants that?) that'd be "future work" in my eyes, @praveksharma (or at least lower priority than many other open issues) and for me not a prerequisite for an approval of this PR.

stevenireeves:
@praveksharma looks like the build with OQS_USE_CUPQC is still failing in this PR. Although I can't see what is causing the failures. Did Neil's comments in the email help?

praveksharma:
> @praveksharma looks like the build with OQS_USE_CUPQC is still failing in this PR. Although I can't see what is causing the failures. Did Neil's comments in the email help?

@stevenireeves Yes, the comments were super helpful. I can succesfully build locally, this is mostly likely an issue with CI config itself that I'm still troubleshooting.

praveksharma:
@stevenireeves cmake in CI is failing with this message:

> CMake Error at /usr/share/cmake-3.28/Modules/CMakeDetermineCUDACompiler.cmake:270 (message):
>   Failed to detect a default CUDA architecture.

I am unable to reproduce this error locally in a docker container provisioned from the same image used in CI. Do you know how to fix this?

stevenireeves:
> > CMake Error at /usr/share/cmake-3.28/Modules/CMakeDetermineCUDACompiler.cmake:270 (message):
> > Failed to detect a default CUDA architecture.
> 
> I am unable to reproduce this error locally in a docker container provisioned from the same image used in CI. Do you know how to fix this?

@praveksharma ah, because CI has no GPU it can't detect a default. Try adding this flag to the CMake command.
`-DCMAKE_CUDA_ARCHITECTURES=80`

stevenireeves:
@praveksharma looks like that CI test is still failing. 

> No CMAKE_CUDA_COMPILER could be found.

 > Tell CMake where to find the compiler by setting either the environment
 > variable "CUDACXX" or the CMake cache entry CMAKE_CUDA_COMPILER to the full
 > path to the compiler, or to the compiler name if it is in the PATH.
 
 
You'll need to tell CMAKE where nvcc is.

stevenireeves:
@praveksharma looks like that did the trick. Any other comments by the reviewers?

SWilson4 (approved):
LGTM, with the caveat that we should eventually move away from a personal repo, as noted by @baentsch above.

baentsch:
> LGTM, with the caveat that we should eventually move away from a personal repo, as noted by @baentsch above.

Agreed: Please do this now or create an issue, ideally assigned, so this is not forgotten.

stevenireeves:
> @baentsch #2053 has been created to track this.
> 
> @stevenireeves do you want to move the upstream to your GitHub account?

@praveksharma 
I think it's fine, if we are just going to refactor copy from upstream so it's irrelevant. If we can it would be great to get this merged today.

stevenireeves:
@praveksharma @baentsch @dstebila anything else that needs to be done to get this in?

praveksharma:
> @praveksharma @baentsch @dstebila anything else that needs to be done to get this in?

I think this is good to merge, just need an additional approval.

stevenireeves:
> I think this is good to merge, just need an additional approval.

@praveksharma looks like we got what we needed

PR #2058, opened 2025-01-29
CI: Check unresolved symbols when compiled with OQS_DLOPEN_OPENSSL

ueno:
As a follow-up of commit 64bceb37fafa9b90cf228965079de9ebd77a83b9, this checks that the library artifacts don't contain any unresolved symbols from libcrypto.so when it is dynamically loaded.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -123,6 +123,12 @@ jobs:
       - name: Build
         run: ninja
         working-directory: build
+      - name: Check the library artifacts
+        if: matrix.name == 'jammy-std-openssl3-dlopen'
+        run: |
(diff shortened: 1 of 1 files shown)


dstebila (approved):

dstebila:
@praveksharma Is the build failure on cuPQC expected here?

praveksharma:
> @praveksharma Is the build failure on cuPQC expected here?

Yes, unfortunately, I think it has to do with CI image itself. Although, I'm not certain why test passed in #2044 but has been failing since. That particular test fails because CMake can't find the CUDA compiler, it shouldn't affect this PR.

dstebila:
> > @praveksharma Is the build failure on cuPQC expected here?
> 
> Yes, unfortunately, I think it has to do with CI image itself. Although, I'm not certain why test passed in #2044 but has been failing since. That particular test fails because CMake can't find the CUDA compiler, it shouldn't affect this PR.

Okay, I'll merge this one now.

PR #2070, opened 2025-02-07
Add DeriveKeyPair API

### LINKED ISSUE #1206: Adding a DeriveKeyPair functionality

SWilson4:
Based on the work of @Eddy-M-K in https://github.com/open-quantum-safe/liboqs/pull/1877.

Closes #1206.


docs/algorithms/kem/bike.md
```diff
@@ -13,11 +13,11 @@
...
-|:---------------:|:----------------------|:-----------------|---------------------:|--------------------------:|--------------------------:|--------------------------:|-----------------------------:|
-|     BIKE-L1     | NA                    | IND-CPA          |                    1 |                      1541 |                      5223 |                      1573 |                           32 |
-|     BIKE-L3     | NA                    | IND-CPA          |                    3 |                      3083 |                     10105 |                      3115 |                           32 |
-|     BIKE-L5     | NA                    | IND-CPA          |                    5 |                      5122 |                     16494 |                      5154 |                           32 |
+|  Parameter set  | Parameter set alias   | Security model   |   Claimed NIST Level |   Public key size (bytes) |   Secret key size (bytes) |   Ciphertext size (bytes) |   Shared secret size (bytes) | Keypair coins (bytes)   | Encapsulation coins (bytes)   |
```
bifurcation, line 16:
```suggestion
|  Parameter set  | Parameter set alias   | Security model   |   Claimed NIST Level |   Public key size (bytes) |   Secret key size (bytes) |   Ciphertext size (bytes) |   Shared secret size (bytes) | Keypair seed (bytes)   | Encapsulation seed (bytes)   |
```

Nit: "Seed" seems to be the terminology that folks are coalescing around.  See, e.g., [this IETF presentation](https://datatracker.ietf.org/meeting/121/materials/slides-121-pquip-fips-issues-with-deploying-ml-kem-and-ml-dsa-04)

bifurcation: Is it necessary to have the encapsulation coins / seed as well?  As far as I am aware, only the deterministic keypair generation is used in real applications; deterministic encapsulation is only done for testing.

docs/algorithms/kem/kyber.md
```diff
@@ -21,11 +21,11 @@
...
-|    Kyber768     | NA                    | IND-CCA2         |                    3 |                      1184 |                      2400 |                      1088 |                           32 |
-|    Kyber1024    | NA                    | IND-CCA2         |                    5 |                      1568 |                      3168 |                      1568 |                           32 |
+|  Parameter set  | Parameter set alias   | Security model   |   Claimed NIST Level |   Public key size (bytes) |   Secret key size (bytes) |   Ciphertext size (bytes) |   Shared secret size (bytes) | Keypair coins (bytes)   | Encapsulation coins (bytes)   |
+|:---------------:|:----------------------|:-----------------|---------------------:|--------------------------:|--------------------------:|--------------------------:|-----------------------------:|:------------------------|:------------------------------|
+|    Kyber512     | NA                    | IND-CCA2         |                    1 |                       800 |                      1632 |                       768 |                           32 | NA                      | NA                            |
```
bifurcation, line 26: Shouldn't these be the same as the ML-KEM entries?  Though I guess it would make sense to introduce the new API service with ML-KEM as the sole example; clearly we can upgrade other KEMs later.

scripts/copy_from_upstream/copy_from_upstream.py
```diff
@@ -200,6 +200,10 @@ def load_instructions(file='copy_from_upstream.yml'):
                 scheme['upstream_location'] = family['upstream_location']
             if (not 'arch_specific_upstream_locations' in scheme) and 'arch_specific_upstream_locations' in family:
                 scheme['arch_specific_upstream_locations'] = family['arch_specific_upstream_locations']
+            if (not 'derandomized_keypair' in scheme) and 'derandomized_keypair' in family:
```
bifurcation, line 203: Nit: I would change "derandomized" to "deterministic".

scripts/copy_from_upstream/src/kem/family/kem_scheme.c
```diff
@@ -133,6 +159,47 @@ extern int libjade_{{ scheme['pqclean_scheme_c'] }}_{{ impl['name'] }}_dec(uint8
...
+    {%- endif %}
+	return (OQS_STATUS) {{ scheme['metadata']['default_keypair_derand_signature'] }}(public_key, secret_key, coins);
+    {%- if scheme['metadata']['implementations']|rejectattr('name', 'equalto', scheme['default_implementation'])|list %}
+#endif
+    {%- endif %}
+    {%- else %}
+	(void)public_key;
+	(void)secret_key;
+	(void)coins;
+	return OQS_ERROR;
```
bifurcation, line 199: Just to confirm -- this provides a default always-error implementation if the back-end KEM doesn't supply one, right?  And likewise below for encap().

bifurcation (approved):
Overall, this looks right to me.  Couple of minor naming things and questions.

bifurcation:
This PR seems pretty mature.  What still needs to be done to merge it?

baentsch:
> This PR seems pretty mature. What still needs to be done to merge it?

Well, for one, all CI should pass. Then review should be requested. Finally, there's the question whether, how and where this should be added to some application layer to prove actual utility. Any input on that @SWilson4 ?

dstebila:
In terms of naming, I think "seed" is better than "coins" since that's the term used in FIPS 203. As for whether to call the function `_internal` or `_derand` or `_deterministic`: I don't think `_internal` is very descriptive, even though that's the FIPS 203 term; so I'd go with `_derand` or `_deterministic`, but don't have a strong preference between them.

SWilson4:
Marking as "Ready for Review" now that CI passes.

This PR exposes the ML-KEM "internal" keygen function via a new OQS KEM API for derandomized key generation.

My main intent with this PR is to enable experimentation with post-quantum crypto in protocols that require the HPKE interface, which includes a "DeriveKeyPair" function that is essentially derandomized key generation (see [RFC 9180](https://www.rfc-editor.org/rfc/rfc9180.html) and [draft-connolly-cfrg-hpke-mlkem](https://datatracker.ietf.org/doc/draft-connolly-cfrg-hpke-mlkem)). It turns out that this also allows users of liboqs to work with ML-KEM seeds (derandomized keygen converts a seed into an expanded keypair).

I've called the API "keypair_derand" since that was the name in the ML-KEM and Kyber codebases, but I'm open to changing this ("keypair_deterministic"? "keypair_derive"? "keypair_from_seed"?) 

Currently, calls to the new API will fail for algorithms other than ML-KEM.

scripts/copy_from_upstream/src/kem/family/kem_scheme.c
SWilson4: Yes, that's correct.

docs/algorithms/kem/bike.md
SWilson4: I agree: I've removed derandomized encaps from the latest iteration of this PR.

SWilson4:
> Well, for one, all CI should pass. Then review should be requested. Finally, there's the question whether, how and where this should be added to some application layer to prove actual utility. Any input on that @SWilson4 ?

@bifurcation Do you have any work on post-quantum MLS to point to here?

docs/algorithms/kem/kyber.md
baentsch: fwiw, I do not think this upgrade is going to happen: `liboqs` effectively switched away from incorporating the original reference implementation of Kyber/ML-KEM to using the ML-KEM-only PQCP implementation, so the support for Kyber will at least "wither away", if not outright removed.

baentsch:
> Currently, calls to the new API will fail for algorithms other than ML-KEM.

This statement I think is worth while documenting with higher visibility (README? Web site?): OQS by now clearly has pretty different levels of support for PQC algorithms also at the API level that I think users should be more prominently made aware of
- APIs only implemented for ML-DSA and ML-KEM
- A generally implemented API supported for all algorithms -- with some caveats:
   - other standardized algorithms are only included at a non-standard implementation level (Sphincs+/SLH-DSA, Falcon)
   - other standardized algorithms with some APIs disabled by default (XMSS, LMS)
   - "adjunct standard" algorithms (better wording welcome) like Frodo, Bike (?)
   - legacy algorithms like ntruprime, Kyber, Dilithium, HQC with limited update support
   - new standards-track proposals at different implementation levels like Cross, Mayo, hopefully UOV

The more I think about it, this seems worthy of a separate issue: This PR only contributes to an incoherence/deviation from the original OQS promise to support all PQC algorithms at equal measure that started earlier and goes deeper than just for this new API.

dstebila:
> The more I think about it, this seems worthy of a separate issue: This PR only contributes to an incoherence/deviation from the original OQS promise to support all PQC algorithms at equal measure that started earlier and goes deeper than just for this new API.

I don't agree that OQS has such a promise. There are many examples in liboqs where different algorithms have had different support or treatment: e.g., how far they are through the standards process (selections/alternates), constant-time support, platform support, large-stack usage. We recently introduced the sign-with-context API because ML-DSA needed it. We shouldn't do this without thought, but I think it is quite reasonable that the more important algorithms (ML-KEM and ML-DSA) merit special treatment. I think it may be more true that consumers of liboqs (OQS Provider, the language wrappers) have taken more of the approach you mentioned.

We have an action item from the last TSC to document algorithm support, similar to PLATFORMS.md.

bifurcation:
> > Well, for one, all CI should pass. Then review should be requested. Finally, there's the question whether, how and where this should be added to some application layer to prove actual utility. Any input on that @SWilson4 ?
> 
> @bifurcation Do you have any work on post-quantum MLS to point to here?

About a year ago, I did a prototype of XWing + MLS based on just copy/pasting the BoringSSL Kyber implementation into [MLSpp](https://github.com/cisco/mlspp).  I'm eager to do a more proper version of that using libOQS (ideally using OpenSSL, but I can go direct as an intermediate step).  This issue has been the major blocker on that.

baentsch:
> a prototype of XWing + MLS based on just copy/pasting the BoringSSL Kyber implementation into [MLSpp](https://github.com/cisco/mlspp).

That confuses me: According to its documentation, MLSpp is an

> Implementation of the proposed [Messaging Layer Security](https://github.com/mlswg/mls-protocol/blob/master/draft-ietf-mls-protocol.md) protocol in C++. Depends on C++17, STL for data structures, and OpenSSL or BoringSSL for crypto.

As both OpenSSL and BoringSSL support ML-KEM "natively" by now, what additional work is needed? What does liboqs add?

dstebila:
> > I don't agree that OQS has such a promise.
> 
> I do recall very vividly hearing very regularly that "OQS does not pick winners" whenever I suggested dropping some support (e.g., testing, platform-specific optimizations, profiling support) for some algorithms: Did I dream this? This may, well, actually must, have changed in light of the apparent drop in community contributions towards OQS. But I don't think this (whether change or not) is goodness and a more impartial support for all possible PQC algorithms would be goodness for OQS to enable true apple-to-apples comparisons between PQC algs.

Indeed we don't pick winners; that language is still in our README. But other people have picked winners, and I am in favour of giving the needs of those algorithms special consideration.

dstebila (approved):

baentsch:
> other people have picked winners, and I am in favour of giving the needs of those algorithms special consideration.

This special consideration comes with a cost, given our limited resources in that it distracts OQS from its original goal of highlighting benefits and drawbacks of not-yet standardized PQC algorithms.

I would whole-heartedly agree to putting special  OQS effort into standardized algorithms if there'd be no alternatives.

But there is: There's substantial (also FOSS) effort outside of OQS supporting standardized algorithms. OpenSSL alone has a much better (more functionally complete, more secure and certifiable, more performant pure C) code base for FIPS 20x algorithms that OQS still needs to catch-up to. And there's many more code bases with that focus.

Thus what benefit is there for OQS to play this "catch-up game"? Wouldn't it be better if OQS were focusing on next-gen PQC (again), not on what's published and dealt with?

Tagging @bifurcation : As you +1'd the above, can you please shed an "outsiders" perspective on this? What's your rationale for preferring `liboqs` over say BoringSSL or OpenSSL for standardized PQC? Any thoughts welcome to put inconsistencies in my thinking straight!

SWilson4:
@open-quantum-safe/liboqs-committers @open-quantum-safe/liboqs-maintainers Could I get some more eyes on this?

bhess (commented):
Thank you @SWilson4 for adding this useful API. The integration looks good to me.

I do have one question: Is there a specific reason why this API has been added only for KEMs and not for Signatures? I'm asking because there is also the use case of generating the key from a seed in the [IETF draft](https://github.com/lamps-wg/dilithium-certificates/blob/a079dcc4fb1c597a378d7c8ff17d688a0dfe9762/draft-ietf-lamps-dilithium-certificates.md?plain=1#L467-L475) for ML-DSA signatures.

praveksharma (approved):

SWilson4:
> Thank you @SWilson4 for adding this useful API. The integration looks good to me.
> 
> I do have one question: Is there a specific reason why this API has been added only for KEMs and not for Signatures? I'm asking because there is also the use case of generating the key from a seed in the [IETF draft](https://github.com/lamps-wg/dilithium-certificates/blob/a079dcc4fb1c597a378d7c8ff17d688a0dfe9762/draft-ietf-lamps-dilithium-certificates.md?plain=1#L467-L475) for ML-DSA signatures.

The original PR was only to support HPKE / MLS, and was made before the ML-KEM final standard was released (and all the discussion about seeds started in earnest). I imagine it would be straightforward to extend it to signatures (albeit only to ML-DSA for now) as well, if people would use the functionality. Does the use case from the IETF draft necessitate going through oqs-provider?

scripts/copy_from_upstream/patches/mlkem-native-ml_kem-derand.patch
```diff
@@ -0,0 +1,130 @@
+5ac78fb3da71b45132ba91256c5c5ac76b18fdbd
```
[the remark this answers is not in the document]
SWilson4: Good point, I will submit a concurrent PR there.

Dimidroll1989:
> > Thank you @SWilson4 for adding this useful API. The integration looks good to me.
> > I do have one question: Is there a specific reason why this API has been added only for KEMs and not for Signatures? I'm asking because there is also the use case of generating the key from a seed in the [IETF draft](https://github.com/lamps-wg/dilithium-certificates/blob/a079dcc4fb1c597a378d7c8ff17d688a0dfe9762/draft-ietf-lamps-dilithium-certificates.md?plain=1#L467-L475) for ML-DSA signatures.
> 
> The original PR was only to support HPKE / MLS, and was made before the ML-KEM final standard was released (and all the discussion about seeds started in earnest). I imagine it would be straightforward to extend it to signatures (albeit only to ML-DSA for now) as well, if people would use the functionality. Does the use case from the IETF draft necessitate going through oqs-provider?

@SWilson4 thank you for adding this API. Regarding signatures I want to mention the use case from [Post-Quantum PGP RFC](https://github.com/openpgp-pqc/draft-openpgp-pqc/blob/51f48d124bb67d635d9f4179fa65a6e532d7efba/draft-ietf-openpgp-pqc.md?plain=1#L793) where both ML-KEM and ML-DSA share same serialization format

scripts/copy_from_upstream/patches/mlkem-native-ml_kem-derand.patch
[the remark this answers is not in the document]
bhess: Thanks!

bhess:
> The original PR was only to support HPKE / MLS, and was made before the ML-KEM final standard was released (and all the discussion about seeds started in earnest). I imagine it would be straightforward to extend it to signatures (albeit only to ML-DSA for now) as well, if people would use the functionality. Does the use case from the IETF draft necessitate going through oqs-provider?

I believe the key encoding/decoding functionality would be implemented in the oqs-provider, while liboqs would still need to provide an API for (re-)generating the keys from the seed. However, I think it's fine to proceed with merging this PR as is focusing only on the KEM API.

bhess (approved):

docs/algorithms/kem/kyber.md
SWilson4: If we get requests to support the new API function for Kyber or other KEMs, we can look into it—in theory it should not be too difficult, but API maintenance may be an issue.

SWilson4:
I've done a rebase to clean up the git history a little and run the extended tests. Once those pass, I will hit the "merge" button.

Regarding naming of the new API: in the absence of a strong consensus, I prefer "derand" as that has been in use by ML-KEM / Kyber implementations for some time and matches the naming conventions of our upstream sources.

PR #2078, opened 2025-02-16
Update CROSS to version 2.0

### LINKED ISSUE #1958: Verification error in CROSS-RDSDPG-128-small
I managed to reproduce this very rare verification error in CROSS-RDSDPG-128-small. I'm still looking for the root cause but I can confirm that it seems totally unrelated to the new fuzzing harness.

_Originally posted by @rtjk in https://github.com/open-quantum-safe/liboqs/issues/1955#issuecomment-2432126812_

### LINKED ISSUE #1995: CROSS broken sigs validate
I’ve run the suggested test by @baentsch and noticed that the CROSS signature scheme seems to fail in some cases, passing the verification of modified signatures.

Tagging @rtjk.

_Originally posted by @bhess in https://github.com/open-quantum-safe/liboqs/pull/1919#discussion_r1848001764_

rtjk:
This pull request updates CROSS from version 1.2 to version 2.0.

CROSS has advanced to the second round of NIST's call for additional signatures,  and for the occasion, we have improved a few things. Here are the most relevant changes for liboqs (for a full list, please refer to the [changelog][cross_resources] submitted to NIST):

-Fixed trivial SUF-CMA forgeries (see #1995) by checking the zero-padding bytes during verification.
-Changed the seed-tree structure, fixing #1958 in the process.
-Resolved the endianness ambiguity in the CSPRNG (see #1961) by explicitly setting the bytes one at a time (we chose this solution over the one in #1983).

The insights we gained from OQS (especially from the liboqs test harness and oqs-demos) were instrumental in some of the improvements in CROSS 2.0. A huge thank you to the OQS maintainers for their great work!

Fixes #1958

Fixes #1995

[cross_resources]: https://www.cross-crypto.com/CROSS_Changelog_v2.0.pdf


rtjk:
Regarding the fuzzing tests from #1955, which led to #1958: Am I correct in understanding that they are enabled only for Dilithium2 and only in `basic.yml`, without being included in any other platform tests?

rtjk:
> As far as I understand, "No" and "Yes" in that order

Doesn't that test enable [only](https://github.com/open-quantum-safe/liboqs/blob/5afca642057faa54878cf6937b46fe6f00b45646/.github/workflows/basic.yml#L130) Dilithium2, though?

baentsch:
> > As far as I understand, "No" and "Yes" in that order
> 
> Doesn't that test enable [only](https://github.com/open-quantum-safe/liboqs/blob/5afca642057faa54878cf6937b46fe6f00b45646/.github/workflows/basic.yml#L130) Dilithium2, though?

It indeed seems that way -- I _guess_ that's because of the short time allotted for a PR CI. If you do a full build and fuzz test run, all SIG algs should be subject to the test, though. But upon checking the [extended weekly test](https://github.com/open-quantum-safe/liboqs/blob/main/.github/workflows/extended.yml), I indeed don't find extended fuzz testing there as I'd expected. Intentional or oversight @SWilson4 ?

baentsch (approved):
Thanks for the update @rtjk (and the friendly words!). Integration LGTM.

tests/KATs/sig/kats.json
```diff
@@ -105,74 +105,74 @@
   },
   "cross-rsdp-128-balanced": {
     "all": "cafdde1c92c9ce6c2cb67d7c6d225a4bbd103ecaece5cc4ebcb4fe6078e03421",
```
bhess, line 107: I believe the "all" hashes should be updated as well, tested with `test_kat_all.py`.

bhess (commented):
Thanks for the quick update, @rtjk! Please see the single comment. It appears that the extended tests (full KAT and constant-time tests) haven't run yet. You can trigger them by adding "[extended tests]" to the commit message.

tests/KATs/sig/kats.json
rtjk: Oops, my bad... I updated those and triggered the extended tests

dstebila (approved):

tests/KATs/sig/kats.json
baentsch: ok to mark resolved @bhess?

bhess (approved):

SWilson4:
> > > As far as I understand, "No" and "Yes" in that order
> > 
> > 
> > Doesn't that test enable [only](https://github.com/open-quantum-safe/liboqs/blob/5afca642057faa54878cf6937b46fe6f00b45646/.github/workflows/basic.yml#L130) Dilithium2, though?
> 
> It indeed seems that way -- I _guess_ that's because of the short time allotted for a PR CI. If you do a full build and fuzz test run, all SIG algs should be subject to the test, though. But upon checking the [extended weekly test](https://github.com/open-quantum-safe/liboqs/blob/main/.github/workflows/extended.yml), I indeed don't find extended fuzz testing there as I'd expected. Intentional or oversight @SWilson4 ?

I believe that extensive fuzzing of liboqs happens as part of [Google's oss-fuzz project](https://github.com/google/oss-fuzz/pull/12408), and that was the motivation for adding the fuzzing tests. We (should) get an email to security@openquantumsafe.org if a fuzzing test fails. The check in our CI is just to ensure we don't break the fuzzing build.

SWilson4:
@nathaniel-brough Please fact-check the above for me!

nathaniel-brough:
@SWilson4 yeah that's right. There is a small amount of fuzzing done on every pull request to ensure the build doesn't break.

Then every night google/oss-fuzz will build the fuzzers and run them on a distributed cluster for a few hours a day.

Every sig algorithm is enabled during the oss-fuzz build.

PR #2080, opened 2025-02-20
Added Kyber implementation details to test_kem.c

### LINKED ISSUE #1786: Enhance test output
Currently, most tests output information about the build and execution environment, e.g., as such:

```
Configuration info
==================
Target platform:  x86_64-Linux-6.5.0-26-generic
Compiler:         gcc (11.4.0)
OQS version:      0.10.1-dev
Git commit:       864be05738585e975540dd3ee1095d580904517f
OpenSSL enabled:  Yes (OpenSSL 3.0.2 15 Mar 2022)
AES:              NI
SHA-2:            OpenSSL
SHA-3:            C
CPU exts active:  ADX AES AVX AVX2 BMI1 BMI2 PCLMULQDQ POPCNT SSE SSE2 SSE3
```

However, `liboqs` increasingly makes more/different implementations available even for one algorithm, such as plain C code, hardware optimized code containing some assembly optimizations or even 100% assembly code built by external tooling. The actual code run is highly dependent on both build _and_ execution platform (if different).

This issue is to suggest adding a marker next to each algorithm name outputting which of the different code bases actually is running, e.g., by way of reference to the source file (path) actually being executed -- but many other options are conceivable.

pablogf-uma:
This PR aims to fix #1786 by including the implementation path in the test_kem.c file for Kyber implementations. Before moving forward with other algorithms, I would like to confirm if this approach is correct and if the referenced implementation path aligns with the information requested in the issue.

### DIFF
--- a/tests/test_kem.c
+++ b/tests/test_kem.c
@@ -149,8 +149,68 @@ static OQS_STATUS kem_test_correctness(const char *method_name) {
 		goto err;
 	}
 
+	// Check KEM implementation
+	const char *implementation = NULL;
+
+#if defined(OQS_ENABLE_LIBJADE_KEM_kyber_512_avx2)
+	if (strcasecmp(method_name, OQS_KEM_alg_kyber_512) == 0) {
+		implementation = "src/kem/kyber/libjade_kyber512_avx2";
+	}
+#elif defined(OQS_ENABLE_LIBJADE_KEM_kyber_512)
+	if (strcasecmp(method_name, OQS_KEM_alg_kyber_512) == 0) {
+		implementation = "src/kem/kyber/libjade_kyber512_ref";
+	}
+#elif defined(OQS_ENABLE_KEM_kyber_512_avx2)
+	if (strcasecmp(method_name, OQS_KEM_alg_kyber_512) == 0) {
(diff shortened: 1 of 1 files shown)


baentsch:
In addition to the comments by @dstebila allow me to state that OQS is a collection of many different algorithms and the amount of code catering to only some should be minimized as it makes maintenance of the library more and more difficult. Created https://github.com/open-quantum-safe/liboqs/pull/2081 to document this recurring issue for good.

pablogf-uma:
Thank you for your feedback! I included the alg_version field instead, as suggested by @dstebila. Let me know if you would like it to be added to any other tests as well. Also, @baentsch, would this information cover what was requested in the original issue?

pablogf-uma:
Thank you for your feedback! I included the alg_version field instead, as suggested by @dstebila. Let me know if you would like it to be added to any other tests as well. Also, @baentsch, would this information cover what was requested in the original issue?

dstebila (approved):

baentsch (approved):

baentsch:
Restarted failing Travis CI to try to remove failing CI marker.... If green, proceed to merge.

baentsch:
Merged. Thanks for the contribution @pablogf-uma !

PR #2086, opened 2025-02-24
Ensure that building against liboqs build directory works

levitte:
liboqsTargets.cmake is supposed to be adjacent liboqsConfig.cmake for the
latter to be functional.  This change ensure that this condition is met in
the build directory, allowing other CMake projects to build against a liboqs
build directory (as should be possible, implied by the use of 'export()').


SWilson4 (approved):
Thanks for the fix!

dstebila (approved):

Issue #1999, opened 2024-11-19
Strengthen signature verification failure tests 


SWilson4:
For background, please see this discussion: https://github.com/open-quantum-safe/liboqs/pull/1919#discussion_r1848001764 and related issue #1995.

It may be desirable to modify or extend our signature tests to take into account "weaker" forgeries as well as specific security claims (EUF-CMA vs SUF-CMA).

rtjk:
If we wanted to thoroughly check trivial forgeries for SUF-CMA, instead of flipping a single random bit we could have the test flip each bit of a signed message, one at a time, and expect a verification error. I made something similar to investigate #1995, would this be a welcome addition or is it out of scope?

baentsch:
> would this be a welcome addition or is it out of scope?

any improvement on testing the quality of code in OQS is welcome!

dstebila:
@rtjk Your proposal above would be nice to include. Is it something you'd be able to spend some cycles on?

rtjk:
> [@rtjk](https://github.com/rtjk) Your proposal above would be nice to include. Is it something you'd be able to spend some cycles on?

Sure! I can adapt the test I wrote for CROSS. Would it make sense to keep this in `test_sig.c`, or would it be better to create a separate file like `test_sig_bitflip.c`?

dstebila:
> > [@rtjk](https://github.com/rtjk) Your proposal above would be nice to include. Is it something you'd be able to spend some cycles on?
> 
> Sure! I can adapt the test I wrote for CROSS. Would it make sense to keep this in `test_sig.c`, or would it be better to create a separate file like `test_sig_bitflip.c`?

Great!

I think it would be okay to do all of it in `test_sig.c`.

PR #2090, opened 2025-02-27
Add bitflip test for trivial SUF-CMA forgeries

rtjk:
This pull request adds a test against trivial SUF-CMA forgeries in signature schemes: for every bit of the signature, flip it and ensure that the verification fails.

As discussed in #1999.


rtjk:
A couple of concerns from this initial draft:

- flipping every single bit like this results in a significant slowdown of `test_sig.c`, on my Skylake laptop it went from basically instantaneous to taking 2 minutes for a single signature (tested with cross-rsdp-128-balanced)

- ideally we would test only signatures declared as SUF-CMA, should we add `bool suf_cma` the OQS_SIG struct? Or maybe replace `bool euf_cma` with an enum specifying the claimed security?

Tagging @dstebila and @SWilson4 for feedback on this.

dstebila:
>     * flipping every single bit like this results in a significant slowdown of `test_sig.c`, on my Skylake laptop it went from basically instantaneous to taking 2 minutes for a single signature (tested with cross-rsdp-128-balanced)

Ouch. Before thinking thoroughly about, in mind signatures are short (few hundred bits) so it's "only" a factor of a few hundred, but of course PQ signatures are much longer, and so it's a factor of several thousand at least. Yeah, that's now a pain point.  Sorry for setting you down the wrong track.

I think Spencer's suggestion of a command-line argument would be a good compromise.

>     * ideally we would test only signatures declared as SUF-CMA, should we add `bool suf_cma` the OQS_SIG struct? Or maybe replace `bool euf_cma` with an enum specifying the claimed security?

I think recording this in the OQS_SIG struct would make sense; for backwards compatibility, I'd go with introducing a new bool rather than changing the name/behaviour of the existing flag.

scripts/copy_from_upstream/copy_from_upstream.py
```diff
@@ -368,7 +368,12 @@ def load_instructions(file='copy_from_upstream.yml'):
                             raise RuntimeError("Found duplicate arch {} in scheme {}".format(arch, scheme))
                         scheme['scheme_paths'][arch] = (os.path.join('repos', location,
                                                                     upstreams[location]['sig_scheme_path'].format_map(scheme)))
+            # assume EUF-CMA for schemes that don't specify a security classification
             scheme['metadata']['euf_cma'] = 'true'
+            scheme['metadata']['suf_cma'] = 'false'
+            if 'claimed-security' in metadata:
+                if metadata['claimed-security'] == "SUF-CMA":
+                    scheme['metadata']['suf_cma'] = 'true'
```
rtjk, line 376: Here's a recap of the claimed security of signature schemes as of now:
-ML-DSA is declared as SUF-CMA in the liboqs docs, but not in the upstream `META.yml`
-CROSS is declared as SUF-CMA both in the docs and in `META.yml`
-all other signatures are declared as EUF-CMA in the docs, but make no claim in `META.yml`
-stateful signatures don't report any claimed-security

From here, I think it should be clarified who has the ultimate responsibility for making the security claim: the liboqs docs or the upstream repo?

If we choose the liboqs docs, then it would be necessary to edit `sig->suf_cma` based on the documentation (e.g., the `claimed-security` field in `docs/algorithms/sig/ml_dsa.yml`).

Also, is it okay to assume EUF-CMA for imported schemes? This was the previous behavior, which I have kept.

Finally, how should we treat stateful signatures? Should I add the bitflip test there too?

dstebila: Thanks @rtjk. Probably simplest just to ask the ML-DSA upstream to add an SUF-CMA field to their META.yml. Yes, okay to assume EUF-CMA for all the others; we would not keep a scheme that is not EUF-CMA.  For stateful signatures, I think one could carry out the similar bit-flip test there, is there a reason not to?

rtjk: On second thought, there is actually no formal proof of SUF-CMA in CROSS at the moment, so I removed the claim from `META.yml`. I also opened a pull request to the ML-DSA upstream to add SUF-CMA [there](https://github.com/pq-crystals/dilithium/pull/101).

The test now works as suggested by @bhess in [#1919 (comment)](https://github.com/open-quantum-safe/liboqs/pull/1919#discussion_r1848674045):
-For EUF-CMA signatures, flip bits of the message and check that the verification fails.
-For SUF-CMA signatures (just ML-DSA for now), flip bits of the signature and check that the verification fails.

I just need to include stateful signatures, and then the PR should be ready for review!

SWilson4 (commented):
It looks like none of the current signature schemes have `suf_cma` set (unless I missed one in my review). Is it possible to add a patch for at least one (e.g., ML-DSA) so that the test logic is executed in CI?

src/sig/dilithium/sig_dilithium_2.c
```diff
@@ -16,6 +16,7 @@ OQS_SIG *OQS_SIG_dilithium_2_new(void) {
 
 	sig->claimed_nist_level = 2;
 	sig->euf_cma = true;
+	sig->suf_cma = false;
```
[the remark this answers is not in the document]
rtjk: I added patches to `META.yml` in Dilithium and ML-DSA, we can remove them once the [upstream pr](https://github.com/pq-crystals/dilithium/pull/101) gets merged

tests/test_sig.c
```diff
@@ -29,7 +29,63 @@ typedef struct magic_s {
...
+	int num_tests = sig->suf_cma ? 2 : 1;
+	for (int test = 0; test < num_tests; test++) {
+		/* select the array to tamper with (message or signature) */
+		uint8_t *tampered_array = (test == 1) ? signature : message;
+		size_t tampered_array_len = (test == 1) ? signature_len : message_len;
+		/* select the number of bitflips */
+		uint64_t bitflips_selected = bitflips_all[test] ? tampered_array_len * 8 : bitflips[test];
+		for (uint64_t i = 0; i < bitflips_selected; i ++) {
+			uint64_t random_bit_index;
+			OQS_randombytes((uint8_t *)&random_bit_index, sizeof(i));
```
SWilson4, line 57:
```suggestion
			OQS_randombytes((uint8_t *)&random_bit_index, sizeof(random_bit_index));
```

tests/test_sig_stfl.c
```diff
@@ -407,7 +407,59 @@ static char *convert_method_name_to_file_name(const char *method_name) {
...
+	int num_tests = sig->suf_cma ? 2 : 1;
+	for (int test = 0; test < num_tests; test++) {
+		/* select the array to tamper with (message or signature) */
+		uint8_t *tampered_array = (test == 1) ? signature : message;
+		size_t tampered_array_len = (test == 1) ? signature_len : message_len;
+		/* select the number of bitflips */
+		uint64_t bitflips_selected = bitflips_all[test] ? tampered_array_len * 8 : bitflips[test];
+		for (uint64_t i = 0; i < bitflips_selected; i ++) {
+			uint64_t random_bit_index;
+			OQS_randombytes((uint8_t *)&random_bit_index, sizeof(i));
```
SWilson4, line 435:
```suggestion
			OQS_randombytes((uint8_t *)&random_bit_index, sizeof(random_bit_index));
```

SWilson4 (commented):
This looks awesome, thanks @rtjk! I spotted one typo in a more thorough look at the bitflip testing logic; I left code suggestions.

To reduce code duplication, is it possible to factor some or all of the bitflip testing functions out into `test_helpers.c`? I apologize for not catching that in my previous review.

rtjk:
> This looks awesome, thanks @rtjk! I spotted one typo in a more thorough look at the bitflip testing logic; I left code suggestions.
> 
> To reduce code duplication, is it possible to factor some or all of the bitflip testing functions out into `test_helpers.c`? I apologize for not catching that in my previous review.

All done :)

SWilson4 (approved):
LGTM, thanks for improving our tests! Thanks also for the bonus refactoring work, it's always appreciated. Feel free to merge once CI passes.

PR #2093, opened 2025-03-03
Update actions/cache to v4.2.2

mkannwischer:
image

Github recently turned off their old caching APIs: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down.
liboqs' CI started failing with the following error message for me: This request has been automatically failed because it uses a deprecated version of `actions/cache: e12d46a63a90f2fae62d114769bbf2a179198b5c`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions.

This commit updates to the newer version of actions/cache that is using the new Github APIs.

Note that Github's blog post is misleading. A lot more versions than v1/v2 are being disabled including v3.3.3 that is used in liboqs - see https://github.com/actions/cache/discussions/1510

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -223,7 +223,7 @@ jobs:
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # pin@v4
       - name: Retrieve OpenSSL330 from cache
         id: cache-openssl330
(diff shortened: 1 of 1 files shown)


baentsch (approved):
Thanks @mkannwischer !

bhess (approved):
Thank you @mkannwischer !

SWilson4 (approved):

PR #2094, opened 2025-03-05
Add UOV 

### LINKED ISSUE #2028: Add UOV sig alg
As per https://github.com/open-quantum-safe/oqs-provider/discussions/605#discussioncomment-11697369

mkannwischer:
This PR adds the UOV signature scheme which is a second-round candidate in the [NIST digital signature competition](https://csrc.nist.gov/projects/pqc-dig-sig/round-2-additional-signatures). 
The implementation is sourced from [pqov](https://github.com/pqov/pqov) and implements the [round 2 specification](https://kannwischer.eu/papers/2025_uov_round2.pdf). 

CI (full tests + extended tests) is passing in my fork of liboqs. 

This PR includes the reference, AVX2, and Neon implementations for all 12 parameter sets. There are additional 12 parameter sets using 4-round AES128 instead of the default 10 rounds. I have not included those yet - if there is interest in them, I'm happy to add them later.

I had to make a few changes to the implementation to make it through CI and I'm porting them back to the main code base here: https://github.com/pqov/pqov/pull/58.
~Once that is merged no further patches are required for the implementation. Once that is merged, I will update this PR and mark it as ready for review.~
The changes have been merged upstream by now.

The import is entirely done via the `copy_from_upstream.py` script. I have integrated the required META files upstream. 

There are two small changes to the liboqs code base: 
1) Allow the `pqov_` namespace (https://github.com/open-quantum-safe/liboqs/pull/2094/commits/377232e5a5d32805ef7e8f150e5d199f662acfce) 
2) Do not rely on SPHINCS+ being the last item in an alphabetically ordered list in `update_docs_from_yaml.py` (https://github.com/open-quantum-safe/liboqs/pull/2094/commits/3fafba1585b94f5016e6e8912a186988c0091b5a)

Resolves #2028.  

* [X] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged.)

This PR was prepared jointly with @potsrevennil.


mkannwischer:
The Travis CI found two interesting problems that I have fixed by now: 
1) Was a `-Waggressive-loop-optimization` warning with gcc 10 or older. gcc11 and later seems to work. I think it is a compiler bug. I refactored the affected code a bit to work around that and funnily that gave quite a speed-up for verification. 

2) There was an endianness assumption in my OQS glue code that's calling AES-CTR. I have fixed that now. 

Unless I broke something in the meantime, CI should be all green now.

zephyr/Kconfig
```diff
@@ -76,4 +76,9 @@ config LIBOQS_ENABLE_SIG_CROSS
 	default y
 	depends on LIBOQS
 
+config LIBOQS_ENABLE_SIG_UOV
+	bool "Enable the MAYO signature algorithm"
```
dstebila, line 80:
```suggestion
	bool "Enable the UOV signature algorithm"
```

dstebila (approved):

dstebila:
> None of the zephyr tests would work for me even for the parameter sets with the smallest memory footprint. I suspect that memory is rather limited in these tests?

Tagging @Frauschi about the Zephyr question.

zephyr/Kconfig
mkannwischer: good catch. Thank you @dstebila!

bhess (commented):
Thank you @mkannwischer for contributing UOV!

One minor thing, we have a config option that builds only the onramp candidates. The UOV variants should be added to this list as well:

https://github.com/open-quantum-safe/liboqs/blob/d4eb7a6d61cfddd20c14a0a7c9c58ccde6d5d8d7/.CMake/alg_support.cmake#L217-L219

The zephyr config is limited to 512K stack size

https://github.com/open-quantum-safe/liboqs/blob/d4eb7a6d61cfddd20c14a0a7c9c58ccde6d5d8d7/zephyr/samples/Signatures/prj.conf#L14

and 256K malloc size

https://github.com/open-quantum-safe/liboqs/blob/d4eb7a6d61cfddd20c14a0a7c9c58ccde6d5d8d7/zephyr/samples/Signatures/prj.conf#L18

I am not certain if these values can be safely increased.

mkannwischer:
> Thank you @mkannwischer for contributing UOV!
> 
> One minor thing, we have a config option that builds only the onramp candidates. The UOV variants should be added to this list as well:
> 
> https://github.com/open-quantum-safe/liboqs/blob/d4eb7a6d61cfddd20c14a0a7c9c58ccde6d5d8d7/.CMake/alg_support.cmake#L217-L219
> 

I added UOV there. 

> The zephyr config is limited to 512K stack size
> 
> https://github.com/open-quantum-safe/liboqs/blob/d4eb7a6d61cfddd20c14a0a7c9c58ccde6d5d8d7/zephyr/samples/Signatures/prj.conf#L14
> 
> and 256K malloc size
> 
> https://github.com/open-quantum-safe/liboqs/blob/d4eb7a6d61cfddd20c14a0a7c9c58ccde6d5d8d7/zephyr/samples/Signatures/prj.conf#L18
> 
> I am not certain if these values can be safely increased.

I think there is no hope that this is going to work with the current implementations. 
The smallest parameter set has 278432-byte public keys and 237896-byte. That already 512 KB. You can compress both, but the current implementations still stores the full expanded keys in memory.
We are working on more stack friendly implementations - but those are not yet ready.

Frauschi:
Regarding the Zephyr problems: The CI tests emulate different Zephyr boards via QEMU. Technically, we can increase the stack and heap sizes for these tests to make them run successfully (as in QEMU, available memory is much larger). 

However, the current values are already quite large considering the actual hardware Zephyr is intended for (smaller microcontrollers). Hence, I am not sure if this would be a real benefit, as UOV (in its current implementation with the memory usage) would never run on a real microcontroller target anyway (as they typically have available RAM in the kilobyte or small megabyte region). So, maybe we disable UOV for Zephyr completely at the moment? This can be revisited once an implementation with less memory usage is available.

mkannwischer:
> Regarding the Zephyr problems: The CI tests emulate different Zephyr boards via QEMU. Technically, we can increase the stack and heap sizes for these tests to make them run successfully (as in QEMU, available memory is much larger).
> 
> However, the current values are already quite large considering the actual hardware Zephyr is intended for (smaller microcontrollers). Hence, I am not sure if this would be a real benefit, as UOV (in its current implementation with the memory usage) would never run on a real microcontroller target anyway (as they typically have available RAM in the kilobyte or small megabyte region). So, maybe we disable UOV for Zephyr completely at the moment? This can be revisited once an implementation with less memory usage is available.

Thanks!
Makes sense to me. That is what is currently implemented, too.

docs/algorithms/sig/uov.md
```diff
@@ -0,0 +1,154 @@
...
+| [Primary Source](#primary-source) | avx2                     | x86\_64                     | Linux,Darwin                    | AVX2                    | True                               | True                                           | False                |
+
+Are implementations chosen based on runtime CPU feature detection? **Yes**.
+
+## OV-III-pkc-skc implementation characteristics
+
+|       Implementation source       | Identifier in upstream   | Supported architecture(s)   | Supported operating system(s)   | CPU extension(s) used   | No branching-on-secrets claimed?   | No branching-on-secrets checked by valgrind?   | Large stack usage?   |
+|:---------------------------------:|:-------------------------|:----------------------------|:--------------------------------|:------------------------|:-----------------------------------|:-----------------------------------------------|:---------------------|
+| [Primary Source](#primary-source) | ref                      | All                         | All                             | None                    | True                               | True                                           | True                 |
```
baentsch, line 136: Is this really correct: Only one single algorithm, and then the plain C code implementation being tagged "Large Stack usage"?

mkannwischer: Good catch. No - this is incorrect. 
None of the implementations has a large stack usage. I had to move all key material to the heap to make it work on x86 MacOS (which I understand has a 512 KB stack limit). 
I had exprimented with this option in the beginning to see if it makes a difference in CI, but later learned this is only for documentation.

Let me fix this.

mkannwischer: Fixed: https://github.com/open-quantum-safe/liboqs/compare/613abe639df29cd82403cd7a8e56e87371143117..56aa83c739fc6d1f30be906e318a8627c69d0d19

dstebila:
We have two approving merges so I think we can go ahead and merge this. However there is a merge conflict in cbom.json; I don't know enough about the CBOM file to know how to resolve this. Can anyone help?

bhess:
> We have two approving merges so I think we can go ahead and merge this. However there is a merge conflict in cbom.json; I don't know enough about the CBOM file to know how to resolve this. Can anyone help?

I'll have a look at the CBOM merge conflict and try to resolve it.

bhess:
> > We have two approving merges so I think we can go ahead and merge this. However there is a merge conflict in cbom.json; I don't know enough about the CBOM file to know how to resolve this. Can anyone help?
> 
> I'll have a look at the CBOM merge conflict and try to resolve it.

Resolved.

docs/algorithms/sig/uov.md
bhess: ok to mark as resolved @baentsch ?

SWilson4 (approved):
LGTM! Thanks for the contribution. Merging now that the CBOM conflict is resolved.

PR #2104, opened 2025-03-18
build: don't search sys/random.h for getentropy

mkroening:
This PR enables proper, portable detection of `getentropy` on non-Linux systems by configuring CMake to only search for `getentropy` in `unistd.h` without also including `sys/random.h`.

`getentropy` through `unistd.h` is [standardized in POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/functions/getentropy.html) and also [documented by Linux](https://man7.org/linux/man-pages/man3/getentropy.3.html).

Linux' man-page says:

> **getentropy**() is also declared in **<sys/random.h>**.  (No feature test macro need be defined to obtain the declaration from that header file.)

On Debian 9 (Stretch, 2017), neither `sys/random.h` nor `unistd.h` define `getentropy`.
Since Debian 10 (Buster, 2019), both `sys/random.h` and `unistd.h` define `getentropy` without requiring `_DEFAULT_SOURCE`.

Alternatively, we could check both `sys/random.h` and `unistd.h`.


dstebila:
Hi Martin! Thanks for the pull request.  

Did the current code trigger a problem in any system?  If so it would be useful to understand that.

I have a question about your comment "Alternatively, we could check both sys/random.h and unistd.h.".  Isn't that's what already in the CMake file?

mkroening:
Awesome! I have rebased and added the DCO.

mkroening:
Some older CMake versions seem not to have `block()`, which I used to constrain the scope of the `_DEFAULT_SOURCE` definition. Since Linux provides the definition through `sys/random.h` anyway, I propose to just remove the `_DEFAULT_SOURCE` definition altogether, which is how it was before, if you agree.

dstebila:
> Some older CMake versions seem not to have `block()`, which I used to constrain the scope of the `_DEFAULT_SOURCE` definition. Since Linux provides the definition through `sys/random.h` anyway, I propose to just remove the `_DEFAULT_SOURCE` definition altogether, which is how it was before, if you agree.

Sure... I am not familiar with what that does anyway, so I am happy to take your judgement here.

mkroening:
Done 👍 

> I am not familiar with what that does anyway.

According to the [Linux man page](https://man7.org/linux/man-pages/man3/getentropy.3.html), `_DEFAULT_SOURCE` should be defined for the getentropy definition to appear in `unistd.h`. This is not required on Debian, though, and the definition is available without a feature test macro through `sys/random.h` anyway.

SWilson4 (approved):
Thanks for the fix!

dstebila (approved):

praveksharma:
Thank you for the work! Merging this now.

PR #2122, opened 2025-04-10
Temporarily disable HQC

dstebila:
Recently a [security flaw in the IND-CCA2 security of HQC](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/Wiu4ZQo3fP8) was announced. A fix from the HQC is forthcoming.  This PR proposes disabling HQC from the default build for the liboqs 0.13.0 release.

Note that copy_from_upstream consistency checks will fail since I have manually overridden a CMakeLists variable to disable the algorithm.

The following text can be included in the liboqs 0.13.0 release notes:

> - HQC: Due to a [recently announced security flaw in the IND-CCA2 security of HQC](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/Wiu4ZQo3fP8), the HQC code in liboqs has been disabled by default. It will be re-enabled in a future release when a revised version of HQC is available.


SWilson4:
> [The PR tests workflow is failing silently on startup](https://github.com/open-quantum-safe/liboqs/actions/workflows/pr.yml)... it looks to me like GitHub Actions has made some recent (as of yet undocumented?) changes that have invalidated our workflow files.

Fix is up here: https://github.com/open-quantum-safe/liboqs/pull/2123

SWilson4 (approved):
LGTM. I think it's OK to merge this before #2123 given that the push tests passed, and any tests that this might cause to fail in the PR run should have failed on push as well.

praveksharma (approved):

bhess (approved):

PR #2123, opened 2025-04-10
Fix PR workflow runs

SWilson4:
PR workflow runs (specifically, the OpenSSF scorecard runs) are failing at startup due to an issue with the GitHub Actions configuration. It looks like GitHub has added a new `models` option that needs to be given a value.

### DIFF
--- a/.github/workflows/commit-to-main.yml
+++ b/.github/workflows/commit-to-main.yml
@@ -15,24 +15,9 @@ jobs:
   scorecard:
     uses: ./.github/workflows/scorecard.yml
     secrets: inherit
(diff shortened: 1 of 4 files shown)


SWilson4:
The PR tests run on this PR, indicating that GitHub was expecting the  `models` option, but it's new enough that actionlint doesn't recognize it. I think this can be avoided by being a little more careful with the permissions set in the scorecard workflow. I will experiment with this; apologies in advance for the noise.

praveksharma (approved):

bhess (approved):

PR #2126, opened 2025-04-24
Update Nix flake inputs

aidenfoxivey:
Updating the Nix flake inputs here as a chore. I don't think there's any urgent security need for it, but I figured it would be good.


SWilson4 (approved):
Thanks for the ongoing maintenance @aidenfoxivey!

Issue #2061, opened 2025-01-29
Further ACVP test vectors for ML-DSA


dstebila:
As noted in https://github.com/open-quantum-safe/liboqs/pull/2051: 

The following ACVP vectors remain untested and will be enabled once upstream changes are available:

- External API Vectors: "pre-hash" implementation.
- Internal API Vectors: With "external mu" implementation.

PR #2131, opened 2025-05-01
Update ACVP vectors to latest release

abhinav-thales:
Updates ACVP vectors for ML-DSA and ML-KEM to 1.1.0.39

New vectors are added for ML-DSA-SignatureGeneration to cover the rejection scenarios during the generation process as well as cases that have rejection count > 64.

As earlier, the TC's related to PreHash and ExternalMu calculation remain untested due to unavailable API support in upstream. Also, captured in https://github.com/open-quantum-safe/liboqs/issues/2061


SWilson4:
Seems like everything is good except one Travis CI glitch (looks like a runner is down). OK to mark as ready for review @abhinav-thales?

bhess (approved):
Thanks for the update @abhinav-thales! LGTM.

SWilson4:
Thanks for the contribution!

PR #2133, opened 2025-05-05
test: Add basic kem fuzz testing

nathaniel-brough:
Adds basic fuzz harness for kem algorithms

 This is part of work towards: #1215


nathaniel-brough:
Looks like the CI is failing due to something unrelated. The disk on the Travis CI seems to be full when fetching build dependencies.

I'll have a closer look over the next week. Unless this is a known issue?

dstebila:
Looks like the CI failure is resolved.

I think the fuzz test itself looks fine. Do we need to code to trigger this to actually run in the CI?

dstebila (approved):

SWilson4 (approved):
LGTM. @nathaniel-brough do you need anything from our end to enable the oss-fuzz integration?

nathaniel-brough:
I don't believe so, but I'll follow up if I do.

Issue #2045, opened 2025-01-21
Create a table of algorithm support levels


dstebila:
Similar to PLATFORMS.md, we should have a table indicating the support levels of the algorithms, supplementing/summarizing the information spread across the algorithm datasheets.

dstebila:
In order to create a table of algorithm support levels, we'll need to have some characteristics that define the different support levels.  Here are some features that come to mind.  Any other ideas? 
 
- degree of standardization
- implementation characteristics
  - verification
  - constant-time
  - platform optimizations
- degree of support from implementation maintainers
- vulnerability response

SWilson4:
In case it helps anyone working on this in the future, I got started on this in this branch: https://github.com/open-quantum-safe/liboqs/compare/sw-alg-datasheet.

xuganyu96:
> In order to create a table of algorithm support levels, we'll need to have some characteristics that define the different support levels. Here are some features that come to mind. Any other ideas?
> 
>     * degree of standardization
> 
>     * implementation characteristics
>       
>       * verification
>       * constant-time
>       * platform optimizations
> 
>     * degree of support from implementation maintainers
> 
>     * vulnerability response

I feel confused about the need for a separate `ALGORITHMS.md`. There is already a set of documents under `docs/kem`, `docs/sig`, and `docs/sig_stfl` that record a lot of the information mentioned above, so it seems like a duplicate?

dstebila:
The primary concern to me is that we record accurately the state of the different algorithms, so users can make informed choices. And then secondary is that we present this information in a helpful location. The thought was that a top-level ALGORITHMS.md file would be more visible (and more concise?) than the quite detailed datasheets under the docs folder. But I also take your point about duplication.

Maybe a first step would be to get that information to the algorithm datasheets (add appropriate fields to the YML files and then update the Markdown generation).  As a second step, if we have the data in the YML files, then we can also have the code generation summarize it into a top-level ALGORITHMS.md file.

baentsch:
The more I think about it, I doubt a separate file is needed:

What about the idea to just update the algorithms documentation in the README, simply stating which algorithm families have upstream maintenance and which ones don't? For all details, pointing to the files in the `docs` folder may be sufficient.
The thing we did in [oqs-demos](https://github.com/open-quantum-safe/oqs-demos) may serve as an example: Support status and maintainer is listed there for each demo, giving users an immediate hint at what they may "get into" when using a demo (algorithm in this case).

[...]

PR #2134, opened 2025-05-05
Continuous Benchmarking using Github Actions

pablo-gf:
This PR aims to provide a new benchmarking approach for liboqs. It uses the [Continuous Benchmarking](https://github.com/marketplace/actions/continuous-benchmark) action from Marketplace, like the [ mlkem-native](https://github.com/pq-code-package/mlkem-native) repository. For speed benchmarking of the current algorithms in the library, 3 new files are included:

- scripts/parse_liboqs_speed.py : Retrieves the benchmarking data from speed_kem and speed_sig and outputs it in a json file that matches the format required by the continuous benchmarking action.

- workflows/kem-bench.yml: Iterates through the different KEM algorithms executing the speed test and gathering its information using parse_liboqs_speed.py. It then pushes the output json file to a gh-pages branch using the Continuous Benchmarking action.

- workflows/sig-bench.yml: Same kem-bench.yml but for signature algorithms.

To complete the benchmarking, it is required to create a new gh-pages branch so that the workflows generate and continuously update a Github page with the visualization of the benchmarking results. I have adapted the html file to include some additional features [here ](https://github.com/pablo-gf/liboqs/blob/gh-pages/dev/bench/index.html). I can include these changes once the new gh-pages branch for liboqs is set up. You can see an example of what the final output should look like [here](https://pablo-gf.github.io/liboqs/dev/bench/).

Let me know if you have any questions or suggestions!

### DIFF
--- a/.github/workflows/kem-bench.yml
+++ b/.github/workflows/kem-bench.yml
@@ -3,7 +3,7 @@ name: kem benchmark
 on:
   workflow_dispatch:
   push:
-    branches: [bench]
+    branches: [main]
 
 permissions:
   deployments: write
--- a/.github/workflows/kem-bench.yml
+++ b/.github/workflows/kem-bench.yml
@@ -1,0 +1,122 @@
+name: kem benchmark
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [bench]
+
+permissions:
+  deployments: write
+  contents: write
+  pull-requests: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      # Checkout repository
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      # Set up dependencies
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cmake ninja-build gcc g++ python3 python3-pip
+          pip3 install py-cpuinfo
+
+      # Build the speed_kem binary only
+      - name: Build speed_kem binary
+        run: |
+          mkdir -p build
+          cd build
+          cmake -GNinja .. -DBUILD_SHARED_LIBS=OFF
+          ninja speed_kem
+
+      # Copy the parse_liboqs_speed.py script
+      - name: Copy parse_liboqs_speed.py
+        run: |
+          cp scripts/parse_liboqs_speed.py build/tests/
+
+      # Upload the built binary and script as an artifact
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: built-binary
+          path: build/tests/
+
+  benchmark:
+    needs: build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix: 
+        algorithm: [ # List of available KEMs to perform the benchmarking on
+          "BIKE-L1",
+          "BIKE-L3", 
+          "BIKE-L5",
+          "Classic-McEliece-348864", 
+          "Classic-McEliece-348864f",
+          "Classic-McEliece-460896", 
+          "Classic-McEliece-460896f",
+          "Classic-McEliece-6688128", 
+          "Classic-McEliece-6688128f",
+          "Classic-McEliece-6960119", 
+          "Classic-McEliece-6960119f",
+          "Classic-McEliece-8192128", 
+          "Classic-McEliece-8192128f",
+          "Kyber512",
+          "Kyber768",
+          "Kyber1024",
+          "ML-KEM-512",
+          "ML-KEM-768",
+          "ML-KEM-1024",
+          "sntrup761",
+          "FrodoKEM-640-AES", 
+          "FrodoKEM-640-SHAKE",
+          "FrodoKEM-976-AES", 
+          "FrodoKEM-976-SHAKE",
+          "FrodoKEM-1344-AES", 
+          "FrodoKEM-1344-SHAKE"
+        ]
+      max-parallel: 1 # No parallel jobs to not compromise the pull-push operations of the benchmarking actions below
+
+    steps:
+      # Ensure the repository is checked out
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      # Download the built binary and script
+      - name: Download artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: built-binary
+          path: build/tests/
+
+      # Set execute permissions for the binary
+      - name: Set execute permissions
+        run: chmod +x build/tests/speed_kem
+
+      # Run speed_kem tests for each algorithm
+      - name: Run speed_kem tests
+        run: |
+          cd build/tests
+          ./speed_kem "${{matrix.algorithm}}" > ${{matrix.algorithm}}_output.txt
+          python3 parse_liboqs_speed.py ${{matrix.algorithm}}_output.txt --algorithm ${{matrix.algorithm}}
+
+      # Push to GitHub pages using continuous-benchmark
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@d48d326b4ca9ba73ca0cd0d59f108f9e02a381c7
+        with:
+          name: ${{matrix.algorithm}}
+          tool: "customSmallerIsBetter"
+          output-file-path: build/tests/${{matrix.algorithm}}_formatted.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          comment-on-alert: true
+          summary-always: true
+          alert-threshold: 50%
+          comment-always: true
--- a/.github/workflows/sig-bench.yml
+++ b/.github/workflows/sig-bench.yml
@@ -1,0 +1,152 @@
+name: sig benchmark
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [bench]
+
+permissions:
+  deployments: write
+  contents: write
+  pull-requests: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      # Checkout repository
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      # Set up dependencies
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cmake ninja-build gcc g++ python3 python3-pip
+          pip3 install py-cpuinfo
+
+      # Build the speed_sig binary only
+      - name: Build speed_sig binary
+        run: |
+          mkdir -p build
+          cd build
+          cmake -GNinja .. -DBUILD_SHARED_LIBS=OFF
+          ninja speed_sig
+
+      # Copy the parse_liboqs_speed.py script
+      - name: Copy parse_liboqs_speed.py
+        run: |
+          cp scripts/parse_liboqs_speed.py build/tests/
+
+      # Upload the built binary and script as an artifact
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: built-sig-binary
+          path: build/tests/
+
+  benchmark:
+    needs: build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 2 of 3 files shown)


.github/workflows/kem-bench.yml
```diff
@@ -0,0 +1,122 @@
+name: kem benchmark
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
```
SWilson4, line 6: I'd like this to be triggered via the [`commit-to-main.yml`](https://github.com/open-quantum-safe/liboqs/blob/main/.github/workflows/commit-to-main.yml#L37-L46)l file so that it fits in with the rest of our CI.

.github/workflows/kem-bench.yml
```diff
@@ -0,0 +1,122 @@
...
+  contents: write
+  pull-requests: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      # Checkout repository
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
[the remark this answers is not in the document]
SWilson4: We'll want to pin this to a specific hash (you can see the other checkout actions in the repo for examples). The same applies for the other occurrences of this message.

.github/workflows/kem-bench.yml
```diff
@@ -0,0 +1,122 @@
...
+    branches: [main]
+
+permissions:
+  deployments: write
+  contents: write
+  pull-requests: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
SWilson4, line 15: Eventually we'll want to have this working on multiple runners. How scalable is the approach taken here?

I walked through the PQCP setup with Matthias and Ry last week; once https://github.com/open-quantum-safe/tsc/pull/180 lands we'll be able to work with a similar setup.

dstebila:
I think it would be good if the build & configuration information that's currently in the expandable "Latest commit build information" is displayed directly at the top of the page.

In the pop-ups that show when you hover over a datapoint, it looks like all the commits have been authored by you. Is that placeholder information? Or should something else be showing here?

dstebila:
> @dstebila I have added the build information at the top, let me know if that works: https://pablo-gf.github.io/liboqs/dev/bench/. As for your second comment, that is placeholder information. The idea is that those pop-ups will contain the details of each commit made to the library starting from the first commit after this continuous benchmarking framework is deployed.

Looks good, thanks!

pablo-gf:
@SWilson4 I fixed the security warnings that popped-up after my last commit. Let me know if you have any comments or suggestions. As I mentioned at the beginning, to make the entire process work we would also need to create a new gh-pages branch so that the workflows generate and continuously update a Github page with the visualization of the benchmarking results.

SWilson4:
Thanks for the updates, @pablo-gf! Are you able to merge this branch into `main` of your fork so we can test the `commit-to-main` flow and see if it's working as expected?

dstebila (approved):

.github/workflows/kem-bench.yml
SWilson4: In the OQS status call yesterday, we agreed to proceed with the default GitHub runner for now so that we can get the site set up. We'll move to AWS runners in the future.

pablo-gf:
Thank you @SWilson4! What's left is to create a branch called gh-pages so that the benchmarking results are posted there.

SWilson4:
> Thank you @SWilson4! What's left is to create a branch called gh-pages so that the benchmarking results are posted there.

Done! I also set up branch protection so that we don't accidentally delete it.

pablo-gf:
> > Thank you @SWilson4! What's left is to create a branch called gh-pages so that the benchmarking results are posted there.
> 
> Done! I also set up branch protection so that we don't accidentally delete it.

Awesome! Should be good to merge now.

SWilson4:
Merging, thanks @pablo-gf for the contribution!

dstebila:
In a [merge today](https://github.com/open-quantum-safe/liboqs/commit/8d9cfd7f33080b2463e145218ec0db47048a2d81#commitcomment-159857632) we got a whole bunch of alerts about speed regression, which I think is related to this continuous benchmarking PR landing. @pablo-gf do you have any idea what's going on here?

pablo-gf:
> In a [merge today](https://github.com/open-quantum-safe/liboqs/commit/8d9cfd7f33080b2463e145218ec0db47048a2d81#commitcomment-159857632) we got a whole bunch of alerts about speed regression, which I think is related to this continuous benchmarking PR landing. @pablo-gf do you have any idea what's going on here?

Yes, @dstebila. There is an option to throw an alert if the algorithm speed decreases a certain value (I believe it's a specific percentage). I can definitely look into that if it's something we are not interested in.

dstebila:
> Yes, @dstebila. There is an option to throw an alert if the algorithm speed decreases a certain value (I believe it's a specific percentage). I can definitely look into that if it's something we are not interested in.

In principle I think we'd be interested in that. But the alerts being thrown in commit I linked to seem to be too sensitive.  Would you be able to check how the thresholds are configured?

pablo-gf:
> > Yes, @dstebila. There is an option to throw an alert if the algorithm speed decreases a certain value (I believe it's a specific percentage). I can definitely look into that if it's something we are not interested in.
> 
> In principle I think we'd be interested in that. But the alerts being thrown in commit I linked to seem to be too sensitive. Would you be able to check how the thresholds are configured?

Yes @dstebila , here is some information from [the page](https://github.com/marketplace/actions/continuous-benchmark):

"This action can raise an alert to the commit when its benchmark results are worse than previous exceeding a specified threshold. By default, this action marks the result as performance regression when it is worse than the previous exceeding 200% threshold. For example, if the previous benchmark result was 100 iter/ns and this time it is 230 iter/ns, it means 230% worse than the previous and an alert will happen. The threshold can be changed by alert-threshold input."

I believe the current parameters are set to `alert-threshold: 50%`. I set it that way to make sure it was working during testing, but let me know the value you would like it to be increased to. I believe the default value in the ml-kem repository is 103%.

SWilson4:
[In the past](https://github.com/open-quantum-safe/liboqs/wiki/Release-process) we deemed 15% as an acceptable variation. I think I would lower it now that we have some stable algorithms in the library—a 14% performance drop might not be cause for alarm for MAYO, but it certainly would be for ML-KEM. How about setting `alert-threshold` to 105% for now? If we find that we're getting a bunch of warnings for more experimental algs, we can raise it further.

pablo-gf:
@SWilson4 Sounds good. Would you like me to create a new PR for that?

SWilson4:
> @SWilson4 Sounds good. Would you like me to create a new PR for that?

That would be great, thanks @pablo-gf.

baentsch:
As a person re-activating GH notifications for the project, I was overwhelmed by the amount of messages this feature just put into my Inbox. So a few questions to @pablo-gf if I may: 
1) Is there anything that can be done to reduce this (amount of notifications)? Say one summary message per merged PR only? Or would I need to locally filter out anything from say "github-actions [bot]"? Doesn't feel "right"... 
2) The 105% for all algorithms seems a sub-optimal choice (see the plethora of messages in https://github.com/open-quantum-safe/liboqs/commit/59bf831b47fb43124f816d594ec801f729b08053#commitcomment-165475327 )
3) Is there anything that can be done to only highlight performance regressions if the code for the specific algorithm has actually been changed? Most of the messages seem like a random (infrastructure-induced?) deviation not worth while looking at -- creating the risk of not looking at _any_ message.
4) Should there be a per-algorithm "trigger setting"? For example, there are algorithms for which there is not exactly a lot of upstream support; highlighting deviations for those (not getting attention/fixes anyway) is pretty pointless, isn't it? Also see discussion at https://github.com/open-quantum-safe/liboqs/issues/2045#issuecomment-3273571216

If none of the optimizations above are possible, may I suggest changing this feature to only run as part of the weekly tests and only for actively supported algorithms?

pablo-gf:
Hello @baentsch! Here are some answers to your questions:

1. I don't think the functionality built on continuous-benchmarking allows to make single messages summarizing the performance of all algorithms.

2. Yes, that is definetely a lot of comments...we can increase the threshold as you see fit. We could also add a specific threshold for each algorithm in case, we would just need to figure out what those thresholds will be.

3. As in point 1, I don't think there is a functionality available for that. However, I thought the idea was to have all algorithms tested continuously in case certain changes in the codebase affect all algorithms (even those whose code has not been changed).

4. I believe we can fix this by specifying the algorithms that we would like to run the continuous benchmarking tests on within the algorithm matrix here: https://github.com/open-quantum-safe/liboqs/blob/59bf831b47fb43124f816d594ec801f729b08053/.github/workflows/kem-bench.yml#L53C7-L81C10. Same thing would apply to signatures.

Still, if this causes too much of a bother, it might be ideal to run the continuous benchmarking in the weekly tests instead as you suggest. Let me know and I can get to it.

Apart from all this, I wanted to seek advice on something: If a new algorithm is added and we want the continuous benchmarking tests to run on it as well, it should be included in the algorithm matrix inside the kem-bench.yml or the sig-bench.yml files. Is there a way to make that a requirement whenever new algorithms are included into liboqs?

baentsch:
Thanks for the responses @pablo-gf ! Looks like option 4 is the most reliable one (assuming my thinking is correct that the deviations reported in recent runs are indeed unexplicable by code changes - TBC: @dstebila: As you merged all of this can we safely assume you did such checks and these are all indeed false alerts?).

>  If a new algorithm is added and we want the continuous benchmarking tests to run on it as well, it should be included in the algorithm matrix inside the kem-bench.yml or the sig-bench.yml files. Is there a way to make that a requirement whenever new algorithms are included into liboqs?

"requirement" is imo too strong a word -- I'd call it a (good) advice. And that (wording) could be added e.g., [here](https://github.com/open-quantum-safe/liboqs/blob/main/CONTRIBUTING.md) and [here](https://github.com/open-quantum-safe/liboqs/blob/main/.github/pull_request_template.md)...

dstebila:
> Thanks for the responses @pablo-gf ! Looks like option 4 is the most reliable one (assuming my thinking is correct that the deviations reported in recent runs are indeed unexplicable by code changes - TBC: @dstebila: As you merged all of this can we safely assume you did such checks and these are all indeed false alerts?).

I do look at them, and if they are above the threshold that triggers the alert but still not too big (e.g., 1.08 instead of 1.05), I generally chalk that down to noise / typical variation, and proceed.

pablo-gf:
> Thanks for the responses @pablo-gf ! Looks like option 4 is the most reliable one (assuming my thinking is correct that the deviations reported in recent runs are indeed unexplicable by code changes - TBC: @dstebila: As you merged all of this can we safely assume you did such checks and these are all indeed false alerts?).
> 
> > If a new algorithm is added and we want the continuous benchmarking tests to run on it as well, it should be included in the algorithm matrix inside the kem-bench.yml or the sig-bench.yml files. Is there a way to make that a requirement whenever new algorithms are included into liboqs?
> 
> "requirement" is imo too strong a word -- I'd call it a (good) advice. And that (wording) could be added e.g., [here](https://github.com/open-quantum-safe/liboqs/blob/main/CONTRIBUTING.md) and [here](https://github.com/open-quantum-safe/liboqs/blob/main/.github/pull_request_template.md)...

Sounds good. Should I update the algorithms matrices with those set as "selected" in the README, then? Also, let me know if you would like me to increase the threshold in case you find the current set up to trigger it too often.

I will include the advice to add new algorithms into the continuous benchmarking platform matrices in the pull_request_template.md file as well.

baentsch:
> I do look at them, and if they are above the threshold that triggers the alert but still not too big (e.g., 1.08 instead of 1.05), I generally chalk that down to noise / typical variation, and proceed.

That does make sense. But where do you draw the line? In the run I referenced above there's also 1.21, 1.17 and 1.12 deviations: Can 20% still be considered noise? I'm not sure. That's also why I'm hard-pressed to answer 

> Also, let me know if you would like me to increase the threshold in case you find the current set up to trigger it too often.

So until we understand what triggers these deviations, or even can remove them (multiple runs, maybe?), I think my suggestion would be to move this to weekly tests -- and look at those results more regularly/carefully instead of not looking at too many messages in every commit/PR we (get used to) consider false alarms.

Issue #2110, opened 2025-03-21
Remove mlkem-native keypair_derand patch


SWilson4:
Whenever we upgrade the mlkem-native implementation to a release that includes https://github.com/pq-code-package/mlkem-native/pull/886, we should remove the [patch file](https://github.com/open-quantum-safe/liboqs/blob/3ca1a36/scripts/copy_from_upstream/patches/mlkem-native-ml_kem-derand.patch) that enables integration of the derandomized keygen API.

PR #2146, opened 2025-05-29
Update mlkem-native to v1.0.0

mkannwischer:
This PR updates mlkem-native to the first stable release v1.0.0 that was released today. 
The release notes can be found [here](https://github.com/pq-code-package/mlkem-native/releases/tag/v1.0.0).

This also removes a patch that was needed for an older version of mlkem-native. Resolves https://github.com/open-quantum-safe/liboqs/issues/2110


mkannwischer:
CI is looking good. The only failure is the macos-15 failure that's being addressed in #2157.

praveksharma (approved):

PR #2153, opened 2025-06-02
tests: Check OQS_STATUS of RNG and fstore functions

Hussain1811:
Check whether OQS_randombytes_switch_algorithm()
and oqs_fstore_init() return OQS_SUCCESS
before proceeding with the next code block.

### DIFF
--- a/tests/speed_kem.c
+++ b/tests/speed_kem.c
@@ -116,9 +116,16 @@ int main(int argc, char **argv) {
 
 	OQS_KEM *single_kem = NULL;
 
-	OQS_randombytes_switch_algorithm(OQS_RAND_alg_openssl);
-
 	OQS_init();
+#ifdef OQS_USE_OPENSSL
+	rc = OQS_randombytes_switch_algorithm(OQS_RAND_alg_openssl);
+	if (rc != OQS_SUCCESS) {
+		printf("Could not generate random data with OpenSSL RNG\n");
+		OQS_destroy();
+		return EXIT_FAILURE;
+	}
+#endif
+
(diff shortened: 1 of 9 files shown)


Hussain1811:
`oqs_fstore_init` seems to be returning `OQS_ERROR`
is this expected?

bhess:
My guess is that `mkdir` fails because the tmp folder already exists. This could be handled gracefully by the code, and still return an OQS success code.

dstebila (approved):

Hussain1811:
Some mac-os, linux-arm-emulated, linux platform tests fail as `OQS_randombytes_switch_algorithm` returns `OQS_ERROR`. Looks like this is because `OQS_USE_OPENSSL` is set to `OFF` in these tests.
Is this expected? The documentation says it should be default ON (https://github.com/open-quantum-safe/liboqs/blob/main/CONFIGURE.md#oqs_use_openssl)

tests/speed_kem.c
```diff
@@ -116,9 +116,14 @@ int main(int argc, char **argv) {
 
 	OQS_KEM *single_kem = NULL;
 
-	OQS_randombytes_switch_algorithm(OQS_RAND_alg_openssl);
-
 	OQS_init();
+	rc = OQS_randombytes_switch_algorithm(OQS_RAND_alg_openssl);
```
bhess, line 121: I think this section should be guarded with an `#ifdef USE_OQS_OPENSSL` to avoid it failing.

bhess (approved):
LGTM

PR #2162, opened 2025-06-10
Adjust constant-time test exception for mlkem-native

### LINKED ISSUE #2161: Constant-time tests failing for ML-KEM-1024
See https://github.com/open-quantum-safe/liboqs/actions/runs/15513187236. Perhaps suppression files need to be updated after https://github.com/open-quantum-safe/liboqs/pull/2146?

Tagging @mkannwischer and @bhess.

mkannwischer:
- Resolves #2161 

The constant-time tests have been failing for mlkem-native as check_sk changed it's name and is no no longer recognized as an exception. This function processes exclusively public data in the secret key and it's, hence, okay to branch both inside the function and depending on the return value.

This commit renames the function in the constant_time exceptions file.

### DIFF
--- a/tests/constant_time/kem/passes/ml_kem
+++ b/tests/constant_time/kem/passes/ml_kem
@@ -1,14 +1,14 @@
 {
    The parts of sk being hashed and compared here are public.
    Memcheck:Cond
-   fun:PQCP_MLKEM_NATIVE_MLKEM*_check_sk
+   fun:mlk_check_sk
    fun:PQCP_MLKEM_NATIVE_MLKEM*_dec
 }
 {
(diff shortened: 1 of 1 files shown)


mkannwischer:
~How do I trigger the constant-time tests again? I thought adding "[extended tests]" to the commit message will do, but that does not seem to be the case?~

Nevermind - it was correct, but it was executing in my fork. The CT tests are passing: https://github.com/mkannwischer/liboqs/actions/runs/15550155982/job/43779406159

bhess (approved):
Thanks for the fix @mkannwischer !

PR #2164, opened 2025-06-12
Adjust HTML file for the continuous benchamarking page

pablo-gf:
This PR is a continuation of #2134, which aims to adjust some details of the generated html file so the layout shows as in https://pablo-gf.github.io/liboqs/dev/bench/ . I could not include these changes to the previous PR because the file was not yet generated.

### DIFF
--- a/dev/bench/index.html
+++ b/dev/bench/index.html
@@ -82,12 +82,23 @@
       .benchmark-chart {
         max-width: 1000px;
       }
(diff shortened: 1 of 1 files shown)


pablo-gf:
> LGTM. Just confirming that this is intended to target gh-pages instead of main?

Yes, that is correct! You can go ahead and merge if it looks good, then.

PR #2168, opened 2025-06-16
Benchmarking comments only on alerts

dstebila:
Reduce the number of comments posted/emailed by benchmark regression testing.

### DIFF
--- a/.github/workflows/kem-bench.yml
+++ b/.github/workflows/kem-bench.yml
@@ -118,4 +118,4 @@ jobs:
           comment-on-alert: true
           summary-always: true
           alert-threshold: 105%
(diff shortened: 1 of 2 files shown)


dstebila:
@pablo-gf Is it okay to change this setting to reduce the number of comments posted? Otherwise I get 50+ emails for every benchmarked commit.

pablo-gf:
@dstebila Of course! No problem. These were all features I was trying during testing, but we can adjust them as needed.

SWilson4:
@open-quantum-safe/liboqs-committers Can we please get another review?

Martyrshot (approved):

PR #2172, opened 2025-06-20
update ACVP to 1.1.0.40

abhinav-thales:
WIP...

Newly introduced Test groups tgId-7 to tgId-12(kem encaps-decaps) serve as basic sanity checks and do not involve any inputs or outputs beyond the respective key.
The functions mlk_check_sk and mlk_check_pk are static in kem.c and not exposed for direct use. The current PR changes ATM does not cover key-check validation TCs.

TODO:
how to add sanity checks TCs


abhinav-thales:
The latest ACVP release (v1.1.0.40) introduces test cases for encapsulation and decapsulation key sanity checks, aligning with the mandatory requirements outlined in Section 7 of the FIPS-203 specification.

However, these new vectors only validate the success or failure of key checks (i.e., true/false), without providing the full set of encapsulation/decapsulation parameters. This makes it challenging to test these cases.

Two potential approaches:
1. Proceed with the latest ACVP vectors but skip the key check test cases for now (as done in the current PR). These can be added later if the relevant key sanity check APIs in ml-kem-native are exposed publicly (currently they are static).
2. Enable testing of **only** corrupted key scenarios by using dummy values for other parameters and relying on the OQS API to detect failures. This would require removing the current key sanity checks in vector_kem.c that were implemented prior to ml-kem-native support.
	a. for decapsulation key check this approach is ok
	b. for encapsulation key check, it seems that invalid key TCs have incorrect key length that would get caught in the sanity check.
		    eg. [invalid encap key](https://github.com/usnistgov/ACVP-Server/blob/master/gen-val/json-files/ML-KEM-encapDecap-FIPS203/internalProjection.json#L1107C18-L1107C2450) for ML-KEM-512 has length of 1216 bytes


What do you suggest @bhess @SWilson4   ? or any other options ?

abhinav-thales:
@SWilson4 are we planning to merge PRs after 0.14.0 ?

dstebila:
> @SWilson4 are we planning to merge PRs after 0.14.0 ?

Yes. I plan to do the release of 0.14.0 after next week's status meeting, barring any objections.

dstebila (approved):

bhess (approved):

abhinav-thales:
@bhess see your approval. so we are ok with approach 1 then mentioned [here](https://github.com/open-quantum-safe/liboqs/pull/2172#issuecomment-2995774988) ?

if yes, I guess its good to create a task for handling the sanity key check TCs later when key check APIs are global/external ?

bhess:
> @bhess see your approval. so we are ok with approach 1 then mentioned [here](https://github.com/open-quantum-safe/liboqs/pull/2172#issuecomment-2995774988) ?
> 
> if yes, I guess its good to create a task for handling the sanity key check TCs later when key check APIs are global/external ?

I agree.

abhinav-thales:
created task for pending tasks:

https://github.com/open-quantum-safe/liboqs/issues/2197

Issue #2178, opened 2025-06-25
Use OQS_MEM_cleanse instead of memset in KEM and signature code


SWilson4:
We should make sure that OQS_MEM_cleanse is being used instead of memset whenever secret objects are zeroed.

Related issues: #1864 and #2163.

PR #2186, opened 2025-07-01
Replace memset calls with OQS_MEM_cleanse

aidenfoxivey:
Should help sort out #2178 w.r.t. Classic McEliece. I just used a bit of `sed`.

### DIFF
--- a/scripts/copy_from_upstream/copy_from_upstream.yml
+++ b/scripts/copy_from_upstream/copy_from_upstream.yml
@@ -9,7 +9,7 @@ upstreams:
     kem_scheme_path: 'crypto_kem/{pqclean_scheme}'
     sig_scheme_path: 'crypto_sign/{pqclean_scheme}'
     patches: [pqclean-dilithium-arm-randomized-signing.patch, pqclean-kyber-armneon-shake-fixes.patch, pqclean-kyber-armneon-768-1024-fixes.patch, pqclean-kyber-armneon-variable-timing-fix.patch,
-    pqclean-kyber-armneon-asan.patch]
+    pqclean-kyber-armneon-asan.patch, pqclean-classic-mceliese.patch]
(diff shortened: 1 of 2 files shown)


aidenfoxivey:
Hmm, this seems to indicate that the upstreams are different - can we safely change them then?

praveksharma:
Changes made to code need to be preserved after each run of `scripts/copy_from_upstream/copy_from_upstream.py`, here they're being overwritten. You'll need to add a patch file to `scripts/copy_from_upstream/patches` and add said patch file to relevant upstream in `scripts/copy_from_upstream/copy_from_upstream.yml` so `copy_from_upstream.py` knows to find it.

Also, have you considered using a C preprocessor macro to "define" `memset` to `OQS_MEM_cleanse`? Doing this would result in a smaller patch file but I don't know if it would work.

aidenfoxivey:
> Also, have you considered using a C preprocessor macro to "define" memset to OQS_MEM_cleanse? Doing this would result in a smaller patch file but I don't know if it would work.

Hmm, that's an interesting idea. I have a gut feeling to prefer the patch file being just replacing instances of `memset` with `OQS_MEM_cleanse` out of avoiding the C preprocessor, but would be happy to give that a go if you think it'd be cleaner.

aidenfoxivey:
Seems to be `error: src/kem/classic_mceliece/pqclean_mceliece348864_avx2/controlbits.c: patch does not apply` - I think maybe I'm trying to patch on the wrong revision?

aidenfoxivey:
oh jeez this doesn't seem to work as well - I'll opt for the #define approach

aidenfoxivey:
hopefully this should work 🤞🏻

praveksharma:
Sorry, I should have been more explicit: copy_from_upstream.py patches the [PQClean repository it pulls McEliece from](https://github.com/open-quantum-safe/liboqs/blob/b5d3dac4ebdfbbc5de1f6ab9fc6a94c3fd47d13d/scripts/copy_from_upstream/copy_from_upstream.yml#L98) before copying the patched code to its final location. So the patch file you need to add needs to be relative to the PQClean repo directory structure and not relative to liboqs directory structure.

[This](https://github.com/open-quantum-safe/liboqs/blob/b5d3dac4ebdfbbc5de1f6ab9fc6a94c3fd47d13d/scripts/copy_from_upstream/copy_from_upstream.yml#L15) is the relevant upstream repo.

PR #2195, opened 2025-07-07
Replace memset calls with OQS_MEM_cleanse in Classic McEliese

aidenfoxivey:
rg --files-with-matches "memset" | xargs gsed -i 's/memset/OQS_MEM_cleanse/g'

- **Add OQS_MEM_aligned_secure_free convenience fn**
- **Rewrite SHA3 aligned frees to zeroize**
- **Add Classic McEliece memory sanitization patch**


aidenfoxivey:
Hmm I seem to have dragged some other commits along with this one.

Issue #2159, opened 2025-06-04
Reenable gcc-14 on macos-15


dstebila:
Undo https://github.com/open-quantum-safe/liboqs/pull/2157 once CI environment has been fixed externally.

PR #2201, opened 2025-07-15
Revert gcc-13 to gcc-14 for MacOS CI

aidenfoxivey:
Closes https://github.com/open-quantum-safe/liboqs/issues/2159


aidenfoxivey:
welp seems it’s still busted

PR #2206, opened 2025-07-17
Fix #2057 and some OpenSSL 3.x compatibility related issues

### LINKED ISSUE #2057: opensslconf.h not used
Liboqs uses openssl flags like `OPENSSL_NO_STDIO`, but does not use openssl's `opensslconf.h` used for publicizing these flags. In `src/common/common.h` or other files something like this could be added:
```
#ifdef OQS_USE_OPENSSL
#include <openssl/opensslconf.h>
#endif
```

erendemirel:
This PR is to resolve the issue #2057 alongside with some related issues I came across.

Issues Fixed:

1. **Issue** #2057 - Missing opensslconf.h Include

2. **OpenSSL 3.x Naming Conflicts**- OpenSSL 3.x defines OSSL_FUNC as a type, conflicting with liboqs's macro definition:

Renamed OSSL_FUNC to OQS_OSSL_FUNC across the codebase.

3. **UOV Algorithm OpenSSL Integration**- UOV signature algorithms were calling OpenSSL functions directly instead of using liboqs abstraction, causing undefined symbols (EVP_MD_CTX_free, EVP_MD_CTX_new) in CI builds:

Updated all UOV config.h files to change fallback from #define _UTILS_OPENSSL_ to #define _UTILS_OQS_, ensuring UOV algorithms use liboqs hash abstraction.
Fixed conditional compilation logic in all UOV utils_hash.c files to explicitly prevent OpensSL usage when _UTILS_OQS_ is defined:
```
  // Changed from:
  #if defined(_UTILS_OPENSSL_)||defined(_UTILS_SUPERCOP_)
  // To:
  #if (defined(_UTILS_OPENSSL_)||defined(_UTILS_SUPERCOP_)) && !defined(_UTILS_OQS_)
```
  
Note: Also wrapped any remaining direct OpenSSL calls with OQS_OSSL_FUNC() to ensure complete symbol isolation.
  
4. **Deprecated OpenSSL Function Calls**- UOV algorithms used deprecated OpenSSL functions removed in OpenSSL 3.x
Updated function calls to resolves CI symbol visibility check failures:
```
EVP_MD_CTX_create() -> EVP_MD_CTX_new()
EVP_MD_CTX_destroy() -> EVP_MD_CTX_free()

```
Non-OpenSSL builds compile successfully.
Functional tests pass.
I expect no breaking changes to public API.

**Note**: This is my first contribution to the repo, so a thorough review might be needed.

No additional tests needed, this PR fixes existing functionality rather than adding new features.

No
No

### DIFF
--- a/src/common/common.c
+++ b/src/common/common.c
@@ -427,7 +427,7 @@ OQS_API void *OQS_MEM_malloc(size_t size) {
 OQS_API void *OQS_MEM_calloc(size_t num_elements, size_t element_size) {
 #if defined(OQS_USE_OPENSSL)
 	return OQS_OSSL_FUNC(CRYPTO_zalloc)(num_elements * element_size,
-	                                OPENSSL_FILE, OPENSSL_LINE);
+	                                    OPENSSL_FILE, OPENSSL_LINE);
 #else
 	return calloc(num_elements, element_size); // IGNORE memory-check
 #endif


erendemirel:
There is a formatting issue, I will have a look later

aidenfoxivey:
CleanShot 2025-07-21 at 20 44 39@2x
It seems that the travis CI build failed here.

Regarding the DCO, the CI requires having `--signoff` run for each commit. The easiest approach is to [squash your commits together](https://stackoverflow.com/questions/5189560/how-do-i-squash-my-last-n-commits-together) and then commit with `--signoff`.

erendemirel:
The CI failure is due to a formatting issue in src/common/common.c. I am aware of the DCO and will update it once the build succeeds.
I retriggered CI to test the issue.

erendemirel:
The PR is now ready to be reviewed

src/common/common.c
```diff
@@ -418,24 +418,24 @@ void OQS_MEM_aligned_secure_free(void *ptr, size_t len) {
...
 	return malloc(size); // IGNORE memory-check
 #endif
 }
 
 OQS_API void *OQS_MEM_calloc(size_t num_elements, size_t element_size) {
 #if defined(OQS_USE_OPENSSL)
-	return OSSL_FUNC(CRYPTO_zalloc)(num_elements * element_size,
-	                                OPENSSL_FILE, OPENSSL_LINE);
+	return OQS_OSSL_FUNC(CRYPTO_zalloc)(num_elements * element_size,
+	                                    OPENSSL_FILE, OPENSSL_LINE);
```
aidenfoxivey, line 430: Looks like an extra tab was added here - is that by intention?

src/common/ossl_helpers.h
```diff
@@ -6,60 +6,72 @@
 extern "C" {
 #endif
 
+#include <oqs/oqs.h>
+
+#if defined(OQS_USE_OPENSSL) || defined(OQS_USE_AES_OPENSSL) || defined(OQS_USE_SHA2_OPENSSL) || defined(OQS_USE_SHA3_OPENSSL)
+
+#include <stdio.h>
```
aidenfoxivey, line 13: Is stdio.h required here?

src/common/ossl_helpers.h
```diff
@@ -6,60 +6,72 @@
...
 
-#define OSSL_FUNC(name) name
+#if defined(OQS_DLOPEN_OPENSSL)
+#define FUNC(ret, name, args, cargs) ret _oqs_ossl_##name args;
+#define VOID_FUNC FUNC
+#include "ossl_functions.h"
+#undef VOID_FUNC
+#undef FUNC
 
+#define OQS_OSSL_FUNC(name) _oqs_ossl_##name
```
aidenfoxivey, line 66: Sharp macro! :)

src/common/sha2/sha2_ossl.c
```diff
@@ -52,105 +54,108 @@ static void SHA2_sha256_inc_init(OQS_SHA2_sha256_ctx *state) {
...
 	unsigned int md_len;
 	if (inlen > 0) {
-		OQS_OPENSSL_GUARD(OSSL_FUNC(EVP_DigestUpdate)((EVP_MD_CTX *) state->ctx, in, inlen));
+		OQS_OPENSSL_GUARD(OQS_OSSL_FUNC(EVP_DigestUpdate)((EVP_MD_CTX *) state->ctx, in, inlen));
 	}
-	OQS_OPENSSL_GUARD(OSSL_FUNC(EVP_DigestFinal_ex)((EVP_MD_CTX *) state->ctx, out, &md_len));
-	OSSL_FUNC(EVP_MD_CTX_free)((EVP_MD_CTX *) state->ctx);
+	OQS_OPENSSL_GUARD(OQS_OSSL_FUNC(EVP_DigestFinal_ex)((EVP_MD_CTX *) state->ctx, out, &md_len));
+	OQS_OSSL_FUNC(EVP_MD_CTX_free)((EVP_MD_CTX *) state->ctx);
+	state->ctx = NULL;
```
aidenfoxivey, line 148: Good choice adding this.

src/sig/uov/pqov_ov_III_avx2/utils_hash.c
```diff
@@ -7,13 +7,14 @@
...
-#if defined(_UTILS_OPENSSL_)||defined(_UTILS_SUPERCOP_)
+#if (defined(_UTILS_OPENSSL_)||defined(_UTILS_SUPERCOP_)) && !defined(_UTILS_OQS_)
 
 #include <openssl/evp.h>
+#include "../../../common/ossl_helpers.h"
 
 
 int hash_init( hash_ctx *ctx ) {
-    ctx->x = EVP_MD_CTX_create();
+    ctx->x = OQS_OSSL_FUNC(EVP_MD_CTX_new)();
```
aidenfoxivey, line 17: Are there any subtle differences between `EVP_MD_CTX_create` and `EVP_MD_CTX_new` w.r.t. their API and function?

src/sig/uov/pqov_ov_III_avx2/utils_hash.c
```diff
@@ -43,7 +44,7 @@ int hash_ctx_copy( hash_ctx *nctx, const hash_ctx *octx ) {
 
 int hash_final_digest( unsigned char *out, size_t outlen, hash_ctx *ctx ) {
     int ok = EVP_DigestFinalXOF(ctx->x, out, outlen);
-    EVP_MD_CTX_destroy(ctx->x);
+    OQS_OSSL_FUNC(EVP_MD_CTX_free)(ctx->x);
```
aidenfoxivey, line 47: Same question as above - is free different from destroy significantly?

erendemirel:
Thanks Aiden, I will review your comments and apply the required changes.

aidenfoxivey:
Wonderful! Thank you for all your hard work here. :)

src/common/ossl_helpers.h
erendemirel: I think so. The OpenSSL wrapper code extensively uses OQS_EXIT_IF_NULLPTR and OQS_OPENSSL_GUARD macros. These macros contain fprintf(stderr, ...) calls

src/sig/uov/pqov_ov_III_avx2/utils_hash.c
erendemirel: No functional difference. It is just for naming consistency with the newer releases of OpenSSL. I can return them back if you wish.

src/sig/uov/pqov_ov_III_avx2/utils_hash.c
erendemirel: No functional difference. free() explicitly releases memory, while destroy() cleans up state but might not release memory directly. Again for compatibility and consistency with the new releases of OpenSSL, and again I can return back the old ones if you wish

src/common/common.c
erendemirel: It wasn't intentional. Updated it. Thank you

erendemirel: Reverted it as it fails astyle test on CI

src/common/common.h
```diff
@@ -44,6 +44,11 @@ extern "C" {
  * handling strategy is developed.
  */
 #ifdef OQS_USE_OPENSSL
+#if defined(__has_include)
+#if __has_include(<openssl/opensslconf.h>)
```
dstebila, line 48: I'm not familiar with `__has_include`, can you explain what this is doing?

src/common/ossl_helpers.c
```diff
@@ -114,7 +118,7 @@ const EVP_MD *oqs_sha256(void) {
 #endif
 	return sha256_ptr;
 #else
-	return OSSL_FUNC(EVP_sha256)();
+	return EVP_sha256();
```
dstebila, line 121: Why is the macro being removed here? And similarly on the rest of the diff in this file?

src/common/common.h
aidenfoxivey: Appears to be a GNUism https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html.

dstebila (changes requested):
For the changes in the UOV files, unfortunately changing them directly in the UOV files causes problems with our mechanism for importing the UOV code from the upstream reference implementation. This is what's causing the [CI failure](https://github.com/open-quantum-safe/liboqs/actions/runs/16534077411/job/46785334723). There are two possible solutions: make the changes back in the original codebase (probably not right for this situation), or make a patchfile in liboqs that will make these changes when the code gets copied in from liboqs. Since you've already made the changes directly in liboqs, it should be relatively straightforward: from your current code, create a patch file that represents the change compared to the previous version of UOV, and then add that patch file to our copy_from_upstream mechanism.

mkannwischer:
> For the changes in the UOV files, unfortunately changing them directly in the UOV files causes problems with our mechanism for importing the UOV code from the upstream reference implementation. This is what's causing the [CI failure](https://github.com/open-quantum-safe/liboqs/actions/runs/16534077411/job/46785334723). There are two possible solutions: make the changes back in the original codebase (probably not right for this situation), or make a patchfile in liboqs that will make these changes when the code gets copied in from liboqs. Since you've already made the changes directly in liboqs, it should be relatively straightforward: from your current code, create a patch file that represents the change compared to the previous version of UOV, and then add that patch file to our copy_from_upstream mechanism.

I'd very much prefer if we can get this solved without patching in liboqs.
I'll happily merge changes upstream as long as they do not change the behavior if _UTILS_OQS_ is not defined.
From looking at the code, that will need a bit of rework by adding some `#if defined(_UTILS_OQS_)` branches with some minor code-duplication, but it is still the cleaner solution.

erendemirel:
Taking a look at your review

erendemirel:
@mkannwischer I raised a PR based on your feedback, please take a look: https://github.com/pqov/pqov/pull/61
@dstebila I will proceed with the  changes you requested once pqov is merged

mkannwischer:
> @mkannwischer I raised a PR based on your feedback, please take a look: [pqov/pqov#61](https://github.com/pqov/pqov/pull/61) @dstebila I will proceed with the changes you requested once pqov is merged

Thanks! Left some comments in the PR. This is not the correct fix.

mkannwischer:
@erendemirel, you write: 
> UOV Algorithm OpenSSL Integration- UOV signature algorithms were calling OpenSSL functions directly instead of using liboqs abstraction, causing undefined symbols (EVP_MD_CTX_free, EVP_MD_CTX_new) in CI builds:

I don't think this is true. If `_UTILS_OQS_` is set (which it should be in LIBOQS), only `OQS_SHA3_shake128/256_*` are used - no direct calls to OpenSSL should be present. Can you please elaborate what the problem is we need to fix here? It seems more likely that `_UTILS_OQS_` is not correctly set in some places.

@dstebila @bhess @ashman-p , we have a proposed change upstream: https://github.com/pqov/pqov/pull/61

This is basically removing the calls to the `OQS_SHA3_shake128/256_*` and replacing it with direct calls to `OQS_OSSL_*`. I don't understand why this would be necessary, so I am objecting this change.

erendemirel:
> @erendemirel, you write:
> 
> > UOV Algorithm OpenSSL Integration- UOV signature algorithms were calling OpenSSL functions directly instead of using liboqs abstraction, causing undefined symbols (EVP_MD_CTX_free, EVP_MD_CTX_new) in CI builds:
> 
> I don't think this is true. If `_UTILS_OQS_` is set (which it should be in LIBOQS), only `OQS_SHA3_shake128/256_*` are used - no direct calls to OpenSSL should be present. Can you please elaborate what the problem is we need to fix here? It seems more likely that `_UTILS_OQS_` is not correctly set in some places.
> 
> @dstebila @bhess @ashman-p , we have a proposed change upstream: [pqov/pqov#61](https://github.com/pqov/pqov/pull/61)
> 
> This is basically removing the calls to the `OQS_SHA3_shake128/256_*` and replacing it with direct calls to `OQS_OSSL_*`. I don't understand why this would be necessary, so I am objecting this change.

You're absolutely right. After reviewing the architecture more carefully, I understand that:

1.  _UTILS_OQS_ should use OQS_SHA3_* functions, not OpenSSL functions.
2. My claim that  "undefined symbol" issue was based on a misunderstanding of how the conditional compilation works
3. When UTILS_OQS is properly set, UOV correctly uses liboqs SHA3 functions as intended

And in addition to these, after some further research on OpenSSL, I realized that:

4. My claim that "OpenSSL 3.x defines OSSL_FUNC as a type" in the PR description is incorrect, OSSL_FUNC is not defined as a type or a macro in OpenSSL 3.x, instead, OpenSSL uses a pattern where function types are defined with the suffix _fn and helper functions are used to retrieve function pointers from OSSL_DISPATCH elements: https://docs.openssl.org/3.3/man7/provider-base/#description.

Therefore, I'll focus on the only legitimate issues, missing opensslconf.h includes and OpenSSL 3.x compatibility. For that, I will apply below corrections:

**In pqov repo**: 
- I'll update the pqov PR to only fix the deprecated OpenSSL functions (EVP_MD_CTX_create/destroy → EVP_MD_CTX_new/free) (but only if you want to have OpenSSl 3.x compatibility change, so let me know)

**In liboqs repo:** 
- Restore consistent OSSL_FUNC usage in ossl_helpers.c (undo direct calls without OSSL_FUNC that I incorrectly removed thinking it is causing astyle tests to fail whereas the problem was with broken UOV import mechanism), I mean this: https://github.com/open-quantum-safe/liboqs/pull/2206#discussion_r2244081838
- I'll update this PR description to remove the incorrect UOV integration claims
- I'll update this PR description to remove the incorrect "OpenSSL 3.x defines OSSL_FUNC as a type" claim

### Questions:
1. How do above corrections(In pqov and liboqs repos) sound to you?
2. Should I keep OQS_ renaming changes(as a good practice to prevent potential future naming conflicts with future OpenSSL vesrions) or should I revert them? A rename turns out to be not necessary(see point 4 above)
3. Should I try breaking down OpenSSL 3.x compatibility into another PR and make this one focus only on the original issue? Let me know if you have any preferences.

@mkannwischer @dstebila @bhess @ashman-p

src/common/common.h
erendemirel: I believe this is the most robust and reliable solution here, the only issue I can think of might be that not all compilers will support it. I can try removing it and finding another way if you didn't like it. 
It is a C preprocessor feature that checks if a header file exists before including it. This prevents compilation errors on systems where opensslconf.h might not be available. The double check ensures compatibility.

src/common/ossl_helpers.c
erendemirel: It is a mistake, I will revert them

dstebila:
>     3. Should I try breaking down OpenSSL 3.x compatibility into another PR and make this one focus only on the original issue? Let me know if you have any preferences.

Let's focus on the original issue in this PR and handle any OpenSSL 3.x compatibility issues in a second PR.

Also, can you please disclose whether you are using AI tools to generate code and/or comments?

erendemirel:
> > ```
> > 3. Should I try breaking down OpenSSL 3.x compatibility into another PR and make this one focus only on the original issue? Let me know if you have any preferences.
> > ```
> 
> Let's focus on the original issue in this PR and handle any OpenSSL 3.x compatibility issues in a second PR.
> 
> Also, can you please disclose whether you are using AI tools to generate code and/or comments?

Okay, I will try to break them down into two separate issues. And no, the only AI usage involved was asking ChatGPT if liboqs has any more OpenSSL 3.x compatibility issues after seeing that it uses deprecated EVP_MD_CTX_{create,destroy} functions, which then resulted in an incorrect claim (of "OpenSSL 3.x defines OSSL_FUNC as a type") and then an unnecessary change, as I unfortunately hadn't confirmed it myself until I checked the OpenSSL documentation properly.
I will update the PR at my next availability inthe following days, but if that doesn't work, I think it would be better to close this PR until I become available again in a couple of months, since it is causing a serious time loss for everyone. Or alternatively a person with more experience and better understanding of the repo can take over/remove this and create a new one.

dstebila:
Closing stale pull request.

PR #2213, opened 2025-07-23
Add Poutine SASL

aidenfoxivey:


praveksharma:
Thank you for looking at this Aiden! I had started working on this but couldn't get the results from the SARIF file to properly integrate with the GitHub security panel. Do you know if that integration works now? Also, where can a user see the results of a Poutine scan?

aidenfoxivey:
> Do you know if that integration works now?

Yes! The PR successfully generates both SARIF files and then uploads them.

> Also, where can a user see the results of a Poutine scan?

A user should be able to see them in the "Security" section of the repository.

aidenfoxivey:
For example, here is it on the main branch of my fork of liboqs. (The action fails because my fork does not have the GitHub secret set to upload, but it does generate SARIF files correctly as you can see in the trace.)

https://github.com/aidenfoxivey/liboqs/actions/runs/16579349957/job/46891457939

praveksharma:
Okay, LGTM!

dstebila:
Hi @aidenfoxivey, there are some [build failures on main](https://github.com/open-quantum-safe/liboqs/actions/runs/16599924326/job/46956414931) after this landed. Any insight into what's going on there?

aidenfoxivey:
> Hi @aidenfoxivey, there are some [build failures on main](https://github.com/open-quantum-safe/liboqs/actions/runs/16599924326/job/46956414931) after this landed. Any insight into what's going on there?

Let me give those a read - should be able to have a fix out shortly.

aidenfoxivey:
Okay, trying out some fixes here https://github.com/aidenfoxivey/liboqs/tree/fix-poutine-scan

praveksharma:
> Okay, trying out some fixes here https://github.com/aidenfoxivey/liboqs/tree/fix-poutine-scan

The OSSF Scorecard job is [restricted](https://github.com/ossf/scorecard-action#workflow-restrictions) to only allow [certain actions](https://github.com/ossf/scorecard-webapp/blob/a2d4e8f2b85b07f8cc51d9e7a67e9152c91f700e/app/server/verify_workflow.go#L166), so I think your approach of creating a separate Poutine analysis job should work. Thank you @aidenfoxivey!

PR #2216, opened 2025-07-28
Icicle-pqc integration

nonam3e:
This PR adds ICICLE PQC ML-KEM support into liboqs, requires NVIDIA GPU to run


src/kem/ml_kem/icicle_ml-kem-1024_icicle_cuda/icicle_ml-kem.cpp
```diff
@@ -0,0 +1,31 @@
+#include <stdint.h>
```
dstebila, line 3: Needs an SPDX-License-Identifier header

src/kem/ml_kem/icicle_ml-kem-512_icicle_cuda/icicle_ml-kem.cpp
```diff
@@ -0,0 +1,31 @@
+#include <stdint.h>
```
dstebila, line 3: Needs an SPDX-License-Identifier header

src/kem/ml_kem/icicle_ml-kem-768_icicle_cuda/icicle_ml-kem.cpp
```diff
@@ -0,0 +1,31 @@
+#include <stdint.h>
```
dstebila, line 3: Needs an SPDX-License-Identifier header

dstebila (changes requested):
Once the license headers are added I think the code is okay. There should also be a line added to PLATFORMS.md. I assume there's no way for us to test this in CI at this point, similar to how we aren't able to test the cuPQC backend; is that right, @praveksharma?

dstebila (approved):

praveksharma:
> I assume there's no way for us to test this in CI at this point, similar to how we aren't able to test the cuPQC backend; is that right, @praveksharma?

That is correct.

praveksharma:
> There should also be a line added to PLATFORMS.md

At present PLATFORMS.md reads: "NVIDIA GPU architectures 70, 75, 80, 86, 89, and 90 with a x86_64 CPU for Linux". @nonam3e, does icicle target any GPU platforms not listed here?

praveksharma (approved):
Apart from the comment about PLATFORMS.md this looks good to me, thank you @nonam3e!

nonam3e:
> > There should also be a line added to PLATFORMS.md
> 
> At present PLATFORMS.md reads: "NVIDIA GPU architectures 70, 75, 80, 86, 89, and 90 with a x86_64 CPU for Linux". @nonam3e, does icicle target any GPU platforms not listed here?

Thank you for the review! ICICLE-PQC supports only NVIDIA GPU architectures 80 and above, so there’s nothing to add to PLATFORMS.md.

praveksharma:
Merging now.

PR #2217, opened 2025-07-29
Fix code scanning workflow

aidenfoxivey:


aidenfoxivey:
woohoo! looks like it works

dstebila:
@praveksharma Can you take a look at this PR?

PR #2196, opened 2025-07-07
Add Classic McEliece sanitization patch

aidenfoxivey:
Signed-off-by: Aiden Fox Ivey <aiden@aidenfoxivey.com>

PR #2218, opened 2025-07-29
Add Classic McEliece sanitization patch 

aidenfoxivey:
Supersedes https://github.com/open-quantum-safe/liboqs/pull/2196

(I accidentally closed 2196)


aidenfoxivey:
oh jeez, seems I had the same problem here as before. I'll sort this out after I fix the poutine predicament.

dstebila (approved):

Issue #2135, opened 2025-05-06
Adding a DeriveEncapsulation functionality


RodriM11:
Version 0.13.0 includes a DeriveKeyPair functionality. I was wondering whether the same functionality could be included for Encapsulation procedures. Every considerations explained in #1206 regarding the interest of this functionality apply here as well and IMO would also be interesting to have (and use it). 
Furthermore, I saw that as part of the merge to include the DeriveKeyPair functionality, @SWilson4 included similar modifications for Encapsulation, but was later deleted. So, hopefully, this inclusion would not be a lot of trouble. 
I could take a swing at including it if desired/necessary.

RodriM11:
Hi @SWilson4 , thanks for the quick response. IMO, for example similar uses would be desirable under the inclusion of this feature:

- deterministic testing of Encaps procedures without toggling the DRBG (concurrence, shared use between different functionalities)
- For non-ephemeral encapsulation procedures or functionalities in which encapsulation procedures might need to be stored, the possibility to store just the seeds to generate such procedure, instead of the (expensive) storage of the encapsulation value (at the cost of having to deterministically generate it from the seeds once it is needed). I think the possibility of this situation alone (which is not at all an uncommon situation, I would like to use the feature if available) should warrant the ability to do so. (One could argue that I could also do that will the DRBG API, but as before, I might not always want to use deterministic generation or might have to concurrent uses of the same library with different requirements, etc...)

SWilson4:
> Hi [@SWilson4](https://github.com/SWilson4) , thanks for the quick response. IMO, for example similar uses would be desirable under the inclusion of this feature:
> 
>     * deterministic testing of Encaps procedures without toggling the DRBG (concurrence, shared use between different functionalities)

For the time being, we would only be able support this feature for ML-KEM, which does have this sort of testing done with [NIST ACVP vectors](https://github.com/open-quantum-safe/liboqs/blob/main/tests/test_acvp_vectors.py). Do you feel that additional testing is needed beyond the ACVP vectors? It would certainly be useful for non-standardized algorithms which don't have exhaustive official test vectors, but there's no standard derandomized encapsulation API for those.

[...]

PR #2221, opened 2025-07-31
Include DeriveEncapsulation functionality (Issue #2135)

RodriM11:
This PR provides the `DeriveEncapsulation` functionality related to issue #2135, i.e. it provides support for a deterministic KEM encapsulation API. This API will only yield an `OQS_SUCCESS` result on ML-KEM, as it is the only PQC algorithm that accounts for said API explicitly.

The `test_kem` file has been expanded to also include the possibility to call `OQS_KEM_encaps_derand`.

* [ ] Does this PR change the input/output behavior of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)

There is no impact of the new `encaps_derand` API on `oqsprovider` right now, as it does not account for `seed` parameters in either `keypair` or `encaps` operations.


dstebila (commented):
Thanks Rodrigo! This looks great! I've triggered the CI tests to run. One question I have is whether there are test vectors available for derandomized encapsulation in the NIST ACVP system.

RodriM11:
Hi @dstebila, thanks for the feedback! 

> One question I have is whether there are test vectors available for derandomized encapsulation in the NIST ACVP system.

AFAIK, `liboqs` already employs ACVP tests for encapsulation, so that work is already covered. It is true though that the current implementation of ACVP vectors test (given in `vectors_kem.c`) employs DRBG toggle to implement derandomized operations. This is likely because `_derand` APIs are somehow new. If desired, now that both `OQS_KEM_keypair_derand` and `OQS_KEM_encaps_derand` are available, I could modify `vectors_kem.c` to employ them.

dstebila:
> Hi @dstebila, thanks for the feedback!
> 
> > One question I have is whether there are test vectors available for derandomized encapsulation in the NIST ACVP system.
> 
> AFAIK, `liboqs` already employs ACVP tests for encapsulation, so that work is already covered. It is true though that the current implementation of ACVP vectors test (given in `vectors_kem.c`) employs DRBG toggle to implement derandomized operations. This is likely because `_derand` APIs are somehow new. If desired, now that both `OQS_KEM_keypair_derand` and `OQS_KEM_encaps_derand` are available, I could modify `vectors_kem.c` to employ them.

With the `_derand` APIs now available, yes, it would be good to have vectors_kem.c employ them, if you're able to make that happen.

RodriM11:
Have added the modifications to 'vectors_kem.c' as well!

praveksharma:
Thank you @RodriM11! This looks good to me with the exception of the patch file. When you get a chance, could you please contribute the changes upstream to mlkem-native's liboqs integration: https://github.com/pq-code-package/mlkem-native/tree/main/integration/liboqs? This would keep the integration in good shape and we could eventually git rid of the patch file.

RodriM11:
Thanks for the review @praveksharma ! Yes, I was planning on contributing the changes to mlkem native repository as soon as this PR is closed. Will also create a PR here to remove the patch file and use the updated native repository when is finished.

PR #2223, opened 2025-08-01
Added dilithium3 derand keypair from seed

paulforbugs:
Added dilithium3 implementation for deterministic keypair generation from seed (please review for security correctness) with enhancement to test suite mirroring the way other algos test deterministic generation

Included Docker runner for test

* [N] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged.)

### DIFF
--- a/src/sig/dilithium/pqcrystals-dilithium_dilithium3_avx2/sign.c
+++ b/src/sig/dilithium/pqcrystals-dilithium_dilithium3_avx2/sign.c
@@ -56,6 +56,7 @@ static inline void polyvec_matrix_expand_row(polyvecl **row, polyvecl buf[2], co
 }
 #endif
 
+
 /*************************************************
 * Name:        crypto_sign_keypair
 *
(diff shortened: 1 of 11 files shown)


test_sig.Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:22.04
```
aidenfoxivey, line 1: Given that there exists Dockerfiles already that are cached to use within liboqs, does it make sense to add another one?

test_sig.sh
aidenfoxivey, line 1: Is the intention to run this within a github workflow file?

src/sig/dilithium/pqcrystals-dilithium_dilithium3_avx2/sign.c
```diff
@@ -56,6 +56,7 @@ static inline void polyvec_matrix_expand_row(polyvecl **row, polyvecl buf[2], co
 }
 #endif
 
+
```
aidenfoxivey, line 59: I suspect this extra newline may not be necessary.

dstebila:
Hi @paulforbugs,

Thanks for working on a contribution. However we will be removing support for Dilithium quite shortly -- we announced this in our [most recent release notes](https://github.com/open-quantum-safe/liboqs/releases/tag/0.14.0).  So this change would be to code that will be shortly removed.

I think there could be general interest in derandomized key pair generation for signatures. I think we'd want to give some consideration to the API, and also check in with some of our upstream projects like [mldsa-native](https://github.com/pq-code-package/mldsa-native) to see what their plan is for derandomized key pair generation.

PR #2225, opened 2025-08-01
Rng improvement

### LINKED ISSUE #2224: RNG improvements
Hello devs,

I'd like to point out the following two issues with the RNG utils:
* `OQS_randombytes_system()` uses the deprecated Crypto API on Win32 instead of the Next Gen Cryptography API's [BCryptGenRandom](https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom).
* The [NIST SP 800-90A](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-90Ar1.pdf) specification defines parameters for the max number of random bits per-call (_max_number_of_bits_per_request_ in Table 3) and max number of requests between reseeds (_reseed_interval_ in Table 3) to ensure that the RNG is not invoked without sufficient seed material. The security guarantees are incomplete if these limits are not enforced.

I am not quite sure how the NIST RNG was meant to be used, but I strongly believe that there should be some way of indicating failure if requests exceed these limits -- either by returning a failure status or exiting with failure if the function must have the current signature (void return).

vibhav950:
This PR implements the following fixes:
* Use [BCryptGenRandom](https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom) as the default Win32 system RNG, and fallback to the legacy Crypto API RNG if the former is not available (versions before Vista or Server 2008 or <= Visual Studio 2008).
* Make `OQS_randombytes_nist_kat` exit with failure if requests exceed the limits specified in [SP 800-90Ar1 Table 3](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-90Ar1.pdf).
    - As per section 10.2.1.2 step 2.1, choosing `ctr_len=32` allows us to increment the counter as a single 32-bit addition to the last 32 bits of the counter `V` in Big-Endian format, thus eliminating the `for` loop.
    - I do realise this requires the caller to be aware of the RNG's behavior and ensure periodic reseeding, but that effort should be worth it for the stronger security guarantees.

Fixes #2224 

* [N] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged.)


dstebila:
Thanks for the contribution!

The Windows-related changes look fine to me.  

However, we don't need the changes related to the NIST RNG, as this is not a general purpose RNG, it is only used for the known answer test vectors in the kat_kem.c and kat_sig.c test files, and is not available for general use in liboqs.

vibhav950:
> However, we don't need the changes related to the NIST RNG, as this is not a general purpose RNG, it is only used for the known answer test vectors in the kat_kem.c and kat_sig.c test files, and is not available for general use in liboqs.

That makes sense -- my bad for not checking up on this before raising the PR!

Reverted the changes to the NIST RNG. Thanks for checking this out!

matlimatli:
Good change. This is the correct way, and looks properly implemented.

(Regarding the changed comment: I'm not sure I would agree with `OQS_randombytes_openssl` being the default though, as it presumes the availability of an external library.)

vibhav950:
> Good change. This is the correct way, and looks properly implemented.

Thanks for checking it out!

> (Regarding the changed comment: I'm not sure I would agree with `OQS_randombytes_openssl` being the default though, as it presumes the availability of an external library.)

My initial way of looking at it was that it first checks for the openssl rng and falls back to the OS' rng if openssl is not available. I guess that notion of "default" does depend on openssl being available, so what you say makes sense. Will revert this change 👍

xuganyu96:
@dstebila I've re-run the CI pipeline and it seems that [26fa86f](https://github.com/open-quantum-safe/liboqs/pull/2225/commits/26fa86fc699ea33da89882d6788a0f4bf7760d4b) fixed the CI pipeline failure. This PR seems good to ship.

dstebila (approved):

Issue #2177, opened 2025-06-25
MSBuild Warnings


1xfakebit:
"C:\Users\Windows\Desktop\github\liboqs\build\ALL_BUILD.vcxproj" (default target) (1) ->
"C:\Users\Windows\Desktop\github\liboqs\build\tests\test_sig.vcxproj" (default target) (130) ->
  C:\Users\Windows\Desktop\github\liboqs\tests\test_sig.c(52,61): warning C4477: 'snprintf' : format string '%ld' requi
res an argument of type 'long', but variadic argument 1 has type 'size_t' [C:\Users\Windows\Desktop\github\liboqs\build
\tests\test_sig.vcxproj]

"C:\Users\Windows\Desktop\github\liboqs\build\ALL_BUILD.vcxproj" (default target) (1) ->
"C:\Users\Windows\Desktop\github\liboqs\build\tests\test_sig_stfl.vcxproj" (default target) (132) ->
  C:\Users\Windows\Desktop\github\liboqs\tests\test_sig_stfl.c(439,61): warning C4477: 'snprintf' : format string '%ld'
 requires an argument of type 'long', but variadic argument 1 has type 'size_t' [C:\Users\Windows\Desktop\github\liboqs
\build\tests\test_sig_stfl.vcxproj]

### Using %zu instead of %ld for snprintf size_t might help to fix this warning.

dstebila:
Thanks for the report. Would you like to submit a PR to fix this?

PR #2239, opened 2025-08-11
Avoid invalid compiler options for MSVC

### LINKED ISSUE #2238: Warnings when building with MSVC
MSVC does not support the compiler flags -mssse3 and -maes, so they cause some warnings when building. There are not any corresponding flags to use, so the best solution is to simply not use these flags when building with MSVC.

```
cl : command line  warning D9002: ignoring unknown option '-maes' [...\src\common\internal.vcxproj]
cl : command line  warning D9002: ignoring unknown option '-mssse3' [...\src\common\internal.vcxproj]
```

There are also some warnings that PR #2181 fixes. For some reason that PR is approved and closed, but not merged.

matlimatli:
The options -mssse3 and -maes are not available when building with MSVC and causes warnings when used.

Fixes #2238


matlimatli:
Also added the fixes for #2177, as its PR was closed prematurely.

PR #2237, opened 2025-08-11
SLHDSA ACVP tests

h2parson:
This PR adds ACVP tests for SLH DSA

This does not address any issue.

The ACVP tests are sourced from NIST and have been run. All passed.

This PR does not change the behaviour of algorithms or the list of algorithms available.

Please note that this pull request will need to be merged AFTER slhdsa has been integrated into libOQS.

Issue #2204, opened 2025-07-15
Create SLH DSA hash shims


h2parson:
SLH DSA currently uses its own internal hash function implementations. However, these will need to be swapped out with libOQS's hash function implementations by adding shims.

It is expected that this will greatly increasing the speed of SLH DSA signing. In this case, the context signing tests in sig_test should be re-enabled for SLH-DSA as well as leak testing (possibly with some remaining restrictions to allow CIs to complete sufficiently quickly). If the CIs are still running too slow at this point, maybe the tests can be isolated into their own GitHub actions jobs.

PR #2245, opened 2025-08-18
Replace SLH-DSA single call hash APIs with shims

h2parson:
This branch is off of the changes made in the PR:
https://github.com/open-quantum-safe/liboqs/pull/2237

PR 2237 for the slhdsa ACVP tests should be merged before merging this branch.

Currently, the slhdsa implementation uses its own SHA2 and SHAKE implementations. This PR will add shims to replace these calls with calls into the libOQS APIs.

This PR begins the process of addressing #2204. This PR adds shims for Sha2_224, Sha2_256, Sha2_384 and Sha2_512 functions in slh dsa. the next step is to add Sha2_512/224 and Sha2_512/256 which should follow a slightly simplified version of the process of adding Sha2_224. Next, the sha3 function used in slh dsa will need to be shimmed. This will complete the process of shimming the single call hash APIs in slh dsa.

This PR should not change the behaviour of a cryptographic algorithm.

This PR will not change the list of algorithms available.


xuganyu96:
@h2parson @dstebila This PR needs a rebase since #2237 landed. I can do the rebase if you are busy with something else, though I will need permission to push this branch. Thank you.

tests/vectors_sig.c
```diff
@@ -13,9 +13,11 @@
 
 #include <oqs/oqs.h>
 
+#include "oqs/common.h"
```
xuganyu96: This seems redundant since `<oqs/oqs.h>` already includes `oqs/common.h`.

tests/vectors_sig.c
```diff
@@ -75,6 +77,16 @@ extern int pqcrystals_ml_dsa_87_ref_verify_internal(const uint8_t *sig,
         const uint8_t *pk);
 #endif
 
+#ifdef OQS_ENABLE_SIG_SLH_DSA
```
xuganyu96: these two function declarations also seem redundant since `slh_dsa.h` was already included:

```c
#include "../src/sig/slh_dsa/slh_dsa_c/slh_dsa.h"
```

tests/vectors_sig.c
```diff
@@ -140,6 +152,65 @@ static void MLDSA_randombytes_free(void) {
 	prng_state.pos = 0;
 }
 
+/* ML_DSA-specific functions */
+static inline bool is_slh_dsa(const char *method_name) {
+	return (0 == strncmp(method_name, "SLH_DSA", 7));
```
xuganyu96, line 158: 7 is a magic number here; consider using either `sizeof` or `strlen` on the string "SLH_DSA" instead?

tests/vectors_sig.c
```diff
@@ -140,6 +152,65 @@ static void MLDSA_randombytes_free(void) {
...
+}
+
+static void SLHDSA_randombytes_init(const uint8_t *entropy_input, const uint8_t *personalization_string) {
+	(void) personalization_string;
+	prng_state.pos = entropy_input;
```
xuganyu96: I know this program is not meant to be thread-safe, so having `prng_state` as some kind of global state is probably fine, but it feels bad.

tests/vectors_sig.c
```diff
@@ -140,6 +152,65 @@ static void MLDSA_randombytes_free(void) {
...
+
+static void SLHDSA_randombytes_free(void) {
+	prng_state.pos = 0;
+}
+
+#ifdef OQS_ENABLE_SIG_SLH_DSA
+static const slh_param_t *SLHDSA_param_set(const char *method_name) {
+	char *id = strstr(method_name, "PURE");
+	if (id != NULL) {
+		id += 5;
```
xuganyu96: 5 is another magic number. I think you mean to search for `"PURE_"` then advance the pointer to after "PURE_"? In which case I would do this instead for better readability:

```c
#define PURE_PREFIX "PURE_"
char *id = strstr(method_name, PURE_PREFIX);
if (id) {
    id += strlen(PURE_PREFIX);  // TODO: define a macro for "PURE_"
} else {
    /* ... */
}
```

tests/vectors_sig.c
```diff
@@ -140,6 +152,65 @@ static void MLDSA_randombytes_free(void) {
...
+	} else {
+		id = strstr(method_name, "PREHASH");
+		id += 8;
+	}
+
+	if (!strcmp(id, "SHA2_128S")) {
```
xuganyu96: there is a `return` in the if-else block, you don't need the `else if`. This would look cleaner:

```c
	if (!strcmp(id, "SHA2_128S")) 
		return &slh_dsa_sha2_128s;
    if (!strcmp(id, "SHA2_128F")) 
		return &slh_dsa_sha2_128f;
    if (!strcmp(id, "SHA2_192S")) 
		return &slh_dsa_sha2_192s;
    if (!strcmp(id, "SHA2_192F")) 
		return &slh_dsa_sha2_192f;
    if (!strcmp(id, "SHA2_256S")) 
		return &slh_dsa_sha2_256s;
    if (!strcmp(id, "SHA2_256F")) 
		return &slh_dsa_sha2_256f;
    if (!strcmp(id, "SHAKE_128S")) 
		return &slh_dsa_shake_128s;
    if (!strcmp(id, "SHAKE_128F")) 
		return &slh_dsa_shake_128f;
    if (!strcmp(id, "SHAKE_192S")) 
		return &slh_dsa_shake_192s;
    if (!strcmp(id, "SHAKE_192F")) 
		return &slh_dsa_shake_192f;
    if (!strcmp(id, "SHAKE_256S")) 
		return &slh_dsa_shake_256s;
    if (!strcmp(id, "SHAKE_256F")) 
		return &slh_dsa_shake_256f;
    return NULL;
```

Also I appreciate using `strncmp` at other places in this file, any reason why `strcmp` is used here?

tests/vectors_sig.c
```diff
@@ -255,6 +332,11 @@ static int sig_ver_vector(const char *method_name,
 	} else if (!strcmp(method_name, "ML-DSA-87")) {
 #ifdef OQS_ENABLE_SIG_ml_dsa_87
 		rc = pqcrystals_ml_dsa_87_ref_verify_internal(sigVer_sig_bytes, sig->length_signature, sigVer_msg_bytes, msgLen, NULL, 0, sigVer_pk_bytes);
+#endif
+	} else if (!strncmp(method_name, "SLH_DSA", 7)) {
```
xuganyu96: I am not sure about the logic of the macro here.
If `OQS_ENABLE_SIG_SLH_DSA` is turned off, then having `method_name` set to `"SLH_DSA_XXX"` should be considered invalid, so the routine should `goto err`.
Instead, this block says if the macro is turned off, and `method_name` is set to `"SLH_DSA_XXX"`, then do nothing. Is this the intended behavior?

tests/vectors_sig.c
```diff
@@ -282,6 +364,8 @@ static int sig_ver_vector(const char *method_name,
 	OQS_SIG_free(sig);
 	return ret;
 
+	/* __attribute__ doesn't work on windows, so "use" sigLen to make compiler happy */
+	(void)sigLen;
```
xuganyu96: Is this because when `OQS_ENABLE_SIG_SLH_DSA` is turned off, `sigLen` is unused?
Regardless, please move it to the beginning of the function. It doesn't make sense to have code after `return`.

tests/vectors_sig.c
```diff
@@ -334,6 +418,8 @@ static int sig_ver_vector_ext(const char *method_name,
...
```
xuganyu96: Same as line 368.

tests/vectors_sig.c
```diff
@@ -379,13 +465,18 @@ static int sig_gen_vector(const char *method_name,
 	} else if (!strcmp(method_name, "ML-DSA-87")) {
 #ifdef OQS_ENABLE_SIG_ml_dsa_87
 		rc = pqcrystals_ml_dsa_87_ref_signature_internal(signature, &sigLen, sigGen_msg, sigGen_msgLen, NULL, 0, prng_output_stream, sigGen_sk);
+#endif
+	} else if (!strncmp(method_name, "SLH_DSA", 7)) {
+#ifdef OQS_ENABLE_SIG_SLH_DSA
```
xuganyu96: Same as line 336

tests/vectors_sig.c
```diff
@@ -550,18 +645,37 @@ int main(int argc, char **argv) {
...
+
+		if (is_ml_dsa(alg_name)) {
+			prng_output_stream = argv[3];
+			kg_pk = argv[4];
+			kg_sk = argv[5];
+
+			if (strlen(prng_output_stream) != 2 * MLDSA_RNDBYTES ||
+			        strlen(kg_pk) != 2 * sig->length_public_key ||
+			        strlen(kg_sk) != 2 * sig->length_secret_key) {
+				printf("lengths bad\n");
```
xuganyu96: A more informative error message should be helpful:

```c
printf("Expected %d bytes for random bytes, found %d\n", 
       MLDSA_RNDBYTES, strlen(prng_output_stream) / 2);
```

tests/vectors_sig.c
xuganyu96, line 1: Most of the comments here are cosmetic. I guess at the end of the day this is an "internal" program so we don't need to guard against every possible memory issue.
I understand that the compiled program `vectors_sig` will take data from command-line input, which makes it tedious if I want to test `vectors_sig` on its own (instead of it being invoked from another Python or shell script). There are some command-line utilities that can parse JSON files (e.g. `jq`), but their usage should definitely be documented so as to make things easier.
Sooner or later, FN-DSA will come out, and the additional rounds of signatures means we might get another signature, so it's better to plan for future ACVP integration.

tests/ACVP_Vectors/SLH-DSA-keyGen-FIPS205/internalProjection.json
xuganyu96, line 1: I know this file already landed, but I would like to challenge storing bulk data files in a git repository, especially for SLH-DSA, whose signature tests vectors are pretty large. Large data files affect git performance.
Another argument against this is that storing the test vectors implies taking responsibility for maintaining them and keeping them up-to-date, which may have unintended consequence down the road.
I personally prefer having the test scripts call `fetch_values.sh` and store the test vectors as temporary files to be cleaned up, not committed.

baentsch: I absolutely second your thoughts @xuganyu96 . Worthwhile creating an issue to track this very sensible clean-up separately?

src/sig/slh_dsa/patches/hash_shims.patch
```diff
@@ -0,0 +1,110 @@
...
+-}
+diff --git a/src/sig/slh_dsa/slh_dsa_c/sha2_api.h b/src/sig/slh_dsa/slh_dsa_c/sha2_api.h
+index 367b5e4f..1e948cf8 100644
+--- a/src/sig/slh_dsa/slh_dsa_c/sha2_api.h
++++ b/src/sig/slh_dsa/slh_dsa_c/sha2_api.h
+@@ -20,9 +20,6 @@ extern "C"
+ 
+   /* === Single-call hash wrappers === */
+ 
+-  /* SHA2-224: Compute 28-byte hash to "md" from "m" which has "m_sz" bytes. */
```
baentsch, line 105: If this PR aims to remove all local SHA2 code and replace it with calls to `liboqs` common code, why does it look like sha2-512 is not being patched in the same way as is sha2-224?

h2parson: Hi Michael, I think this patch file may be deceptive. It can be seen in the changes applied above that the definitions for some of the other sha2 variants were applied earlier in the patch:
https://github.com/open-quantum-safe/liboqs/pull/2245/files/c043ef2660cbed78d3d05c57423f2a0a3f1a4865#diff-f9299022c96d07afa8646fe6de438aa6726cc9e288c62a1f9e838decf4dfce64R66

These changes can be seen in the sha2_api.h file:

image

However, this does bring up an important point that this PR is not complete. It only provides shims for 4 sha2 variants. There are still some truncated sha2 variants and sha3 which need shims. The purpose of this PR is to provide an example for what these shims might look like.

@xuganyu96 the changes you have suggested could be implemented when the rest of the shims are added. Thank you for your feedback.

baentsch: > this does bring up an important point that this PR is not complete

Would it then be prudent to mark this "Draft" until complete?

tests/ACVP_Vectors/SLH-DSA-keyGen-FIPS205/internalProjection.json
baentsch: Thanks for creating https://github.com/open-quantum-safe/liboqs/issues/2279 @xuganyu96 - marking this comment resolved here, then.

xuganyu96:
@h2parson Would you be okay with me taking over the remaining work on this project? I am thinking of moving this work onto a branch within this repository and closing this pull request, then raising a separate PR from the new branch. This way current committers can commit as needed, and your prior commits will be preserved.

cc: @dstebila

baentsch:
> @h2parson Would you be okay with me taking over the remaining work on this project?

I'd offer support for this proposal by then being reviewer that you then shouldn't/couldn't be any more @xuganyu96 (but surely would welcome input on that from @h2parson !)

src/sig/slh_dsa/patches/hash_shims.patch
h2parson: done!

h2parson:
> @h2parson Would you be okay with me taking over the remaining work on this project? I am thinking of moving this work onto a branch within this repository and closing this pull request, then raising a separate PR from the new branch. This way current committers can commit as needed, and your prior commits will be preserved.
> 
> cc: @dstebila

That sounds good to me! I will not have time to work on this, so I would be glad if somebody could get this merged. Please reach out if you have any questions about what has been done so far.

PR #2246, opened 2025-08-18
Fix permissions for poutine_analysis job

dstebila:
Try to fix https://github.com/open-quantum-safe/liboqs/actions/runs/17043292457

### DIFF
--- a/.github/workflows/supplychain.yml
+++ b/.github/workflows/supplychain.yml
@@ -1,6 +1,7 @@
 name: Scorecard supply-chain security
 
-permissions: {}
+permissions:
+  contents: read
 
 on:
   # For Branch-Protection check. Only the default branch is supported. See
--- a/.github/workflows/commit-to-main.yml
+++ b/.github/workflows/commit-to-main.yml
@@ -22,6 +22,7 @@ jobs:
     permissions:
       id-token: write
       security-events: write
+      contents: read
 
   basic-downstream:
     uses: ./.github/workflows/downstream-basic.yml


praveksharma (approved):

.github/workflows/supplychain.yml
```diff
@@ -1,6 +1,7 @@
 name: Scorecard supply-chain security
 
-permissions: {}
+permissions:
+  contents: read
```
xuganyu96, line 4: Hi, first time contributor and GitHub Action newbies here.

I'm not sure if this will fix the issue. It seems like "Scorecard supply-chain security" runs fine on its own, but doesn't have the `contents: read` permission when invoked as a nested job under "Main branch test".

I think we also need to add the permission to "Main branch test".

Also, maybe we can add more triggers to "Main branch test" so it can be tested?

```yaml
-- in commit-to-main.yml
on:
  workflow_dispatch:
```

dstebila: > I'm not sure if this will fix the issue. It seems like "Scorecard supply-chain security" runs fine on its own, but doesn't have the `contents: read` permission when invoked as a nested job under "Main branch test".
> 
> I think we also need to add the permission to "Main branch test".

Oh, good point. Yes, we'll need to do that. I've pushed a commit that does that.

> Also, maybe we can add more triggers to "Main branch test" so it can be tested?
> 
> ```yaml
> -- in commit-to-main.yml
> on:
>   workflow_dispatch:
> ```

This I don't quite understand.

dstebila: I understand now from our meeting: this would allow manual triggering of the workflow to test it, rather than waiting for a commit to main to trigger it. That would be useful, but possibly on other workflows as well, so let's consider that in a separate issue/PR.

PR #2250, opened 2025-08-19
Update license info for ML-KEM

### LINKED ISSUE #2249: License information for ML-KEM is not up-to-date
According to https://github.com/pq-code-package/mlkem-native/blob/main/LICENSE, the license for the ML-KEM source is

> All source code in mlkem/* and dev/* is licensed under your choice
of the Apache-2.0 license OR the ISC license OR the MIT license.

The licensing information in the Readme however only states that it is `Apache License v2.0`, while the algorithm description states `CC0-1.0 or Apache-2.0`.

Furthermore, there is no license file in any of the ml_kem subdirectories.

matlimatli:
Update the licensing information for ML-KEM and add license files for the ml_kem subdirectories with other licenses than plain MIT. (Fixes #2249)

For the mlkem-native files, the LICENSE text is taken from the mlkem-native repository, with the ingress modified in order to not refer to non-existing foldersfiles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ liboqs includes some third party libraries or modules that are licensed differen
 - `src/kem/kyber/pqcrystals-*`: public domain (CC0) or Apache License v2.0
 - `src/kem/kyber/pqclean_*`: public domain (CC0), and public domain (CC0) or Apache License v2.0, and public domain (CC0) or MIT, and MIT
 - `src/kem/kyber/libjade_*` public domain (CC0) or Apache License v2.
-- `src/kem/ml_kem/mlkem-native_*`: Apache License v2.0
+- `src/kem/ml_kem/mlkem-native_*`: MIT or Apache License v2.0 or ISC License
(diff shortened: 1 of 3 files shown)


matlimatli:
I realized that it was not possible to just add LICENSE files (caused issues with upstream verification). Perhaps the SPDX strings in the source files is good enough? If so, it might be worth reviewing the statement in the current top-level README:

> liboqs includes some third party libraries or modules that are licensed differently; the corresponding subfolder contains the license that applies in that case

dstebila:
> I realized that it was not possible to just add LICENSE files (caused issues with upstream verification). Perhaps the SPDX strings in the source files is good enough? If so, it might be worth reviewing the statement in the current top-level README:
> 
> > liboqs includes some third party libraries or modules that are licensed differently; the corresponding subfolder contains the license that applies in that case

Thanks for the fixes. Yes, adding the files directly to the directories won't work because of how we copy code from upstream. We could use the patch mechanism to add those files to those directories, but I think with the SPDX headers and the corrections you've given, it should be clear to anyone looking at the source code how those files are licensed.

ashman-p (approved):

PR #2252, opened 2025-08-21
Change 64 bit add to 32 bit add to wrap on 32 bit counter for AES-CTR…

### LINKED ISSUE #2234: AES-CTR implementation using AES-NI wraps on 64 bit counter
**Describe the bug**
In the AES256-CTR implementation with AES-NI instructions, the counter is 64 bits rather than 32 bits. The software implementation has a counter with 32 bits.

**To Reproduce**
Call AES256_CTR_inc_stream_iv with a nonce ending in {0xFF, 0xFF, 0xFF, 0xFF} with at least 2 blocks of output. The software and hardware implementations should have different outputs.

**Expected behavior**
The software and hardware implementations are expected to match, with a counter of 32 bits.

**Screenshots**

 - OS: [e.g. Ubuntu 20] Redhat 9
 - OpenSSL version [e.g., 3.0.2]
 - Compiler version used [e.g., clang 9.0.0]
 - Build variables used [e.g., "-DOQS_ALGS_ENABLED=STD"] With and without AES-NI
 - liboqs version [e.g. 0.7.2 or main branch] main branch

**Additional context**
This is because the increment uses _mm_add_epi64 instead of _mm_add_epi32 in oqs_aes256_ctr_enc_sch_ni and oqs_aes256_ctr_enc_sch_upd_blks_ni.

max-p-log-p:
… AES-NI implementation

In the AES256-CTR implementation with AES-NI instructions, the counter is 64 bits rather than 32 bits. The software implementation has a counter with 32 bits. The software and hardware implementations are expected to match, with a counter of 32 bits.

Fixes #2234 

### DIFF
--- a/src/common/aes/aes256_ni.c
+++ b/src/common/aes/aes256_ni.c
@@ -184,18 +184,18 @@ void oqs_aes256_ctr_enc_sch_upd_blks_ni(void *schedule, uint8_t *out, size_t out
 
 	while (out_blks >= 4) {
 		__m128i nv0 = _mm_shuffle_epi8(ctx->iv, mask);
-		__m128i nv1 = _mm_shuffle_epi8(_mm_add_epi64(ctx->iv, _mm_set_epi64x(1, 0)), mask);
-		__m128i nv2 = _mm_shuffle_epi8(_mm_add_epi64(ctx->iv, _mm_set_epi64x(2, 0)), mask);
-		__m128i nv3 = _mm_shuffle_epi8(_mm_add_epi64(ctx->iv, _mm_set_epi64x(3, 0)), mask);
+		__m128i nv1 = _mm_shuffle_epi8(_mm_add_epi32(ctx->iv, _mm_set_epi64x(1, 0)), mask);
+		__m128i nv2 = _mm_shuffle_epi8(_mm_add_epi32(ctx->iv, _mm_set_epi64x(2, 0)), mask);
+		__m128i nv3 = _mm_shuffle_epi8(_mm_add_epi32(ctx->iv, _mm_set_epi64x(3, 0)), mask);
 		aes256ni_encrypt_x4(schedule, nv0, nv1, nv2, nv3, out);
-		ctx->iv = _mm_add_epi64(ctx->iv, _mm_set_epi64x(4, 0));
+		ctx->iv = _mm_add_epi32(ctx->iv, _mm_set_epi64x(4, 0));
 		out += 64;
 		out_blks -= 4;
 	}
 	while (out_blks >= 1) {
 		__m128i nv0 = _mm_shuffle_epi8(ctx->iv, mask);
 		aes256ni_encrypt(schedule, nv0, out);
-		ctx->iv = _mm_add_epi64(ctx->iv, _mm_set_epi64x(1, 0));
+		ctx->iv = _mm_add_epi32(ctx->iv, _mm_set_epi64x(1, 0));
 		out += 16;
 		out_blks--;
 	}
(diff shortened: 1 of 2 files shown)


max-p-log-p:
The build test fails when installing dependencies, but I was able to build it locally

aidenfoxivey:
> The build test fails when installing dependencies, but I was able to build it locally

Are you referring to the Travis CI? If so, you can fully disregard that.

src/common/aes/aes256_ni.c
```diff
@@ -184,18 +184,18 @@ void oqs_aes256_ctr_enc_sch_upd_blks_ni(void *schedule, uint8_t *out, size_t out
 
 	while (out_blks >= 4) {
 		__m128i nv0 = _mm_shuffle_epi8(ctx->iv, mask);
-		__m128i nv1 = _mm_shuffle_epi8(_mm_add_epi64(ctx->iv, _mm_set_epi64x(1, 0)), mask);
-		__m128i nv2 = _mm_shuffle_epi8(_mm_add_epi64(ctx->iv, _mm_set_epi64x(2, 0)), mask);
-		__m128i nv3 = _mm_shuffle_epi8(_mm_add_epi64(ctx->iv, _mm_set_epi64x(3, 0)), mask);
+		__m128i nv1 = _mm_shuffle_epi8(_mm_add_epi32(ctx->iv, _mm_set_epi64x(1, 0)), mask);
```
dstebila, line 187: Does the `_mm_set_epi64x` call also need to be switched to a `32` version to match?

dstebila: I'm not very familiar with the details of these intrinsics, so treat this as question informed by pattern match, not deep technical understanding.

max-p-log-p: It will give the same result as _mm_set_epi32(0, i, 0, 0), where i is the amount being incremented. Since the value it is being added to has had the 1st and 2nd 64 bit registers swapped using _mm_shuffle_epi8, it is the same as adding (0, 0, 0, i) which matches the software implementation, but it may be more readable to use the _mm_set_epi32 function.

xuganyu96:
I am okay with this PR as it is.

I was originally thinking if we should test for wrapping behaviors, but it seems like AES-CTR is not meant to handle counter wrapping, so it doesn't make sense to test for undefined behaviors.

Maybe it makes sense to instead test that AES-NI and C implementations agree? This is probably not trivial given how underlying implementations are hidden behind a single set of AES API in `aes.h`. I don't think the engineering efforts are worth it. Again this is testing for UB's.

dstebila:
> I am okay with this PR as it is.
> 
> I was originally thinking if we should test for wrapping behaviors, but it seems like AES-CTR is not meant to handle counter wrapping, so it doesn't make sense to test for undefined behaviors.
> 
> Maybe it makes sense to instead test that AES-NI and C implementations agree? This is probably not trivial given how underlying implementations are hidden behind a single set of AES API in `aes.h`. I don't think the engineering efforts are worth it. Again this is testing for UB's.

Yes, testing the AES-NI and C implementations against each other could be tricky. I think if you wanted to do comparison testing then we could do what we are doing with hash functions, and have a Python wrapper where we test against the implementation available in Python (which I think is OpenSSL underneath).

dstebila:
Also @xuganyu96 do you have Github permissions for code review? Looking at the permissions in https://github.com/open-quantum-safe/tsc/blob/main/config.yaml I think you should, but I also couldn't tag you in the reviewer dropdown, so I'm not sure.

xuganyu96:
> Also @xuganyu96 do you have Github permissions for code review? Looking at the permissions in https://github.com/open-quantum-safe/tsc/blob/main/config.yaml I think you should, but I also couldn't tag you in the reviewer dropdown, so I'm not sure.

I don't see any options for approving a PR, so it seems I just don't have the permissions.

edit: never mind, I found the right button!

aidenfoxivey:
Happy to review this later if desired.

Issue #2257, opened 2025-08-29
ICICLE ML-KEM missing derandomized encapsulation


xuganyu96:
The ICICLE impl of ML-KEM was merged into `main` before `DerivedEncapsulation` was merged. Currently `src/kem/ml_kem/kem_ml_kem_xxx.c` does not contain de-randomized encapsulation. This causes `copy_from_upstream.py` to modify `src/kem/ml_kem/kem_ml_kem_xxx.c`, which then causes CI failures.

**To Reproduce**
```bash
# From $LIBOQS_DIR at commit 68b0d1
cd scripts/copy_from_upstream
./copy_from_upstream.py -d copy
git status --porcelain
```

The `git status --porcelain` command returns:

```
 M src/kem/ml_kem/kem_ml_kem_1024.c
 M src/kem/ml_kem/kem_ml_kem_512.c
 M src/kem/ml_kem/kem_ml_kem_768.c
```

Instead, it should return an empty string.

**Environment (please complete the following information):**
 - OS: MacOS 15.6.1 (Apple Silicon)
 - OpenSSL OpenSSL 3.4.1
 - Apple clang version 17.0.0 (clang-1700.0.13.5)
 - liboqs@68b0d1

xuganyu96:
I would like a second opinion on the scope of this fix.

The **minimal fix** would be to "just commit the change", which means to add the `enc_derand` functions without providing any implementation. This is the case with `PQCLEAN_MLKEM1024_CUDA_crypto_kem_enc_derand`, which does not have a declaration nor an implementation.

The next step up would be to provide some implementation, but without testing. This is doable because ICICLE's ML-KEM encap API accepts a coin.

*edit:* or I can do an implementation that just "fails" and add some compilation warning?

The next step up would be to provide complete implementation and testing, although I am not sure how much efforts would be needed. I assume we don't have GPU infrastructure to run the ICICLE code, right?

The first two options seem like reasonable fixes, although I personally prefer the minimal fix because I don't like committing untested code.

cc: @dstebila @aidenfoxivey

xuganyu96:
[Stackoverflow](https://stackoverflow.com/questions/20186848/can-i-compile-a-cuda-program-without-having-a-cuda-device) says that it is possible to install the CUDA toolkit without the presence of a CUDA device (i.e. a GPU). If I provide an implementation that actually uses ICICLE's API, then I can maybe at least check that the thing compiles.

xuganyu96:
[...]

PR #2258, opened 2025-09-02
Patch ICICLE ML-KEM for deterministic encapsulation

xuganyu96:
This PR partially addresses #2257 

ICICLE's ML-KEM was merged before deterministic encapsulation landed, so `copy_from_upstream.py` ends up adding non-existent ICICLE functions. This PR adds a patch that implements `icicle_ml_kem_xxx_enc_derand` in the ICICLE integration, then calls the correct `icicle_ml_kem_xxx_enc_derand` API from within OQS's generated source code.

Review guide:
---------------
For review, start with `copy_from_upstream.yml` and `icicle-mlkem-enc-derand.patch`. Everything else is the result of applying these patches.

There is currently no straightforward way to test correctness (which requires the presence of a CUDA device), but that's not the point of this PR anyways. The main purpose of this PR is to resolve the CI failure resulting from inconsistencies between the current state of the code base and the upstreams.

For reviewers: does this PR count as "changing the behavior of a public API?" There is no previous `icicle_ml_kem_xxx_enc_derand`, so this PR technically "adds" these functions.

### DIFF
--- a/docs/algorithms/kem/ntru.md
+++ b/docs/algorithms/kem/ntru.md
@@ -14,14 +14,14 @@
 
 ## Parameter set summary
 
-|   Parameter set    | Parameter set alias   | Security model   |   Claimed NIST Level |   Public key size (bytes) |   Secret key size (bytes) |   Ciphertext size (bytes) |   Shared secret size (bytes) | Keypair seed size (bytes)   |
-|:------------------:|:----------------------|:-----------------|---------------------:|--------------------------:|--------------------------:|--------------------------:|-----------------------------:|:----------------------------|
-| NTRU-HPS-2048-509  | NA                    | IND-CCA2         |                    1 |                       699 |                       935 |                       699 |                           32 | NA                          |
-| NTRU-HPS-2048-677  | NA                    | IND-CCA2         |                    3 |                       930 |                      1234 |                       930 |                           32 | NA                          |
-| NTRU-HPS-4096-821  | NA                    | IND-CCA2         |                    5 |                      1230 |                      1590 |                      1230 |                           32 | NA                          |
-| NTRU-HPS-4096-1229 | NA                    | IND-CCA2         |                    5 |                      1842 |                      2366 |                      1842 |                           32 | NA                          |
-|   NTRU-HRSS-701    | NA                    | IND-CCA2         |                    3 |                      1138 |                      1450 |                      1138 |                           32 | NA                          |
-|   NTRU-HRSS-1373   | NA                    | IND-CCA2         |                    5 |                      2401 |                      2983 |                      2401 |                           32 | NA                          |
+|   Parameter set    | Parameter set alias   | Security model   |   Claimed NIST Level |   Public key size (bytes) |   Secret key size (bytes) |   Ciphertext size (bytes) |   Shared secret size (bytes) | Keypair seed size (bytes)   | Encapsulation seed size (bytes)   |
+|:------------------:|:----------------------|:-----------------|---------------------:|--------------------------:|--------------------------:|--------------------------:|-----------------------------:|:----------------------------|:----------------------------------|
+| NTRU-HPS-2048-509  | NA                    | IND-CCA2         |                    1 |                       699 |                       935 |                       699 |                           32 | NA                          | NA                                |
+| NTRU-HPS-2048-677  | NA                    | IND-CCA2         |                    3 |                       930 |                      1234 |                       930 |                           32 | NA                          | NA                                |
+| NTRU-HPS-4096-821  | NA                    | IND-CCA2         |                    5 |                      1230 |                      1590 |                      1230 |                           32 | NA                          | NA                                |
+| NTRU-HPS-4096-1229 | NA                    | IND-CCA2         |                    5 |                      1842 |                      2366 |                      1842 |                           32 | NA                          | NA                                |
+|   NTRU-HRSS-701    | NA                    | IND-CCA2         |                    3 |                      1138 |                      1450 |                      1138 |                           32 | NA                          | NA                                |
+|   NTRU-HRSS-1373   | NA                    | IND-CCA2         |                    5 |                      2401 |                      2983 |                      2401 |                           32 | NA                          | NA                                |
 
 ## NTRU-HPS-2048-509 implementation characteristics
 


xuganyu96:
@dstebila This pull request now blocks every future pull requests because of the "check upstrema code" test in the CI pipeline and should be given review priority.

cc: @baentsch this PR blocks #2261

dstebila:
Thanks Bruce! You'll need to push a new version with the DCO sign-off (see [instructions in the failed DCO job](https://github.com/open-quantum-safe/liboqs/pull/2258/checks?check_run_id=49431815046)). But then yes, let's get this merged.

scripts/copy_from_upstream/patches/icicle-mlkem-enc-derand.patch
```diff
@@ -0,0 +1,102 @@
...
+     return icicle::pqc::ml_kem::decapsulate<icicle::pqc::ml_kem::Kyber512Params>((std::byte*)sk, (std::byte*)ct, config, (std::byte*)ss) == icicle::eIcicleError::SUCCESS ? OQS_SUCCESS : OQS_ERROR;
+diff --git a/icicle_cuda/ml-kem-768/icicle_ml-kem.cpp b/icicle_cuda/ml-kem-768/icicle_ml-kem.cpp
+index 33b38cb..793afb6 100644
+--- a/icicle_cuda/ml-kem-768/icicle_ml-kem.cpp
++++ b/icicle_cuda/ml-kem-768/icicle_ml-kem.cpp
+@@ -25,6 +25,11 @@ OQS_STATUS icicle_ml_kem_1024_enc(uint8_t *ct, uint8_t *ss, const uint8_t *pk) {
+     return icicle::pqc::ml_kem::encapsulate<icicle::pqc::ml_kem::Kyber1024Params>((std::byte*)coins, (std::byte*)pk, config, (std::byte*)ct, (std::byte*)ss) == icicle::eIcicleError::SUCCESS ? OQS_SUCCESS : OQS_ERROR;
+ }
+ 
++OQS_STATUS icicle_ml_kem_1024_enc_derand(uint8_t *ct, uint8_t *ss, const uint8_t *pk, const uint8_t *coins) {
```
baentsch, line 95: I don't quite understand the reference to Kyber1024 params in a folder labelled ml-kem-768: Intentional?

baentsch:
@xuganyu96 Without any understanding of the CUDA code I only glanced over it from a "logics" perspective (see single question just posted). But as it's not tested/testable anyway, I guess we can rely on shouts of anger of real users if something's wrong :)

> For reviewers: does this PR count as "changing the behavior of a public API?" There is no previous icicle_ml_kem_xxx_enc_derand, so this PR technically "adds" these functions.

IMO, no. This text only is meant to refer to new toplevel functions externally announced/documented/used. This is not the case here.

scripts/copy_from_upstream/patches/icicle-mlkem-enc-derand.patch
xuganyu96: Thank you for noticing this peculiarity. I took a look at [`main@59bf831`](https://github.com/open-quantum-safe/liboqs/blob/59bf831b47fb43124f816d594ec801f729b08053/src/kem/ml_kem/icicle_ml-kem-768_icicle_cuda/icicle_ml-kem.cpp) and it seems like the the implementation was incorrect to begin with.

**edit**: It seems like the issue is was inherited from upstream [upstream](https://github.com/ingonyama-zk/icicle-liboqs/blob/main/icicle_cuda/ml-kem-768/icicle_ml-kem.cpp). I personally don't think it is OQS's responsibility for fixing upstream bugs with our patches. Instead, I can raise a PR to the upstream repository.

Do you think we should stick with "ML-KEM-1024 in `ml-kem-768.cpp`", or should I try to fix this?

cc: @dstebila

baentsch: > should I try to fix this?

imo, yes: One of the main benefits of OQS is to alert upstreams/contributors of possible errors.

dstebila: Thanks for catching that. Yes, let's inform upstream of possible errors.

xuganyu96: Given that the main point of this pull request is to fix CI failures and unblock other PRs, I am inclined to first patch the "ML-KEM-1024 in `ml-kem-768.cpp`" on our side. I will concurrently raise a PR with upstream, and when the upstream is fixed, I can raise another PR to revoke the patch mentioned above.

On top of this, `copy_from_upstream.py` is now also modifying `docs/algorithms/kem/ntru.md` because NTRU was merged before deterministic encapsulation. I don't think these two issues can be addressed separately and will have to be fixed in a single PR (i.e. this PR).

Any thoughts or suggestions?

baentsch: This then begs the question: Shall we move this forward (approve) regardless of possible errors? Given the nature of the project and as this seems to have been the case before/doesn't make the state worse than it is (untested, possibly buggy), I'm inclined towards that to make progress -- but I'd prefer an update before a next release (https://github.com/open-quantum-safe/liboqs/issues/2267).

baentsch: > I don't think these two issues can be addressed separately and will have to be fixed in a single PR (i.e. this PR).

This sounds ugly.

> Any thoughts or suggestions?

Wouldn't it be possible to fix whatever broke by the NTRU merge first in a PR dedicated to that, merge that and then rebase this PR? This would enable clean removal of algorithm-specific features/commits, say if NTRU shall "go" again in the future (an imo not completely unlikely event, right?).

baentsch: Sorry, disregard my last comment @xuganyu96 : It's just a generated documentation update, right(?) so will be auto-fixed (deleted) if/when NTRU is withdrawn again in some future. So PR is OK for me as-is.

xuganyu96: > Wouldn't it be possible to fix whatever broke by the NTRU merge first in a PR dedicated to that, merge that and then rebase this PR?

Unfortunately not. A single PR fixing `docs/algorithms/kem/ntru.md` will not pass CI pipeline because `copy_from_upstream.py` will then modify `src/kem/ml_kem/ml_kem_xxx.c` because ICICLE integration was also merged before deterministic encryption. I think to get "proper integration by `copy_from_upstream`" we can't have a partial fix.

dstebila: Sounds like a bit of a chicken-and-egg problem. Okay with me to get this landed with a quick patch of upstream and then have a separate issue to track removing that patch once upstream is updated.

xuganyu96:
@dstebila @baentsch *Oh the spaghetti monster grew larger!* Now this PR is failing CI pipeline because it is missing the fix from [NTRU not initializing `encaps_seed`](https://github.com/open-quantum-safe/liboqs/pull/2266).

I am afraid that we just have to ship these two PRs as they to restore `main` to a green-light state before applying further changes.

**edit:** I've manually triggered a [basic check](https://github.com/open-quantum-safe/liboqs/actions/runs/17682772751) and a [platform check](https://github.com/open-quantum-safe/liboqs/actions/runs/17682871942) on a branch that has both PRs, and they passed.

baentsch:
> I've manually triggered a [basic check](https://github.com/open-quantum-safe/liboqs/actions/runs/17682772751) and a [platform check](https://github.com/open-quantum-safe/liboqs/actions/runs/17682871942) on a branch that has both PRs, and they passed.

Why then not land ("CI-green") #2266 first and re-base this PR afterwards to keep code changes for ICICLE and NTRU separated? Ah, I see: They are already now mixed in _both_ PRs :-(. If indeed there is a need to merge CI-failing PRs, then I'd strongly suggest to at least keep them clean in the sense that each addresses/touches only the code it mentions in the commit message (such as to be able to understand/possibly roll back later what's been done on a per-algorithm basis). A spaghetti monster persisted to GH history sounds even worse than one in code, no?

Issue #2156, opened 2025-06-02
Deprecate and remove Dilithium


dstebila:

SWilson4:
A deprecation notice will be included in the release notes for 0.14.0 (see https://github.com/open-quantum-safe/liboqs/pull/2180). The code will need to be removed after the release is published.

xuganyu96:
@dstebila Does deprecation mean to wipe Dilithium (and SPHINCS+ for #2155) clean from the repository? Are these two deprecation tasks still blocking 0.15 release?

dstebila:
For Dilithium, we would remove it from the 0.15.0 release (since we advertised that in the 0.14.0 release notes and did not receive any objections.)

For SPHINCS+, our normal practice would be to keep it in the 0.15.0 release but advertise in the release notes that this is the last release containing it, and then remove it after that.

xuganyu96:
Search for "dilithium" in the code base:

```bash
grep -ril \
    --exclude-dir=.git \
    --exclude-dir="ml_kem" \
    --exclude-dir="ml_dsa" \
    --exclude-dir=build \
    --exclude-dir=.pytest_cache \
    "dilithium" .
```

Summary of change (WIP):
- **copy_from_upstream** items are not trivial to modify; a lot of source code is generated by `copy_from_upstream.py` and should not be manually changed

```
------- COPY FROM UPSTREAM --------
./scripts/copy_from_upstream/copy_from_upstream.yml
./scripts/copy_from_upstream/patches/pqcrystals-dilithium-SUF-CMA.patch
./scripts/copy_from_upstream/patches/pqclean-dilithium-arm-randomized-signing.patch
./scripts/copy_from_upstream/patches/pqcrystals-dilithium-ref-shake-aes.patch
./scripts/copy_from_upstream/patches/pqcrystals-dilithium-yml.patch
./scripts/copy_from_upstream/patches/pqcrystals-ml_dsa.patch
./scripts/copy_from_upstream/patches/pqcrystals-dilithium-avx2-shake-aes.patch
./tests/kat_sig.c
./src/sig/sig.c
./src/sig/sig.h
./src/sig/dilithium/pqcrystals-dilithium_dilithium5_avx2/*
./src/sig/dilithium/sig_dilithium_5.c
./src/sig/dilithium/CMakeLists.txt
./src/sig/dilithium/pqcrystals-dilithium_dilithium2_ref/*
[...]

PR #2262, opened 2025-09-08
Remove Dilithium

xuganyu96:
Remove Dilithium from `liboqs` code base #2156.


baentsch:
@xuganyu96 As discussed in the call this week here's the code for auto-triggering downstream testing on alg changes: https://github.com/open-quantum-safe/liboqs/blob/main/.github/workflows/downstream-release.yml: oqs-provider branch must be labelled "liboqs-branch-name-tracker".

PR #2275, opened 2025-09-15
Remove Dilithium

### LINKED ISSUE #2156: Deprecate and remove Dilithium

xuganyu96:
This pull request removes Dilithium source code, tests, configurations, and documentation from the repository. This pull request closes #2156 . See reviewer's guide below for details:

* [YES] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [YES] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged.)

Reviewer's guide to "removing Dilithium"
----------------------------------------
The north star of this pull request is a `grep` command that recursively searched through the entire repository for the keyword "dilithium":

```bash
grep -ril \
    --exclude-dir .git \
    --exclude-dir "ml_kem" \
    --exclude-dir "ml_dsa" \
    --exclude-dir build \
    --exclude-dir .pytest_cache \
    --exclude-dir patches \
    --exclude ml_dsa.yml \
    --exclude ml_dsa.md \
    --exclude RELEASE.md \
    --exclude copy_from_upstream.yml \
    "dilithium" .
```

- `-i` makes the search case insensitive
- `.git`, `.pytest_cache`, `build` are excluded because they are build/VCS artifacts
- `ml_kem` is excluded because `src/kem/ml_kem` mentions Dilithium only as a academic reference/bibliography; similarly we don't care about mentions of Dilithium in ML-DSA code and/or documentation
- `copy_from_upstream.yml` is excluded because many ML-DSA integrations have patches that mention Dilithium. CI pipeline will confirm that this does not affect `copy_from_upstream.py`

With `main` at `1ae1b92302` piping the `grep` command into `wc -l` gives `194`. With this pull request, the number of matches is 1.
Here is a summary of the changes (in this order) that got us to the state of this pull request:
1. `41b34ee` deleted two sections from `copy_from_upstream.yml` (the upstream `pqcrystals-dilithium` and the signature scheme `dilithium`). It also deleted everything under `src/sig/dilithium`. Re-running `copy_from_upstream.py` then produced downstream changes in the files below. After this commit, `grep` produces 20 matches.
    - `.CMake/alg_support.cmake`
    - `CMakeLists.txt`
    - `src/CMakeLists.txt`
    - `src/oqsconfig.h.cmake`
    - `tests/kat_sig.c`
1. `c2ac9ab` deleted `Dilithium` from `tests/KATs/sig/kats.json`. `grep` 19 matches
1. `a2c7f66` deleted `Dilithium` from `tests/constant_time/sig`. `grep` at 14
1. `329f601` deleted `docs/algorithm/sig/dilithium.yml` and `docs/algorithm/sig/dilithium.md`. Re-running `copy_from_upstream.py` then modified `README.md`, and `cbom.json`. `grep` produces 11 matches.
1. `2273d55` updated the documentation in `FUZZING.md`. 10 matches.
1. `9f214d9` removed Dilithium entries from the "licensing" section in `README.md`. Still at 10 matches because `README.md` contains one more mention of Dilithium saying "Dilithium-AES has been deprecated", which I chose not to remove.
1. `6d7175c` updated the example in `CONFIGURE.md` to use ML-KEM-768 and ML-DSA-44 instead of Kyber-768 and Dilithium-2. 9 matches.
1. `232493c` updated the C++ linking test to use ML-DSA-44 instead of Dilithium-2. This is done with a straightforward `s/dilithium_2/ml_dsa_44/g` search and replace.
1. `db3b29e` and `7d1fd49` replaced Dilithium with ML-DSA in CI pipeline and in Zephyr's examples, respectively.


cpp/sig_linking_test.cpp
```diff
@@ -2,6 +2,18 @@
...
+ * g++ -g -I${LIBOQS_DIR}/build/include \
+ *     -L${LIBOQS_DIR}/build/lib -loqs \
+ *     -lcrypto -std=c++11 \
+ *     -o ${LIBOQS_DIR}/build/tests/example_sig \
+ *     ${LIBOQS_DIR}/cpp/sig_linking_test.cpp \
+ * && ${LIBOQS_DIR}/build/tests/example_sig
+ *
+ * `-lcrypto` requires libcrypto to be in dynamic linker's path.
+ * If installed with Homebrew, it is likely located at:
+ * /opt/homebrew/Cellar/openssl@3/x.y.z/lib
```
bhess, line 16: Are these changes related to the Dilithium removal?

scripts/copy_from_upstream/copy_from_upstream.yml
```diff
@@ -55,14 +55,6 @@ upstreams:
...
     kem_scheme_path: '.'
     patches: [icicle-mlkem-enc-derand.patch]
-  -
-    name: pqcrystals-dilithium
-    git_url: https://github.com/pq-crystals/dilithium.git
-    git_branch: master
-    git_commit: 3e9b9f1412f6c7435dbeb4e10692ea58f181ee51
-    sig_meta_path: '{pretty_name_full}_META.yml'
-    sig_scheme_path: '.'
-    patches: [pqcrystals-dilithium-yml.patch, pqcrystals-dilithium-ref-shake-aes.patch, pqcrystals-dilithium-avx2-shake-aes.patch, pqcrystals-dilithium-SUF-CMA.patch]
```
bhess, line 65: The patch files listed here should be deleted.

bhess (commented):
Thanks @xuganyu96 for taking care of the Dilithium removal. I've added two comments inline.

cpp/sig_linking_test.cpp
xuganyu96: They are not related to Dilithium removal; I just added them so it is easier to build and run the linking test by itself. I can remove them if they are distracting or wrong.

bhess: Great, just wanted to confirm it wasn’t a merge conflict or something similar.

scripts/copy_from_upstream/copy_from_upstream.yml
bhess: Thx

dstebila:
> LGTM. Only thought (not meant to stop this PR): Worthwhile adding a HISTORY section somewhere stating the algorithms that OQS no longer supports (plus maybe a rationale for each alg (family) removal)?

This would be a long list and require a fair bit of digging back in Git history to properly reconstruct.  An alternative would be to have a clearly stated policy somewhere on what our algorithm inclusion policy is.  I wouldn't hold this PR for that.

baentsch:
>  I wouldn't hold this PR for that.

Fine. It would have been good to hold the merge to follow through our stated questionnaire, though: The "Yes" above meant that a fitting downstream PR should (have) be(en) ready by the time of merge. This did not happen. In consequence, all downstream PRs and "main" CI fail now :-( Time to do away with the questionnaire and go back to dealing with problems only as they arise? That's not really efficient, though and impacts many people (at least active maintainers and PR contributors).

dstebila:
> > I wouldn't hold this PR for that.
> 
> Fine. It would have been good to hold the merge to follow through our stated questionnaire, though: The "Yes" above meant that a fitting downstream PR should (have) be(en) ready by the time of merge. This did not happen. In consequence, all downstream PRs and "main" CI fail now :-( Time to do away with the questionnaire and go back to dealing with problems only as they arise? That's not really efficient, though and impacts many people (at least active maintainers and PR contributors).

I just didn't think to go back and check that before merging, my mistake; sorry. Maybe we can reformulate the text/questionnaire slightly -- Github's PR list will show how many checkboxes remain unchecked. So maybe it should be less of a questionnaire, and more of a "remove the lines that don't apply, and for the ones that do apply, leave the checkboxes unchecked until they are resolved"?

PR #2281, opened 2025-09-23
Fix typo in test_kem

### LINKED ISSUE #2280: Weekly CI failing due to code typo
### Description

A typo in test_kem.c is causing a build failure in some of the weekly CI jobs; see e.g. https://github.com/open-quantum-safe/liboqs/actions/runs/17886747773/job/50861396745

### Expected behaviour

Weekly CI should run cleanly

### liboqs version

HEAD

### Environment

CI

### Use of generative AI

_No response_

### Additional information

_No response_

dstebila:
Fixes #2280

### DIFF
--- a/tests/test_kem.c
+++ b/tests/test_kem.c
@@ -232,8 +232,8 @@ static OQS_STATUS kem_test_correctness(const char *method_name, bool derand) {
 		if (kem->length_encaps_seed > 0) {
 			OQS_randombytes(encaps_seed, kem->length_encaps_seed);
 		}
-		rc = OQS_KEM_encaps_derand(kem, ciphertext, shared_secret_e, public_key, encaps_seed)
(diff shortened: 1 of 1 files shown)


dstebila:
> The fix makes perfect sense. I wonder how this could slip through PR CI, though? Worth while investigating @dstebila @xuganyu96 ?

The second modified line of code in the PR is a macro that is only used for constant time testing, and thus defined away to nothing when compiled in baseline CI; but it is followed by a semicolon, so there was still a semicolon present after the first line of code from the compiler's perspective in baseline CI.  But when constant time testing was enabled in the weekly tests, that macro is rendered to actual text, which now means that the first line is missing a semicolon. I suppose in principle one could make a build configuration in baseline CI to try to catch this, but I don't think it's worth the effort.

PR #2283, opened 2025-09-24
Re-enable all ACVP tests

bhess:
It appears that #2237 inadvertently disabled running all ACVP tests in test_acvp_vectors.py, which I missed during review.

This PR re-enables running all ACVP tests.

### DIFF
--- a/tests/test_acvp_vectors.py
+++ b/tests/test_acvp_vectors.py
@@ -339,5 +339,4 @@ def test_acvp_vec_slh_dsa_sig_ver(sig_name):
 
 if __name__ == "__main__":
     import sys
-    test_acvp_vec_slh_dsa_sig_gen("SLH_DSA_SHA2_224_PREHASH_SHA2_128S")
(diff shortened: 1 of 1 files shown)


baentsch (approved):
Another case in point speaking [for more thorough than just more reviews](https://github.com/orgs/open-quantum-safe/discussions/2268)? As a thought for future improvement: (Automatically?) Check for commented code lines in PRs (possible at all?) as possible causes for problems.

dstebila:
Isn't it the case that pytest automatically grabs all methods starting with `test_` the relevant directory?  I thought main was only called if you ran the file directly from the command line, not the pytest harness. But regardless yes let's fix this line.

bhess:
> Isn't it the case that pytest automatically grabs all methods starting with `test_` the relevant directory? I thought main was only called if you ran the file directly from the command line, not the pytest harness. But regardless yes let's fix this line.

Yes, I think you're right, so all CI ran properly. I was running it directly from the command line when noticing it.

PR #2285, opened 2025-09-25
Fixes for Weekly Test Failures

### LINKED ISSUE #2287: Zephyr CI image disk size
[Zephyr tests have been failing lately](https://github.com/open-quantum-safe/liboqs/actions/runs/18102120087). The error message complains that the `zephyrproject-rtos/ci:v0.27.4` image takes up too much disk space.

I tried pulling this image on an AWS EC2 instance (Amazon Linux 2023). After the pull, `docker image ls` returns the following:

```
REPOSITORY                      TAG       IMAGE ID       CREATED         SIZE
hello-world                     latest    1b44b5a3e06a   7 weeks ago     10.1kB
ghcr.io/zephyrproject-rtos/ci   v0.27.4   3ec9ae3c7931   11 months ago   16.1GB
```

Since the standard GitHub Action runner only has [14 GB of disk space](https://docs.github.com/en/actions/reference/runners/github-hosted-runners), the disk size limit makes sense. What does not make sense it why this issue is coming up now. Further investigation is needed. This is probably a high priority since all future pull requests will fail.

bhess:
This PR addresses failures in the weekly test runs.

Issues discovered and fixed in the process:

1. Constant-time tests timing out because of:
- Repeated signature tests
test_sig performs repeated tests, which significantly slow down execution, especially for slower algorithms like SLH-DSA.
- All SLH-DSA variants were tested (156 variants, if I'm counting correctly). Many are now skipped (except the pure variants and one set of pre-hashed variants)
- No valgrind "suppression files" were defined for SLH-DSA. Added a blanket one under issues.json for future triaging.

3. Address Sanitizer failures with SLH-DSA
Failures are due to incorrect 'addrnd' byte requests in the SLH wrapper code:
- The SLH wrapper code always requested 32 bytes, regardless of the variant.
- According to FIPS 205, the correct 'addrnd' size is defined by parameter n (16, 24, or 32 bytes).
- This mismatch caused buffer overflows in ACVP tests, which supply custom RNG buffers sized to n.

4. Timeouts in ARM Emulated tests
- SLH-DSA variants were causing timeouts in ARM emulation.
- Mitigated by skipping SLH-DSA variants (except the pure variants and one set of pre-hashed variants)

5. Valgrind AVX512 incompatibility
- Github runners might now support AVX512. To avoid incompatibility (and errors) with valgrind in the CT-tests, I've set the `OPT_OPT_TARGET` to `haswell` instead of `auto`, to avoid that AVX512 code is generated.

Fixes #2287

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -12,8 +12,6 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          # macos-13 runs on x64; the others run on aarch64
-          - macos-13
           - macos-14
           - macos-15
         CMAKE_ARGS:


bhess:
It looks like the Zephyr builds have started failing recently with a "no space left on device" error during container loading. I haven’t found any direct link to changes in liboqs, and it’s not just the CI in this PR that’s affected. @Frauschi, do you happen to have any insights into what might be causing this?

xuganyu96:
@dstebila @bhess see #2287 for CI failures

dstebila:
@ryjones Have you seen any weirdness in other proejcts lately around Github CI failures involving the Zephyr real-time operating system, or disk size limits? We're trying to diagnose some weird failures that popped up here seemingly without any changes on our side.

ryjones:
@dstebila what I could do is make a large runner with more disk to try it out on. is this arm64 or x86-64?

dstebila:
> @dstebila what I could do is make a large runner with more disk to try it out on. is this arm64 or x86-64?

The [relevant Github workflow file](https://github.com/open-quantum-safe/liboqs/blob/dd942d434463872193ba282250532ac0be15f1e3/.github/workflows/zephyr.yml#L11C5-L11C26) has `runs-on: ubuntu-22.04`. I think this means x86-64? IIRC it would explicitly say if it was an arm64 image, right?

ryjones:
@bhess take a look at https://github.com/open-quantum-safe/liboqs/pull/2289/files - you also need to fix the actionlint.yml file to allow this runner type, or skip this check

bhess:
> @bhess take a look at https://github.com/open-quantum-safe/liboqs/pull/2289/files - you also need to fix the actionlint.yml file to allow this runner type, or skip this check

Thanks @ryjones . This seems to work! See https://github.com/open-quantum-safe/liboqs/pull/2285/commits/2aa95f1fea64619554a6cfc6d68395d382f2c85a with the updated actionlint.yaml.

ryjones:
@bhess I see [macos-13](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/) has been deprecated. [I suggest removing it](https://github.com/open-quantum-safe/liboqs/blob/2aa95f1fea64619554a6cfc6d68395d382f2c85a/.github/workflows/macos.yml#L16) to prevent failures during brownouts etc.

bhess:
Removed the runs on macos-13 as suggested by @ryjones, and all CI is passing now.

dstebila (approved):

ryjones:
not for this PR, but the push tests and pr tests really need to see each other as we're running every test twice, and given these are large runners, it is burning 2x the money needed :)

PR #2286, opened 2025-09-26
More readable algorithm support in README.md

xuganyu96:
# Updated algorithm support table in README.md
The existing README.md lists the set of supported algorithms in a flat list. This list lacks certain important information such as standardization status and source of implementation. It is also difficult to read, especially for algorithms with many variants.

This PR updates the Python script for rendering the algorithm support section in README.md. This effort originated from #2045 but did not fully accomplish the goal set out in the issue, so this PR will NOT close the issue.

**Summary of changes**:
- Two data fields `standardization-status` and `spec-url` have been added to each algorithm's YAML datasheet under `docs/algorithms`. The only exception is `slh_dsa.yaml`, which generated from a Jinja2 template, so the template is also modified.
- A new Python script `update_alg_support_table.py`. This script is not meant to run by itself; instead it provides functions that are invoked in `update_docs_from_yaml.py`, which in turn is invoked by `copy_from_upstream.py`. I made it a separate script instead of adding to existing scripts because I want to maintain higher code readability, but I don't want to change existing scripts. (FYI I used a code formatter)
- The modified README

**Some additional notes**:
- I'd appreciate a second pair of eyes on `spec-url`. Some of them have multiple sources (the same spec can be sourced from NIST website and from the submitter's website), and some of them have updated versions not submitted to NIST (for example HQC, CROSS)
- I am not sure how to label Kyber and SPHINCS+ so as to not create confusion with ML-KEM and SLH-DSA. Should NTRU/NTRU-Prime be considered "Not selected", or should we reference NTT/Japanese efforts to standardize that?
- I did not link each algorithm family to the corresponding markdown document. This is because Doxygen cannot handle these links. Instead, the "status" section contains instruction about "details can be found under `docs/algorithms`"


docs/algorithms/kem/kyber.yml
```diff
@@ -15,6 +15,8 @@ auxiliary-submitters:
 crypto-assumption: Module LWE+R with base ring Z[x]/(3329, x^256+1)
 website: https://pq-crystals.org/
 nist-round: 3
+standardization-status: Selected by NIST
```
baentsch: I'd disagree with this wording: AFAIK, Kyber (round 3 submission) is not identical with ML-KEM. So what about "Selected by NIST as the basis for standardization in ML-KEM (see below)"?

docs/algorithms/sig/falcon.yml
```diff
@@ -16,6 +16,8 @@ auxiliary-submitters:
 crypto-assumption: hardness of NTRU lattice problems
 website: https://falcon-sign.info
 nist-round: 3
+standardization-status: Selected by NIST
```
baentsch: Again, don't agree as-is. What about "Selected by NIST as basis for upcoming standardization"

docs/algorithms/sig/sphincs.yml
```diff
@@ -23,6 +23,8 @@ auxiliary-submitters:
 crypto-assumption: hash-based signatures
 website: https://sphincs.org/
 nist-round: 3
+standardization-status: Selected by NIST
```
baentsch: Same comment as above.... "basis for...."

docs/algorithms/kem/ntru.yml
```diff
@@ -16,6 +16,9 @@ auxiliary-submitters:
 - Zhenfei Zhang
 crypto-assumption: NTRU in Z[x]/(q, x^n-1) with prime n and power-of-two q
 website: https://ntru.org/
+# TODO: under NTT (Nippon Telegraph and Telephone) consideration?
+standardization-status: Not selected
```
baentsch: So why not write "Not selected by NIST. Under consideration for standardization by NTT" (link available?)?

xuganyu96:
@baentsch Thank you for the feedback. I've rephrased standardization statuses for the various Round 3 candidates since they are the most confusing.

For Kyber/SPHINCS+, the status will say "Selected by NIST as basis for ML-KEM/SLH-DSA"
For Falcon/HQC, it will say "Selected by NIST for upcoming standardization"
For NTRU, it will say "Not selected by NIST, under standardization consideration by NTT", with links to NIST and NTT websites respectively
For NTRU Prime, it will say "Not selected by NIST". From NTT's website it looks like NTT is only working with NTRU, not NTRU Prime

docs/algorithms/kem/bike.yml
```diff
@@ -21,6 +21,7 @@ principal-submitters:
 crypto-assumption: QC-MDPC (Quasi-Cyclic Moderate Density Parity-Check)
 website: http://bikesuite.org/
 nist-round: 4
+standardization-status: Not selected by [NIST](https://csrc.nist.gov/csrc/media/Projects/post-quantum-cryptography/documents/round-4/submissions/BIKE-Round4.zip)
```
bhess: Since the spec version referenced below is 5.1, it might be more appropriate to reference the corresponding version of the specification (https://bikesuite.org/files/v5.1/BIKE_Spec.2022.10.10.1.pdf)

docs/algorithms/sig/sphincs.yml
```diff
@@ -23,6 +23,8 @@ auxiliary-submitters:
 crypto-assumption: hash-based signatures
 website: https://sphincs.org/
 nist-round: 3
+standardization-status: Selected by [NIST](https://csrc.nist.gov/CSRC/media/Projects/post-quantum-cryptography/documents/round-3/submissions/SPHINCS-Round3.zip)
```
bhess: The spec of version 3.1 (as indicated below) is available here: https://sphincs.org/data/sphincs+-r3.1-specification.pdf. The NIST page might still refer to version 3.0

docs/algorithms/kem/frodokem.yml
```diff
@@ -17,6 +17,7 @@ crypto-assumption: learning with errors (LWE)
 website: https://frodokem.org/
 nist-round: 3
 spec-version: NIST Round 3 submission
+standardization-status: Not selected by [NIST](https://csrc.nist.gov/CSRC/media/Projects/post-quantum-cryptography/documents/round-3/submissions/FrodoKEM-Round3.zip)
```
dstebila:
```suggestion
standardization-status: Under [ISO](https://frodokem.org/) consideration
```

baentsch: I'm not sure I understand the linkage of the word "ISO" to the website frodokem.org: Can someone please explain? Pointing to frodokem.org is something I'd expect for "spec version" -- but then a specific pointer to the actual spec documented at frodokem.org would be more suitable for users than a generic top-level link. Regarding standardization in turn, I'd expect a  specific iso.org location users can read up on the current standardization state of FrodoKEM.

dstebila: The ISO draft document is behind a paywall (https://www.iso.org/standard/86890.html).  We as the FrodoKEM team post our version of the FrodoKEM part of the document on the frodokem.org website.

PR #2289, opened 2025-09-30
chore: move CI to large runner for Zephyr

ryjones:

### DIFF
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -8,7 +8,7 @@ on: [workflow_call, workflow_dispatch]
 jobs:
 
   zephyr_test:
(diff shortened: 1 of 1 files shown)


.github/workflows/zephyr.yml
```diff
@@ -8,7 +8,7 @@ on: [workflow_call, workflow_dispatch]
 jobs:
 
   zephyr_test:
-    runs-on: ubuntu-22.04
+    runs-on: oqs-x64
     container: ghcr.io/zephyrproject-rtos/ci:v0.27.4
```
LadyE215, line 12: https://github.com/open-quantum-safe/liboqs/pull/2285

PR #2296, opened 2025-10-08
update uov_ov_iii_pkc_skc to use its own verify function

openforge-kt:
UOV_OV_III_PKC_SKC algorithm currently uses the verify function of UOV_OV_III_PKC instead of it's own (which already exists), causing per-algorithm builds to fail unless it's built together with the entire UOV family or UOV_OV_III_PKC, this just changes it to use it's own verify function instead.

### DIFF
--- a/src/sig/uov/sig_uov_ov_III_pkc_skc.c
+++ b/src/sig/uov/sig_uov_ov_III_pkc_skc.c
@@ -34,7 +34,7 @@ OQS_SIG *OQS_SIG_uov_ov_III_pkc_skc_new(void) {
 
 extern int pqov_uov_III_pkc_skc_ref_keypair(uint8_t *pk, uint8_t *sk);
 extern int pqov_uov_III_pkc_skc_ref_signature(uint8_t *sig, size_t *siglen, const uint8_t *m, size_t mlen, const uint8_t *sk);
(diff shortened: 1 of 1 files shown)


openforge-kt:
Ah I see, I had a misunderstanding of how all this works, sorry about that, I think this PR should be closed, and the fix needs to be applied to the source project, more specifically, https://github.com/pqov/pqov/blob/main/integration/liboqs/OV-III-pkc-skc_META.yml#L27 it looks to me like this was just an oversight, I can make a PR if @mkannwischer wishes, but if my understanding is correct, this is the only place where the error occurs. If they agree, this can be closed

mkannwischer:
Good catch - this is a small mistake I made when writing the META files: https://github.com/pqov/pqov/blob/0ec8862a073646c0e8c3173fe38fa54453ad1a0c/integration/liboqs/OV-III-pkc-skc_META.yml#L27

Note that the verification of pkc is the same as for pkc_skc (the skc stands for secret key compression which only affects signing + keygen). That's why this was not caught by any tests. 
It should still be fixed. I'll fix it upstream and then re-import shortly.

baentsch:
@openforge-kt OK to close this as #2299 should contain the fix (a second pair of eyes on that PR welcome!). Thanks @mkannwischer for the fix.

PR #2298, opened 2025-10-12
Add build option for randomized ML-DSA signing

M-AlNoaimi:
This PR adds an option to enable randomized signing for ML-DSA, as requested in [open-quantum-safe/liboqs-python#128](https://github.com/open-quantum-safe/liboqs-python/issues/128).

I've added a `OQS_ENABLE_SIG_ML_DSA_RANDOMIZED_SIGNING` build flag (defaulting to `OFF`) and a conditional test to verify that signatures are non-deterministic when the flag is enabled.

A small update to the CONFIGURE.md was done as well to note the flag.

### DIFF
--- a/tests/test_sig.c
+++ b/tests/test_sig.c
@@ -296,7 +296,8 @@ static OQS_STATUS sig_test_repeat_signing(const char *method_name, bool is_rando
 	}
 
 	if (signature1_len != signature2_len) {
-		fprintf(stderr, "ERROR: Signatures have different lengths.\n");
+		fprintf(stderr, "ERROR: %s signing for %s produced signatures of different lengths (%zu vs %zu). This should not happen.\n",
+		        is_randomized ? "Randomized" : "Deterministic", sig->method_name, signature1_len, signature2_len);
 		goto err;
 	}
 


ashman-p (dismissed):

CONFIGURE.md
```diff
@@ -62,7 +62,13 @@ To enable `XMSS` stateful signature, set `OQS_ENABLE_SIG_STFL_XMSS` to `ON`, the
 
 For a full list of such options and their default values, consult [.CMake/alg_support.cmake](https://github.com/open-quantum-safe/liboqs/blob/master/.CMake/alg_support.cmake).
 
-**Default**: Unset.
```
baentsch, line 65: Why is this deleted?

tests/test_sig.c
```diff
@@ -244,6 +244,101 @@ static OQS_STATUS sig_test_correctness(const char *method_name, bool bitflips_al
...
+	uint8_t *signature1 = NULL;
+	size_t signature1_len;
+	uint8_t *signature2 = NULL;
+	size_t signature2_len;
+	OQS_STATUS rc, ret = OQS_ERROR;
+
+	sig = OQS_SIG_new(method_name);
+	if (sig == NULL) {
+		// ML-DSA is not enabled, so we can't test it.
+		return OQS_SUCCESS;
```
baentsch: I conceptually don't like a test to succeed if an API call errors. Can you please add a (CI) test for this feature so I can understand why this is necessary in this way?

tests/test_sig.c
```diff
@@ -349,6 +444,16 @@ int main(int argc, char **argv) {
 	}
 #endif
 
+#if defined(OQS_ENABLE_SIG_ML_DSA_RANDOMIZED_SIGNING)
+	if (strncmp(alg_name, "ML-DSA", 6) == 0) {
+		rc = sig_test_randomized_signing(alg_name);
```
baentsch: Ahh, here's the test. But then I do not understand why the absence of ML-DSA is not tested for here such as to make the "succeed-on-error" condition above go away (?) Actually, I wonder how it can happen at all given this define can only be set if ML-DSA is present.

baentsch:
Oops -- seeing just now that @ashman-p approved this PR. In that case I assume I am unaware of some things and am willing to retract my review: In order for me to do so, would you please address my questions, Norm such as to not have my possible lack of understanding cause undue work to the PR's author?

CONFIGURE.md
M-AlNoaimi: I tried to follow similar styling as the rest of the options, I think the line in question was a placeholder?

tests/test_sig.c
M-AlNoaimi: Good call, I did think a CI test would probably be a good idea but I didn't wanna add to the already huge amount of GH actions. If you think it should be added, I can have a go at it.

tests/test_sig.c
M-AlNoaimi: Yeah that's another good note, are you good with removing the succeed on error path and instead assert that ML-DSA is present if the test is compiled? Or what would you prefer?

M-AlNoaimi:
> Thanks for the PR @M-AlNoaimi -- I'd be glad to get feedback to my single comments before approving this. And then I have 2 basic questions: 1) How and where does this change the actual logic in ML-DSA?

These changes only introduces a build time option and a test, AFAIK the signing logic is only affected if the new option is enabled, in which case the ML-DSA impl will use randomized signing.

> How shall users of the library recognize whether this config option has been set? Doing it this way seems like inviting a maintenance nightmare ("User: ML-DSA doesn't work! Maintainer: Which build flag did you set? User: How should I know? I only have a binary.")

I guess right now the only way they'd be able to tell was at build time, but I don't think it should be much of an issue if the default behaviour is set to OFF, and the new option keeps it that way as suggested on [liboqs-python #128](https://github.com/open-quantum-safe/liboqs-python/issues/128#issuecomment-3292318378.) So I think if it's off by default, current usage would stay the same.

bhess:
Thanks for the PR!

> 1. How and where does this change the actual logic in ML-DSA?

I have the same question. From what I can tell, the OQS_ENABLE_SIG_ML_DSA_RANDOMIZED_SIGNING macro isn’t currently used in the ML-DSA implementation to control the switch between randomized and deterministic signing.

As far as I understand, the ML-DSA implementation already defaults to randomized signing. Is there any evidence that this wasn't the case?
The behavior is controlled by the following parts of the code:

https://github.com/open-quantum-safe/liboqs/blob/52169a1edfaf8b5f1593ef3a002e2553f33271a0/src/sig/ml_dsa/pqcrystals-dilithium-standard_ml-dsa-65_ref/config.h#L5
and https://github.com/open-quantum-safe/liboqs/blob/52169a1edfaf8b5f1593ef3a002e2553f33271a0/src/sig/ml_dsa/pqcrystals-dilithium-standard_ml-dsa-65_ref/sign.c#L229-L234

tests/test_sig.c
baentsch: > instead assert that ML-DSA is present if the test is compiled

that sounds sensible.

CONFIGURE.md
baentsch: Not really: It documents (documented) the default of all options "OQS_ENABLE_KEM_ALG/OQS_ENABLE_SIG_ALG/OQS_ENABLE_SIG_STFL_ALG"

baentsch:
> So I think if it's off by default, current usage would stay the same.

Correct. But this PR makes the option available (thanks for the latest commit making it _actually_ available!), so this can be used to change the behaviour of the library and there should be a way to query this also in a binary. Otherwise this invites a maintenance nightmare.

tests/test_sig.c
```diff
@@ -244,6 +244,101 @@ static OQS_STATUS sig_test_correctness(const char *method_name, bool bitflips_al
 	return ret;
 }
 
+#if defined(OQS_ENABLE_SIG_ML_DSA_RANDOMIZED_SIGNING)
+// Test that two signatures of the same message are different.
+static OQS_STATUS sig_test_randomized_signing(const char *method_name) {
```
bhess: I like this test, but it would be good to extend it to also cover the deterministic mode. Specifically, verifying that signing the same message twice produces identical signatures. This would ensure the OFF switch is functioning as intended.

tests/test_sig.c
```diff
@@ -298,12 +296,24 @@ static OQS_STATUS sig_test_randomized_signing(const char *method_name) {
 	}
 
 	if (signature1_len != signature2_len) {
-		printf("Signatures have different lengths, so they are not identical.\n");
-	} else if (memcmp(signature1, signature2, signature1_len) == 0) {
-		fprintf(stderr, "ERROR: Two signatures of the same message are identical.\n");
+		fprintf(stderr, "ERROR: Signatures have different lengths.\n");
```
baentsch: Misleading error message

tests/helpers.py
```diff
@@ -193,6 +193,17 @@ def available_use_options_by_name():
 def is_use_option_enabled_by_name(name):
     return name in available_use_options_by_name()
 
+def is_ml_dsa_randomized_signing_enabled():
```
baentsch, line 196: This approach feels "brittle" (but I may be paranoid here): For example, would it be possible to build everything with "-DOQS_ENABLE_SIG_ML_DSA_RANDOMIZED_SIGNING" (passed to the compiler during a re-build)? If so, this function will fail to return the right value. And/or would it be sensible to have `cmake` variable and `C` define have different names instead? And then there's still the possibility that header file and binary under test don't match up, e.g., because of some user-specific LD_LIBRARY_PATH setup... 
I still feel it would be good for users to have a way (e.g., static variable) to query which way (randomized or deterministic) the library under test/use has been built; this test function could use the same facility to pick which way to test (and test the user-level "build-type" facility at the same time).

baentsch:
@M-AlNoaimi fyi we discussed this PR in our status call and agreed to not land this as-is as it is using an algorithm code base that neither OQS nor the upstream supports any longer, namely "PQCrystals".

That said, we would be very glad if you could raise a new PR adding this capability using the ML-DSA code base OQS does support, namely mldsa-native: Would you be willing to do this? Would you need any support beyond some of the still valid comments above?

M-AlNoaimi:
> That said, we would be very glad if you could raise a new PR adding this capability using the ML-DSA code base OQS does support, namely mldsa-native: Would you be willing to do this? Would you need any support beyond some of the still valid comments above?

Hi @baentsch, thanks for the clarification and for raising this in the status call. I understand the concern with the current state, I agree this should be implemented against the mldsa-native implementation instead.

The earlier comments were very helpful, i’ll take those into account when working on the new PR. I’ll start very soon, thanks again!

xuganyu96:
@M-AlNoaimi Since `pqcrystals-dilithium-standard_ml-dsa-xx` implementations have all been deprecated in favor of `mldsa-native`, the code change in this pull request is no longer relevant. Would you be okay with closing this pull request? 

If you feel inclined to adapt your work (explicit "deterministic signing" and/or "randomized signing" API) to the new implementations, you can raise a separate pull request.

PR #2299, opened 2025-10-14
Re-import PQOV after fixing OV-III-pkc_skc integration

mkannwischer:
This re-imports PQOV after fixing a wrong reference: OV-III-pkc_skc was using the OV-III-pkc verification routine.
The reason that this worked is that skc does not impact verification (pkc_skc has the same verification routine as pkc).

See #2296 

The other changes that were made upstream (the OpenSSL stuff), are dead code for the OQS integration as we are using the OQS hashing functions.


baentsch:
> The other changes that were made upstream (the OpenSSL stuff), are dead code for the OQS integration as we are using the OQS hashing functions.

Thanks for pointing that out @mkannwischer - I already wondered. May I ask why this code then gets copied over to OQS in the first place? Wouldn't it be possible to just drop these whole files? Would make reviewing here easier.

mkannwischer:
> > The other changes that were made upstream (the OpenSSL stuff), are dead code for the OQS integration as we are using the OQS hashing functions.
> 
> Thanks for pointing that out @mkannwischer - I already wondered. May I ask why this code then gets copied over to OQS in the first place? Wouldn't it be possible to just drop these whole files? Would make reviewing here easier.

It's not the entire file that is dead code. There is a pre-processor conditional
```c
#elif defined(_UTILS_OQS_)
#include <oqs/sha3.h>

int hash_init( hash_ctx *ctx ) {
    #if defined(_HASH_SHAKE128_)
    OQS_SHA3_shake128_inc_init( ctx );
    #else
    OQS_SHA3_shake256_inc_init( ctx );
    #endif
    return 0;
}
...
```
that glues in the OQS SHA-3 API.
I'll give it some thought if we can do this cleaner in the next iteration.

PR #2303, opened 2025-10-22
Remove ACVP test vectors from repository

### LINKED ISSUE #2279: Fetch ACVP test vectors from source instead of storing them
ACVP test vectors stored under `tests/ACVP_Vectors` are fetched from NIST repositories on GitHub using the `fetch_values.sh` script. Since we do not own and are not responsible for theses vectors, we should not commit them into this repository. Storing large bulk data files in a git repository also negatively affects git performance. Instead, we should fetch the test vectors as needed during test runtime.

xuganyu96:
This PR closes #2279 

ACVP test vectors will now be fetched from `https://raw.githubusercontent.com/usnistgov/ACVP-Server/refs/tags/v1.1.0.40/gen-val/json-files/` at test time using Python's `requests` library. Previously stored JSON files and `fetch_values.sh` scripts were removed. I also ran a [Python code formatter](https://github.com/psf/black).

Some considerations for reviewers:
- Speaking from principles, external resources (i.e. the JSON file) *should* be set up as `pytest.fixture` so that a single instance of a resource is not fetched repeatedly across tests. However, from past experience I found fixtures to be a pain to debug, and all but one JSON file is used exactly once, so I chose to not use fixtures
- Currently the JSON files are fetched from the website instead of [GitHub API](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28). Should this be a concern?

* [**NO**] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [**NO**] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged. Also, make sure to update the list of algorithms in the continuous benchmarking files: .github/workflows/kem-bench.yml and sig-bench.yml)

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -142,29 +142,31 @@ def is_sig_stfl_enabled_by_name(name):
                     return True
     return False
 
+# TODO: this function needs a refactor because the scan logic currently applies
+# re.findall to all args/kwargs in the wrapped function. If the function takes
+# a non-string argument, re.findall will raise a TypeError, which is
+# undesirable. It can also cause incorrect skipping if the wrapped function
+# takes some string argument that unintentionally matches the regex.
+# Instead, I prefer replacing this decorator with a regular function that returns
+# a boolean indicating whether a KEM/SIG/STFL_SIG name matches the env var
+# SKIP_ALGS, then let the caller decide whether to call pytest.skip.
 def filtered_test(func):
-    funcname = func.__name__[len("test_") :]
+    funcname = func.__name__[len("test_"):]
 
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
-        if ("SKIP_ALGS" in os.environ) and len(os.environ["SKIP_ALGS"]) > 0:
-            for algexp in os.environ["SKIP_ALGS"].split(","):
+        if ('SKIP_ALGS' in os.environ) and len(os.environ['SKIP_ALGS'])>0:
+            for algexp in os.environ['SKIP_ALGS'].split(','):
                 for arg in args:
-                    if isinstance(arg, str) and len(re.findall(algexp, arg)) > 0:
+                    if len(re.findall(algexp, arg))>0:
                         pytest.skip("Test disabled by alg filter")
                 for arg in kwargs:
-                    if (
-                        isinstance(kwargs[arg], str)
-                        and len(re.findall(algexp, kwargs[arg])) > 0
-                    ):
+                    if len(re.findall(algexp, kwargs[arg]))>0:
                         pytest.skip("Test disabled by alg filter")
-        if ("SKIP_TESTS" in os.environ) and (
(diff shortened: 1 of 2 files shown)


baentsch (dismissed):
From just looking at the code, this seems OK to me. Thanks @xuganyu96 !

> I chose to not use fixtures

Sounds sensible to me.

> Currently the JSON files are fetched from the website instead of [GitHub API](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28). Should this be a concern?

My only concern would have been that that website may be unreachable when CI runs -- but given this is also on GH and not some US gov server (that may be subject to shutdown (in various meanings of the word :), I'd say it's highly unlikely for this to be a problem. My take for now: Let's see how this performs and whether we need to improve later (eg, using [GH caching](https://docs.github.com/en/actions/concepts/workflows-and-actions/dependency-caching)).

ashman-p:
I also share @baentsch concerns about NIST server accessibility. It adds another variable to trouble shoot in case of a test failure.

xuganyu96:
> I also share @baentsch concerns about NIST server accessibility. It adds another variable to trouble shoot in case of a test failure.

The test vectors are hosted on GitHub, not NIST server. I am open to discuss backup solutions, but for now I consider the cost of maintaining local copies of test vector to significantly outweigh the risk of test vectors becoming unavailable.

xuganyu96:
> I also share @baentsch concerns about NIST server accessibility. It adds another variable to trouble shoot in case of a test failure.

The test vector JSON files are currently hosted on GitHub, not NIST website. It is true that GitHub might go down, but I think that this risk is insignificant, especially since if GitHub goes down, GitHub Actions will likely go down, in which case we are running into much bigger problems than a few ACVP test vectors being offline.

cc: @ashman-p @baentsch @dstebila

ashman-p:
> > I also share @baentsch concerns about NIST server accessibility. It adds another variable to trouble shoot in case of a test failure.
> 
> The test vector JSON files are currently hosted on GitHub, not NIST website. It is true that GitHub might go down, but I think that this risk is insignificant, especially since if GitHub goes down, GitHub Actions will likely go down, in which case we are running into much bigger problems than a few ACVP test vectors being offline.
> 
> cc: @ashman-p @baentsch @dstebila

Thanks, Bruce. I understand now. I have no further issues with this.

xuganyu96:
@open-quantum-safe/liboqs-committers @open-quantum-safe/oqs-committers  This pull request is getting a bit long in the tooth. A second review would be much appreciated.

dstebila (approved):

Issue #2304, opened 2025-10-23
Remove Alpine support -- or add it throughout the CI stack


baentsch:
As mentioned in https://github.com/open-quantum-safe/ci-containers/pull/97#pullrequestreview-3368190792, why does `liboqs` still support Alpine in CI? Wouldn't it suffice to only run on Ubuntu? This would
a) be in line with what is stated in PLATFORMS.md
b) eliminate the need to improve the ci-containers workflow to include building alpine docker images (currently not done, necessitating manual work to update the alpine images, e.g., to make CI pass on https://github.com/open-quantum-safe/liboqs/pull/2303).
c) generally ease contribution and maintenance

As git-blame points to @dstebila as the person introducing Alpine 4 years ago, can you maybe comment? Would you agree it'd be easier to use just using one docker image for all tests? Or is there a reason for prioritizing Alpine over Ubuntu in some tests? In that case, we should update PLATFORMS.md and the build-and-push logic in ci-containers.

baentsch:
Tagging @dstebila @xuganyu96 for input. If none is received, I'd suggest changing all CI jobs over to drop anything other than Ubuntu:latest.

hippi777:
alpine uses musl, while ubuntu is glibc only... musl is nearer to standards (posix/whatever else?), so it can be good for compatibility but statically linking musl or any other c lib on ubuntu should yield the same results (i believe), even if anything else uses a dynamically linked glibc... also, alpine is small, so even ubuntu could be questioned... :D

(note that im not much familiar with this code base, im just around for learning and seeing news...)

xuganyu96:
> alpine uses musl, while ubuntu is glibc only... musl is nearer to standards (posix/whatever else?), so it can be good for compatibility but statically linking musl or any other c lib on ubuntu should yield the same results (i believe), even if anything else uses a dynamically linked glibc... also, alpine is small, so even ubuntu could be questioned... :D
> 
> (note that im not much familiar with this code base, im just around for learning and seeing news...)

Seems like a fair point. Perhaps this is the reason why Alpine was added to the list of test targets in the beginning.

[...]

PR #2305, opened 2025-10-23
[DO NOT MERGE] Remove platform tests running on Alpine Linux

xuganyu96:
Alpine Linux is not listed in PLATFORMS.md, and the CI pipeline over in the `ci-containers` repository does not automatically build/push the Alpine Linux CI image. Perhaps it is time to sunset this set of tests.

See #2304 and [`ci-containers` issue 78](https://github.com/open-quantum-safe/ci-containers/issues/78)


xuganyu96:
> That's a bit "drastic": Have you checked we don't lose test coverage with this change? If we'd do, a better approach would be to replace ci-alpine-amd64 with ubuntu:latest, no?

I agree with this assessment. I will edit the PR to indicate that this is not meant to be merged, only for evaluation purposes only.

PR #2310, opened 2025-10-30
Add missing LMS macros. Correct typo.

### LINKED ISSUE #2121: Many LMS variants disabled with -DOQS_ENABLE_SIG_STFL_LMS=ON
**Describe the bug**
I am trying to benchmark statefull hash-based signature schemes with the existing speed_sig_stfl benchmark. However, a lot of LMS-variants are disabled - even when using CMake options as described.

**Expected behavior**
speed_sig_stfl should support all implemented LMS versions

**Screenshots**
./speed_sig_stfl --algs
XMSS-SHA2_10_256
XMSS-SHA2_16_256
XMSS-SHA2_20_256
XMSS-SHAKE_10_256
XMSS-SHAKE_16_256
XMSS-SHAKE_20_256
XMSS-SHA2_10_512
XMSS-SHA2_16_512
XMSS-SHA2_20_512
XMSS-SHAKE_10_512
XMSS-SHAKE_16_512
XMSS-SHAKE_20_512
XMSS-SHA2_10_192
XMSS-SHA2_16_192
XMSS-SHA2_20_192
XMSS-SHAKE256_10_192
XMSS-SHAKE256_16_192
XMSS-SHAKE256_20_192
XMSS-SHAKE256_10_256
XMSS-SHAKE256_16_256
XMSS-SHAKE256_20_256
XMSSMT-SHA2_20/2_256
XMSSMT-SHA2_20/4_256
XMSSMT-SHA2_40/2_256
XMSSMT-SHA2_40/4_256
XMSSMT-SHA2_40/8_256
XMSSMT-SHA2_60/3_256
XMSSMT-SHA2_60/6_256
XMSSMT-SHA2_60/12_256
XMSSMT-SHAKE_20/2_256
XMSSMT-SHAKE_20/4_256
XMSSMT-SHAKE_40/2_256
XMSSMT-SHAKE_40/4_256
XMSSMT-SHAKE_40/8_256
XMSSMT-SHAKE_60/3_256
XMSSMT-SHAKE_60/6_256
XMSSMT-SHAKE_60/12_256
LMS_SHA256_H5_W1
LMS_SHA256_H5_W2
LMS_SHA256_H5_W4
LMS_SHA256_H5_W8
LMS_SHA256_H10_W1
LMS_SHA256_H10_W2
LMS_SHA256_H10_W4
LMS_SHA256_H10_W8
LMS_SHA256_H15_W1
LMS_SHA256_H15_W2
LMS_SHA256_H15_W4
LMS_SHA256_H15_W8 (disabled)
LMS_SHA256_H20_W1 (disabled)
LMS_SHA256_H20_W2 (disabled)
LMS_SHA256_H20_W4 (disabled)
LMS_SHA256_H20_W8 (disabled)
LMS_SHA256_H25_W1 (disabled)
LMS_SHA256_H25_W2 (disabled)
LMS_SHA256_H25_W4 (disabled)
LMS_SHA256_H25_W8 (disabled)
LMS_SHA256_H5_W8_H5_W8
LMS_SHA256_H10_W4_H5_W8
LMS_SHA256_H10_W8_H5_W8 (disabled)
LMS_SHA256_H10_W2_H10_W2 (disabled)
LMS_SHA256_H10_W4_H10_W4 (disabled)
LMS_SHA256_H10_W8_H10_W8 (disabled)
LMS_SHA256_H15_W8_H5_W8 (disabled)
LMS_SHA256_H15_W8_H10_W8 (disabled)
LMS_SHA256_H15_W8_H15_W8 (disabled)
LMS_SHA256_H20_W8_H5_W8 (disabled)
LMS_SHA256_H20_W8_H10_W8 (disabled)
LMS_SHA256_H20_W8_H15_W8 (disabled)
LMS_SHA256_H20_W8_H20_W8 (disabled)

 - OS: Ubuntu Server 24.04.2 LTS
 - OpenSSL version: 3.0.13
 - Compiler version used: cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
 - Build variables used: -GNinja -DOQS_ENABLE_SIG_STFL_XMSS=ON -DOQS_ENABLE_SIG_STFL_LMS=ON -DOQS_HAZARDOUS_EXPERIMENTAL_ENABLE_SIG_STFL_KEY_SIG_GEN=ON
 - liboqs version: main

**Additional context**

ashman-p:
Fixes #2121

### DIFF
--- a/tests/test_kat.py
+++ b/tests/test_kat.py
@@ -45,7 +45,7 @@ def test_sig_stfl(sig_stfl_name):
     katfile = helpers.get_katfile("sig_stfl", sig_stfl_name)
     if not katfile: pytest.skip("KATs file is missing")
     if not os.path.exists(katfile): # Skip if the file does not exist
-        pytest.skip(f'File not found: {katfile}')
+        pytest.xfail(f'KAT not available: {sig_stfl_name}')
     output = helpers.run_subprocess(
         [helpers.path_to_executable('kat_sig_stfl'), sig_stfl_name, katfile],
     )


tests/test_cmdline.py
```diff
@@ -40,11 +53,16 @@ def test_sig_stfl(sig_stfl_name):
...
+        if sig_stfl_name not in names_to_allow:
+            pytest.skip('Test skipped')
+        else:
+            helpers.run_subprocess(
+                [helpers.path_to_executable('test_sig_stfl'), sig_stfl_name],
+            )
 
 if __name__ == "__main__":
     import sys
+    print("NMA Debug In main\n")
```
ashman-p:
```suggestion
```

tests/test_cmdline.py
```diff
@@ -40,11 +53,16 @@ def test_sig_stfl(sig_stfl_name):
...
+    print (sys.argv)
```
ashman-p:
```suggestion
```

tests/test_kat.py
```diff
@@ -44,6 +44,8 @@ def test_sig_stfl(sig_stfl_name):
     if not(helpers.is_sig_stfl_enabled_by_name(sig_stfl_name)): pytest.skip('Not enabled')
     katfile = helpers.get_katfile("sig_stfl", sig_stfl_name)
     if not katfile: pytest.skip("KATs file is missing")
+    if not os.path.exists(katfile): # Skip if the file does not exist
+        pytest.skip(f'File not found: {katfile}')
```
baentsch: Is this good test practice? IMO a test should fail if a precondition fails (KAT file present in this case). This logic here seems to let the test pass instead (by skipping a test without KAT file). Skipping tests makes sense if an algorithm indeed is unavailable (3 lines above).

tests/test_speed.py
```diff
@@ -23,11 +23,26 @@ def test_sig(sig_name):
 @helpers.filtered_test
 @pytest.mark.parametrize('sig_stfl_name', helpers.available_sig_stfls_by_name())
 def test_sig(sig_stfl_name):
+    # Define the list of LMS varients to allow
```
baentsch, line 26: This seems to be a duplicate of the list in test_cmdline. Suggest to move to a common file to ensure both stay aligned/there is only one list to maintain.

tests/test_speed.py
```diff
@@ -23,11 +23,26 @@ def test_sig(sig_name):
...
+                        "LMS_SHA256_H15_W2",
+                        "LMS_SHA256_H15_W4",
+                        "LMS_SHA256_H5_W8_H5_W8"
+                        "LMS_SHA256_H10_W4_H5_W8",]
     kats = helpers.get_kats("sig_stfl")
     if not(helpers.is_sig_stfl_enabled_by_name(sig_stfl_name)):
         pytest.skip('Not enabled')
     elif sig_stfl_name.find("_10")==-1 and sig_stfl_name.find("H10")==-1:
         pytest.skip('Test skipped')
+    elif sig_stfl_name not in names_to_allow:
```
baentsch, line 44: If I get this right, this basically means, no SHBS alg other than the ones listed above can ever be tested for its performance, even by a person with ample time at their hands and a keen interest in such algorithm. Is that so and is that intentional?

tests/test_cmdline.py
```diff
@@ -31,6 +31,19 @@ def test_sig(sig_name):
 @helpers.filtered_test
 @pytest.mark.parametrize('sig_stfl_name', helpers.available_sig_stfls_by_name())
 def test_sig_stfl(sig_stfl_name):
+    # Define the list of LMS varients to allow
+    names_to_allow = ["LMS_SHA256_H5_W1", "LMS_SHA256_H5_W2",
```
baentsch, line 35: ~Should this list be documented somewhere, e.g., in the README and/or docs/algorithms/sig_stfl as otherwise users can easily complain about certain (many) SHBS algs not being available for testing?~

baentsch (commented):
Thanks for this update @ashman-p -- beyond the functional fixes, can I assume the goal of this PR also is to reduce testing runtime? Please see single comments.

tests/test_speed.py
ashman-p: They are separated to allow for customization of the list for different tests.

tests/test_speed.py
ashman-p: This change allows current CI to run without timing out or the need to extend the entire pipeline test timeout settings.
The test can be executed manually for any desired variant.

tests/test_cmdline.py
baentsch: Disregard the question above: I misunderstood the placement of the algorithm limitation in `names_to_allow`, mea culpa. Indeed, all algorithms can still be tested -- just not using this Python script. If the goal is to reduce test runtime, why not use a mechanism like [this](https://github.com/open-quantum-safe/oqs-provider/blob/main/CONFIGURE.md#oqs_skip_tests) to dynamically enable/disable tests as per the specific tests needs (disable most SIG_STFL tests in (fast) PR testing, but run all in weekly regression testing)? This would be more flexible than hard coding an algorithm list in a specific test script, no?

tests/test_kat.py
ashman-p: I think that 'skipping' a test (and logging the event) is not the same as saying a test was successfully executed.
All these are skipped because a KAT isn't available. I can create one, but then the question becomes how valuable is that exercise.

tests/test_cmdline.py
ashman-p: Thanks @baentsch. These tests take so long to run that I cannot imagine a case where we would want to run them in the pipeline. Anyone wanting to test them would do it manually.

ashman-p:
Correct. After the fix to enable the additional variants CI experienced timeouts.

src/oqsconfig.h.cmake
xuganyu96, line 1: The formatting here seems rather inconsistent. Some lines have `OQS_ENABLE_SIG_STFL_xxx` and `1` separated by one space or two spaces. Some lines are separated by an extra line break. I can empathize if you want to make some aesthetic separation, but please at least make them consistent.

tests/test_cmdline.py
```diff
@@ -40,9 +53,12 @@ def test_sig_stfl(sig_stfl_name):
             [helpers.path_to_executable('test_sig_stfl'), sig_stfl_name, katfile],
         )
     else:
-        helpers.run_subprocess(
-            [helpers.path_to_executable('test_sig_stfl'), sig_stfl_name],
-        )
+        if sig_stfl_name not in names_to_allow:
+            pytest.skip('Test skipped')
+        else:
```
xuganyu96: `pytest.skip` will cause the function to exit. This `else` is unnecessary.

tests/test_kat.py
xuganyu96: I agree with @baentsch: KAT not being available should be distinguished from skipping. I personally consider "skip" to be a consequence of user choice, such as with build flags and `SKIP_ALGS`. KAT not being available is not a user choice but an external circumstance.
I would like to propose using `pytest.xfail`, which will show up differently from `pytest.skip`, thus providing a note to developers that there is a pending item.

tests/test_cmdline.py
```diff
@@ -40,9 +53,12 @@ def test_sig_stfl(sig_stfl_name):
             [helpers.path_to_executable('test_sig_stfl'), sig_stfl_name, katfile],
         )
     else:
-        helpers.run_subprocess(
-            [helpers.path_to_executable('test_sig_stfl'), sig_stfl_name],
-        )
+        if sig_stfl_name not in names_to_allow:
+            pytest.skip('Test skipped')
```
xuganyu96: Consider providing more details in the this message. For example, "skipping slow stateful sig tests; use command line directly if needed."

tests/test_speed.py
```diff
@@ -23,11 +23,26 @@ def test_sig(sig_name):
...
+        pytest.skip('Test skipped')
```
xuganyu96, line 45: Same as above.

xuganyu96 (commented):
Made a few cosmetic suggestions. Will approve once I finish running the test suites myself.

tests/test_kat.py
ashman-p: Thanks for the suggestion. I will use pytest.xfail.

src/oqsconfig.h.cmake
ashman-p: Removed the spacing.

ashman-p:
> Made a few cosmetic suggestions. Will approve once I finish running the test suites myself.

Thanks for reviewing.

PR #2312, opened 2025-11-01
Add the path to the binary include dir

mmarock:
Purpose of this pull request is to allow building/using this library with cmake FetchContent.
FetchContent downloads/builds the given dependency and introduces its targets into the parent project.
In that case custom targets may build custom code and directly link to the liboqs oqs target.
But currently the built include dir is not forwarded in the build interface and it may not be used.

Sample cmake and main.c is included. After applying the change in this pull request this will work.
[CMakeLists.txt](https://github.com/user-attachments/files/23286725/CMakeLists.txt)
[main.c](https://github.com/user-attachments/files/23286726/main.c)

No issue is fixed, I don't know if this is considered an issue?

No change in a cryptographic algorithm.

No AI used.


mmarock:
Hello @baentsch,
thank you for your good feedback :-).
I tried changing the INCLUDE_DIRECTORIES property but it did not work, I assume it is because this property is
for *building* the target. The property for *using* this target would be INTERFACE_INCLUDE_DIRECTORIES.
But there is a problem since this is a list. So I have to use set_property with the APPEND flag so that I do not overwrite
existing entries.

I think you have a good point about using and building the library and I can offer two solutions which seem to work and do not pollute the library build:

Either:
```
set_property(TARGET oqs APPEND PROPERTY
                        INTERFACE_INCLUDE_DIRECTORIES "$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>") 
```

Or:
```
target_include_directories(oqs
                                                INTERFACE
                                                "$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>"
)
```

I like using the provided functions a bit more but I'd like to hear you opinion :-)

baentsch:
Thanks for giving this a try @mmarock ! And indeed, you're absolutely right using INTERFACE_INCLUDE_DIRECTORIES. My personal preference would be the first option if that works: Just syntactically, it signals that this is about setting a property for users. The second option --to me personally-- at first glance may look like just a normal build option for the target.

baentsch:
Thanks @mmarock for the update. The code LGTM so I triggered the CI run. If it passes, I'll approve. You might need to re-base before merge, though.

baentsch (approved):

PR #2323, opened 2025-11-19
Fix small README issues

samvandevelde:
This PR fixes tiny issues with the README.md file I just spotted while checking out PQ-libraries.
- commit 1: fixes hyperlink
- commit 2: removes information about dagger symbol (†) as this symbol is no longer used since commit [796aa12](https://github.com/open-quantum-safe/liboqs/commit/796aa120b1ed4e41b2fe11a22084eb34ab77c348).

-> No AI has been used for this tiny fix.

### DIFF
--- a/scripts/update_alg_support_table.py
+++ b/scripts/update_alg_support_table.py
@@ -20,7 +20,6 @@ ALG_SUPPORT_HEADER = [
     "Primary implementation",
 ]
 COMMIT_HASH_LEN = 7
-STD_STATUS_KEYWORDS = ["NIST", "IRTF", "ISO"]
 
 
 def format_upstream_source(source: str) -> str:
@@ -48,22 +47,7 @@ def format_upstream_source(source: str) -> str:
         output += f"@{commit}"
     return f"[`{output}`]({url})"
 
-def format_std_status_with_spec_url(std_status, spec_url=None) -> str:
-    """Embed the specification URL in the standardization status.
 
-    Looks for certain key words (e.g. NIST) to include in the hyperlink.
-    If none found, embed the whole standardification status.
-    """
-    if spec_url is None:
-        return std_status
-    separator = " "
-    words = std_status.split(separator)
-    for i, word in enumerate(words):
-        if word in STD_STATUS_KEYWORDS:
-            words[i] = f"[{word}]({spec_url})"
-            return separator.join(words)
-    return f"[{std_status}]({spec_url})"
-        
 def render_alg_support_tbl(doc_dir: str, anchor_alg_name: bool = False) -> str:
     """Render a markdown table summarizing the algorithms described by YAML data
     sheets stored in the specified doc directory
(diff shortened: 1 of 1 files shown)


xuganyu96:
I am the author of the algorithm support table. Thank you for catching this issue and proposing a contribution!

Unfortunately this fix is not sufficient. The algorithm support table in the top-level `README.md` is generated by code. You can find the logic under `scripts/update_alg_support_table.py`. The standardization status data can be found under `docs/algorithms/<kem|sig|stfl_sig>/*.yml` files. Furthermore, `slh_dsa.yml` is generated from `src/sig/slh_dsa/templates/slh_dsa_docs_yml_template.jinja` by `copy_from_slh_dsa_c.py`. A complete fix would probably also involve fixing the Jinja template.

Please let me know if you need help. Cheers.

samvandevelde:
Didn't catch the README was generated, was just a quick PR from skimming all the available libs so didn't get further into it than README and src dir. I do feel responsible now for a proper fix, will ping you when the work's done. Thanks for the catch!

scripts/update_alg_support_table.py
```diff
@@ -47,7 +48,22 @@ def format_upstream_source(source: str) -> str:
         output += f"@{commit}"
     return f"[`{output}`]({url})"
 
+def format_std_status_with_spec_url(std_status, spec_url=None) -> str:
+    """Embed the specification URL in the standardization status.
 
+    Looks for certain key words (e.g. NIST) to include in the hyperlink.
+    If none found, embed the whole standardification status.
+    """
+    if spec_url is None:
```
xuganyu96: Style suggestion: the "Pythonic way" would be:

```python
if not spec_url:
    return std_status
# ...
```

scripts/update_alg_support_table.py
xuganyu96, line 1: I see what you are trying to do. This is neat! However, I am afraid it is overkill.

Instead, I would suggest modifying `src/sig/slh_dsa/templates/slh_dsa_docs_yml_template.jinja`'s `standardization-status` to have similar formatting as [`docs/algorithms/sig/ml_dsa.yml`](https://github.com/open-quantum-safe/liboqs/blob/39ea281fd0fd9e68bf24106c3a2b12aff9a0ba63/docs/algorithms/sig/ml_dsa.yml#L16), then re-run `copy_from_upstreams.py` with:

```bash
# from project root
python ./scripts/copy_from_upstream/copy_from_upstream.py
```

samvandevelde: Thanks for the feedback! I did just that: reverted the overkill code, and just changed the jinja template. I initially refused to go this way because it felt like removing data, but now I see that I was what we in dutch call "mosquito-sifting" (worrying too much about the little things).

LMK if there is any other change you would like to see!

KR

samvandevelde:
Dear @baentsch and  @dstebila ,

All tests of this branch have been passing for a month, and xuganyu96 already approved of this branch. However, a 2nd review is required before merging can happen.

I was cleaning up small stuff before christmas when I stumbled upon this PR I forgot about; hence the ping. If there are still changes you would like to see before merging, be sure to let me know!

Merry christmas and a happy new year,
Sam

baentsch (approved):
LGTM. Thanks @samvandevelde and sorry for not having taken a look earlier. Seasons Greetings!

baentsch:
Merged. Thanks for the contribution @samvandevelde !

PR #2325, opened 2025-11-25
Re-produce ML-KEM integration

xuganyu96:
This pull request implements enough OQS Builder features and `oqsbuildfile.yml` content to reproduce ML-KEM integrations, including:
- All sub-directories under `src/kem/ml_kem`
- The generated `src/kem/ml_kem/CMakeLists.txt` file

The result builds successfully on my development machine and passes local unit tests.

### DIFF
--- a/oqsbuilder/oqsbuilder.py
+++ b/oqsbuilder/oqsbuilder.py
@@ -43,6 +43,102 @@ class CryptoPrimitive(enum.Enum):
                 return "stfl_sig"
 
 
+class Upstream:
+    """A git repository containing some source files"""
+
+    def __init__(
+        self,
+        url: str,
+        branch_or_tag: str | None = "main",
+        commit: str | None = None,
+        patches: list[str] | None = None,
+    ):
+        """
+        :param url: required, URL of the remote repository
+        :param branch_or_tag: optional, check out the specified branch or tag after clone
+        :param commit: optional, check out the specified commit after clone. This
+            will overwrite branch_or_tag. At least one of these two should be supplied.
+        :param patches: path to a single patch file, or a list of many patch files;
+            if a list of patches is passed in, then all of them will be applied in
+            a single "git apply" command. Empty patch list will be ignored.
+        """
+        if not url:
+            raise ValueError("url cannot be None")
+        if (not branch_or_tag) and (not commit):
+            raise ValueError("branch_or_tag and commit cannot both be None")
+        self.url = url
+        self.branch_or_tag = branch_or_tag
+        self.commit = commit
+        self.patches = patches or []
+
+    @staticmethod
+    def from_dict(data: dict):
+        return Upstream(
+            data["git_url"],
+            data.get("git_branch", None),
+            data.get("git_commit", None),
+            data.get("patches", None),
+        )
+
+    def clone(self, parentdir: str, dstdirname: str, dryrun: bool = False) -> str:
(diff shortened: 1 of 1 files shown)


oqsbuilder/README.md
```diff
@@ -4,14 +4,331 @@ This is the replacement for `copy_from_upstream.py`
...
+#endif /* OQS_DIST_BUILD */
+		return (OQS_STATUS) PQCP_MLKEM_NATIVE_MLKEM768_AARCH64_keypair_derand(public_key, secret_key, seed);
+#if defined(OQS_DIST_BUILD)
+	} else {
+		return (OQS_STATUS) PQCP_MLKEM_NATIVE_MLKEM768_C_keypair_derand(public_key, secret_key, seed);
+	}
+#endif /* OQS_DIST_BUILD */
+```
+
+Valid CPU features: `avx2`, `bmi2`, `popcnt`, `asimd` (will translate to `arm_neon`)
```
bhess, line 54: Is this a conclusive list of CPU features supported by the builder script? If not, it might be best to document that these are only examples.

oqsbuilder/oqsbuildfile.yml
```diff
@@ -1,106 +1,548 @@
...
+    enable_by:
+  aarch64:
+    enable_by:
+  cuda:
+    enable_by: OQS_USE_CUPQC
+  icicle_cuda:
+    enable_by: OQS_USE_ICICLE
+
+copies:
+  mlkem-native_aarch64:
```
bhess, line 419: I worry that specifying each upstream file explicitly will significantly bloat this file and shift maintenance burden to liboqs whenever upstreams add/remove/rename files. For the mlkem-native case, the upstream repository currently maintains this in their "integrations" folder, which allows them to test the liboqs integration in their CI.

bhess (commented):
Thanks @xuganyu96 for this work on improving the upstream integration.

I want to make sure I'm understanding the design correctly. It looks like with the new oqsbuilder, liboqs would be responsible for maintaining the list of all files from upstream codebases in oqsbuildfile.yml. Previously, upstream repositories maintained this metadata themselves (e.g., mlkem-native's [*_META.yml](https://github.com/pq-code-package/mlkem-native/blob/main/integration/liboqs/ML-KEM-512_META.yml) files), which also allowed them to test the liboqs integration in their own CI.

liboqs currently has two integration patterns:

1. Upstream-maintained metadata: copy_from_upstream where upstreams maintain integration metadata (e.g., mlkem-native, pqclean imports, onramp signatures)
2. Direct/manual integrations: liboqs maintains everything (e.g., frodo, bike)

I'm wondering if oqsbuilder could support both patterns? Specifically:

- For upstreams that already maintain liboqs integration metadata, we consume that directly (preferred approach)
- For direct integrations, we maintain the metadata in oqsbuildfile.yml (or preferrably, algorithm-specific yaml-files)

This would preserve the benefits of upstream-maintained metadata (reduced maintenance burden, upstream CI testing) while still supporting the flexibility we need for direct integrations, and having the oqsbuilder as the component to handle all integrations. What are your thoughts on this approach?

xuganyu96:
@bhess Thank you for your feedback.

> I'm wondering if oqsbuilder could support both patterns?

At a high level I agree that this would be nice, though the devil lies in implementation details. I am happy to discuss specific approaches after returning from the holidays.

PR #2330, opened 2025-12-05
Fix incorrect arg register update in avx512 Keccak

mdcornu:
Fix for a bug in the AVX512 Keccak implementation where the wrong registers were being updated after extracting bytes from the state. This is currently not an issue as it only occurs when fewer than 8 bytes are extracted using the multi-buffer API, and both ML-KEM and ML-DSA always squeeze data in multiples of 8 bytes, so this code path currently isn’t exercised in practice.


dstebila:
Thanks Marcel!

PR #2333, opened 2025-12-30
Docs: List all optimized implementations

mkannwischer:
The list of optimized implementations is currently misleading. For example,
https://openquantumsafe.org/liboqs/algorithms/kem/ml-kem.html lists mlkem-native as the source for optimized implementations followed by a list including cupqc-cuda and icicle-icicle_cuda, but excluding the optimized implementations from mlkem-native.

This commit fixes that by listing ALL optimized implementations instead of just those that come from non-primary upstreams.
To avoid listing non-optimized implementations, a default field is added to the ymls to detect those (naming is inconsistent across various schemes, so this appears to be the most robust approach).


dstebila (approved):

dstebila:
Thanks Matthias!

PR #2338, opened 2026-01-07
Derand Implementations

knightcode:
This PR adds implementations for the `_derand` methods of the API for most of the algorithms.

**No**

**No**

Cursor's Composer was consulted for suggestions when making the first edits to one variant of a given algorithm. Once the solution was determined for that variant. A Cursor agent was instructed to mimic the changes for the other variants of that algorithm.

### DIFF
--- a/src/kem/ntru/kem_ntru.h
+++ b/src/kem/ntru/kem_ntru.h
@@ -10,12 +10,13 @@
 #define OQS_KEM_ntru_hps2048509_length_secret_key 935
 #define OQS_KEM_ntru_hps2048509_length_ciphertext 699
 #define OQS_KEM_ntru_hps2048509_length_shared_secret 32
-#define OQS_KEM_ntru_hps2048509_length_keypair_seed 0
-#define OQS_KEM_ntru_hps2048509_length_encaps_seed 0
+#define OQS_KEM_ntru_hps2048509_length_keypair_seed 32
+#define OQS_KEM_ntru_hps2048509_length_encaps_seed 32
 OQS_KEM *OQS_KEM_ntru_hps2048509_new(void);
 OQS_API OQS_STATUS OQS_KEM_ntru_hps2048509_keypair(uint8_t *public_key, uint8_t *secret_key);
 OQS_API OQS_STATUS OQS_KEM_ntru_hps2048509_keypair_derand(uint8_t *public_key, uint8_t *secret_key, const uint8_t *seed);
 OQS_API OQS_STATUS OQS_KEM_ntru_hps2048509_encaps(uint8_t *ciphertext, uint8_t *shared_secret, const uint8_t *public_key);
+OQS_API OQS_STATUS OQS_KEM_ntru_hps2048509_encaps_derand(uint8_t *ciphertext, uint8_t *shared_secret, const uint8_t *public_key, const uint8_t *seed);
 OQS_API OQS_STATUS OQS_KEM_ntru_hps2048509_decaps(uint8_t *shared_secret, const uint8_t *ciphertext, const uint8_t *secret_key);
 #endif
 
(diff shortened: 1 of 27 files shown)


RodriM11:
I have only reviewed a handful of implementations of the PR, but I have a couple of comments: 

1. In the cases of (at least) HQC and FrodoKEM, the implementation of the randomized and derand interfaces are duplicated (i.e. the randomized does not generate random material and then called the derand interface), so code gets duplicated. I have further seen that, in some cases, that is because the `derand` interface does not simply execute the deterministic bits of the code, but is actually deriving the actual randomness material from the seed using a hash, which is a cryptographic addition that, on itself, should be analyzed (see e.g. [here](https://github.com/open-quantum-safe/liboqs/blob/68a463cac7455ecfefd5636227023aea3b2d2f1e/src/kem/frodokem/external/kem.c#L27)). In any case, this could also be grouped into a single "functional" block.
2. In the case of (at least) FrodoKEM and HQC, the length of the values for the seed material in both interfaces are not "according" to the specification (or what a somehow "transparent" extension of the specification would render). This is because this PR assumes (in, at least, some cases) a random material of 32 bytes and expands pseudorandomly using a hash function to the required lengths, instead of asking for the full length of random material to be employed. As denoted above, this is a major cryptographic modification of the specification with plausible wide-ranging implications (choice of length of the seed material, choice of hash, etc...) that, IMO, should not be agreed.

knightcode:
1. I don't claim to be an expert on the algorithms themselves and will implement any recommended changes.
2. I didn't want to disrupt the randomized functions at all, so I *did* make duplicate code. With whoever's blessing I can refactor to have the randomized ones call into the derand ones.
3. I wasn't sure how much the input should be trusted. Hashing it ensures its in a high entropy space and has the proper length. But, if that's the responsibility of the caller, ok.

educate me and let me know what you'd prefer!

RodriM11:
> I don't claim to be an expert on the algorithms themselves and will implement any recommended changes.

First, I want to note that it is likely that the PR itself, regardless of the way it is finally implemented, will not get accepted. That is because people might claim (rightfully so) that since there is no standard or specification on how to implement `derand` interfaces for the remaining algorithms, no implementation should be adopted.

Having said so, IMO the only way this inclusion **could** be considered acceptable is to implement the `derand` interfaces as the ones receiving as input exactly the amount of randomness material declared by each specification.

> I wasn't sure how much the input should be trusted. Hashing it ensures its in a high entropy space and has the proper length.

This on itself is a very interesting (and years old) discussion. IMO, in the absence of an specification formally declaring what to do (like in the case of ML-KEM), no cryptographic decisions should be taken, and hashing it is a very cryptographic one.

saitomst:
@knightcode Thank you for your efforts to improve the functional support including NTRU algorithms. However, I fully agree with the points raised by @RodriM11.

### About the general design direction

As far as I understand, the  DeriveKeyPair functionality is currently implemented only for ML-KEM (and kyber?), as discussed in the issue: https://github.com/open-quantum-safe/liboqs/issues/1206

So I believe we first need to clarify the intended status of this feature within the liboqs project:
* Is this mechanism considered experimental, intended only for ML‑KEM?
* Or is DeriveKeyPair functionality something liboqs expects all KEMs to support in the future?

I think this clarification is necessary before extending the functionality more broadly.

### Handling of randomness and seed material

I also agree with @RodriM11 that a derandomized interface should not take a fixed‑size seed and expand it internally using a hash or any other mechanism, especially in the absence of clear specifications or cryptographic analysis.
In my view, the derand functions should require the caller to provide cryptographically secure random coins of the appropriate length, and the responsibility for preparing such randomness should rest with the caller.

### NTRU-specific concerns

Regarding NTRU in particular:

The `pqclean_*` directories contain core cryptographic implementations imported from the upstream. These should be modified as little as possible. Operations such as seed expansion are not part of the core algorithm and should not be introduced into these directories.

For maintainability and ease of tracking upstream updates, we should avoid duplicating functions with independent modifications here.
Looking at ML‑KEM —the only current implementation of DeriveKeyPair functionality— its structure looks that follows the design principles mentioned by @RodriM11. I believe NTRU (and other KEMs) should follow the same model if deterministic support were to be introduced.

xuganyu96:
I agree with @RodriM11's assessment. The modifications proposed in this pull request deviate too much from each algorithm's specification to be suitable for liboqs.

knightcode:
You don't want a partial update?

xuganyu96:
> You don't want a partial update?

No. Implementation modifications with no basis in written specification are out of scope for this project. We also don't have the technical capacity to independently verify whether this pull request's code change is either correct or secure.

PR #2348, opened 2026-01-16
OQSBuilder

xuganyu96:
This pull request reproduces ML-KEM integration from mlkem-native, cupqc, and icicle.

OQSBuilder aims to improve upon `copy_from_upstream.py` by clarifying the data flow from `oqsbuildfile.yml` (previously `copy_from_upstream.yml`) into Python data structures, then into actions such as cloning upstream repositories and rendering internal CMakeLists.txt/source/header files from Jinja templates.

Reviewers might find it easier to start with `oqsbuilder/__main__.py` and read through the code.

Feedback on design choices is very much appreciated.

NOTE: I have not gotten implementing fetching file list from remote `META.yml` file, but this item is on the roadmap.


scripts/oqsbuildfile.yml
```diff
@@ -0,0 +1,1366 @@
...
+  x86_64:
+    enable_by:
+  aarch64:
+    enable_by:
+  cuda:
+    enable_by: OQS_USE_CUPQC
+  icicle_cuda:
+    enable_by: OQS_USE_ICICLE
+
+copies:
```
dstebila, line 1110: This is going to get quite large. Would it be useful to break `copies` out into its own file? Or possibly one YML file for each entry within `copies`, e.g., `mldsa-native_x86_64`, ...

scripts/oqsbuilder.py
```diff
@@ -0,0 +1,1347 @@
...
+- 🔨 Re-produce existing builds
+    - ✅ ML-KEM
+        - ✅ render family-level `CMakeLists.txt`
+        - ✅ render family-level header file (e.g. `kem_ml_kem.h`)
+        - ✅ render family-level source file (e.g. `kem_ml_kem_512.c`)
+        - 🔨 render `.CMake/alg_support.cmake`
+        - 🔨 render documentation
+    - 🔨 !!! ML-DSA !!!
+- ✅ Parse `oqsbuildfile.yml` into structured data instead of Python dictionary
+- 🔨 Consider documenting oqsbuildfile's schema in Python code
```
dstebila, line 17: Yes, I think some documentation of oqsbuildfile.yml's schema will be necessary so that others will know how to use it. Would a Markdown file be better than in the Python code? 

There are tools for codifying a scheme for a YAML file and also validating a YAML file against a schema, e.g., https://github.com/yaml-schema/yaml-schema.  I don't know if that's overkill.

Issue #2311, opened 2025-10-30
Refactor SLH-DSA C integration


xuganyu96:
SLH-DSA (specifically the [slhdsa-c](https://github.com/pq-code-package/slhdsa-c/) impl) was integrated into `liboqs` with the 0.15.0 release. However, a lot of integration logic is inconsistent with all other algorithm/implementation integrations, and integration metadata is not stored in the same `copy_from_upstream.yml` datasheet as all other implementations.

This issue tracks the ongoing effort to refactor the existing SLH-DSA C integration scaffolding. I am aiming to avoid any changes that affects external users.

Issue #2254, opened 2025-08-27
Support for prehashed ML-DSA variants


bhess:
[FIPS 204](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf) specifies two features related to pre-hashing in ML-DSA that are currently not supported in liboqs: "externalMu" and "HashML-DSA".

### FIPS context
Defined in [Algorithm 7, line 6](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf):  
  > message representative that may optionally be computed in a different cryptographic module  

This means `mu` can be computed externally and passed into the signature process instead of the raw message.

### Current upstream state
The pq-crystals upstream code does not provide an option to pass `mu`. It is always computed internally, see [here](https://github.com/pq-crystals/dilithium/blob/444cdcc84eb36b66fe27b3a2529ee48f6d8150c2/ref/sign.c#L110-L116).

### Future upstream state when adopting mldsa-native
The upcoming `mldsa-native` implementation does support external `mu`, see [here](https://github.com/pq-code-package/mldsa-native/blob/17368828d884eec727657505c51afeb3845d89c2/mldsa/sign.c#L498-L507).

### Integration options (once `mldsa-native` is released and integrated)

1. Add algorithm variants
   - Example: `ML-DSA-44-externalMu`  
   - Reuse existing signing API and interpret the message input as `mu` for these variants.  
   - Pros: No API changes; aligns with liboqs pattern of algorithm-agnostic API.
   - Cons: Slight ambiguity in API semantics (message parameter is overloaded).

2. Introduce a new API  
   - Example: `OQS_SIG_sign_with_external_mu(...)`  
   - Pros: Explicit interface.
   - Cons: Heavy API change for an algorithm-specific feature.

My personal preference: Option 1, as `externalMu` is algorithm-specific and adding a new global API is difficult to justify.

### FIPS context
Defined in [Algorithm 4](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf).

Steps:
  - Accept a pre-hashed message as input.
  - Prepend:
    - A different domain separation byte than pure ML-DSA
    - The OID of the hash function.
- This makes HashML-DSA incompatible with standard ML-DSA.

Note: SLH-DSA (FIPS 205) has a similar pre-hash option, so this pattern is not unique to ML-DSA.

### Upstream state
[...]

PR #2352, opened 2026-01-25
Add external-mu sign gen ACVP tests

abhi-dev-engg:
The internal signature generation API tests using the mu parameter were previously disabled due to lack of support in pq-crystals.
With the addition of this support in mldsa-native, these tests are now enabled.

**NOTE:** The "prehash" API tests are still disabled as I found they are not completely supported yet. Please correct me if I am wrong.
OR 
I could call into "mld_sign_signature_pre_hash_internal" by calculating the "hash" in the test file. Not sure if this is the correct way as the API needs to be exposed "externally"


dstebila:
Thanks very much, Abhi!

> NOTE: The "prehash" API tests are still disabled as I found they are not completely supported yet. Please correct me if I am wrong.
> OR
> I could call into "mld_sign_signature_pre_hash_internal" by calculating the "hash" in the test file. Not sure if this is the correct way as the API needs to be exposed "externally"

Tagging @mkannwischer for an opinion on this matter.

tests/test_acvp_vectors.py
```diff
@@ -218,19 +218,25 @@ def test_acvp_vec_ml_dsa_sig_gen(sig_name):
...
-                    message = testCase["message"]
+                    if variant["signatureInterface"] == "internal":
+                        if variant["externalMu"]:
+                            message = testCase["mu"]
+                            extmu = "1"
+                        else:
+                            message = testCase["message"]
+                            extmu = "0"
+                    else:
+                        message = testCase["message"]
```
xuganyu96, line 254: My LSP is freaking out about `extmu` being possibly unbound. I understand that `variant["signatureInterface"] == "internal"` ensures that `extmu` is always defined where it is needed, but it would be very nice to make this error go away. I would also love to see the `message, extmu` logic being abstracted instead of repeated between `test_acvp_vec_ml_dsa_sig_gen` and `test_acvp_vec_ml_dsa_sig_ver`. For example:

```python
def get_msg_extmu(testCase, variant) -> tuple[str, str | None]:
    if variant["signatureInterface"] == "internal":
        if variant["externalMu"]:
            return testCase["mu"], "1"
        else:
            return testCase["message"], "0"
    return testCase["message"], None
```

Now this entire chunk can be replaced with

```python
message, extmu = get_msg_extmu(testCase, variant)
```

xuganyu96 (changes requested):

tests/test_acvp_vectors.py
abhi-dev-engg: This test file follows a single-flow structure throughout without separate helper functions, so I kept the same pattern here for consistency. For that reason, I’m not sure this comment applies.

Also, throughout these ACVP tests we already rely on the JSON structure being valid, so this code follows the same assumption model used elsewhere in the file.

tests/vectors_sig.c
```diff
@@ -1031,4 +1036,4 @@ int main(int argc, char **argv) {
 	} else {
 		return EXIT_SUCCESS;
 	}
-}
+}
```
xuganyu96: Please add a new line at the end of the source file

xuganyu96 (changes requested):

tests/vectors_sig.c
abhi-dev-engg: done

mkannwischer:
> Thanks very much, Abhi!
> 
> > NOTE: The "prehash" API tests are still disabled as I found they are not completely supported yet. Please correct me if I am wrong.
> > OR
> > I could call into "mld_sign_signature_pre_hash_internal" by calculating the "hash" in the test file. Not sure if this is the correct way as the API needs to be exposed "externally"
> 
> Tagging @mkannwischer for an opinion on this matter.

HashML-DSA is fully supported my mldsa-native via the `signature_pre_hash_internal` and `verify_pre_hash_internal` APIs - those are external APIS. For those APIs, the pre-hashing itself needs to be done by the caller.
If you want to leave the pre-hashing to mldsa-native, so far we only support SHAKE256 as pre-hash hash functions via the `signature_pre_hash_shake256` and `verify_pre_hash_shake256` API.
If other hash functions are required, please open an issue in mldsa-native and we will see if we can add support. However, we expect that most consumers already have pre-existing hash implementations and would rather want to use the `*_pre_hash_internal` functions.

abhi-dev-engg:
> > Thanks very much, Abhi!
> > > NOTE: The "prehash" API tests are still disabled as I found they are not completely supported yet. Please correct me if I am wrong.
> > > OR
> > > I could call into "mld_sign_signature_pre_hash_internal" by calculating the "hash" in the test file. Not sure if this is the correct way as the API needs to be exposed "externally"
> > 
> > 
> > Tagging @mkannwischer for an opinion on this matter.
> 
> HashML-DSA is fully supported my mldsa-native via the `signature_pre_hash_internal` and `verify_pre_hash_internal` APIs - those are external APIS. For those APIs, the pre-hashing itself needs to be done by the caller. If you want to leave the pre-hashing to mldsa-native, so far we only support SHAKE256 as pre-hash hash functions via the `signature_pre_hash_shake256` and `verify_pre_hash_shake256` API. If other hash functions are required, please open an issue in mldsa-native and we will see if we can add support. However, we expect that most consumers already have pre-existing hash implementations and would rather want to use the OQS_SIG functions.

The support via the internal APIs is understood. I assumed a new API that is aligned with Algorithm 4 as per FIPS-204 would be the final API. But , I also too believe performing hash inside the DSA API is completely redundant.

If agreed, i could add a new signature prehash API reference under the structure "OQS_SIG" that would be mapping to "pre_hash_internal" APIs of the native package. The user could pre-calculate the hash of the message and then call into "OQS_SIG_prehash_sign". Need to say handling in slh-dsa though.

What say @bhess @dstebila ?

tests/vectors_sig.c
abhi-dev-engg: @xuganyu96 could you pls resolve this if ok and approve the workflow as well ?
The code-formatting issue seems to fixed now at least in my [fork](https://github.com/abhi-dev-engg/liboqs/tree/dsa_update)

dstebila:
> If agreed, i could add a new signature prehash API reference under the structure "OQS_SIG" that would be mapping to "pre_hash_internal" APIs of the native package. The user could pre-calculate the hash of the message and then call into "OQS_SIG_prehash_sign". Need to say handling in slh-dsa though.
> 
> What say @bhess @dstebila ?

Let's not do this in this PR.  If memory serves there have been some concerns that our current approach for the prehash variants in SLH-DSA is inconvenient: there are a huge number of variants, for each prehash algorithm option.  There was a start of a discussion in #2254 about prehashed API for ML-DSA, and that might also come into play in #2311 for SLH-DSA.

abhi-dev-engg:
> 
> Let's not do this in this PR. If memory serves there have been some concerns that our current approach for the prehash variants in SLH-DSA is inconvenient: there are a huge number of variants, for each prehash algorithm option. There was a start of a discussion in #2254 about prehashed API for ML-DSA, and that might also come into play in #2311 for SLH-DSA.

makes sense. I will try to test the "prehash" tests via the "internal" APIs in the meantime. Hope that's ok ?
This can be tracked via separate WIP PR : https://github.com/open-quantum-safe/liboqs/pull/2355

abhi-dev-engg:
@bhess : could you pls review this PR once you have a moment ?

bhess (approved):
Thanks @abhi-dev-engg for extending the ACVP tests ssupporting externalMu. LGTM.

abhi-dev-engg:
@xuganyu96 could you pls provide approval as well pls if all is ok?

xuganyu96 (approved):

abhi-dev-engg:
@dstebila could you pls review and provide approval/merge if all is ok ?

abhi-dev-engg:
@xuganyu96 could you pls help to merge this PR please ? its been stuck over a week now.

baentsch:
Thanks for the contribution @abhi-dev-engg !

abhi-dev-engg:
> Thanks for the contribution @abhi-dev-engg !

no worries @baentsch , pls tag me for any topic related to kem/dsa in future

PR #2366, opened 2026-02-18
Add API support for External-MU Signature[ML-DSA]

abhi-dev-engg:
Add External API support for "External MU" Signature generation and verification.

The following changes were done as a part of this PR:

- updated copy_from_upstream.py to dynamically generate extmu metadata and map API calls, minimizing deviations from upstream logic.
- configured the build system to compile extmu variants strictly as dispatcher wrappers.
- enforced a mandatory 64-byte input check in the sig_scheme.c wrapper to ensure memory safety when processing arbitrary inputs.
- bypassed extmu variants in KAT generation and generic memory benchmarks to accommodate their specialized API requirements.


abhi-dev-engg:
Hi @bhess ,

I was trying to implement the external mu OQS APIs based on approach 1 suggested [here](https://github.com/open-quantum-safe/liboqs/issues/2254) by you. However, in my (limited) understanding it seems difficult adding a "new variant" without having support from the upstream repo(see one error below).

Tried adding the new variants "ml_dsa_xx_extmu" in '_copy_from_upstream.yml_' under ml-dsa umbrella leading to '_FileNotFoundError: [Errno 2] No such file or directory: 'repos/mldsa-native/integration/liboqs/ML-DSA-44-EXTMU_META.yml_'.

Additionally, the extmu and the existing variants shall be using the same set of source files, which could lead to linking errors due to "same definition" if both the variant are enabled.

Please let me know your opinion or please guide me incase I am in the wrong direction.

Thanks.

bhess:
Hi @abhi-dev-engg, thanks for working on enabling externalMu.

In my view, the preferred approach is to add the external-mu API definition directly to the upstream YAML file (for example, in [ML-DSA-44_META.yml](https://github.com/pq-code-package/mldsa-native/blob/main/integration/liboqs/ML-DSA-44_META.yml), signature_signature_extmu/signature_verify_extmu).

By doing this, we can then adapt the copy-from-upstream script to automatically generate the code for the -externalmu variant. This keeps the integration easier to maintain if upstream APIs change. Let me know if that makes sense to you!

abhi-dev-engg:
Hi @bhess , thank you for your reply.

I agree with your approach. would try to do the same and yes it requires "some" changes in copy_from_upstream.py for this to work.

Let me spend some time over this and get back.

abhi-dev-engg:
> Hi @abhi-dev-engg, thanks for working on enabling externalMu.
> 
> In my view, the preferred approach is to add the external-mu API definition directly to the upstream YAML file (for example, in [ML-DSA-44_META.yml](https://github.com/pq-code-package/mldsa-native/blob/main/integration/liboqs/ML-DSA-44_META.yml), signature_signature_extmu/signature_verify_extmu).
> 
> By doing this, we can then adapt the copy-from-upstream script to automatically generate the code for the -externalmu variant. This keeps the integration easier to maintain if upstream APIs change. Let me know if that makes sense to you!

Hi @bhess , the first version is ready now. Pls have a look at your convenience.

docs/algorithms/sig/ml_dsa.yml
```diff
@@ -16,7 +16,7 @@ nist-round: FIPS204
 standardization-status: Standardized by [NIST](https://csrc.nist.gov/pubs/fips/204/final)
 spec-version: ML-DSA
 primary-upstream:
-  source: https://github.com/pq-code-package/mldsa-native/commit/f48f164cefb07f4ffa519ddda7cee670b8ee3517
+  source: https://github.com/abhi-dev-engg/mldsa-native/commit/b3198696c21f4b28508ad2231e130768f622076a
```
xuganyu96: I am not comfortable with moving the source of implementation from official PQCA repository to a personal fork.

I can see that you are trying to change the schema of the upstream META.yml file. If a change in the schema of META.yml files is necessary, I would recommend opening a separate pull request to `mldsa-native`.

Alternatively, you can set only the `_extmu` implementations to upstream from your fork, which means that in `copy_from_upstream.yml` there will be a separate upstream (e.g. `mldsa-native-fork`) from the official PQCA upstream.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
```diff
@@ -68,286 +68,318 @@ OQS_SIG *OQS_SIG_{{ family }}_{{ scheme['alias_scheme'] }}_new(void) {
 	sig->keypair = OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_keypair;
 	sig->sign = OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_sign;
 	sig->verify = OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_verify;
-    {%- if 'api-with-context-string' in default_impl and default_impl['api-with-context-string'] %}
+	{%- if 'api-with-context-string' in default_impl and default_impl['api-with-context-string'] %}
```
xuganyu96: I find these diffs rather strange. It looks like nothing changed, but git is reporting a diff. I suspect that there are invisible character, or that the line break characters are different (i.e. Windows using CRLF while Unix-like systems use LF). Please resolve these empty diffs.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
```diff
@@ -68,286 +68,318 @@ OQS_SIG *OQS_SIG_{{ family }}_{{ scheme['alias_scheme'] }}_new(void) {
...
-    {%- endif %}
+	{%- endif %}
 }
 
 OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_sign(uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *secret_key) {
-    {%- for impl in scheme['metadata']['implementations'] if impl['name'] != scheme['default_implementation'] %}
-    {%- if loop.first %}
+{%- if scheme.is_extmu %}
+	/* extmu variants require exactly 64 bytes of input message representative */
+	if (message_len != 64) {
```
xuganyu96: While ML-DSA's external mu is fixed at 64 bytes, we don't know if the same applies to future signature schemes. It might be better to abstract the size of external mu.

I recommend documenting the size of external mu in the integration YAML file `META.yml`, define a macro `ML_DSA_EXT_MU_LEN` in the header file `src/sig/ml_dsa/sig_ml_dsa.h`, then use the macro here.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
```diff
@@ -68,286 +68,318 @@ OQS_SIG *OQS_SIG_{{ family }}_{{ scheme['alias_scheme'] }}_new(void) {
...
-    {%- if scheme['metadata']['implementations']|rejectattr('name', 'equalto', scheme['default_implementation'])|list %}
+	{%- if scheme['metadata']['implementations']|rejectattr('name', 'equalto', scheme['default_implementation'])|list %}
 #endif
-    {%- endif %}
+	{%- endif %}
 }
 {%- else %}
 
 OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_sign_with_ctx_str(uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *ctx_str, size_t ctx_str_len, const uint8_t *secret_key) {
+{%- if scheme.is_extmu %}
```
xuganyu96, line 371: I think the `sign_with_ctx_str`/`verify_with_ctx_str` API's and the external-mu API's are mutually exclusive. If the user is explicitly calling `sign_with_ctx_str` API, it is definitely not trying to use external-mu.

tests/test_sig.c
```diff
@@ -355,7 +358,7 @@ int main(int argc, char **argv) {
 	char no_thread_sig_patterns[][MAX_LEN_SIG_NAME_]  = {"MAYO-5", "cross-rsdp-128-small", "cross-rsdp-192-small", "cross-rsdp-256-balanced", "cross-rsdp-256-small", "cross-rsdpg-192-small", "cross-rsdpg-256-small", "SNOVA_37_17_2", "SNOVA_56_25_2", "SNOVA_49_11_3", "SNOVA_37_8_4", "SNOVA_24_5_5", "SNOVA_60_10_4", "SNOVA_29_6_5"};
 	int test_in_thread = 1;
 	for (size_t i = 0 ; i < sizeof(no_thread_sig_patterns) / MAX_LEN_SIG_NAME_; ++i) {
-		if ( (strncmp(alg_name, "SLH_DSA", 7) == 0) || (strstr(alg_name, no_thread_sig_patterns[i]) != NULL) ) {
+		if ((strncmp(alg_name, "SLH_DSA", 7) == 0) || (strstr(alg_name, no_thread_sig_patterns[i]) != NULL) ) {
```
xuganyu96: Remove the space between the two closing parentheses, too, to ensure consistent style.

xuganyu96 (changes requested):
Thank you for the contribution.

In addition to the individual code change comments I've made, I have a bigger concern with the strategy used to implement the external-mu API. From what I have read, your strategy for implementing external mu is to treat it as a separate implementation of ML-DSA whose `sign` and `verify` behave differently from the non-external-mu implementations of ML-DSA. This feels inconsistent with how liboqs intends "different implementations of the same scheme" to behave. The expectation is that the same function across different implementations of the same scheme (e.g. `x86` impl of ML-DSA's sign versus aarch64's impl, versus the portable impl) should have identical external behavior, such as "passing all of the same KAT". This is not true for the "sign" and "verify" function under the external-mu implementations.

Because external-mu sign/verify have different external behaviors from non-external-mu sign-verify, I feel strongly that they should be separate functions instead of separate implementations. I am thinking of something like this:

```c
// list of OQS API
OQS_SIG *OQS_SIG_ml_dsa_44_new(void);
OQS_API OQS_STATUS OQS_SIG_ml_dsa_44_keypair(uint8_t *public_key, uint8_t *secret_key);
OQS_API OQS_STATUS OQS_SIG_ml_dsa_44_sign(uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *secret_key);
OQS_API OQS_STATUS OQS_SIG_ml_dsa_44_verify(const uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *public_key);
OQS_API OQS_STATUS OQS_SIG_ml_dsa_44_sign_with_ctx_str(uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *ctx, size_t ctxlen, const uint8_t *secret_key);
OQS_API OQS_STATUS OQS_SIG_ml_dsa_44_verify_with_ctx_str(const uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *ctx, size_t ctxlen, const uint8_t *public_key);
/* ext-mu functions: replace (msg, msglen) with mu
 * return OQS_ERROR if no implementation of this scheme can do external-mu
 */
OQS_API OQS_STATUS OQS_SIG_ml_dsa_44_sign_extmu(uint8_t *signature, size_t *signature_len, const uint8_t *mu, const uint8_t *secret_key);
OQS_API OQS_STATUS OQS_SIG_ml_dsa_44_verify_extmu(const uint8_t *mu, const uint8_t *signature, size_t signature_len, const uint8_t *public_key);
```

cc: @dstebila @bhess @baentsch for additional inputs

bhess:
Thanks @abhi-dev-engg. I’ll take a more detailed look.

> Because external-mu sign/verify have different external behaviors from non-external-mu sign-verify, I feel strongly that they should be separate functions instead of separate implementations. I am thinking of something like this:

This would be Option 2 from the associated issue https://github.com/open-quantum-safe/liboqs/issues/2254.
My stated preference is Option 1, as implemented in this PR, because externalMu is an algorithm‑specific feature of ML‑DSA. Introducing a global API for something so specific feels difficult to justify. The context strings, in comparison, is a feature applicable to multiple signature algorithms.

docs/algorithms/sig/ml_dsa.yml
abhi-dev-engg: I would be creating a PR to merge this into mldsa-native once this PR is reviewed and then replacing this with the official git repo as currently done. I needed a way to test out this feature hence this change.

My bad, I should have mentioned it earlier. The PR description is updated with this info.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: Ext-MU is specific to ML_DSA and scheme.is_extmu is set to True only in such a case in the upstream script.

defining a macro breaks the overall format of "sig_xx.c" used everywhere else, hence hardcoded here.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: Agreed, which is why sig_with_ctx_support is set to false. I only included them because the liboqs architecture requires all schemes to define these API pointers for consistency. They safely reject any call where ctx isn't NULL, identical to the pattern used in other schemes - [here](https://github.com/open-quantum-safe/liboqs/blob/main/src/sig/mayo/sig_mayo_1.c#L130)

abhi-dev-engg:
> 
> Because external-mu sign/verify have different external behaviors from non-external-mu sign-verify, I feel strongly that they should be separate functions instead of separate implementations. I am thinking of something like this:

That approach is valid, but it requires expanding the core OQS_SIG struct. This means we'd have to update every signature scheme across the entire library just to mark the new functions as unsupported. The variant approach is much less invasive since extmu only applies to ML-DSA right now.

dstebila:
After discussing with Bruce this afternoon, I support option 1's approach.

tests/test_sig.c
abhi-dev-engg: fixed

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: yeah that's strange. have fixed it now

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
bhess: I agree with @xuganyu96. I would also prefer having this defined in the YAML file. Even though externalMu is ML‑DSA–specific, defining the message‑length restriction in the YAML could be useful more generally and also serves as documentation.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
```diff
@@ -220,14 +227,21 @@ OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_sign(uint8_t *sig
...
     {%- endif %}
     {%- if scheme['metadata']['implementations']|rejectattr('name', 'equalto', scheme['default_implementation'])|list %}
 #endif
     {%- endif %}
 }
 
 OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_verify(const uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *public_key) {
+{%- if scheme.is_extmu %}
+	/* extmu variants require exactly 64 bytes of input message representative */
+	if (message_len != 64) {
```
bhess: Same comment here, consider defining this in the yml file.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
```diff
@@ -277,6 +291,12 @@ OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_verify(const uint
 {%- set default_impl = scheme['metadata']['implementations'] | selectattr("name", "equalto", scheme['default_implementation']) | first %}
 {%- if 'api-with-context-string' in default_impl and default_impl['api-with-context-string'] %}
 OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_sign_with_ctx_str(uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *ctx_str, size_t ctx_str_len, const uint8_t *secret_key) {
+{%- if scheme.is_extmu %}
+	if (message_len != 64) {
```
bhess: same comment as above

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
```diff
@@ -312,6 +332,12 @@ OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_sign_with_ctx_str
 }
 
 OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_verify_with_ctx_str(const uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *ctx_str, size_t ctx_str_len, const uint8_t *public_key) {
+{%- if scheme.is_extmu %}
+	if (message_len != 64) {
```
bhess: same comment as above

scripts/copy_from_upstream/copy_from_upstream.py
```diff
@@ -675,8 +695,7 @@ def process_families(instructions, basedir, with_kat, with_generator, with_libja
                                 scheme['scheme'], str(ke), impl['name']))
                         pass
 
-
-            if with_kat:
+            if with_kat and not scheme.get('is_extmu', False):
```
bhess, line 719: I don’t fully understand why is_extmu is excluded here. Couldn’t the KAT hash for externalMu also be defined in the upstream YAML file?

scripts/copy_from_upstream/copy_from_upstream.yml
```diff
@@ -57,9 +57,9 @@ upstreams:
     patches: []
   -
     name: mldsa-native
-    git_url: https://github.com/pq-code-package/mldsa-native.git
-    git_branch: v1.0.0-alpha2
-    git_commit: f48f164cefb07f4ffa519ddda7cee670b8ee3517
+    git_url: https://github.com/abhi-dev-engg/mldsa-native.git
```
bhess: I suppose this will be changed back to the pqcp repo before merging, right?

tests/KATs/sig/kats.json
```diff
@@ -35,14 +35,23 @@
     "all": "54322d25c05b96941aec73eb3ee27a975d38a34bfb53ef4cb9fd4a8e9cab554c",
     "single": "9a196e7fb32fbc93757dc2d8dc1924460eab66303c0c08aeb8b798fb8d8f8cf3"
   },
+  "ML-DSA-44-extmu": {
```
bhess: The "all" entry appears to be missing.
Also, out of curiosity, how were the KAT hashes generated? I’m asking because they don’t seem to originate from the upstream KAT.

scripts/copy_from_upstream/copy_from_upstream.yml
abhi-dev-engg: yes, correct

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: fixed, mu length is added in the yaml file now

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: fixed, added in yml file

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: fixed, added in yml file

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: fixed, added in yml file

tests/KATs/sig/kats.json
abhi-dev-engg: yeah my bad, during my initial "incorrect" commits it fetched value from the corresponding non-mu variant and was added and I mistakingly pushed it as well :(

kat for extmu variants are removed completely now.

scripts/copy_from_upstream/copy_from_upstream.py
abhi-dev-engg: Hi @bhess , 

I disabled KAT tests for mu variants as [kat_sig.c](https://github.com/open-quantum-safe/liboqs/blob/main/tests/kat_sig.c) tests varying message lengths, but extmu strictly requires exactly 64 bytes for the message.

Since NIST KATs expect varying lengths, I disabled them as msglen is constant here.

how would you prefer to handle this for external-mu APIs?

tests/test_acvp_vectors.py
xuganyu96, line 1: The ACVP vectors have external-mu test cases, but they are skipped here. Please get the ACVP test cases to pass. Given the absence of KAT's, I think ACVP test cases are required.

```bash
python -m pytest --verbose tests/test_acvp_vectors.py -k "test_acvp_vec_ml_dsa_sig_gen[ML-DSA-44-extmu]"
```

src/sig/ml_dsa/CMakeLists.txt
```diff
@@ -86,4 +86,31 @@ if(OQS_ENABLE_SIG_ml_dsa_87_aarch64)
     set(_ML_DSA_OBJS ${_ML_DSA_OBJS} $<TARGET_OBJECTS:ml_dsa_87_aarch64>)
 endif()
 
+if(OQS_ENABLE_SIG_ml_dsa_44_extmu)
```
xuganyu96, line 89: I have two concerns with the diff's here:

One is that `x86_64` and `aarch64` implementations are missing.

More importantly, the source files are missing for build targets `ml_dsa_<44|65|87>_extmu_ref`. The following build will fail to compile:

```bash
cmake -GNinja -DOQS_MINIMAL_BUILD="SIG_ml_dsa_44_extmu" ..
ninja
```

The CI pipeline reports success because ExtMu targets and non-ExtMu targets are always compiled, so ExtMu targets calling `mldsa-native API's get to piggyback off non-ExtMu targets.

Maybe copying mldsa-native source code nine more times (3 external-mu times 3 architectures) is just way too much duplicate code in a single repository. Maybe consider reusing the source files from the non-ExtMu targets.

xuganyu96 (changes requested):
Please raise a pull request to `pq-code-package/mldsa-native` for the `META.yml` code change and link it here. We cannot merge this pull request with upstream pointing to a fork.

tests/test_acvp_vectors.py
abhi-dev-engg: They are not skipped and are testable in this PR. The json files from ACVP server don't include the variant "ML-DSA_xx-extmu", they are covered under the regular variant with 'externalMu' set as true.

have handled this accordingly , refer the below section in 'test_acvp_vectors.py' 

`variant["signatureInterface"] == "internal" and variant["externalMu"]:`

Also , refer to the newly added functions in vector_sig.c - 'sig_ver_vector_extmu' & 'sig_gen_vector_extmu' testing these cases via OQS APIs

abhi-dev-engg:
> Please raise a pull request to `pq-code-package/mldsa-native` for the `META.yml` code change and link it here. We cannot merge this pull request with upstream pointing to a fork.

I would prefer doing this as last step of this PR once all the review comments are in and fixes are approved. There is no point IMO in creating a mldsa-native PR now with reviews pending.

src/sig/ml_dsa/CMakeLists.txt
abhi-dev-engg: Great catch! I initially structured it this way to avoid duplicating the mldsa-native source code and triggering duplicate symbol errors, but I completely overlooked how that approach would break the minimal build and drop the optimized architectures.

I've updated the CMake template to handle this cleanly:

- If an extmu variant is enabled (even in a minimal build), it now forces the base target to compile the shared mldsa-native source files (including the x86_64 and aarch64 optimizations).

- The extmu target then only compiles its own C wrapper (sig_xx_extmu.c) and piggybacks off those base object files.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: could you pls resolve this if ok ?

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: could you pls resolve this if ok ?

tests/test_sig.c
abhi-dev-engg: could you pls resolve this if ok ?

abhi-dev-engg:
> Please raise a pull request to `pq-code-package/mldsa-native` for the `META.yml` code change and link it here. We cannot merge this pull request with upstream pointing to a fork.

Created PR for mldsa-native/main :https://github.com/pq-code-package/mldsa-native/pull/976. This PR description is updated with this info.

Since the upstream of mldsa-native points to v1.0.0-alpha2 tag, I checked with mldsa-native maintainers regarding merging changes to this tag, however they mentioned of a release starting of march.

Please find the conversation here : https://github.com/orgs/pq-code-package/discussions/202#discussioncomment-15932113.

abhi-dev-engg:
hi @xuganyu96 ,

could you pls have a look at the failing TCs, it seems to be not related to changes in this PR and look like generic failure affecting other PRs as well ?

[Failure 1](https://github.com/open-quantum-safe/liboqs/actions/runs/22436088694/job/64966853968?pr=2366) & [Failure 2](https://github.com/open-quantum-safe/liboqs/actions/runs/22436088694/job/64972903865?pr=2366)

xuganyu96:
> hi @xuganyu96 ,
> 
> could you pls have a look at the failing TCs, it seems to be not related to changes in this PR and look like generic failure affecting other PRs as well ?
> 
> [Failure 1](https://github.com/open-quantum-safe/liboqs/actions/runs/22436088694/job/64966853968?pr=2366) & [Failure 2](https://github.com/open-quantum-safe/liboqs/actions/runs/22436088694/job/64972903865?pr=2366)

image


Unfortunately coveralls.io is down (it's always DNS!). If the site does not come back up in a day or two we will consider some workarounds.

abhi-dev-engg:
Hi @xuganyu96 , @bhess 

The provided review comments have been addressed, could you pls have a look and resolve if all is ok ?
Thanks !

abhi-dev-engg:
The mldsa-native PR for the yaml changes is merged.

PR : https://github.com/pq-code-package/mldsa-native/pull/976

cc : @dstebila @xuganyu96 @bhess

docs/algorithms/sig/ml_dsa.yml
```diff
@@ -16,7 +16,7 @@ nist-round: FIPS204
 standardization-status: Standardized by [NIST](https://csrc.nist.gov/pubs/fips/204/final)
 spec-version: ML-DSA
 primary-upstream:
-  source: https://github.com/pq-code-package/mldsa-native/commit/f48f164cefb07f4ffa519ddda7cee670b8ee3517
+  source: https://github.com/abhi-dev-engg/mldsa-native/commit/d562dad07d0679a4cbe41860c5b981c630bc6484
```
xuganyu96: Please change sources back to `pq-code-package/mldsa-native`.

scripts/copy_from_upstream/copy_from_upstream.yml
xuganyu96: Please change sources back to pq-code-package/mldsa-native.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
```diff
@@ -181,6 +181,12 @@ OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_keypair(uint8_t *
 }
 
 OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_sign(uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *secret_key) {
+{%- if scheme.is_extmu %}
+	if (message_len != OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_length_mu) {
+		return OQS_ERROR;
+	}
+	(void)message_len;
```
xuganyu96: What is the purpose of `(void) message_len`? I thought `message_len` is already used in the if statement above.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
```diff
@@ -220,14 +226,20 @@ OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_sign(uint8_t *sig
...
 #endif
     {%- endif %}
 }
 
 OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_verify(const uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *public_key) {
+{%- if scheme.is_extmu %}
+	if (message_len != OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_length_mu) {
+		return OQS_ERROR;
+	}
+	(void)message_len;
```
xuganyu96: Same as above.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
```diff
@@ -277,6 +289,12 @@ OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_verify(const uint
 {%- set default_impl = scheme['metadata']['implementations'] | selectattr("name", "equalto", scheme['default_implementation']) | first %}
 {%- if 'api-with-context-string' in default_impl and default_impl['api-with-context-string'] %}
 OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_sign_with_ctx_str(uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, const uint8_t *ctx_str, size_t ctx_str_len, const uint8_t *secret_key) {
+{%- if scheme.is_extmu %}
+	if (message_len != OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_length_mu) {
+		return OQS_ERROR;
+	}
+	(void)message_len;
```
xuganyu96: Same as above.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
```diff
@@ -312,6 +330,12 @@ OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_sign_with_ctx_str
 }
 
 OQS_API OQS_STATUS OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_verify_with_ctx_str(const uint8_t *message, size_t message_len, const uint8_t *signature, size_t signature_len, const uint8_t *ctx_str, size_t ctx_str_len, const uint8_t *public_key) {
+{%- if scheme.is_extmu %}
+	if (message_len != OQS_SIG_{{ family }}_{{ scheme['scheme'] }}_length_mu) {
+		return OQS_ERROR;
+	}
+	(void)message_len;
```
xuganyu96: Same as above.

xuganyu96 (changes requested):

docs/algorithms/sig/ml_dsa.yml
abhi-dev-engg: auto generated when "copy_from_upstream" is run. see below comment

scripts/copy_from_upstream/copy_from_upstream.yml
abhi-dev-engg: IMO cannot be done as it will migrate mldsa-native from v1.0.0-alpha2 release to an intermediate release at the moment.

As mentioned by mldsa-native maintainers [here](https://github.com/orgs/pq-code-package/discussions/202#discussioncomment-15932113), mldsa-native will happen soon and the "upstream" would be replaced by the maintainers. We can merge this PR post that.

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: The upstream extmu functions drop the length parameter since mu is strictly fixed to 64 bytes, so I added the if statement to enforce this and prevent fuzzing errors. Because message_len is now actively evaluated in that check, the compiler's unused variable warning is already satisfied. I will remove the redundant (void) cast

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: refer above reply

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: refer above reply

scripts/copy_from_upstream/src/sig/family/sig_scheme.c
abhi-dev-engg: refer above reply

xuganyu96 (commented):
I am satisfied with the code change as it is now, with the exception of changing the source of mldsa-native from Abhi'personal fork back to a full release of mldsa-native from pq-code-package. Per @abhi-dev-engg's suggestion, we can wait for an mldsa-native release before merging this pull request.

abhi-dev-engg:
> I am satisfied with the code change as it is now, with the exception of changing the source of mldsa-native from Abhi'personal fork back to a full release of mldsa-native from pq-code-package. Per @abhi-dev-engg's suggestion, we can wait for an mldsa-native release before merging this pull request.

Thank you @xuganyu96 . could you pls "resolve" the review comments in such a case pls. helps me to differentiate b/w fixed & pending reviews.

Hi @bhess , could you pls re-review the changes and resolve if all is ok ?

scripts/copy_from_upstream/copy_from_upstream.py
bhess: Ok, I suppose known-answer tests for externalMu are covered by the ACVP tests?

bhess (commented):
Looks good. To confirm I'm following: we wait for the upstream release, update the input location to the upstream repo, and then merge. I'll approve once these changes are in.

abhi-dev-engg:
> Looks good. To confirm I'm following: we wait for the upstream release, update the input location to the upstream repo, and then merge. I'll approve once these changes are in.

yes, that's correct. The PR for the yaml changes in mldsa-native repo is already merged.
As a part of the mldsa-native release, the upstream yaml would be updated by the maintainers. Once that is done, I would rebase my changes upon that and then we can merge.

abhi-dev-engg:
> Ok, I suppose known-answer tests for externalMu are covered by the ACVP tests?

yes, ACVP tests cover external mu cases. I have modified vector_sig to use the "OQS_Sig" APIs for external-mu tc as well.

scripts/copy_from_upstream/copy_from_upstream.py
abhi-dev-engg: Yes, correct. ACVP tests cover external-mu

Mastmeetu:
@abhi-dev-engg any updates by when you are planning to complete this ?

I read through this and it is almost done and was waiting for mL-DSA merge which is now done.

Sorry for bothering and thanks in advance

abhi-dev-engg:
> @abhi-dev-engg any updates by when you are planning to complete this ?
> 
> I read through this and it is almost done and was waiting for mL-DSA merge which is now done.
> 
> Sorry for bothering and thanks in advance

sure, will try to get this done, would start on it by Fri. 
pls check back on 22.06. changing it to DRAFT in the meantime

abhi-dev-engg:
@Mastmeetu PR is upto date. only rebasing it against main was needed.

Mastmeetu:
> @Mastmeetu PR is upto date. only rebasing it against main was needed.
> 
> 

Thank you so much @abhi-dev-engg for the quick work.

Hi @bhess @xuganyu96 , this PR was reviewed by you earlier and approved for correctness. The pending was upstream changes which are included now.

Could you please have a look and approve ?

Thanks in advance!

xuganyu96:
> only rebasing it against main was needed.

@abhi-dev-engg Please rebase on top of main. Thank you.

abhi-dev-engg:
> > only rebasing it against main was needed.
> 
> @abhi-dev-engg Please rebase on top of main. Thank you.

@xuganyu96 done

Mastmeetu:
Confirming if this feature would be part of next release ?

xuganyu96:
> Confirming if this feature would be part of next release ?

Planned for 0.17.0 release.

Mastmeetu:
> > Confirming if this feature would be part of next release ?
> 
> 
> 
> Planned for 0.17.0 release.

Was assuming this release as you asked for a rebase. Sigh

xuganyu96:
> Was assuming this release as you asked for a rebase. Sigh

0.16 was already late compared to the normal release cadence. It also has a large number of major API/algorithm changes that piled on enormous amount of work for downstream projects like [oqs-provider](https://github.com/open-quantum-safe/oqs-provider). Hence the decision to draw this admittedly arbitrary line on what to include.

I understand that this pull request has been there for a while and will prioritize it as soon as 0.16 release lands. When this pull request lands, we can work out how you can check out a specific commit so you don't have to wait for 0.17.

xuganyu96 (approved):

Mastmeetu:
@xuganyu96 i understand. No worries at all. Thank you for your time and response

ashman-p (approved):
@abhi-dev-engg, Thanks for the contribution.
We have received some cases/issues concerning buffer overwrite/read. I did not see any new logic here that adds such problems (thanks).
In general, it may be good to add a note stating that we depend on the caller to do the "right thing" (allocate the correct buffer size matching the expected length parameters).
Although, this probably should be an OQS-wide statement.

PR #2383, opened 2026-03-17
Move compiler optimization level to CMAKE_BUILD_TYPE

xuganyu96:
This pull request simplifies how users specify compiler optimization level.

Previously, `compiler_opts.cmake` applies a blunt `add_compile_option(-O3)` globally. This actually duplicates C flags set by `CMAKE_BUILD_TYPE`. For example:

```bash
cmake -GNinja -DCMAKE_BUILD_TYPE="Release" ..
cat compile_commands.json
```
The flag `-O3` will be specified twice:

```json
{
  "directory": "/Users/ganyuxu/opensource/open-quantum-safe/liboqs/build",
  "command": "/usr/bin/cc -I/Users/ganyuxu/opensource/open-quantum-safe/liboqs/build/include -I/Users/ganyuxu/opensource/open-quantum-safe/liboqs/src -I/usr/local/include -O3 -DNDEBUG -std=gnu11 -arch arm64 -fPIC -fvisibility=hidden -march=armv8-a+crypto -Wa,--noexecstack -O3 -fomit-frame-pointer -o src/CMakeFiles/oqs.dir/kem/kem.c.o -c /Users/ganyuxu/opensource/open-quantum-safe/liboqs/src/kem/kem.c",
  "file": "/Users/ganyuxu/opensource/open-quantum-safe/liboqs/src/kem/kem.c",
  "output": "/Users/ganyuxu/opensource/open-quantum-safe/liboqs/build/src/CMakeFiles/oqs.dir/kem/kem.c.o"
}
```

In this pull request, `add_compile_option(-O3)` is removed from `compile_opts.cmake` (except for debug builds used in constant-time tests). Instead, `CMAKE_BUILD_TYPE` is given an explicit default `Release`. I also added documentation describing how developers can get other optimization levels like "RelWithDebInfo" (`-O2 -g -DNDEBUG`) and "MinSizeRel" (`-Os -DNDEBUG`).

Last but not lease, `compiler_opts.cmake` is run through an autoformatter to improve readability. The previous version has lots of confusing and inconsistent indents that make reading it unnecessarily difficult.

cc: @kaminuma 

* [No] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged. Also, make sure to update the list of algorithms in the continuous benchmarking files: .github/workflows/kem-bench.yml and sig-bench.yml)

No AI is involved in this pull request.


xuganyu96:
For reviewers: code change to SNOVA is described in [this commit message](https://github.com/open-quantum-safe/liboqs/pull/2383/changes/68c9a7d16ce44807d836e5dad7b3967a2b29d2f7)

dstebila:
Here's what Claude's code review of the PR found; suggested changes made in [3f6e3e6](https://github.com/open-quantum-safe/liboqs/pull/2383/commits/3f6e3e60481f039eb011be0e42ac2ec005ea6590).

**Multi-config generator handling for `CMAKE_BUILD_TYPE` default**

On multi-configuration generators (Xcode, Visual Studio, Ninja Multi-Config), `CMAKE_BUILD_TYPE` is not used — the configuration is selected at build time via `--config` and CMake reads `CMAKE_CONFIGURATION_TYPES` instead. The previous version of this PR unconditionally defaulted `CMAKE_BUILD_TYPE` to `Release` when unset, which on a multi-config generator would set the variable to `Release` but have no effect on the actual build. Worse, the `OQS_DEBUG_BUILD` derivation immediately below keys off `CMAKE_BUILD_TYPE`, so it would always compute `OFF` regardless of which configuration the user actually built — meaning a `Debug` build in Xcode/VS would silently miss `-g3`, sanitizers, and the other debug-only compile options in `compiler_opts.cmake`.

The fix guards the default with `GENERATOR_IS_MULTI_CONFIG` so the variable is only set on single-config generators (Ninja, Makefile), where it's actually meaningful:

```cmake
get_property(_isMultiConfig GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
if(NOT _isMultiConfig AND NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type" FORCE)
endif()
```

Note that the `OQS_DEBUG_BUILD` mechanism itself remains unsuitable for multi-config generators — properly fixing that would require converting the conditional `add_compile_options` calls in `compiler_opts.cmake` to per-config generator expressions (`$<$<CONFIG:Debug>:...>`), which is out of scope here. `CONFIGURE.md` has been updated to note that the `Release` default only applies to single-config generators.

dstebila (approved):

dstebila:
Bruce, would you like to merge this? Or anything further to do?

xuganyu96:
> Bruce, would you like to merge this? Or anything further to do?

Yes, please merge this as a standalone enhancement to liboqs's build system.

PR #2384, opened 2026-03-18
Fix Reading beyond buffer end

ashman-p:


src/sig_stfl/xmss/sig_stfl_xmss.h
```diff
@@ -425,6 +425,7 @@ OQS_API OQS_STATUS OQS_SIG_STFL_alg_xmss_shake256_h20_256_sigs_total(unsigned lo
 #define OQS_SIG_STFL_alg_xmssmt_sha256_h20_2_length_sk (5998 + XMSS_OID_LEN)
 #define OQS_SIG_STFL_alg_xmssmt_sha256_h20_2_length_pk (64 + XMSS_OID_LEN)
 #define OQS_SIG_STFL_alg_xmssmt_sha256_h20_2_length_signature 4963
+#define OQS_SIG_STFL_alg_xmss_mt_sha256_h20_2_length_signature OQS_SIG_STFL_alg_xmssmt_sha256_h20_2_length_signature
```
xuganyu96: What is the difference between `xmssmt_sha256_h20_2` and `xmss_mt_sha256_h20_2`?

ashman-p: No difference at all. The new name just made it a little easier for use in the macro.

xuganyu96: I find having two distinct macros `OQS_SIG_STFL_alg_xmssmt_sha256_h20_2_length_signature` and `OQS_SIG_STFL_alg_xmss_mt_sha256_h20_2_length_signature` that mean the same thing rather confusing. Is it possible to either provide better documentation or stick with one of them?

xuganyu96: > The new name just made it a little easier for use in the macro.

Would it make sense to migrate to the easier-to-use macro and remove the old ones? Having duplicate macros feels confusing to the user.

ashman-p: @xuganyu96 For XMSS, OQS_SIG_STFL_alg_xmss##xmss_v##_length_signature resolves (e.g. _sha256_h10 → OQS_SIG_STFL_alg_xmss_sha256_h10_length_signature). For XMSSMT, the macro expands to OQS_SIG_STFL_alg_xmss_mt_sha256_h20_2_length_signature while the header defines OQS_SIG_STFL_alg_xmssmt_sha256_h20_2_length_signature. 

The duplicate name makes it so that the macro constant resolves for both XMSS and XMSSMT.

I can add a comment that refers to the duplicate name as an alias if you think it would help clear up the confusion?

xuganyu96: > For XMSSMT, the macro expands to OQS_SIG_STFL_alg_xmss_mt_sha256_h20_2_length_signature while the header defines OQS_SIG_STFL_alg_xmssmt_sha256_h20_2_length_signature.

I am still confused about this point. Let's take this parameter set `XMSSMT_SHA256_H20_2`. The source file is `src/sig_stfl/xmss/sig_stfl_xmssmt_sha256_h20_2.c`, which looks like this:

```c
// SPDX-License-Identifier: (Apache-2.0 OR MIT) AND CC0-1.0

#include "sig_stfl_xmss_xmssmt.c"

// ======================== XMSSMT-SHA2_20/2_256 ======================== //

XMSS_ALG(mt, mt_sha256_h20_2, MT_SHA256_H20_2)
```

Looking at the `XMSS_ALG` macro in `sig_stfl_xmss_xmssmt.c`, the signature length is defined by:

```c
#define XMSS_ALG(mt, xmss_v, XMSS_V) \
/* ... */
        sig->length_signature = OQS_SIG_STFL_alg_xmss##xmss_v##_length_signature; \
/* ... */
```

So the macro should always expand to `OQS_SIG_STFL_alg_xmssmt_sha256_h20_2_length_signature` (no underscore between `xmss` and `mt`).

What am I missing here?

ashman-p: OK, I got you now. Reverting this change.

tests/test_sig_stfl.c
```diff
@@ -399,6 +399,36 @@ static char *convert_method_name_to_file_name(const char *method_name) {
...
+	if (sig == NULL) {
+		return OQS_ERROR;
+	}
+
+	uint8_t pk[68] = {0};
+	pk[3] = 0x01; //The Actual public key can be any value.
+
+	uint8_t message[] = "test";
+	uint8_t malicious_sig[10] = {0};  // Only 10 bytes!
+
```
baentsch, line 437: "magic" constants: 68, 3, 10: Why these? Why not others? This seems to be a very narrow "invalidity" test (that could be made to pass by equally very narrow code paths). What about making these "magic numbers" available to fuzzing tests instead (or in addition)?

ashman-p: Sounds like a good idea @baentsch. I will look into adding fuzzing in addition.

ashman-p: @baentsch i created an issue (XMSS Signature Verification Fuzzing #2399) to address this.

tests/test_sig_stfl.c
```diff
@@ -1022,6 +1051,7 @@ void *test_create_keys(void *arg) {
 void *test_correctness_wrapper(void *arg) {
 	struct thread_data *td = arg;
 	td->rc = sig_stfl_test_correctness(td->alg_name, td->katfile, td->bitflips_all, td->bitflips);
+	td->rc2 = test_invalid_sig(td->alg_name);
```
dstebila: `rc2` is only being checked in the non-pthreads codepath (lines 1291-1302) but not the pthreads codepath (lines 1286-1288).

tests/test_sig_stfl.c
```diff
@@ -399,6 +399,36 @@ static char *convert_method_name_to_file_name(const char *method_name) {
...
+	}
+
+	uint8_t pk[68] = {0};
+	pk[3] = 0x01; //The Actual public key can be any value.
+
+	uint8_t message[] = "test";
+	uint8_t malicious_sig[10] = {0};  // Only 10 bytes!
+
+	// This triggers the bug via proper API
+	OQS_STATUS status = OQS_SIG_STFL_verify(sig, message, sizeof(message), malicious_sig, 10, pk);
```
dstebila: Do you really want sizeof (which includes the `\0` nul string terminator) or strlen (which does not include it)?

dstebila (changes requested):

tests/test_sig_stfl.c
ashman-p: Updated.

tests/test_sig_stfl.c
baentsch: In my understanding of a review process it is people raising comments as part of a review should also be the ones marking them resolved if they find a good answer to their comment is given; anything else leads the concept of writing review comments ad absurdum, wouldn't you agree, @ashman-p ?

In this case, I see a reminder issue for my question created, but no trace/link back to this in the code base, explaining that the "magic constants" are just place holders for now (ideally with a "TBD" marker, so future readers of the code understand they're just meant to be temporary shortcuts).

tests/test_sig_stfl.c
baentsch: Same here -- I _really_ don't think it is good (n)etiquette to mark other's comments resolved -- particularly when done without any explanation.

ashman-p: > Same here -- I _really_ don't think it is good (n)etiquette to mark other's comments resolved -- particularly when done without any explanation.

Noted.

xuganyu96:
Please fix unit test failures. Thank you.

tests/test_sig_stfl.c
baentsch: Thanks for re-opening @ashman-p and for creating the issue. I'd have either of two suggestions how to resolve this comment now: Either create an explicit link in the comment above referencing #2399 (with a ToDo marker) OR review and merge https://github.com/open-quantum-safe/liboqs/pull/2400 that aims to resolve this right away -- which is a nice testament to FOSS cooperation...

baentsch (changes requested):
The core code change LGTM and is a good safety check: Thanks @ashman-p for that. Approved from my side as soon as the single comments get resolved,

tests/test_sig_stfl.c
ashman-p: You are welcome @baentsch. With regards to moving #2400 forward, I am not sure what my action item would be since the PR has not been approved as yet other than to wait.
Would you say that this current PR's approval depends on #2400? 
Do you find it interesting that #2400 seems to run without exposing the topic of this PR?

baentsch: > You are welcome @baentsch. With regards to moving #2400 forward, I am not sure what my action item would be since the PR has not been approved as yet other than to wait. 

You could review it, no?

> Would you say that this current PR's approval depends on #2400? 

No; if it were proven to show as resolved #2399, this comment would close as my concern would have been resolved by an issue that got resolved; otherwise, my suggestion for this PR to land is to add a pointer in this PR code to #2399 such as for the desire for improvement is not getting lost when looking at the code.

> Do you find it interesting that #2400 seems to run without exposing the topic of this PR?

As far as I can see it didn't run but failed; I did not yet check to see whether it is due merely for formatting reasons or whether the new test indeed doesn't trigger fuzzing as intended. I'd think a person with more experience in XMSS would be best placed to judge this.

xuganyu96 (approved):

tests/test_sig_stfl.c
dstebila: I've fixed the code formatting in #2400 and have restarted the CI. Norm, if you can look at this from a content perspective, that would be great.

ashman-p: I have suggested an update to #2400, which was accepted. I supplied the diffs and am awaiting the merge.

ashman-p: > Thanks for re-opening [@ashman-p](https://github.com/ashman-p) and for creating the issue. I'd have either of two suggestions how to resolve this comment now: Either create an explicit link in the comment above referencing [#2399](https://github.com/open-quantum-safe/liboqs/issues/2399) (with a ToDo marker) OR review and merge [#2400](https://github.com/open-quantum-safe/liboqs/pull/2400) that aims to resolve this right away -- which is a nice testament to FOSS cooperation...

@baentsch , #2400 is currently in progress and I think it will address your concern. Let me know otherwise?

dstebila:
@ashman-p My code review with Claude identified a few things still to fix, which I've done; see the commit message in commit https://github.com/open-quantum-safe/liboqs/pull/2384/changes/53ad5be890001740454341fb62eea35174c9a29d.

dstebila (dismissed):

dstebila:
Apparently a confluence of preprocessor macro and other options meant that certain code paths were rarely tested, resulting in several undiscovered bugs in the test harness. See commit messages above for details. I think they are resolved and this is ready to merge.

dstebila (approved):

dstebila:
@ashman-p Have you had a chance to look over the changes that I made? If you are satisfied with them then I think we can go ahead with merging.

ashman-p:
> @ashman-p Have you had a chance to look over the changes that I made? If you are satisfied with them then I think we can go ahead with merging.

@dstebila, Sorry for the delayed response. I had looked at the changes and agree. However, I was unable to approve because the PR tool blocks me as an author. I should have communicated this before. 
So, someone else (@xuganyu96, @baentsch) needs to click the approve button.

dstebila:
> > @ashman-p Have you had a chance to look over the changes that I made? If you are satisfied with them then I think we can go ahead with merging.
> 
> @dstebila, Sorry for the delayed response. I had looked at the changes and agree. However, I was unable to approve because the PR tool blocks me as an author. I should have communicated this before. So, someone else (@xuganyu96, @baentsch) needs to click the approve button.

I can override and approve following our practice of letting maintainers merge with 1 approving PR at their discretion.

tests/test_sig_stfl.c
baentsch: #2400 just now landed and this comment thus is now addressed and thus is resolved. Pointless comment though as this PR has already been merged. Makes me feel a bit like my comments are unnecessary -- at least when given outside of reviews -- which I do intentionally to not "collect GH points" and thus create the impression of activity that isn't real.

PR #2386, opened 2026-03-19
Add missing ACVP tests [ML-KEM]

abhi-dev-engg:
This PR introduces the previously missing encapsulation and decapsulation key check ACVP tests. These tests were likely disabled earlier because the “sanity-check” APIs were "static"  in v1.0.0 version. With the release of v1.2.0, these APIs are now global, enabling execution of these missing tests.

Also, updates ACVP version to latest.


xuganyu96:
> To be reviewed/merged after mlkem-native v1.1.0 integration.

@abhi-dev-engg mlkem-native v1.2.0 landed with #2470. Does that unblock this pull request?

abhi-dev-engg:
> > To be reviewed/merged after mlkem-native v1.1.0 integration.
> 
> @abhi-dev-engg mlkem-native v1.2.0 landed with #2470. Does that unblock this pull request?

@xuganyu96  Thanks for the hint. yes, it unblocks now. have rebased main on my branch now.
Also updated ACVP version to 1.1.0.42

tests/test_acvp_vectors.py
```diff
@@ -193,6 +193,44 @@ def test_acvp_vec_kem_encdec_val(kem_name):
 
     assert variantFound == True
 
+@helpers.filtered_test
+@pytest.mark.parametrize("kem_name", helpers.available_kems_by_name())
+def test_acvp_vec_kem_key_check(kem_name):
+    if not (helpers.is_kem_enabled_by_name(kem_name)):
+        pytest.skip("Not enabled")
+    if not (kem_name in ml_kem):
```
xuganyu96, line 201: Since this test only applies to ML-KEM, consider only parameterize `kem_name` with ML-KEM names to begin with so we can remove a non-trivial number of skipped tests.

xuganyu96 (changes requested):
Hi @abh-dev-env, thank you for the contribution.

Key checks are indeed nice additions to the test suite, though my main concern lies with the approach taken in `tests/vectors_kem.c`, which makes use of functions that are internal to liboqs. liboqs intends to expose a consistent API across all KEMs, so even though `check_pk` and `check_sk` are public API for mlkem-naitve, they should be treated as private API in liboqs. In fact, this pull request only tests the portable implementations of `check_pk`/`check_sk`, but there are also x86_64/aarch64 implementations (e.g. on my M1 MacBook, compiling liboqs with default settings gives me a symbol called `PQCP_MLKEM_NATIVE_MLKEM768_AARCH64_check_sk`). I think the correct approach is to expand liboqs' KEM API set to include key checks, then use individual implementation's underlying key check routine to implement them. This way, `vectors_kem.c` will only make calls to OQS API's.

tests/test_acvp_vectors.py
abhi-dev-engg: I have tried to keep the uniformity as done for other functions in this file. The same approach is done [here](https://github.com/open-quantum-safe/liboqs/blob/main/tests/test_acvp_vectors.py#L159) and [here](https://github.com/open-quantum-safe/liboqs/blob/main/tests/test_acvp_vectors.py#L114) for instances

abhi-dev-engg:
> Hi @abh-dev-env, thank you for the contribution.
> 
> Key checks are indeed nice additions to the test suite, though my main concern lies with the approach taken in `tests/vectors_kem.c`, which makes use of functions that are internal to liboqs. liboqs intends to expose a consistent API across all KEMs, so even though `check_pk` and `check_sk` are public API for mlkem-naitve, they should be treated as private API in liboqs. In fact, this pull request only tests the portable implementations of `check_pk`/`check_sk`, but there are also x86_64/aarch64 implementations (e.g. on my M1 MacBook, compiling liboqs with default settings gives me a symbol called `PQCP_MLKEM_NATIVE_MLKEM768_AARCH64_check_sk`). I think the correct approach is to expand liboqs' KEM API set to include key checks, then use individual implementation's underlying key check routine to implement them. This way, `vectors_kem.c` will only make calls to OQS API's.

Thanks for the review and comment @xuganyu96. I kinda agree here with your approach, however I am not sure extending KEM APIs is a good idea here. The check_sk/pk APIs are specific to ML-KEM and extending to OQS_KEM_XX APIs for ML-KEM while disabling it for all the other KEM algos is not a good idea IMO.

This is a similar approach that exists at the moment for ML-DSA as well where the "internal-signature" & "external-mu" signature are tested via the internal APIs. check [here](https://github.com/open-quantum-safe/liboqs/blob/main/tests/vectors_sig.c#L24)

Let me know your opinion.

xuganyu96:
> I kinda agree here with your approach, however I am not sure extending KEM APIs is a good idea here. The check_sk/pk APIs are specific to ML-KEM and extending to OQS_KEM_XX APIs for ML-KEM while disabling it for all the other KEM algos is not a good idea IMO.

Considering that deterministic keygen and encap are not universally implemented (not all KEM's support deterministic routines, nor do all impl's of a KEM support deterministic routines), I think "not all KEM's have key checks" is a weak argument against implementing key checks.

edit: there are many approaches we can take to address "not all KEM's have key check": assume all keys are valid by default, use a status code to indicate "key check not implemented", etc.

cc: @dstebila @bhess @baentsch @RodriM11 for additional inputs.

dstebila:
> > I kinda agree here with your approach, however I am not sure extending KEM APIs is a good idea here. The check_sk/pk APIs are specific to ML-KEM and extending to OQS_KEM_XX APIs for ML-KEM while disabling it for all the other KEM algos is not a good idea IMO.
> 
> Considering that deterministic keygen and encap are not universally implemented (not all KEM's support deterministic routines, nor do all impl's of a KEM support deterministic routines), I think "not all KEM's have key checks" is a weak argument against implementing key checks.
> 
> edit: there are many approaches we can take to address "not all KEM's have key check": assume all keys are valid by default, use a status code to indicate "key check not implemented", etc.
> 
> cc: @dstebila @bhess @baentsch @RodriM11 for additional inputs.

I don't have an opinion on this right now. I'd be curious to know how consumers would/wouldn't use it.

ashman-p:
> I don't have an opinion on this right now. I'd be curious to know how consumers would/wouldn't use it.
@dstebila, are you asking how "deterministic keygen" would be used?

dstebila:
> > I don't have an opinion on this right now. I'd be curious to know how consumers would/wouldn't use it.
>
> @dstebila, are you asking how "deterministic keygen" would be used?

I thought the question was about the check key functions, but perhaps I've misunderstood.

baentsch:
> edit: there are many approaches we can take to address "not all KEM's have key check": assume all keys are valid by default, use a status code to indicate "key check not implemented", etc.
> cc: @dstebila @bhess @baentsch @RodriM11 for additional inputs.

As you asked, just for the record, one more time: I still don't think that adding code paths and APIs that are exercised/used by only some algorithms is a good idea: The big advantage of `liboqs` is that it's making all algorithms useable in the same way. If users have to guard/check API calls to cater for algorithm specific features, this benefit falls away. If it's only a single algorithm (supporting the feature), this is even worse. If it's an algorithm that's standardized, i.e., that in my opinion doesn't belong into OQS to begin with, it's counterproductive indeed: It's making everyone's life unnecessarily complicated: Users have to learn new things/write algorithm-specific code and the OQS core team has to add special code paths, document & test them, discuss etc -- all detracting it from supporting progress, i.e., adding new algorithms behind the _one-and-the-same/algorithm agnostic API_ that has served the OQS community pretty well over a decade.

If users want to utilize standardized algorithms, why not point them to standard crypto libraries, e.g., `openssl`? Those already have well-documented APIs with well-known ways how to use algorithm-specific features. I don't see any value in `liboqs` only now beginning to reinvent that wheel: Long discussions, new tests, new code paths, consequently new bugs, new user questions leading to more discussions: Wasted effort all around.

If users want to experiment, compare-and-contrast new algorithms "apples-to-apples" they should be able to have and use a library with APIs permitting this, i.e., `liboqs` -- and the OQS team's priority should (remain to) be to focus on enabling this, not be detracted trying to play catch-up to features readily available in (by now many) other FOSS libraries.

xuganyu96:
> I still don't think that adding code paths and APIs that are exercised/used by only some algorithms is a good idea

Fair point.

Since expanding liboqs' common API is a no-go at this moment, and I am not okay with unit tests using what liboqs would consider private internals, I am afraid the key check tests from ACVP are not a good fit for liboqs at this moment. Let me know if you think otherwise @abhi-dev-engg, or feel free to close this pull request. Thank you.

dstebila:
> Since expanding liboqs' common API is a no-go at this moment, and I am not okay with unit tests using what liboqs would consider private internals, I am afraid the key check tests from ACVP are not a good fit for liboqs at this moment. Let me know if you think otherwise @abhi-dev-engg, or feel free to close this pull request. Thank you.

If memory serves, we've had unit tests call into private API functions before. I can't remember the exact use case, but I think that what's the PRIVATE keyword was used for in tests/CMakeLists.txt and src/CMakeLists.txt...

abhi-dev-engg:
> > Since expanding liboqs' common API is a no-go at this moment, and I am not okay with unit tests using what liboqs would consider private internals, I am afraid the key check tests from ACVP are not a good fit for liboqs at this moment. Let me know if you think otherwise @abhi-dev-engg, or feel free to close this pull request. Thank you.
> 
> If memory serves, we've had unit tests call into private API functions before. I can't remember the exact use case, but I think that what's the PRIVATE keyword was used for in tests/CMakeLists.txt and src/CMakeLists.txt...

That's correct, the same has been done for ML-DSA "internal-api" ACVP tests since quite a while, 0.12.0 if I am not mistaken.

Testing the APIs(internal/external) only strengthens the confidence in code IMO as these individual tests may cover corner cases that may otherwise may not available in regular encaps/decaps tests. Also, since this is all done in "test" folder, I don't see any harm here.

That said, pls feel free to close the PR if the decision is otherwise.

xuganyu96 (approved):
> That's correct, the same has been done for ML-DSA "internal-api" ACVP tests since quite a while, 0.12.0 if I am not mistaken.

Thanks for pointing that out.

baentsch:
> > That's correct, the same has been done for ML-DSA "internal-api" ACVP tests since quite a while, 0.12.0 if I am not mistaken.
> 
> Thanks for pointing that out.

Err, do dubious decisions in the past justify repeating them? Shouldn't those rather be re-visited to retain a clear, logical, and thus arguably more safe and maintainable architecture?

dstebila:
> > > That's correct, the same has been done for ML-DSA "internal-api" ACVP tests since quite a while, 0.12.0 if I am not mistaken.
> > 
> > 
> > Thanks for pointing that out.
> 
> Err, do dubious decisions in the past justify repeating them? Shouldn't those rather be re-visited to retain a clear, logical, and thus arguably more safe and maintainable architecture?

I don't think it's a mistake to have test code that calls internal functions to provide greater assurance.

abhi-dev-engg:
> > Err, do dubious decisions in the past justify repeating them? Shouldn't those rather be re-visited to retain a clear, logical, and thus arguably more safe and maintainable architecture?
> 
> I don't think it's a mistake to have test code that calls internal functions to provide greater assurance.

I agree with @dstebila. I don't see an architecture or design issue here.

If this PR were introducing a KEM-specific external API, I would agree with the concerns raised. This PR only **tests** a mandatory check introduced by FIPS 203 using ACVP test vectors. The intent is simply to validate the required logic, not to expose new functionality or alter the public API.

xuganyu96:
> > > > That's correct, the same has been done for ML-DSA "internal-api" ACVP tests since quite a while, 0.12.0 if I am not mistaken.
> > > 
> > > 
> > > Thanks for pointing that out.
> > 
> > 
> > Err, do dubious decisions in the past justify repeating them? Shouldn't those rather be re-visited to retain a clear, logical, and thus arguably more safe and maintainable architecture?
> 
> I don't think it's a mistake to have test code that calls internal functions to provide greater assurance.

A similar issue was re-iterated recently [here](https://github.com/open-quantum-safe/liboqs/pull/2522#issue-5106322425). Even though I approved this pull request testing mlkem-native internal API's, I still think it is against best practices for OQS tests to be assuming the existence and expected behaviors of implementation-specific API's, and I still advocate for surfacing implementation-specific API's up to OQS public API if testing them is helpful with providing additional assurance.

baentsch:
> A similar issue was re-iterated recently https://github.com/open-quantum-safe/liboqs/pull/2522#issue-5106322425. Even though I approved this pull request testing mlkem-native internal API's, I still think it is against best practices for OQS tests to be assuming the existence and expected behaviors of implementation-specific API's, and I still advocate for surfacing implementation-specific API's up to OQS public API if testing them is helpful with providing additional assurance.

I couldn't agree more.

> I don't think it's a mistake to have test code that calls internal functions to provide greater assurance.

Greater assurance of what exactly? That an upstream-specific test passes? Or in other words, is it sensible to run single-upstream-specific test APIs by a library supposedly delivering all algorithms and implementations it carries to the same (in this case, test-)assurance level? This PR/code seems to be a case of merely duplicating efforts (if it's upstream-specific it should be/have been tested there already) and intransparent/not comparable with others, i.e., via an API every alg (or at the very least, every implementation of the same type) has; basically @xuganyu96 's point.

I think this code excerpt illustrates the point pretty clearly:

```
#ifdef OQS_ENABLE_KEM_ml_kem_512
	if (!strcmp(method_name, OQS_KEM_alg_ml_kem_512)) {
		result = is_pk ?
		         PQCP_MLKEM_NATIVE_MLKEM512_C_check_pk(key) :
		         PQCP_MLKEM_NATIVE_MLKEM512_C_check_sk(key);
```

This to me reads like a one-off test for not only one algorithm but indeed for only one upstream of that algorithm.

Tagging @mkannwischer : Is this test adding additional assurance over what your upstream testing already covers?

Finally, this code seems to be outside of any `copy_from_upstream` logic and will break immediately when a new implementation (or possibly, even already on a new version) comes in via that path, no?

mkannwischer:
> Greater assurance of what exactly? That an upstream-specific test passes? Or in other words, is it sensible to run single-upstream-specific test APIs by a library supposedly delivering all algorithms and implementations it carries to the same (in this case, test-)assurance level? This PR/code seems to be a case of merely duplicating efforts (if it's upstream-specific it should be/have been tested there already) and intransparent/not comparable with others, i.e., via an API every alg (or at the very least, every implementation of the same type) has; basically @xuganyu96 's point.
> 
> I think this code excerpt illustrates the point pretty clearly:
> 
> ```
> #ifdef OQS_ENABLE_KEM_ml_kem_512
> 	if (!strcmp(method_name, OQS_KEM_alg_ml_kem_512)) {
> 		result = is_pk ?
> 		         PQCP_MLKEM_NATIVE_MLKEM512_C_check_pk(key) :
> 		         PQCP_MLKEM_NATIVE_MLKEM512_C_check_sk(key);
> ```
> 
> This to me reads like a one-off test for not only one algorithm but indeed for only one upstream of that algorithm.
> 
> Tagging @mkannwischer : Is this test adding additional assurance over what your upstream testing already covers?

We are running the exact same test upstream. The only extra assurance that this is adding is that the OQS build configuration does not break the test on the platforms OQS is tested on. If this is really useful for anyone that probably means we should be improving the test upstream. Please report any test gaps upstream and we are happy to try closing them.

I think OQS consumers should be discouraged from calling into implementation APIs directly. If consumers need additional APIs, they should be added as OQS APIs. Key validation APIs like this are not specific to ML-KEM, so I don't think it would be too crazy (although the additional benefit for OQS consumers might be too small to justify the effort). 

The motivation to add the test seems to be to be able to tick all the boxes for ACVP. I'm not convinced that this is a goal of OQS, but others may disagree. If it is considered important, one could go for a compromise: It is possible to perform this test by just calling the encaps and decaps APIs as they will also perform key validation internally. That would be a way to still run the tests while still going through the OQS API. It's probably not good enough for actual certification, but (again) I don't think that is a goal of OQS.

> 
> Finally, this code seems to be outside of any `copy_from_upstream` logic and will break immediately when a new implementation (or possibly, even already on a new version) comes in via that path, no?

Yes, this can break at any time and makes our live harder. It don't think v2.0.0 breaks it, though.

PR #2394, opened 2026-04-02
fix: build on windows clang

Nelonn:

Some options is not available in pure clang like google uses (not clang-cl) and CMake has great property called `MSVС`, its true only when compiler is MSVC or clang-cl


xuganyu96:
> Some options is not available in pure clang like google uses (not clang-cl) and CMake has great property called `MSVС`, its true only when compiler is MSVC or clang-cl

Sorry that we are not entirely familiar with Windows build tools. Could you please explain:
- What options are not available in pure clang? What do you mean by "pure clang" on Windows?
- What specific toolchain is this pull request trying to accommodate?

Nelonn:
> > Some options is not available in pure clang like google uses (not clang-cl) and CMake has great property called `MSVС`, its true only when compiler is MSVC or clang-cl
> 
> Sorry that we are not entirely familiar with Windows build tools. Could you please explain:
> 
>     * What options are not available in pure clang? What do you mean by "pure clang" on Windows?
> 
>     * What specific toolchain is this pull request trying to accommodate?

Pure clang is clang with GNU-like command line. Not clang-cl (Clang with MSVC-like command line). Pure clang doesn't support: `/wd4996 /wd4244` cause its MSVC only. I'm unable to build liboqs with pure clang and done this changes, other toolchains will work unchanged

xuganyu96:
> Pure clang is clang with GNU-like command line. Not clang-cl (Clang with MSVC-like command line). Pure clang doesn't support: `/wd4996 /wd4244` cause its MSVC only. I'm unable to build liboqs with pure clang and done this changes, other toolchains will work unchanged

@Nelonn Thank you for the context.

I found some similar code from `aws/aws-lc` (which seems to inherit lots of stuff from [Google's BoringSSL](https://boringssl.googlesource.com/boringssl/+/6b34d549451ca98d1ab758bf509e2a03b4eedc4a%5E%21/#F0)), specifically:
- [this commit from 2016](https://github.com/aws/aws-lc/commit/6b34d549451ca98d1ab758bf509e2a03b4eedc4a) adds a similar snippet that identifies Windows using the WIN32 CMake variable
- [this later commit from 2021](https://github.com/aws/aws-lc/commit/66e35105235735af9299b08ee5f9e5fbe4b31be4) which uses the CMake variable `MSVC` instead of `WIN32`

@Nelonn Do you think following the example above (using only `MSVC`) makes sense for your use case? Is `WIN32` redundant when `MSVC` already implies Windows?

Nelonn:
> > Pure clang is clang with GNU-like command line. Not clang-cl (Clang with MSVC-like command line). Pure clang doesn't support: `/wd4996 /wd4244` cause its MSVC only. I'm unable to build liboqs with pure clang and done this changes, other toolchains will work unchanged
> 
> @Nelonn Thank you for the context.
> 
> I found some similar code from `aws/aws-lc` (which seems to inherit lots of stuff from [Google's BoringSSL](https://boringssl.googlesource.com/boringssl/+/6b34d549451ca98d1ab758bf509e2a03b4eedc4a%5E%21/#F0)), specifically:
> 
>     * [this commit from 2016](https://github.com/aws/aws-lc/commit/6b34d549451ca98d1ab758bf509e2a03b4eedc4a) adds a similar snippet that identifies Windows using the WIN32 CMake variable
> 
>     * [this later commit from 2021](https://github.com/aws/aws-lc/commit/66e35105235735af9299b08ee5f9e5fbe4b31be4) which uses the CMake variable `MSVC` instead of `WIN32`
> 
> 
> @Nelonn Do you think following the example above (using only `MSVC`) makes sense for your use case? Is `WIN32` redundant when `MSVC` already implies Windows?

I suppose that WIN32 is actually redundant

xuganyu96:
@Nelonn Please fix the [DCO errors](https://github.com/open-quantum-safe/liboqs/pull/2394/checks?check_run_id=70479719744), and rebase your branch. After that I can run the test pipeline and approve this pull request. Thank you.

Nelonn:
Idk for what purpose its needed, but I done that

xuganyu96 (approved):

PR #2397, opened 2026-04-06
Fix: Limit pytest processes to prevent OOM on low-RAM systems

3-LUMAnOVA-3:
Hello OQS team,
While building and testing liboqs on a low-resource environment (4GB RAM), the run_tests target triggered an OQS_MEM_malloc failed error during the Wycheproof ML-DSA vectors validation.
This was caused by the --numprocesses=auto flag, which spawned too many parallel intensive tests for the available memory, causing an Out-Of-Memory (OOM) crash.
This patch limits the pytest parallel processes to 2 in tests/CMakeLists.txt to ensure stability and prevent potential memory core dumps on smaller edge/IoT devices during testing.
RowBinWeb
(Salutations from Québec  ⚜️)

        =====================================================================================


xuganyu96 (changes requested):
Thank you for identifying this issue and your willingness to contribute!

I understand that parallelizing memory-intensive tests can cause unexpected test failures and agree that we can do something about making the test suite more flexible. My main concern is with the approach taken in this pull request: hardcoding the number of parallel test jobs to 2 will slow down the test pipeline on resource-abundant systems. Such is the case with our CI pipeline, where memory is not constrained, and slowing down pytest means we will be paying extra for CPU hours.

I would like to propose an alternative approach: define a CMake variable that can accept a value from the command line such as `cmake -DOQS_PYTEST_NUMPROCESSES=1`, then pass to the `pytest` command. This variable can default to `auto` so that existing test infrastructure will work just as before, but users on memory-constrained system will have an easy entrypoint for controlling the number of parallel tests.

tests/CMakeLists.txt
```diff
@@ -240,7 +240,7 @@ endif()
 add_custom_target(
     run_tests
     # skip long KAT tests
-    COMMAND ${CMAKE_COMMAND} -E env OQS_BUILD_DIR=${CMAKE_BINARY_DIR} ${PYTHON3_EXEC} -m pytest --verbose --numprocesses=auto --ignore=scripts/copy_from_upstream/repos --ignore=tests/test_kat_all.py
+    COMMAND ${CMAKE_COMMAND} -E env OQS_BUILD_DIR=${CMAKE_BINARY_DIR} ${PYTHON3_EXEC} -m pytest --verbose --numprocesses=${OQS_PYTEST_NUMPROCESSES} --ignore=scripts/copy_from_upstream/repos --ignore=tests/te>
```
xuganyu96: > `--ignore=tests/te>`

This seems to be a typo.

xuganyu96 (commented):
Please fix the DCO errors. Otherwise LGTM.

3-LUMAnOVA-3:
@xuganyu96 Hello! The typo has been fixed and the DCO is now properly signed. Thank you for your review, let me know if anything else is needed.

xuganyu96 (approved):

baentsch (approved):

3-LUMAnOVA-3:
Hi @dstebila, I'm just checking in on this PR. It has been approved by @baentsch and @xuganyu96, and all 93 checks are passing. Please let me know if there is anything else I should address to get this merged.                               (Salutations du Québec ⚜️)

dstebila:
> Hi @dstebila, I'm just checking in on this PR. It has been approved by @baentsch and @xuganyu96, and all 93 checks are passing. Please let me know if there is anything else I should address to get this merged. (Salutations du Québec ⚜️)

Thanks for following up.  I've merged it!

Issue #2401, opened 2026-04-09
armhf tests failing (weekly tests)


bhess:
### Description

The weekly and extended CI tests have been [failing](https://github.com/open-quantum-safe/liboqs/actions/runs/23991172854/job/69971081777) for some time. The failures originate from the qemu‑emulated armhf test jobs.

The armhf CI container is based on Debian buster, which has been end‑of‑life since [June 2024](https://endoflife.date/debian). As a result, several required Python dependencies are not available in the base image, causing the armhf tests to fail consistently.

Proposed options
1. Migrate armhf tests to a supported base image, such as a current Debian release or Ubuntu.
2. Remove the armhf tests from CI entirely.

### liboqs version

main

bhess:
Resolved by removing armhf CI and move it from Tier 2 to Tier 3 (#2406)

PR #2403, opened 2026-04-09
Pull MAYO update from upstream (fixing issue reported by @programsurf)

### LINKED ISSUE #2392: MAYO signing function returns `MAYO_OK` after 256 failed `sample_solution()` attempts
### Description

Hi, the MAYO signing function (`mayo_sign_signature`) initializes `ret = MAYO_OK` at line 362, then enters a 256-iteration rejection sampling loop calling `sample_solution()`. If none of the 256 attempts find a valid solution, execution falls through to the signature assembly code which runs unconditionally:

```c
int ret = MAYO_OK;                                    // line 362
// ...
for (int ctr = 0; ctr <= 255; ++ctr) {
    // ...
    if (sample_solution(p, A, y, r, x, ...)) {
        break;                                         // found valid solution
    } else {
        memset(Mtmp, 0, sizeof(Mtmp));
        memset(A, 0, sizeof(A));
    }
}
// No check whether the loop succeeded — falls through regardless

for (int i = 0; i <= param_k - 1; ++i) {             // line 477
    // assembles signature from x (garbage if loop failed)
    mat_mul(sk.O, x + i * param_o, Ox, ...);
    mat_add(vi, Ox, s + i * param_n, ...);
}
encode(s, sig, param_n * param_k);
// ...
return ret;                                            // still MAYO_OK
```

`ret` is never updated on loop exhaustion. The liboqs wrapper translates `MAYO_OK` to `OQS_SUCCESS`, so the caller has no way to know the signature is garbage.

This affects all 12 MAYO variants in `src/sig/mayo/pqmayo_mayo-{1,2,3,5}_{opt,avx2,aarch64}/mayo.c`. The upstream source is https://github.com/PQCMayo/MAYO-C.

### Expected behaviour

When all 256 `sample_solution()` attempts fail, the function should return an error (`MAYO_ERR`) instead of assembling a signature from invalid data and returning `MAYO_OK`.

A simple fix:

```c
int found = 0;
for (int ctr = 0; ctr <= 255; ++ctr) {
    if (sample_solution(...)) { found = 1; break; }
    // ...
}
if (!found) { ret = MAYO_ERR; goto err; }
```

### liboqs version

0.15.0 (commit ab0e07c3)

### Environment

- Architecture: x86_64
- OS: Ubuntu 24.04 LTS
- Found by source code analysis, not runtime testing

### Use of generative AI

This bug was identified during a cross-library PQC implementation audit using a combination of automated pattern scanning (Patch2Claim) and manual source code review. The issue description was drafted with the assistance of Claude (Anthropic). The source code analysis and vulnerability classification were verified manually.

### Additional information

With valid keys, this is practically unreachable — MAYO targets < 2^{-128} failure probability per signing attempt. However, a cryptographic signing function should never silently return success with invalid output. With malformed keys, all 256 attempts can fail.

This should also be reported upstream to https://github.com/PQCMayo/MAYO-C since liboqs vendors this code.

bhess:
Pulls in the latest MAYO update from upstream, which includes the fix from PQCMayo/MAYO-C#10.

Fixes #2392.

Note: armhf CI failures are unrelated and tracked separately in #2401.


bhess:
cc @programsurf who reported the issue.

dstebila (approved):

PR #2407, opened 2026-04-14
Update HQC implementation to 2025-08-22 version

### LINKED ISSUE #2118: Update HQC implementation to 2025/02/19 version
Hi all,

**Describe the bug**
The liboqs implementation of HQC is not consistent with the [2025/02/19 update](https://pqc-hqc.org/doc/hqc-reference-implementation_2025-02-19.zip) of the reference implementation. Namely, the changes introduced in the [2024/10/30 update](https://pqc-hqc.org/doc/hqc-submission_2024-10-30.zip) are not included:

- The order of variable sampling in `hqc_pke_encrypt` from `hqc.c` has been modified. The reference implementation now generates `r2`, `e`, `r1` instead of `r1`, `r2`, `e`. For example, in [hqc-128/hqc.c](https://github.com/open-quantum-safe/liboqs/blob/6337a8424deae09aa835ddd920faff83a8f0e1d7/src/kem/hqc/pqclean_hqc-128_clean/hqc.c#L93).
- The randomness `theta` in `kem.c` is now computed using the first 32 bytes of the public key, instead of the entire key. Concretely, the reference implementation uses `SALT_SIZE_BYTES * 2`. For example, in [hqc-128/kem.c](https://github.com/open-quantum-safe/liboqs/blob/6337a8424deae09aa835ddd920faff83a8f0e1d7/src/kem/hqc/pqclean_hqc-128_clean/kem.c#L61).
- Not mentioned in the history of specification updates is the fact that HQC dropped the `x` argument in `hqc_secret_key_from_string`. While the value is not used in `hqc_pke_decrypt`, the fact that both `x` and `y` are derived from a common SHAKE256 state means that the secret key is interpreted differently. Removing the first call to `vect_set_random_fixed_weight` works in my tests. For example, in [hqc-128/parsing.c](https://github.com/open-quantum-safe/liboqs/blob/6337a8424deae09aa835ddd920faff83a8f0e1d7/src/kem/hqc/pqclean_hqc-128_clean/parsing.c#L94).

For some context, I am the developer of [crypto-condor](https://github.com/quarkslab/crypto-condor), a tool to test the conformity of cryptographic implementations. I discovered these differences while adding HQC to our supported primitives, since I used the reference implementation and the included NIST KAT script to generate the test vectors. The liboqs implementation was not passing the `secret == decaps(encaps(pk))` test, which pointed to a difference in how keys are read. I started by testing the Python bindings from [liboqs-python](https://github.com/open-quantum-safe/liboqs-python) but ended up using the shared library directly to check the issue came from the source. By applying the changes described, liboqs passes the tests for decapsulation and `decaps(encaps(pk))`.

**To Reproduce**

1. Compile liboqs as a shared library. I followed the instructions in [liboqs-python](https://github.com/open-quantum-safe/liboqs-python).

```bash
git clone --depth=1 https://github.com/open-quantum-safe/liboqs
cmake -S liboqs -B liboqs/build -DBUILD_SHARED_LIBS=ON
cmake --build liboqs/build --parallel 8
cmake --build liboqs/build --target install
```

2. Create a sample program that uses HQC in any of its parameter sets.
4. Test HQC using the KATs from the reference implementation.
5. The returned values do not match the reference values.

**Expected behavior**
I expect the liboqs implementation to pass the KATs from the reference implementation.

**Screenshots**
n/a

 - OS: Fedora 40
 - OpenSSL version: 3.2.4
 - Compiler version used: gcc 14.2.1
 - Build variables used: "-DBUILD_SHARED_LIBS=ON"
 - liboqs version: main branch

**Additional context**
PQClean currently provides the 2023/04/30 version of HQC, as noted in [META.yml](https://github.com/PQClean/PQClean/blob/448c71a8f590343e681d0d0cec94f29947b0ff18/crypto_kem/hqc-128/META.yml), so the issue comes from upstream. I opened the issue here as this is the implementation I actually tested: compiling the shared library and creating a test harness was very easy, so kudos for ease of use!

xuganyu96:
This pull request replaces the PQClean implementation of HQC with the reference implementation, originally hosted on [GitLab](https://gitlab.com/pqc-hqc/hqc). Fixes #2118 

All code change is developed on a [personal fork](https://github.com/xuganyu96/pqc-hqc/tree/liboqs-integration). After reaching sufficient maturity, the code change is git-diff'ed into a patch file. This admittedly will make future development more tedious, so let me know if we should continue using a personal fork. We can also reach out to the team to merge our code change.

```bash
# from xuganyu96/pqc-hqc at branch liboqs-integration
# 161cd4f is the base commit
git diff 161cd4f --patch > $LIBOQS_DIR/scripts/copy_from_upstream/patches/pqchqc-liboqs-integration.patch
```

# Code change
There are several significant changes to the original implementation:

**Namespacing** was added to the original implementation. Every non-static symbol has a `PQCHQC_NAMESPACE` macro applied so that a user-defined prefix will be attached (e.g. `PQCHQC_HQC1_C_` for the portable implementation of HQC-1). Several symbols that were not used across different source files were additionally declared `static`.

**OQS common primitives** replaced the included implementation of SHA3/Keccak and PRNG.

**Name of algorithms** have changed from HQC-128/192/256 to HQC-1/3/5.

**Other**:
- Some debug logging facility added to `src/common`
- Reorganized the `test_kat.py` test so that on initial failure, the test binary is re-run to capture debugging output

# Known limitations
- `x86_64` optimized implementations were not included
- Branching-on-secret was not validated against valgrind
- On GitHub Actions, some HQC unit tests intermittently fail, but not consistently

*No LLM/AI was used in this pull request*

* [**YES**] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [**YES**] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged. Also, make sure to update the list of algorithms in the continuous benchmarking files: .github/workflows/kem-bench.yml and sig-bench.yml)


xuganyu96:
# Further modification to original implementation
Replacing original` impl's in-house `fips202` module with OQS_SHA3 introduced some memory leaks that caused the memory sanitizer failures:

```
==38468== 256 bytes in 1 blocks are definitely lost in loss record 20 of 20
==38468==    at 0x4846828: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==38468==    by 0x116D69: OQS_MEM_aligned_alloc (in /__w/liboqs/liboqs/build/tests/test_kem)
==38468==    by 0x58EF5B: SHA3_sha3_256_inc_init (in /__w/liboqs/liboqs/build/tests/test_kem)
==38468==    by 0x270CD6: PQCHQC_HQC5_C_hash_j (in /__w/liboqs/liboqs/build/tests/test_kem)
==38468==    by 0x27059D: PQCHQC_HQC5_C_crypto_kem_dec (in /__w/liboqs/liboqs/build/tests/test_kem)
==38468==    by 0x115450: kem_test_correctness (in /__w/liboqs/liboqs/build/tests/test_kem)
==38468==    by 0x115A31: test_wrapper (in /__w/liboqs/liboqs/build/tests/test_kem)
==38468==    by 0x4E10AA3: start_thread (pthread_create.c:447)
==38468==    by 0x4E9DA63: clone (clone.S:100)
```

Proposed fix:
- In `hash_i`, `hash_h`, `hash_g`, `hash_j`, call `ctx_release` before the function exits
- In `xof_xxx`, add an API `xof_ctx_release` that calls the underlying `ctx_release`, then call `xof_ctx_release` for each `xof_init`
- `prng` has been deprecated in favor of OQS's own `randombytes` API, so they should be safely removed

edit: after fix [address sanitizers passed](https://github.com/open-quantum-safe/liboqs/actions/runs/24476220243/job/71530161908?pr=2407)

dstebila:
Analysis by Claude about Windows failure:

Root cause is in the BITMASK macro.

  The bug

  src/kem/hqc/xuganyu96-hqc_hqc-*_ref/parameters.h:12:
  #define BITMASK(a, size)  ((1UL << (a % size)) - 1)

  On Windows x86_64 (LLP64) unsigned long is 32 bits. On Linux/macOS x86_64 (LP64) it's 64 bits. So 1UL << 37 is fine everywhere except Windows, where it's undefined behavior (shift ≥ operand width) and MSVC typically emits an x86 shl that masks the
  count to 5 bits, giving 1<<5 = 32 → BITMASK = 31 instead of 0x1FFFFFFFFF.

  Why only HQC-5:

  ┌─────────┬─────────┬──────────────┬─────────────┐
  │ Variant │ PARAM_N │ PARAM_N % 64 │ Windows OK? │
  ├─────────┼─────────┼──────────────┼─────────────┤
  │ HQC-1   │ 17669   │ 5            │ yes         │
  ├─────────┼─────────┼──────────────┼─────────────┤
  │ HQC-3   │ 35851   │ 11           │ yes         │
  ├─────────┼─────────┼──────────────┼─────────────┤
  │ HQC-5   │ 57637   │ 37           │ no          │
  └─────────┴─────────┴──────────────┴─────────────┘

  The mask is applied at:
  - gf2x.c:139 inside reduce() — called from vect_mul(), which runs in keygen/encaps
  - vector.c:205 in vect_set_random() — generates h from the seed

  So every vect_mul result and every sampled h has its top 37 bits zeroed on Windows, producing corrupted keys/ciphertexts — exactly what your debug fprintf of s, y, h in hqc_pke_keygen should expose (look at s[VEC_N_SIZE_64 - 1] and h[VEC_N_SIZE_64 -
  1]).

  Why "when SLH-DSA is disabled" is a red herring

  .github/workflows/windows.yml:19,52 unconditionally passes -DOQS_ENABLE_SIG_SLH_DSA=OFF. That's just the Windows CI config — not causal. The bug fires on Windows regardless of SLH-DSA.

  Fix

  Change 1UL → 1ULL (or UINT64_C(1)) in all three parameters.h files. Also consider (a) % (size) parens, but the UL→ULL is the actual fix.

xuganyu96:
Narrowing down the source of [KAT failures on Windows](https://github.com/open-quantum-safe/liboqs/actions/runs/24524483691/job/71690877251).

- seed is identical
- public keys differ by a few coefficients. I added the following debugging statements:

```c
void hqc_pke_keygen(uint8_t *ek_pke, uint8_t *dk_pke, uint8_t *seed) {
    uint8_t keypair_seed[2 * SEED_BYTES] = {0};
    uint8_t *seed_dk = keypair_seed;
    uint8_t *seed_ek = keypair_seed + SEED_BYTES;
    shake256_xof_ctx dk_xof_ctx = {0};
    shake256_xof_ctx ek_xof_ctx = {0};

    uint64_t x[VEC_N_SIZE_64] = {0};
    uint64_t y[VEC_N_SIZE_64] = {0};
    uint64_t h[VEC_N_SIZE_64] = {0};
    uint64_t s[VEC_N_SIZE_64] = {0};

    // Derive keypair seeds
    hash_i(keypair_seed, seed);

    // Compute decryption key
    xof_init(&dk_xof_ctx, seed_dk, SEED_BYTES);
    vect_sample_fixed_weight1(&dk_xof_ctx, y, PARAM_OMEGA);
    vect_sample_fixed_weight1(&dk_xof_ctx, x, PARAM_OMEGA);
    xof_ctx_release(&dk_xof_ctx);

    // Compute encryption key
    xof_init(&ek_xof_ctx, seed_ek, SEED_BYTES);
    vect_set_random(&ek_xof_ctx, h);
    xof_ctx_release(&ek_xof_ctx);
    vect_mul(s, y, h);
#if DEBUG_KEYGEN
    fprintf(stderr, "var,val\n");
    char varname[64] = {0};
    for (int i = 0; i < VEC_N_SIZE_64; i++) {
        snprintf(varname, sizeof(varname), "s[%d]", i);
        fprintf(stderr, "%s,%016llX\n", varname, s[i]);
        snprintf(varname, sizeof(varname), "y[%d]", i);
        fprintf(stderr, "%s,%016llX\n", varname, y[i]);
        snprintf(varname, sizeof(varname), "h[%d]", i);
        fprintf(stderr, "%s,%016llX\n", varname, h[i]);
    }
#endif
    vect_add(s, x, s, VEC_N_SIZE_64);
#if DEBUG_KEYGEN
    for (int i = 0; i < VEC_N_SIZE_64; i++) {
        snprintf(varname, sizeof(varname), "s[%d]", i);
        fprintf(stderr, "%s,%016llX\n", varname, s[i]);
        snprintf(varname, sizeof(varname), "x[%d]", i);
        fprintf(stderr, "%s,%016llX\n", varname, x[i]);
    }
#endif

    // Parse encryption key to string
    memcpy(ek_pke, seed_ek, SEED_BYTES);
    memcpy(ek_pke + SEED_BYTES, s, VEC_N_SIZE_BYTES);

    // Parse decryption key to string
    memcpy(dk_pke, seed_dk, SEED_BYTES);

#ifdef VERBOSE
    printf("\n\nseed_dk: ");
    for (int i = 0; i < SEED_BYTES; ++i) printf("%02x", seed_dk[i]);
    printf("\n\nseed_ek: ");
    for (int i = 0; i < SEED_BYTES; ++i) printf("%02x", seed_ek[i]);
    printf("\n\ny: ");
    vect_print(y, VEC_N_SIZE_BYTES);
    printf("\n\nx: ");
    vect_print(x, VEC_N_SIZE_BYTES);
    printf("\n\nh: ");
    vect_print(h, VEC_N_SIZE_BYTES);
    printf("\n\ns: ");
    vect_print(s, VEC_N_SIZE_BYTES);
#endif

    // Zeroize sensitive data
    memset_zero(keypair_seed, sizeof keypair_seed);
    memset_zero(x, sizeof x);
    memset_zero(y, sizeof y);
    memset_zero(&dk_xof_ctx, sizeof dk_xof_ctx);
}
```

Coefficients:
[expected.log](https://github.com/user-attachments/files/26796388/expected.log)
[windows.log](https://github.com/user-attachments/files/26796393/windows.log)

xuganyu96:
> Analysis by Claude about Windows failure:
> 
> Root cause is in the BITMASK macro.
> 
> The bug
> 
> src/kem/hqc/xuganyu96-hqc_hqc-*_ref/parameters.h:12: #define BITMASK(a, size) ((1UL << (a % size)) - 1)
> 
> On Windows x86_64 (LLP64) unsigned long is 32 bits. On Linux/macOS x86_64 (LP64) it's 64 bits. So 1UL << 37 is fine everywhere except Windows, where it's undefined behavior (shift ≥ operand width) and MSVC typically emits an x86 shl that masks the count to 5 bits, giving 1<<5 = 32 → BITMASK = 31 instead of 0x1FFFFFFFFF.
> 
> Why only HQC-5:
> 
> ┌─────────┬─────────┬──────────────┬─────────────┐ │ Variant │ PARAM_N │ PARAM_N % 64 │ Windows OK? │ ├─────────┼─────────┼──────────────┼─────────────┤ │ HQC-1 │ 17669 │ 5 │ yes │ ├─────────┼─────────┼──────────────┼─────────────┤ │ HQC-3 │ 35851 │ 11 │ yes │ ├─────────┼─────────┼──────────────┼─────────────┤ │ HQC-5 │ 57637 │ 37 │ no │ └─────────┴─────────┴──────────────┴─────────────┘
> 
> The mask is applied at:
> 
>     * gf2x.c:139 inside reduce() — called from vect_mul(), which runs in keygen/encaps
> 
>     * vector.c:205 in vect_set_random() — generates h from the seed
> 
> 
> So every vect_mul result and every sampled h has its top 37 bits zeroed on Windows, producing corrupted keys/ciphertexts — exactly what your debug fprintf of s, y, h in hqc_pke_keygen should expose (look at s[VEC_N_SIZE_64 - 1] and h[VEC_N_SIZE_64 - 1]).
> 
> Why "when SLH-DSA is disabled" is a red herring
> 
> .github/workflows/windows.yml:19,52 unconditionally passes -DOQS_ENABLE_SIG_SLH_DSA=OFF. That's just the Windows CI config — not causal. The bug fires on Windows regardless of SLH-DSA.
> 
> Fix
> 
> Change 1UL → 1ULL (or UINT64_C(1)) in all three parameters.h files. Also consider (a) % (size) parens, but the UL→ULL is the actual fix.

Annoyingly, applying Claude's suggestion resolved the failures. 😢

xuganyu96:
Resolving [this CI pipeline failure](https://github.com/open-quantum-safe/liboqs/actions/runs/24533614894/job/71724235277?pr=2407#logs). This is a KAT checksum inconsistency. Based on the output:

- `pk`, `sk` are identical
- `ct` is inconsistent with expected output, but not entirely

RodriM11:
Hi @xuganyu96 ! Thank you for the immense amount of work in order to see this PR through, I greatly appreciate it. I have a general question:

The source implementation maintains a certain amount of "common" code (e.g. `code.c`, `gf.c` and so on), while this PR includes each file (at first sight, being the same) on each of the parametrizations. Is there any reason why this was done with the files appearing on each folder?

xuganyu96:
> Is there any reason why this was done with the files appearing on each folder?

The short reason is simplicity of configuration files and ease of development.

I am aware that only recently I have pushed the [MQOM contributor](https://github.com/open-quantum-safe/liboqs/pull/2361#pullrequestreview-3916184829) to de-duplicate files, but I am currently prioritizing pushing HQC across the finish line so we can release 0.16. I am happy to come back to this.

RodriM11:
> I am aware that only recently I have pushed the https://github.com/open-quantum-safe/liboqs/pull/2361#pullrequestreview-3916184829 to de-duplicate files, but I am currently prioritizing pushing HQC across the finish line so we can release 0.16. I am happy to come back to this.

That decision seems sensible to me, seeing as we are gearing towards a release soon and `HQC` is, IMO, one of the top priorities. So the PR LGTM if we ensure we handle de-duplicating the code once `0.16.0` lands. 

Thanks for the work and for seeing this through @xuganyu96 !

xuganyu96:
It would be ideal to reach out to the HQC team for an "official" liboqs integration effort.

In case that is not feasible, I think it makes sense to move [the fork of official repository](https://github.com/xuganyu96/pqc-hqc/tree/liboqs-integration) under `open-quantum-safe` organization to facilitate future contributions.

baentsch:
> It would be ideal to reach out to the HQC team for an "official" liboqs integration effort.
> 
> In case that is not feasible, I think it makes sense to move [the fork of official repository](https://github.com/xuganyu96/pqc-hqc/tree/liboqs-integration) under `open-quantum-safe` organization to facilitate future contributions.

In my opinion, it would not only be ideal, but necessary -- unless of course you now consider yourself an expert in the HQC code @xuganyu96 : Do you? The discussion above (relying on Claude "knowledge/advice") doesn't quite create that impression, but I may be jumping to conclusions here.

Finally, moving a fork into the oqs namespace in my opinion creates the impression and expectation that OQS becomes/is the final "source of truth" and maintainer of last resort of an algorithm. I would think we should do that only if we have absolute certainty we understand the code and logic inside-out and can maintain it going forward under all circumstances (even if single members leave the project). I'm not sure this applies here and suggest having a TSC discussion about this.

RodriM11 (dismissed):

src/common/common.c
```diff
@@ -450,3 +450,82 @@ OQS_API char *OQS_MEM_strdup(const char *str) {
...
+			line_len += 2;
+			if (offset + 1 < buflen &&
+			        line_len + 1 <= DEBUGLOGGING_LINE_WIDTH) {
+				fprintf(f, " ");
+				line_len += 1;
+			}
+		} else {
+			fprintf(f, "\n");
+			line_len = 0;
+		}
```
dstebila: When this branch is taken, `offset` is still incremented, but `buf[offset]` is never printed, so I think that character will get lost.

dstebila (changes requested):
PR looks good to me, except for one small quirk in the debug print code which I've noted. Claude's review identified a potential fix for the intermittent hash failures on AVX512, which I've put into a [separate PR on top of this PR](https://github.com/open-quantum-safe/liboqs/pull/2426) which can be considered separately, or discarded in favour of the fix already here.

src/common/common.c
xuganyu96: See [this commit msg](https://github.com/open-quantum-safe/liboqs/commit/23db66392db0276a36d9b37ce8fb9b7468171463)

xuganyu96:
> PR looks good to me, except for one small quirk in the debug print code which I've noted. Claude's review identified a potential fix for the intermittent hash failures on AVX512, which I've put into a [separate PR on top of this PR](https://github.com/open-quantum-safe/liboqs/pull/2426) which can be considered separately, or discarded in favour of the fix already here.

Per May 19's team discussion, SHA3-512 AVX512VL fix will ship in a separate pull request #2442. I cherry-picked the commits from that pull request into this branch so CI can validate it.

dstebila (dismissed):

dstebila (approved):

RodriM11 (approved):

RodriM11:
Thanks for the work in seeing this through, @xuganyu96 !

PR #2412, opened 2026-04-21
tests: add fuzz harness for LMS stateful signature verification (#2398)

### LINKED ISSUE #2398: LMS Signature Verification Fuzzing
### Description

To implement fuzz testing for the stateful hash-based signature (HBS) verification algorithms in liboqs for specific parameter sets for LMS. Testing the entire set of variants would be prohibitive. The suggestion is to test LMS_SHA256_H5_W2 and LMS_SHA256_H5_W8_H5_W8.

	•	Target Function: The fuzzer must target the OQS_SIG_STFL_verify function (or the specific internal LMS verification routine) within liboqs.
	•	Algorithm Scope: The testing must specifically include and be limited to the following parameter sets:
		•	LMS_SHA256_H5_W2
		•	LMS_SHA256_H5_W8_H5_W8
	•	Input Handling: The fuzzer should provide mutated data for the signature, message, and public_key inputs.

_No response_

_No response_

### Use of generative AI

_No response_

Vishnu2707:
Closes #2398

## What this PR does

Adds `fuzz_test_sig_stfl_lms.c` — a libFuzzer harness targeting
`OQS_SIG_STFL_verify` for the two parameter sets specified in the issue:
- `LMS_SHA256_H5_W2`
- `LMS_SHA256_H5_W8_H5_W8`

## Approach

- Keypairs loaded from KAT vectors in `tests/KATs/sig_stfl/lms/` to
  avoid expensive LMS key generation overhead per iteration
- Each iteration mutates the signature, message, and public_key inputs
  fed to `OQS_SIG_STFL_verify`
- CMakeLists.txt updated to register the new target under the existing
  `OQS_BUILD_FUZZ_TESTS` condition

## Local verification

Built and run locally on Apple M-series (arm64, macOS 15.3.1) using
LLVM clang 22 with `-fsanitize=fuzzer,address`:

Done 30,429,177 runs in 16 second(s)

No crashes or memory errors detected.

## Checklist
- [ ] Does this PR change the input/output behaviour of a cryptographic algorithm? No
- [ ] Does this PR change the list of algorithms available? No


ashman-p:
@Vishnu2707, thanks again for the contribution.

ashman-p (dismissed):

tests/fuzz_test_sig_stfl_lms.c
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	/* Start from KAT values, then overwrite with fuzz bytes */
+	memcpy(mutated_pk, kp->public_key, pk_len);
+	memcpy(mutated_sig, kp->signature, sig_len);
+	memcpy(mutated_msg, kp->message, msg_len);
+
+	size_t offset = 0;
+
+	if (offset < fuzz_len) {
```
baentsch: A question for clarification, @Vishnu2707 : Why do you chose to always fuzz the PK first? This implies that sigs or msgs hardly ever get fuzzed assuming fuzz data is not _pretty_ long (which I doubt it is -- do you have insight on this @ashman-p ?). So wouldn't it be much better for sensible fuzz coverage to also select which array to "fuzz-overwrite" first based on fuzz input (along the lines of algorithm selection), thus ensuring msgs or sigs regularly get "fuzzed first"?

Vishnu2707:
Good catch @baentsch. With libFuzzer starting from a minimal corpus, the sequential fill means early iterations almost exclusively hit the public key, sig and msg won't see meaningful mutation until the corpus grows large enough to exceed the pk size. A selector byte using modulo 3 (similar to the algorithm selector) would distribute mutation across all three fields from the first iteration. I'll implement this.

Vishnu2707:
Implemented - added a selector byte (modulo 3) to randomise which field is mutated first each iteration. pk, sig, and msg now each get equal access to fuzz bytes from the first iteration regardless of corpus size. 30.4M iterations locally, no crashes. Pushed.

tests/fuzz_test_sig_stfl_lms.c
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	/* Start from KAT values, then overwrite with fuzz bytes in selected order */
+	memcpy(mutated_pk, kp->public_key, pk_len);
+	memcpy(mutated_sig, kp->signature, sig_len);
+	memcpy(mutated_msg, kp->message, msg_len);
+
+	/*
+	 * field_selector % 3 controls which field is overwritten first:
+	 *   0 → pk, sig, msg   1 → sig, pk, msg   2 → msg, pk, sig
```
baentsch: Thanks for this improvement @Vishnu2707 ! Not sure if I'm splitting hairs now, but this logic (for "1") still on average seems to "favour" pk fuzzing and "disadvantages" msg fuzzing. I'd have expected 0:pk,sig,msg; 1:sig,msg,pk; 2: msg,pk,sig

Vishnu2707:
Apologies @baentsch  — I missed that the ordering wasn't a true rotation, I was thinking in terms of "sig gets a turn first" rather than ensuring each field gets equal position across all three cases. Fixed to a proper round-robin: 0: pk→sig→msg, 1: sig→msg→pk, 2: msg→pk→sig. Each field now occupies each position exactly once. 33M iterations locally, no crashes. Pushed. 

Thanks for catching this.

ashman-p:
@Vishnu2707 Thanks for adding this. It performs the requested task and more, in that each input is fuzzed.

ashman-p (approved):

baentsch (approved):
Thanks very much @Vishnu2707 for the PR and the latest improvement! Good resolution for #2398 and good contribution.

Issue #2064, opened 2025-02-04
Compile a list of upstream contacts


SWilson4:
For handling of security issues, it would be good to have contacts for each upstream algorithm source. These contacts could potentially assist with, among other things, assessment of bug reports and coordination of security releases.

In addition to the official point of contact for each upstream, it would be good to include any team members with whom OQS already has a working relationship.

dstebila:
Being done as part of https://github.com/open-quantum-safe/liboqs/pull/2413

Issue #2341, opened 2026-01-08
Discuss impact of PQClean wind-down


dstebila:
As per https://github.com/PQClean/PQClean/issues/604#issuecomment-3724342593, the PQClean project will be winding down over the next 6 months. We rely on PQClean's implementations for:

- Classic McEliece
- HQC: soon to be replaced with another implementation #2328 
- Kyber: only for aarch64
- Falcon: eventually to be replaced with FN-DSA
- SPHINCS+: soon to be removed #2339

baentsch:
Thanks for the summary - so it seems OQS may want to check with users whether keeping McEliece is desired. Keeping an architecture-specific Kyber doesn't look necessary as long as a reference implementation is still around -- _if_ there is a need to keep Kyber (is there such a demand?). And the rest will be gone soon.

Concrete proposal: Announce upcoming removal of McEliece and Kyber support in the next `liboqs` release to see whether anyone complains.

dstebila:
> Thanks for the summary - so it seems OQS may want to check with users whether keeping McEliece is desired. Keeping an architecture-specific Kyber doesn't look necessary as long as a reference implementation is still around -- _if_ there is a need to keep Kyber (is there such a demand?). And the rest will be gone soon.
> 
> Concrete proposal: Announce upcoming removal of McEliece and Kyber support in the next `liboqs` release to see whether anyone complains.

I've revived the discussion in #1989 about Kyber support to see where things stand on usage of Kyber.

Regarding Classic McEliece, as far as I know it is still on track for ISO standardization, and so would be in scope for OQS to continue to include.

baentsch:
> Regarding Classic McEliece, as far as I know it is still on track for ISO standardization, and so would be in scope for OQS to continue to include.

In that case, does that mean that OQS will effectively (have to) become maintainer of this code base (resolve security issues, follow required changes during the standardization process, etc.) when the upstream's gone? Allow me to highlight again #2064 to capture such dependencies/obligations on OQS and/or upstreams (where available).

thomwiggers:
It's probably possible to vendor https://lib.mceliece.org/, but you'd need to do some work to adapt it to the OQS build system and so on.

Guiliano99:
Hello,

[...]

PR #2413, opened 2026-04-21
Create a table of algorithm support levels

### LINKED ISSUE #2045: Create a table of algorithm support levels

dstebila:
The goal is for each algorithm family to have clear information about the level of support provided by the OQS team and by the upstream maintainers of the original implementation.

Main approach:
- Extend the summary table in the top-level README.md with two columns for these two characteristics
- Add a top-level ALGORITHMS.md file that documents the definitions of levels of support along these two axes and includes more detailed family-level tables than in README.md, though not as detailed as the algorithm datasheets
- Update the algorithm datasheet YAML files to contain this information, which is then used to populate the above two tables with templates; adds a CI test confirming synchronization between YAML source files and generated files

The level of upstream maintenance will have to be confirmed by the upstream team, and as per the description in ALGORITHMS.md is an input factor into defining an OQS tier.  I have pre-populated the OQS tier based on my own opinions and what I expect the answers will be for some of the upstream maintenance levels, but that could change as responses do/don't come in. Tags Classic McEliece as "No active maintenance" based on PQClean wind-down (#2341).

We'll need to do an outreach to each upstream team to confirm.  Landing of this PR would be accompanied by creating a tracking issue for that outreach; I'll post template for that in a comment to this PR.

The PR is split into two commits: source-of-truth files (scripts, YAML, workflow, tests, hand-maintained text) in commit 1, and the regenerated output (README table fragment, ALGORITHMS.md tables, per-algorithm pages) in commit 2.

Fixes #2045.

**Generative AI usage**: Used for generation of code and some documentation. I've reviewed and edited.


dstebila:
Issue to be created for tracking outreach to upstream teams:

Parent tracking issue for per-upstream outreach driven by #2045. Each upstream confirms which maintenance label applies to their project. Responses are recorded in the corresponding `docs/algorithms/*.yml` `upstream-maintenance` / `upstream-statement-url` fields. Closes #2064 when all upstreams have responded or been marked `No active maintenance`.

## Outreach message (copy into each child issue)

> Hello <upstream team>,
>
> The [Open Quantum Safe project](https://github.com/open-quantum-safe/liboqs) integrates your implementation of <algorithm> into `liboqs`. As part of [issue #2045](https://github.com/open-quantum-safe/liboqs/issues/2045), we are publishing a user-facing support-level table that records, per algorithm, what level of maintenance the upstream implementation team has committed to. So that users know what to expect, we would like you to pick the label that best describes your project's current state.
>
> **Options** (full definitions in [ALGORITHMS.md](https://github.com/open-quantum-safe/liboqs/blob/main/ALGORITHMS.md#upstream-maintenance)):
>
> - **Actively maintained** — named responsive contact, commitment to investigating security reports and releasing fixes, and a public written statement we can link to.
> - **Best effort** — named contact, attempts to respond when able, no timing commitment.
> - **No active maintenance** — no responsive contact, or a publicly stated wind-down.
>
> Could you:
>
> 1. Tell us which label you would like applied?
> 2. Ideally, point us to a permalink on your repository (README, SUPPORT.md, or similar) containing a short written statement describing this commitment, so we can cite it?
>
> Thanks,
>
> Douglas Stebila (on behalf of OQS)

## Upstreams to contact

- [x] `pq-code-package/mlkem-native` — ML-KEM
- [x] `pq-code-package/mldsa-native` — ML-DSA
- [x] `pq-code-package/slhdsa-c` — SLH-DSA
- [x] `awslabs/bike-kem` — BIKE
- [x] `microsoft/PQCrypto-LWEKE` — FrodoKEM
- [x] `pq-crystals/kyber` — Kyber
- [x] NTRU upstream — NTRU
- [x] `openssh/openssh-portable` (sntrup761) — NTRU-Prime
- [x] `CROSS-signature/CROSS-lib-oqs` — CROSS
- [x] `PQCMayo/MAYO-C` — MAYO
- [x] `mqom/mqom-v2` — MQOM
- [x] `vacuas/SNOVA` — SNOVA
- [x] `pqov/pqov` — UOV
- [x] `cisco/hash-sigs` — LMS
- [x] `XMSS/xmss-reference` — XMSS
- [x] ~~`PQClean/PQClean`~~ — Classic McEliece, Falcon, HQC: already cited as `No active maintenance` via PQClean's public wind-down.

RodriM11:
thanks for the work @dstebila ! I like the idea of classification of algorithms in tiers and think that, if we are going to base future support / prioritization decisions on it, we should establish some objective (as much as possible) criteria for classifying them, so we can justify future decisions to anyone that may ask. A simple classification that comes to mind would be:
- Tier 3 for algorithms discarded / rejected.
- Tier 2 for algorithms under consideration for standardization (by any standardization body)
- Tier 1 for standards.

This classification would change the following tiers:

- Classic McEliece: 3 -> 2
- NTRU: 3 -> 1
- CROSS: 3 -> 2
- Falcon: 3 -> 2
- MAYO: 3 -> 2
- ML-DSA: 2 -> 1
- SLH-DSA: 3 -> 1
- MQOM: 3 -> 2
- UOV: 3 -> 2
- SNOVA: 3 -> 2
- LMS: 2 -> 1
- XMSS: 2 -> 1

But any other proposal for classification would be very much appreciated.

dstebila:
> thanks for the work @dstebila ! I like the idea of classification of algorithms in tiers and think that, if we are going to base future support / prioritization decisions on it, we should establish some objective (as much as possible) criteria for classifying them, so we can justify future decisions to anyone that may ask. A simple classification that comes to mind would be:
> 
>  * Tier 3 for algorithms discarded / rejected.
>  * Tier 2 for algorithms under consideration for standardization (by any standardization body)
>  * Tier 1 for standards.

I see what you're going for, but I think this information is already conveyed in the "standardization status" column of the table. The tier list was meant to convey a different piece of information, which is what our level of support is for that algorithm. The tier list is also meant to reflect the *current* level of support, not an aspirational level of support – ideally we would we would get all standards to tier 1, but I would say that our ML-KEM support is substantially different than our other algorithms, at present.

RodriM11:
> I see what you're going for, but I think this information is already conveyed in the "standardization status" column of the table. The tier list was meant to convey a different piece of information, which is what our level of support is for that algorithm

OK, so it means the level of support we expect to receive from the code-owners from each algorithm, correct?

dstebila:
> > I see what you're going for, but I think this information is already conveyed in the "standardization status" column of the table. The tier list was meant to convey a different piece of information, which is what our level of support is for that algorithm
> 
> OK, so it means the level of support we expect to receive from the code-owners from each algorithm, correct?

The "upstream maintainers" column reflects what support the upstream maintainers of the implementation have indicated they think they can provide. The "OQS tier" column reflects what we (OQS) think we can provide, based on the support by upstream maintainers and our own team.

RodriM11:
Ok, understood. Thanks @dstebila !

ALGORITHMS.md
```diff
@@ -0,0 +1,61 @@
...
+
+- **Actively maintained.** There is a named, responsive upstream contact or team. Upstream commits to investigating security reports in a timely manner and incorporating fixes. Upstream releases updates that OQS can track. Upstream has a public written statement describing this commitment, which OQS links to.
+- **Best effort.** There is a named upstream contact who acknowledges OQS's integration. Upstream will attempt to respond to security reports and incorporate fixes when able, with no commitment on timing or release cadence. Users should not depend on timely upstream fixes.
+- **No active maintenance.** There is no responsive upstream contact, or upstream has publicly stated wind-down or end-of-life. Fixes will not necessarily be incorporated from upstream. Users considering this algorithm should plan accordingly (alternatives, migration, or maintaining their own fork).
+- **TBD.** OQS has not yet obtained a confirmed statement from the upstream. This is not a judgment about the upstream; it means the conversation has not happened or not concluded.
```
baentsch: Is it wise to use one label for two things that are fundamentally different (for users): 1) No conversation has happened and 2) A conversation has not concluded? I'd thus strongly suggest splitting "TBD" into at least "under discussion" and "not discussed". And honestly, the latter is a failure on our end that we ought to fix: Another TSC discussion, please: Someone should reach out to all upstream sources to get a discussion going. I think we should not publish this list with any designation other than one of the first three.

ALGORITHMS.md
```diff
@@ -0,0 +1,61 @@
...
+
+## OQS support tier
+
+The **OQS tier** column records OQS's own support commitment for each algorithm. Tiers are self-assigned by the OQS technical governance team via PR review, and are structurally parallel to the platform tiers defined in [PLATFORMS.md](PLATFORMS.md). Promotion or demotion between tiers is proposed via PR and promotion requires the algorithm to have spent reasonable time at its current tier. Tier assignments are not a recommendation for production use of any algorithm in `liboqs`; see the project-wide disclaimer in [README.md](README.md).
+
+- **Tier 1 — Core.** OQS's highest level of commitment and scrutiny. Standardized by a recognized body (NIST, ISO, IETF); upstream marked `Actively maintained` with a public support statement; covered by the OQS security response process; constant-time tested on all Tier 1 platforms where applicable; built and tested on every Tier 1 platform in CI, every parameter set; at least two OQS committers familiar with the code; has an identified entry in `CODEOWNERS`.
```
baentsch: Can I ask for a clear definition of "familiarity" in this text? Does it mean true understanding in the sense that any bug can be found and fixed, i.e., code and algorithm knowledge at authorship level or is it more like "we once looked at it and think we understand it"?

Also, I think this tier(ing) mixes two things here that shouldn't be mixed: One is standardization state and one is OQS (team) capability. Personally, a true-to-the-word "OQS support tier" should only document the latter; the former should be irrelevant here, ie. be covered in a separate column, namely "standardization level". Personally, I'd actually love to have more Tier 1 algorithms that are _not_ standardized (yet): This way, the OQS team would know and understand those new algs and help them get adoption and possibly, standardization: That is a mission OQS can fulfill; collecting already standardized algs is something other projects (and companies) already do: At the risk of repeating myself: We cannot compete with those entities and should not waste our energies on such futile things.

Then I again have to object to using the term "constant-time tested" here for two reasons: 
1) If this is "Tier 1" than what is code that has explicitly been _developed_ to be CT-safe? Tier 0? 2) From a user's perspective, reading OQS considers some code in the highest tier, explicitly stating "CT tested", wouldn't the user rightfully assume the code _to be_ CT? I do not think we can give this guarantee, so should also avoid creating that impression. I'd thus suggest completely dropping CT from this list; maybe we could say "receives the currently best level of functional and non-functional testing on all Tier-1 platforms".

ALGORITHMS.md
dstebila: I don't think there's a need to split up TBD.

ALGORITHMS.md
dstebila: I can see a point to separating standardization from support tier.

I don't understand what you're getting at with respect to constant-time. We do constant-time testing in CI via our test harness. We have suppression files that track constant-time issues, and algorithms with such issues do not get a CT checkmark in our Markdown files nor in this summary. Why wouldn't we include that in our definition of tier 1?

ALGORITHMS.md
baentsch: Hence my argumentation --which obviously didn't get my point across-- and my consequent wish to discuss in the TSC.

ALGORITHMS.md
baentsch: CT: A Tier 1 OQS category in that regard for me would mean it's code that OQS _ensured_ has been developed to be CT; testing for that property is a good further step, but it's not as good as knowing that (and how) that property has been designed into the code (e.g., https://research.redhat.com/blog/article/the-need-for-constant-time-cryptography/ for "lackluster" CT testing ; for C code, OQS could ensure use of things like https://github.com/openssl/openssl/blob/master/include/internal/constant_time.h by code claiming "Tier 1" -- and passing the tests, of course).

Any comment on a definition of "familiarity" also welcome.

dstebila: > CT: A Tier 1 OQS category in that regard for me would mean it's code that OQS _ensured_ has been developed to be CT; testing for that property is a good further step, but it's not as good as knowing that (and how) that property has been designed into the code (e.g., https://research.redhat.com/blog/article/the-need-for-constant-time-cryptography/ for "lackluster" CT testing ; for C code, OQS could ensure use of things like https://github.com/openssl/openssl/blob/master/include/internal/constant_time.h by code claiming "Tier 1" -- and passing the tests, of course).

While that is potential goal one could have, it is not a goal that we have ever had. If you want to revise all our current constant-time documentation in light of that, that's a separate issue.

baentsch: I completely agree: OQS never had that ambition/capacity. What I'm trying to state here is that users are not going to understand this, though, if we list "CT assurance" in this "Tier 1" definition: They'll think "Explicitly stated CT assurance stated in Tier 1 is as good as what other FOSS crypto libs assert in this space". Hence my suggestion to drop that mention and state only ["receives the currently best level of functional and non-functional testing on all Tier-1 platforms"](https://github.com/open-quantum-safe/liboqs/pull/2413#discussion_r3142528297), maybe even tuned down to read "...currently best possible level given time and resources in the OQS team...".

dstebila: We have many instances of constant-time related language throughout the project documentation, and this simply summarizes that. This text says that Tier 1 algorithms are "constant-time tested". Is there anything false about that statement?

There is a disclaimer about the overall level of assurance that the project offers in the sentence immediately before this: "Tier assignments are not a recommendation for production use of any algorithm in `liboqs`; see the project-wide disclaimer in README.md.".  I guess we can put more such disclaimers into the Tier 1 text explicitly, but we can't have every other sentence be a disclaimer.

baentsch: > This text says that Tier 1 algorithms are "constant-time tested". Is there anything false about that statement?

No, technically this is correct from a writer's perspective. I do take into consideration what (not necessarily technically adept) readers may understand by this statement, though, namely a positive assurance that the code for sure possesses that property and thus allows such reader (doing a check-list crossing exercise only) to use the code productively. And that assurance is one that should not be inferred.

> I guess we can put more such disclaimers into the Tier 1 text explicitly, but we can't have every other sentence be a disclaimer.

Agree to both statements. Something along the lines of "Tier 1 represents the highest level of assurances this project can provide within its limited resources; all statements below should be taken considerate of our recommendation to not use any code within `liboqs` for production purposes; see the project-wide disclaimer in README.md.". If OK, I'm fine leaving the technical statement regarding CT.

dstebila: Addressed in [8f2a2f5](https://github.com/open-quantum-safe/liboqs/pull/2413/commits/8f2a2f502e8a4e61a1b4d2ddaddb94c0736dc66a), please take a look.

RodriM11:
All the upstreams have now been contacted. With these, we can address @baentsch (rightful) comment that all upstream teams to be asked before merging this PR. Subsequent PRs to include the responses for each of the teams will be generated once they answer.

dstebila:
> All the upstreams have now been contacted. With these, we can address @baentsch (rightful) comment that all upstream teams to be asked before merging this PR. Subsequent PRs to include the responses for each of the teams will be generated once they answer.

Thanks for sending all the emails, Rodrigo!

ALGORITHMS.md
dstebila: Marking as resolved based on https://github.com/open-quantum-safe/liboqs/pull/2413#issuecomment-4435089528

baentsch (dismissed):
Thanks @dstebila @RodriM11 for moving this forward allowing users to get a good picture where we (and they) are at with each algorithm!

ALGORITHMS.md
RodriM11: Do we still maintain the "has not happened or not concluded" or do we shift it to "has not concluded"?

baentsch: I'd obviously vote for the latter.

dstebila: Fixed in [68e4d23](https://github.com/open-quantum-safe/liboqs/pull/2413/commits/68e4d235d4d8bdf10c3ba01d8ebaf695dbed8303)

Issue #2229, opened 2025-08-07
LMS Issues with Proposed Fixes


zwang610:
I have been testing the LMS_SHA256_H5_W4, LMS_SHA256_H10_W4, and LMS_SHA256_H20_W4 implementations in liboqs and identified six issues. I’ve temporarily resolved each on my local machine and plan to submit PRs. 

Please share your feedback on the proposed fixes and address the questions. Thank you!

###  1. Missing Initialization in `oqs_deserialize_lms_key`
**Symptom:** When calling `oqs_deserialize_lms_key(OQS_SIG_STFL_SECRET_KEY *sk, const uint8_t *sk_buf, const size_t sk_len, void *context)` with no auxiliary data in `sk_buf`/`sk_len`, the fields `sk->secret_key_data->len_aux_data` and `sk->secret_key_data->aux_data` remain uninitialized. A subsequent call to `oqs_secret_lms_key_free(sk)` can trigger a segmentation fault.

**Proposed fix:** Explicitly initialize `len_aux_data = 0` and `aux_data = NULL` in `oqs_deserialize_lms_key` if no aux data is present.

###  2. Disabled LMS Variants with `DOQS_ENABLE_SIG_STFL_LMS=ON`
**Symptom:** Even with `DOQS_ENABLE_SIG_STFL_LMS=ON`, many LMS parameter sets are disabled. They are not listed in `src/oqsconfig.h.cmake`, matching user‐reported issue #2121.

**Questions:** Are those variants ready to release? Should we add them to the CMake config?

**Local workaround:** I added LMS_SHA256_H20_W4 to `src/oqsconfig.h.cmake` before building. Functionality is correct, but performance compared to the reference [hash-sigs](https://github.com/cisco/hash-sigs) is **unreasonably slow**.

**GenKey**
- liboqs (single-threaded default): 12 min
- liboqs (16 threads patched): 8 min
- hash-sigs (single-threaded): 5 min
- hash-sigs (16 threads): 20 sec

**Sign**
- liboqs (single-threaded default): 12 min
- liboqs (16 threads patched): 9 min
- hash-sigs (single-threaded, no aux): 5 min
- hash-sigs (16 threads, no aux data): 20 sec
- hash-sigs (single-threaded, with aux): <1 sec
- hash-sigs (16 threads, with aux data): <1 sec

[...]

PR #2416, opened 2026-05-03
LMS Enhancements

ashman-p:
Partially addresses #2229

Missing Initialization in oqs_deserialize_lms_key
Indexing Bug in hss_generate_working_key

### DIFF
--- a/src/sig_stfl/lms/external/hss_verify_inc.c
+++ b/src/sig_stfl/lms/external/hss_verify_inc.c
@@ -48,18 +48,28 @@ bool hss_validate_signature_init(
     uint_fast32_t levels = (uint_fast32_t)get_bigendian( signature, 4 ) + 1;
         /* +1 because what's in the signature is levels-1 */
     signature += 4; signature_len -= 4;
-    if (levels < MIN_HSS_LEVELS || levels > MAX_HSS_LEVELS ||
-                               levels != (uint_fast32_t)get_bigendian( public_key, 4 )) {
+    uint_fast32_t pub_levels = (uint_fast32_t)get_bigendian( public_key, 4 );
+    public_key += 4;
+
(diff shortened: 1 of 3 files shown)


ashman-p:
In the course of testing changes to the indexing issue in hss_generate.c, test vectors were used from 
https://github.com/post-quantum-cryptography/KAT/tree/main/LMS.
LMS by nature defaults to a single-level tree (as opposed to HSS; multilevel trees). The number of levels is embedded into the public key and can be set to 0 for LMS. The code change to hss_verify_inc.c was to allow these test vectors to be used.

src/sig_stfl/lms/external/hss_verify_inc.c
```diff
@@ -48,18 +48,28 @@ bool hss_validate_signature_init(
...
+    if (levels < MIN_HSS_LEVELS || levels > MAX_HSS_LEVELS ) {
         ctx->status = info->error_code = hss_error_bad_signature;
         return false;
     }
-    uint_fast32_t pub_levels = (uint_fast32_t)get_bigendian( public_key, 4 );
+
+    /*
+    * The level for public keys for single level trees sometimes have a value of zero.
+    * in this case, bump  public key level to 1.
+    */
```
dstebila, line 62: Seems a bit weird. Are 0-level public keys a quirk in the implementation we use or actually legit in the spec and in the wild?

ashman-p: @dstebila LMS and HSS (multi-level LMS) share the same key format. However, since LMS is a single-level structure, the value is not included. The LMS code handles the absence by treating it as a zero, but now it also converts it as an explicit value of level = 1.

dstebila (approved):

PR #2417, opened 2026-05-05
Build script to build liboqs

charishma1407:
### Purpose
This PR adds a build script (build_liboqs.sh) to simplify the compilation process for liboqs. The script automates the CMake configuration and build steps, making it easier for developers to build the library with common options.

### Changes
- Added build_liboqs.sh script to automate the build process such as installing the dependencies and building.
- Script handles CMake configuration and compilation
- Provides a convenient wrapper for common build operations

### Testing
The script has been tested locally and successfully builds the liboqs library.

### Checklist
1. Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?
       No - This PR only adds a build convenience script and does not modify any cryptographic algorithms.
2.  Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API?
       No - This PR does not modify the algorithm list or any APIs. It only adds a build helper script.

### AI Assistance
This contribution was created with the use of generative AI for **builld-script-test.yml** workflow


baentsch (commented):
Thanks for this PR @charishma1407 ! I like it as it makes life easier for users -- and is even more complete than the similar build script we have in the even more user-focused [oqs-provider](https://github.com/open-quantum-safe/oqs-provider/blob/main/scripts/fullbuild.sh) project.

That said, some questions:
- Who is the target audience for this? `liboqs` is a library that not too many users would (arguably, should :) build for themselves: One could argue that not having a build script excludes people from building the project that may have a propensity to "shoot themselves into the foot" having/wrongly using such script.
- Why no Windows support?
- Why no CI (testing)?
- Who would maintain this going forward (e.g., when new config options or algs get added/changed)?

bhess (commented):
Hi @charishma1407, thanks for the PR, I agree this is a useful convenience for users. To echo and build on @baentsch's last points: my main concern is long‑term maintainability. The script hard‑codes build options and algorithms but isn't exercised in CI, so as new options or algorithms are added it could easily drift out of sync. Adding CI coverage for the script, or otherwise deriving its options from a single source of truth, would help reduce that risk. On the other hand, this might add quite some complexity that counters the goal to simplify the build.

baentsch:
Thanks @charishma1407 for adding thorough CI testing! This addresses those parts of the questions of @bhess and myself -- but it has one drawback: It requires an awful lot of compute cycles covering many --but surely not all-- current option-configs; also it will conceptually never be able to address the maintenance question (new config options added in the future).

So allow me to make a "wild" proposal: Would it be conceivable that a) at minimum this PR adds a feature to the script to issue a build-time warning if it discovers that any config file it depends upon and that are checked in (e.g., CONFIGURE.md, CMakeLists.txt) is more current than the build script? Even better (but I don't know whether it's possible) would be b) that such "build-script-out-of-date" check (in CI) triggers a Copilot "agent" creating an "AI-written" update to the build script taking the latest config options into account and ideally, even generating a PR for that?

charishma1407:
Hi @baentsch and @bhess I made changes according to the comments

1. Created the workflow .yml file to run a CI test, and reduced it from 13 to 4 jobs to prevent the cycle.
2. Updated the build script with the stale code check in it.
3. As integrating it to copilot, I am not sure that copilot has the feature of creating the PR on its own.
4. Build script can make dynamically but one problem with that is it make the script more complex and also it looses it's simplicity.

build_liboqs.sh
```diff
@@ -0,0 +1,587 @@
...
+    exit 0
+}
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Function to check if script is outdated by validating CMake options
+check_script_staleness() {
+    # Skip check if disabled via environment variable (useful for CI)
+    if [ "${SKIP_STALENESS_CHECK:-}" = "1" ]; then
+        return 0
```
baentsch: Could you please move this check to the end of the routine when you know whether there's staleness or not? I see the point that it may be goodness to pass the test in CI changing other things (incl. config files) -- but I'd much rather like this script issue a clear warning then (also visible in the CI logs) if staleness occurred (but still pass so that CI can continue). That way, conscientious reviewers (of CI logs) could notice the need to update the build script during a PR changing config options. Keeping this disablement as-is, no warning at all is issued but only at runtime to users -- which is a bit late.

charishma1407: Addressed in the latest commit.

.github/workflows/build-script-test.yml
```diff
@@ -0,0 +1,198 @@
+name: Build Script Test
+
+permissions:
+  contents: read
+
+on:
+  push:
+    branches: [ main, charishma_build_script ]
```
baentsch: This one of these branches seems superfluous, no?

.github/workflows/build-script-test.yml
```diff
@@ -0,0 +1,198 @@
...
+          - os: ubuntu-latest
+            test-name: shared-no-openssl
+            script-args: "--shared --no-openssl"
+            description: "Shared library without OpenSSL"
+          
+          # Cross-platform verification - one macOS test
+          - os: macos-latest
+            test-name: default-build
+            script-args: ""
+            description: "Verify script works on macOS"
```
baentsch, line 53: Why not test another config option as part of this MacOS test, too - enhances coverage somewhat "free of charge".

build_liboqs.sh
```diff
@@ -0,0 +1,607 @@
...
+    -h, --help                          Show this help message
+    
+    Build Configuration:
+    --shared                            Build shared library (BUILD_SHARED_LIBS=ON)
+    --build-type TYPE                   Set build type: Debug, Release, MinSizeRel, RelWithDebInfo (default: Release)
+    --install-prefix PATH               Set installation prefix (CMAKE_INSTALL_PREFIX)
+    
+    Algorithm Selection:
+    --algs-enabled SET                  Algorithm set: STD, NIST_R4, NIST_SIG_ONRAMP, All (default: All)
+    --minimal-build "ALG1;ALG2"         Build only specified algorithms (e.g., "KEM_ml_kem_768;SIG_ml_dsa_44")
```
baentsch: Indeed only two algs? Or can there be multiple ";" separated ones?

build_liboqs.sh
```diff
@@ -0,0 +1,607 @@
...
+    --enable-kem-ALG                    Enable specific KEM algorithm
```
baentsch: What happens with other KEM algos not specified? Will they be automatically not built once one KEM is spec'd? At least extend documentation minimally on this item; same for SIG's and stfl's

build_liboqs.sh
```diff
@@ -0,0 +1,607 @@
...
+    --use-icicle                        Use ICICLE GPU acceleration (OQS_USE_ICICLE=ON)
+    
+    CPU Features (for non-dist builds):
+    --use-adx                           Use ADX instructions (OQS_USE_ADX_INSTRUCTIONS=ON)
+    --use-aes                           Use AES instructions (OQS_USE_AES_INSTRUCTIONS=ON)
+    --use-avx                           Use AVX instructions (OQS_USE_AVX_INSTRUCTIONS=ON)
+    --use-avx2                          Use AVX2 instructions (OQS_USE_AVX2_INSTRUCTIONS=ON)
+    --use-avx512                        Use AVX512 instructions (OQS_USE_AVX512_INSTRUCTIONS=ON)
+    
+    Advanced Options:
```
baentsch, line 77: General comment: What are default settings?

build_liboqs.sh
```diff
@@ -0,0 +1,607 @@
...
+        # Issue warning but allow CI to continue
+        if [ "${SKIP_STALENESS_CHECK:-}" = "1" ]; then
+            echo -e "${YELLOW}⚠️  Note: Staleness check is disabled (SKIP_STALENESS_CHECK=1)${NC}"
+            echo -e "${YELLOW}   The above warnings are informational only and will not fail the build.${NC}"
+            echo -e "${YELLOW}   Please review and update the build script if needed.${NC}"
+            echo ""
+        fi
+    else
+        # No staleness detected, but check is disabled - no need to warn
+        if [ "${SKIP_STALENESS_CHECK:-}" = "1" ]; then
```
baentsch: This code looks superfluous.

build_liboqs.sh
```diff
@@ -0,0 +1,607 @@
...
+    --permit-unsupported-arch           Permit compilation on unsupported architecture (OQS_PERMIT_UNSUPPORTED_ARCHITECTURE=ON)
+    --strict-warnings                   Enable strict compiler warnings (OQS_STRICT_WARNINGS=ON)
+    --enable-constant-time-test         Enable constant-time testing (OQS_ENABLE_TEST_CONSTANT_TIME=ON)
+    --use-coverage                      Enable code coverage (USE_COVERAGE=ON)
+    --use-sanitizer TYPE                Enable sanitizer: Address, Memory, Undefined, Thread, Leak
+    
+    Stateful Signatures:
+    --enable-xmss                       Enable XMSS stateful signatures (OQS_ENABLE_SIG_STFL_XMSS=ON)
+    --enable-lms                        Enable LMS stateful signatures (OQS_ENABLE_SIG_STFL_LMS=ON)
+    --enable-stfl-key-sig-gen           Enable stateful key/sig generation (HAZARDOUS, see docs)
```
baentsch: OK, I see the big RED warning output below; but I wonder whether this option should be left away to not make enablement of this feature too simple now. Or maybe just output an echo statement saying how this could be activated by setting the env var (together with pointer to implications).

build_liboqs.sh
```diff
@@ -0,0 +1,607 @@
...
+        MAJOR_VERSION=$(echo $PYTHON_VERSION | cut -d. -f1)
+        if [ "$MAJOR_VERSION" -ge 3 ]; then
+            PYTHON_CMD="python"
+            echo "✓ Found python: $(python --version)"
+        fi
+    fi
+    
+    if [ -z "$PYTHON_CMD" ]; then
+        echo -e "${RED}Error: Python 3 is not installed. Please install Python 3 first.${NC}"
+        echo "You can install it via Homebrew: brew install python3"
```
baentsch: why doesn't this script try to run `brew` itself? It does the same for other UNIXs (via `apt`)

build_liboqs.sh
```diff
@@ -0,0 +1,607 @@
...
+    install_brew_package_if_missing cmake
+    install_brew_package_if_missing ninja
+    install_brew_package_if_missing openssl@3
+    install_brew_package_if_missing wget
+    install_brew_package_if_missing doxygen
+    install_brew_package_if_missing graphviz
+    install_brew_package_if_missing astyle
+    
+    # Install Python test dependencies
+    install_python_test_deps_macos
```
baentsch: Why something MacOS specific?

baentsch (commented):
Basically LGTM, see single comments (Yes, I should have grouped them into one coherent review :-()

.github/workflows/build-script-test.yml
charishma1407: Yeah this one I am using this _charishma_build_script_ just for testing the workflow, in the final commit it will be removed.

build_liboqs.sh
charishma1407: Yes, when using --minimal-build, ONLY the specified algorithms are built - all other algorithms (KEMs, SIGs, and SIG_STFLs not in the list) are automatically disabled. This is the intended behavior for minimal builds to reduce binary size and build time.

.github/workflows/build-script-test.yml
charishma1407: Added another config option for MacOS test.

build_liboqs.sh
charishma1407: Updated the default settings in the script

build_liboqs.sh
charishma1407: Updated in the latest commit on this code.

build_liboqs.sh
charishma1407: I removed this flag feature from the script if a user wants to use they can use using CMAKE from the sciript

build_liboqs.sh
charishma1407: Removed this code and implemented direct code using brew install

README.md
```diff
@@ -113,6 +113,42 @@ This project is not commercially supported. All guidelines and goals for liboqs
 
 ### Linux and Mac
 
+#### Automated Build Script
+
+For a simplified build process, you can use the provided build script that automatically detects your OS, installs dependencies, and builds liboqs:
```
dstebila:
```suggestion
For a simplified build process on Linux or Mac, you can use the provided build script that automatically detects your OS, installs dependencies, and builds liboqs:
```

.github/workflows/build-script-test.yml
```diff
@@ -0,0 +1,203 @@
...
+          echo "Options found in build_liboqs.sh:"
+          cat script_options.txt
+      
+      - name: Compare coverage
+        shell: bash
+        run: |
+          echo "Checking if build script covers major CMake options..."
+          # Check for key options that should be in the script
+          echo "Checking for OQS_USE_OPENSSL..."
+          if grep -q "OQS_USE_OPENSSL" build_liboqs.sh; then
```
dstebila: I think `grep -q` will match substrings as well; is this the desired behaviour?

build_liboqs.sh
```diff
@@ -0,0 +1,498 @@
...
+    # Report missing options
+    if [ ${#MISSING_OPTIONS[@]} -gt 0 ]; then
+        echo -e "${YELLOW}⚠️  Warning: Found CMake options not exposed in this build script:${NC}"
+        
+        # Group and display missing options (limit to first 10 to avoid clutter)
+        local count=0
+        for opt in "${MISSING_OPTIONS[@]}"; do
+            if [ $count -lt 10 ]; then
+                echo -e "${YELLOW}   - $opt${NC}"
+                ((count++))
```
dstebila:
```suggestion
                count=$((count+1))
```

build_liboqs.sh
```diff
@@ -0,0 +1,498 @@
...
+        sudo apt update
+        sudo apt install -y astyle cmake gcc ninja-build libssl-dev python3-pytest python3-pytest-xdist unzip xsltproc doxygen graphviz python3-yaml valgrind
+        
+    elif [[ "$ID" == "nixos" ]]; then
+        # NixOS
+        echo -e "${GREEN}Detected OS: NixOS${NC}"
+        echo ""
+        
+        echo "Entering Nix development environment..."
+        nix develop
```
dstebila: Have you tested the nix version? Claude says the following (but I don't use nix so I have no way of verifying whether this is a legitimate complaint or not):

> `nix develop` spawns an interactive subshell and then falls through to cmake/ninja in the parent shell outside the nix env — the NixOS path almost certainly doesn't work as intended. Either `exec nix develop -c ./build_liboqs.sh --inside-nix` style, or document that users should `nix develop` themselves first.

build_liboqs.sh
```diff
@@ -0,0 +1,498 @@
...
+        exit 1
+    fi
+    
+    echo "Installing dependencies..."
+    brew install cmake ninja openssl@3 wget doxygen graphviz astyle python3
+    
+    # Install Python test dependencies using pip3
+    echo ""
+    echo "Installing Python test dependencies..."
+    pip3 install --break-system-packages pytest pytest-xdist pyyaml
```
dstebila: We should not do `--break-system-packages` for pip.

build_liboqs.sh
```diff
@@ -0,0 +1,498 @@
...
+elif [[ -f /etc/os-release ]]; then
+    # Source the os-release file
+    . /etc/os-release
+    
+    if [[ "$ID" == "ubuntu" ]] || [[ "$ID" == "debian" ]] || [[ "$ID_LIKE" == *"debian"* ]]; then
+        # Ubuntu/Debian
+        echo -e "${GREEN}Detected OS: $NAME${NC}"
+        echo ""
+        
+        echo "Installing dependencies..."
```
dstebila: Should warn the user that this is going to call sudo and that they will be prompted for their password.

README.md
```diff
@@ -113,6 +113,42 @@ This project is not commercially supported. All guidelines and goals for liboqs
...
+```
+
+Common examples:
+
+```bash
+# Build shared library with debug symbols
+./build_liboqs.sh --shared --build-type Debug
+
+# Minimal build with only ML-KEM-768 and ML-DSA-44
+./build_liboqs.sh --minimal-build "KEM_ml_kem_768;SIG_ml_dsa_44"
```
dstebila: Warn the user that if they don't quote a string with semicolons then the shell will view it as a command terminator.

.github/workflows/build-script-test.yml
```diff
@@ -0,0 +1,203 @@
...
+            exit 1
+          fi
+          echo "✓ Build artifacts verified"
+      
+      - name: Run basic library test
+        run: |
+          cd build
+          # Check if test executables exist and run a quick test
+          if [ -f "tests/test_kem" ]; then
+            ./tests/test_kem || echo "KEM test skipped or not applicable"
```
dstebila: Here (and other lines): If I'm not mistaken, using `|| echo "KEM test skipped"` has the effect of skipping a failing test, does it not?

dstebila (changes requested):

dstebila:
>     * Who would maintain this going forward (e.g., when new config options or algs get added/changed)?

I'm also worried about maintainability going forward -- not that the script is unmaintainable, just that we will forget to update it. I think the staleness check is a good step. But probably we need some documentation reminding us? Maybe in the PR template? Or somewhere else?

build_liboqs.sh
```diff
@@ -0,0 +1,536 @@
...
+    fi
+    
+    # Extract all option() declarations from CMake files
+    local CMAKE_OPTIONS=()
+    
+    # Parse main CMakeLists.txt
+    if [ -f "$CMAKE_FILE" ]; then
+        while IFS= read -r line; do
+            [ -n "$line" ] && CMAKE_OPTIONS+=("$line")
+        done < <(grep -E '^option\(OQS_|^option\(USE_|^option\(BUILD_' "$CMAKE_FILE" 2>/dev/null | sed 's/option(//;s/ .*//' || true)
```
xuganyu96: This regex `'^option\(OQS_|^option\(USE_|^option\(BUILD_'` only matches `option` at the beginning of the line.
It does not match `option` that is indented (e.g. `OQS_USE_OPENSSL`), nor does it match `cmake_dependent_option` (e.g. `OQS_USE_AES_OPENSSL`).

build_liboqs.sh
```diff
@@ -0,0 +1,536 @@
...
+    
+    # Check if Homebrew is installed
+    if ! command -v brew &> /dev/null; then
+        echo -e "${RED}Error: Homebrew is not installed!${NC}"
+        echo "Please install Homebrew first: https://brew.sh"
+        exit 1
+    fi
+    
+    echo "Installing dependencies..."
+    brew install cmake ninja openssl@3 wget doxygen graphviz astyle python3
```
xuganyu96: I'm not comfortable with this approach. The script assumes the user wants Python installed via Homebrew and attempts a user-level pip install, only falling back to a virtual environment on failure. The same concern applies to other dependencies (for example, I personally compile and install OpenSSL from source for development). A build script should not make these decisions on the user's behalf.

xuganyu96 (changes requested):
Thank you @charishma1407 for your first contribution!

I am not satisfied with this pull request for three reasons:
1. There is some non-trivial logic in `check_script_staleness` that is incorrect and difficult to validate or test (see comments)
1. The approach to dependency management is too inflexible (see comments)
1. The help text introduces another piece of duplicate documentation and adds to maintenance overhead

However, I do recognize the pain points that this build script is trying to address:
1. It is difficult to find a complete list of build options and what each does
2. Some dependencies are not clearly documented, such as `pytest-xdist`

The first point can be addressed by improving the help text in `CMakeLists.txt` and `.CMake/alg_support.cmake`.
As the top-level [`README.md`](https://github.com/open-quantum-safe/liboqs/blob/4956ce72a18ccebb6b6468fa3cef4ba9759b3cf2/README.md?plain=1#L138) already pointed out, running CMake with the `-LAH` flag will print out a complete and documented list of options
and I want to lean on existing tooling as much as possible.
Similarly, CMake has built-in commands like `find_library`, `find_package`, `find_program` that can locate system packages, libraries, and binaries.
It would be better to try them before resorting to a separate bash script.
Finally, we can move Python dependencies into a top-level `requirements.txt` and add the instructions to the documentation.

charishma1407:
Hi @xuganyu96 
Thank you for the detailed feedback on the script. I've addressed all three concerns:

1. Added an option only to build not installing the dependencies.
2. If a user selects for default build it will be installing the default dependencies and build with default settings.
3. Created requirements.txt as you suggested
4. Updated the regex for more accurate.
5. In the script there is another option where user can directly give -D and build option apart from the given options.
6. Script help text focuses on common use cases

This is script is not intended to remove the dependency on manual installation or CMAKE usage. As documented user can also go with the manual installation steps also.

baentsch:
@dstebila @xuganyu96 Did your concerns get sufficiently addressed by the new commits? How do you suggest moving this PR to conclusion?

xuganyu96:
> @dstebila @xuganyu96 Did your concerns get sufficiently addressed by the new commits? How do you suggest moving this PR to conclusion?

Upon re-reviewing, I find the maintenance overhead substantial, while the benefits are dubious. I think building liboqs is fairly straightforward: use cmake to generate build command, then use `ninja` to build. Build option discovery can be managed using `cmake -LAN`, so the help text in the build script is a duplicate of the descriptions already in the CMakeLists.txt files. My stand is to reject this pull request.

bhess:
> > @dstebila @xuganyu96 Did your concerns get sufficiently addressed by the new commits? How do you suggest moving this PR to conclusion?
> 
> Upon re-reviewing, I find the maintenance overhead substantial, while the benefits are dubious. I think building liboqs is fairly straightforward: use cmake to generate build command, then use `ninja` to build. Build option discovery can be managed using `cmake -LAN`, so the help text in the build script is a duplicate of the descriptions already in the CMakeLists.txt files. My stand is to reject this pull request.

There is also ccmake, which I only recently learned about and have found quite useful for configuring CMake builds. I think it would be useful to add something about such "build simplifiers" to CONFIGURE.md.

image

I share the original concern regarding the long-term maintainability of the build script. I appreciate the work that has gone into this PR and the subsequent revisions, but I'm also not convinced that the additional maintenance burden is justified by the benefits, especially given the existing cmake configuration tools.

I do, however, find the dependency installation assistance quite useful. Would it be an option to keep just that part of the script?

charishma1407:
Thank you everyone for taking the time to review this PR and share your feedback. As you have mentioned on the maintenance issue I'm going to go ahead and close this PR for now.
Thanks again for your time and guidance!

PR #2420, opened 2026-05-12
Dependabot configuration and dependency updates

dstebila:
Fixes https://github.com/open-quantum-safe/liboqs/security/dependabot/28. Fixes https://github.com/open-quantum-safe/liboqs/security/dependabot/29. Fixes https://github.com/open-quantum-safe/liboqs/security/dependabot/30.

Adds a Dependabot config file. This will bring more things into scope for Dependabot than just the default config. In particular, it will watch for GitHub Actions updates, and help us address https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,21 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+
+  - package-ecosystem: "pip"
+    directories:
+      - "/scripts/copy_from_upstream"
+      - "/.github/workflows"
+    schedule:
+      interval: "weekly"
+    groups:
+      pip:
+        patterns:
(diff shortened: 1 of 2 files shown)


xuganyu96 (approved):
Code change looks good to me. Just a quick question:

> help us address https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

How does Node 20 impact our GitHub action workflows?

dstebila:
> Code change looks good to me. Just a quick question:
> 
> > help us address https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
> 
> How does Node 20 impact our GitHub action workflows?

I don't know exactly, but Dependabot indicated that it thought this would affect us (perhaps some of the actions we use are affected; and we're pinned to older versions, so we'll need to consciously update). My hope would be that by getting Dependabot configured to track actions we will get this (and other similar things) taken care of more automatically for us.

PR #2328, opened 2025-12-01
Hqc update reference implementation 

vsoftco:
Updates HQC to https://gitlab.com/pqc-hqc/hqc v5.0.0. Minimal changes in the upstream code.

PR #2426, opened 2026-05-13
Fix Keccak dispatch race and cross-backend split

dstebila:
Claude Code thinks that this might be the cause of the [intermittent SHA3 failures](https://github.com/open-quantum-safe/liboqs/pull/2407#issuecomment-4308479620). The suggestion and fix seem plausible to me but I don't have access to an AVX512 machine to test it out.

Claude's report:

> `Keccak_Dispatch()` in `xkcp_sha3.c` (and the equivalent in `xkcp_sha3x4.c`) overwrote the global `sha3_default_callbacks` struct field-by-field on the first SHA3 call when AVX512VL was available. Two problems:
> 
> 1. **Non-atomic struct copy.** The assignment `sha3_default_callbacks = sha3_avx512vl_callbacks` is a member-wise copy of a struct of function pointers. `pthread_once` only serialized callers of `Keccak_Dispatch` itself, not concurrent SHA3 operations on other threads, which could observe a partially-overwritten vtable.
> 
> 2. **Cross-backend split of a single in-flight call.** The very call that triggered dispatch had already entered through the xkcp `_inc_init`. After the swap, subsequent `_inc_absorb` / `_inc_finalize` on the same context routed through the newly-installed AVX512VL table, so one in-flight context was initialized by one backend and absorbed/finalized by another.
> 
> This is consistent with the intermittent SHA3-512 failures observed in HQC's `hash_g` on certain AVX512-capable CI runners but not others.
> 
> The fix moves the top-level backend selection out of `Keccak_Dispatch` and into a one-time `init_callbacks()` in `sha3.c` / `sha3x4.c`, guarded by `pthread_once`, which swaps the `callbacks` **pointer** (a single aligned-pointer store) to `sha3_avx512vl_callbacks` when AVX512 is present. `Keccak_Dispatch` still sets the xkcp-internal `Keccak_*_ptr` function pointers for its own use, but no longer touches the top-level table. `OQS_SHA3_set_callbacks` now triggers the one-time init before assigning the user's value, so auto-detection cannot later clobber a user-provided table.


xuganyu96:
@dstebila I've cherry-picked [the commit](https://github.com/open-quantum-safe/liboqs/pull/2426/commits/7a70751f3a54c2fd20eecdbfe57f415e0458b73d) that fix the SHA3-512 incremental absorption error into the HQC integration pull request #2328. This raises a new question of why this error is only intermittent, which might have something to do with the rest of the code change proposed here, but ~~I currently don't have the capacity to investigate~~ this might be out-of-scope for 0.16 release. I'd like this pull request to be closed to clean up the list of open PRs ahead of 0.16 release, then raise a separate issue for investigating potential race condition in `Keccak_Dispatch`.

xuganyu96:
Closing this pull request as AVX512VL SHA3 failures were already addressed. The possibility of a race condition in Keccak Dispatch logic remains, but given that it is not causing any problems, I find the troubleshooting efforts to not be worth the reward at this moment.

PR #2429, opened 2026-05-13
[Security] kem/hqc: fix secret-dependent break in compute_elp (CT hardening)

Chennamma-Hotkar:
## Summary
The Berlekamp loop in `compute_elp()` contained an early `break` on the
final iteration (`mu == 2*PARAM_DELTA-1`), skipping updates to `pp`,
`d_p`, `X_sigma_p`, `deg_sigma_p` and `d`. This creates a timing side
channel — the last iteration executes less work than all others, leaking
information through execution time.

## Fix
Replaced the `break` with a constant-time `not_last` mask:
```c
uint16_t not_last = (uint16_t)(-((uint16_t)(mu - (2 * PARAM_DELTA - 1)) != 0));
```
All updates are ANDed with `not_last` so every iteration executes
identical work — results are frozen on the final pass rather than skipped.

Applied to all three security levels: HQC-128, HQC-192, HQC-256.

## Evidence
- valgrind --tool=memcheck: ERROR SUMMARY: 0 errors
- test_kem HQC-128/192/256: shared secrets are equal
- KAT vector matches upstream submission

## Verification

Valgrind memcheck with secret key tainted via VALGRIND_MAKE_MEM_UNDEFINED:
Screenshot 2026-05-14 032512


dstebila:
Thanks for the contribution. We will shortly be removing the current version of the HQC code and replacing it with a new version (see #2407). As far as I can tell, the identified bug also applies to the new version, so we would want to address that. Since we pull our HQC implementation from the HQC team's repository, the best approach would be to report the bug at the HQC repository and fix it there, and then we inherit that automatically when we pull in the next update.

Chennamma-Hotkar:
Thanks for the feedback. I'll report this upstream to the HQC repository and submit a fix there. Should I close this PR or keep it open until the upstream fix lands?

dstebila:
> Thanks for the feedback. I'll report this upstream to the HQC repository and submit a fix there. Should I close this PR or keep it open until the upstream fix lands?

Let's close this one, and then please re-open a new one when the upstream fix lands.

Chennamma-Hotkar:
Understood! Fix has been submitted upstream at https://gitlab.com/pqc-hqc/hqc/-/merge_requests/9 - will re-open here once it lands.

PR #2430, opened 2026-05-14
Fix latent warnings exposed by refreshed CI container images

dstebila:
## Summary

The CI Docker images (`openquantumsafe/ci-alpine-amd64:latest` and `openquantumsafe/ci-ubuntu-latest:latest`) were refreshed recently, bumping GCC (9.3.0 → 14.2.0 on Alpine, to 15.2.0 on Ubuntu), Clang (to 21.1.8), and CMake to ≥ 3.24. The newer CMake means `CMAKE_COMPILE_WARNING_AS_ERROR` (set via `OQS_STRICT_WARNINGS`) now actually adds `-Werror`, which exposed two latent issues plus reformatting needs:

- **`Guard trailing_1_bits with OQS_ALLOW_LMS_KEY_AND_SIG_GEN`** — the static helper in `hss_keygen.c` is only called from inside the `OQS_ALLOW_LMS_KEY_AND_SIG_GEN` block, so the `alpine-no-stfl-key-sig-gen` job now trips `-Werror=unused-function`. Wrap the definition in the same `#ifdef`.
- **`Re-run clang format due to new CI images`** — clang-format output drifted with the new Clang version.
- **`Stop promoting warnings to errors for upstream XKCP code`** — Clang 21 errors on `-Wimplicit-int-conversion` in upstream `KeccakP-1600-opt64.c`'s `COMPL` macro. Set `COMPILE_WARNING_AS_ERROR=OFF` on the four `xkcp_low` OBJECT targets so future compiler tightening on imported code doesn't break us again.

These commits were originally landed on the github-actions dependabot branch (#2422); splitting them here because they're unrelated to the action bumps.

## Test plan

- [x] `alpine-no-stfl-key-sig-gen` config builds cleanly with refreshed Alpine image (GCC 14.2.0)
- [x] `ubuntu-latest` clang fuzz build (`-DOQS_STRICT_WARNINGS=ON -DOQS_BUILD_FUZZ_TESTS=ON`) builds cleanly with Clang 21.1.8
- [ ] CI passes on this PR


dstebila:
I'm working on fixes...

ashman-p (dismissed):
Looks good assuming all tests succeed.

dstebila:
> Looks good assuming all tests succeed.

Here's hoping :)

PR #2431, opened 2026-05-14
Implement value barrier and apply to FrodoKEM

xuganyu96:
# Summary
This pull request implements a `OQS_MEM_BLACK_BOX` macro, which will serve as a frontend for optimization barriers. For now, we will provide one implementation that specifically works with GNUC/Clang toolchain. For other tool chains such as Windows, there will be a compile-time warning.

The black box is then applied to FrodoKEM's `ct_select` to prevent compiler from reasoning about `selector` when inlining `ct_select`.
# Validation
Manually verify the absence of conditional memory reads from the compiled assembly with the following build commands:

```bash
# Apple Silicon, Homebrew Clang 18, MacOS 26.4
cmake -GNinja \
    -DCMAKE_C_FLAGS="-DOQS_DISABLE_MEM_BLACK_BOX" \
    -DCMAKE_C_COMPILER="/opt/homebrew/bin/clang-18" \
    -DCMAKE_BUILD_TYPE="MinSizeRel" \
    -DCMAKE_OSX_DEPLOYMENT_TARGET="26.0" \
    -DOQS_MINIMAL_BUILD="KEM_frodokem_640_aes" \
    .. 
    && ninja \
    && objdump -d lib/liboqs.a > liboqs.a.S
```

- With `-DOQS_DISABLE_MEM_BLACK_BOX`, `<OQS_KEM_frodokem_640_aes_decaps>` contains a `csel` instruction.
- Without `-DOQS_DISABLE_MEM_BLACK_BOX`, `<OQS_KEM_frodokem_640_aes_decaps>` does not contain `csel` instruction. This is true for all FrodoKEM parameters (with the `OQS_MINIMAL_BUILD` options removed).
- Without `-DOQS_DISABLE_MEM_BLACK_BOX`, but modifying source code to force fallback, `<OQS_KEM_frodokem_640_aes_decaps>` does not contain `csel` either.

*edit:* removed fallback optimization barrier implementation and automated test binaries. See [this comment](https://github.com/open-quantum-safe/liboqs/pull/2431#issuecomment-4593646573) for details.


dstebila (dismissed):

src/common/common.h
```diff
@@ -19,6 +19,79 @@
...
+    } while (0)
+
+#else
+/* The fallback implementation: pure C, no inline assembly, portable */
+
+#pragma message("OQS_MEM_BLACK_BOX: unrecognised compiler. "                   \
+                "Falling back to volatile round-trip. "                        \
+                "Verify generated assembly for constant-time correctness.")
+
+static inline void oqs_black_box_fallback(volatile void *p,
```
bhess: Is there any reason `oqs_black_box_fallback` is declared here when it's defined immediately after?

src/common/common.h
```diff
@@ -19,6 +19,79 @@
...
+ */
+#define IGNORE_UNUSED_FUNC
+#endif
+
+/**
+ * Value barrier: prevent compiler from optimizing on secret values
+ */
+#if defined(OQS_DISABLE_MEM_BLACK_BOX)
+
+#warning "DANGER: OQS_DISABLE_MEM_BLACK_BOX is for internal testing only"
```
bhess, line 40: `#warning` is a C23 feature (see: https://cppreference.com/c/preprocessor), so it may not be portable across older compilers. Could we use a more portable way to emit a preprocessor warning?

CONFIGURE.md
```diff
@@ -248,6 +249,34 @@ When this option is set to `ON`, the additional option `OQS_ENABLE_TEST_CONSTANT
 
 **Default**: `OFF`.
 
+## OQS_ENABLE_KEM_FO_CACHE_ORACLE
```
bhess: Without reading the explanation below, this option sounds like it enables KEM cache oracles in liboqs (which would be concerning), whereas it only controls building a test. Perhaps a more explicit name like `OQS_ENABLE_KEM_FO_CACHE_ORACLE_TEST` would avoid confusion.

bhess (commented):
Thanks @xuganyu96 for adding this very useful feature.

I haven’t reviewed the FO cache oracle test in detail, but will this be run in CI? If not, is there a plan to run it regularly?

Regarding the countermeasures:

> Use inline assembly `__asm__ volatile("" : "+r"(v) :)` where available  
> Use a volatile round-trip that reads and writes every byte in the input buffer as a fallback

I have two comments/questions:
1. It might be good to document with a rationale or related work why they are effective.
2. The fallback option (reading and writing every byte) seems potentially expensive if it is used with larger data structures than with the selector byte. Were there more efficient alternatives considered?

See also the comments inline.

xuganyu96:
@bhess 

> will this be run in CI? If not, is there a plan to run it regularly?

I don't plan to run this on GitHub Actions because the cache timing test must run on bare metal for cache eviction to be meaningful and for timing measurements to be accurate. I've added a line to `CONFIGURE.md` explaining this.

> It might be good to document with a rationale or related work why they are
> effective.

I borrowed the inline ASM optimization barrier from mlkem-native. Here is [one example](https://github.com/pq-code-package/mlkem-native/blob/31369e41678734a2771389be9a02b946f6da1a2f/mlkem/src/verify.h#L116). I am not sure if there is any official documents recommending this pattern, but it does seem to be the gold standard across many prominent projects.

> The fallback option (reading and writing every byte) seems potentially
> expensive if it is used with larger data structures than with the selector
> byte. Were there more efficient alternatives considered?

The fallback option is meant to cover for Windows builds, since MSVC does not support inline assembly in the same way GNU/LLVM does. I plan to follow up in a subsequent patch to add an appropriate MSVC backend for optimization barrier. This round trip is just an unfortunate temporary band-aid.

src/common/common.h
xuganyu96: I need a declaration to put `IGNORE_UNUSED_FUNC`. I cannot put it at the implementation: clang will raise warning:

```
/Users/ganyuxu/Workspace/open-quantum-safe/liboqs/build/include/oqs/common.h:71:20: error: GCC does not allow 'unused' attribute in this position on a function definition [-Werror,-Wgcc-compat]
   71 |         size_t sz) IGNORE_UNUSED_FUNC {
      |                    ^
/Users/ganyuxu/Workspace/open-quantum-safe/liboqs/build/include/oqs/common.h:28:43: note: expanded from macro 'IGNORE_UNUSED_FUNC'
   28 | #define IGNORE_UNUSED_FUNC __attribute__((unused))
      |                                           ^
```

Which becomes compiler error with `OQS_STRICT_WARNINGS=ON`.

src/common/common.h
xuganyu96: I don't plan on supporting `OQS_DISABLE_MEM_BLACK_BOX` as a user-facing feature unless there is a strong case for it. This flag is for internal use only for now.

CONFIGURE.md
xuganyu96: Renamed to `OQS_ENABLE_KEM_FO_CACHE_ORACLE_DETECTOR`

bhess:
Thanks @xuganyu96 for the update.

> > 
> 
> I borrowed the inline ASM optimization barrier from mlkem-native. Here is [one example](https://github.com/pq-code-package/mlkem-native/blob/31369e41678734a2771389be9a02b946f6da1a2f/mlkem/src/verify.h#L116). I am not sure if there is any official documents recommending this pattern, but it does seem to be the gold standard across many prominent projects.
> 
> >

Can we document this somewhere (e.g., as code comment, reference to mlkem-native). Might be helpful to keep the reference.

tests/kem_fo_cache_oracle.c
```diff
@@ -0,0 +1,526 @@
...
+#define NUM_ROUNDS 10
+#define MEMALIGN 4096
+#define ONE_BILLION 1000000000
+#define CSV_HEADER "epoch,sample,good,probe,ctrl\n"
+#define CSV_DATA "%d,%d,%d,%" PRIu64 ",%" PRIu64 "\n"
+
+#define ARGS_HELP_TEXT                                                         \
+    "Usage: %s <kem_name> <probe_loc>\n"                                       \
+    "Arguments:\n"                                                             \
+    "    kem_name: FrodoKEM-640-AES\n"                                         \
```
baentsch: Why just FrodoKEM-640-AES? Worthwhile listing all algs this could be applied to (be arguments)?

tests/kem_fo_cache_oracle.py
```diff
@@ -0,0 +1,580 @@
...
+- ``sample`` – integer sample index within the epoch
+- ``good``   – 1 if the ciphertext passed to decapsulation was valid, else 0
+- ``probe``  – cycle count for reading the rejection symbol (target)
+- ``ctrl``   – cycle count for reading a control memory location (baseline)
+
+Usage
+-----
+.. code-block:: bash
+
+    python timing_analysis.py samples.csv
```
baentsch: This Usage sample seems to disagree with the script's name. Intentional?

xuganyu96:
Thank you @bhess and @baentsch for the feedback.

After re-considering the points raised in this pull request, I am feeling less confident about merging kem_fo_cache_oracle into the code base. Merging them into the code base carries an implicit warranty that OQS (and specifically, me) is responsible for problems that users might face while using these tools, and I don't feel comfortable claiming this responsibility over kem_fo_cache_oracle.c, which is honestly speaking a crude PoC that I took from the original author, nor kem_fo_cache_oracle.py, which is AI-generated without my scrutiny.

A similar argument applies to the memory roundtrip fallback implementation of the optimization barrier, which I wrote without fully understanding how and when non-GNU/Clang toolchain can guarantee secret-independent memory access pattern, let alone the performance implications.

Thus, I would like to propose stripping down the scope of this pull request to only two things: an optimization barrier that targets GNU/Clang toolchain, and its application to FrodoKEM. The cache timing tests will be removed entirely, though I promise to polish them for later deployment. The fallback memory round trip will also be replaced with a compile-time warning about using non-GNU/Clang toolchain and its security implications.

I would also appreciate some feedback on how this pull request should or should not block 0.16 release. My current approach is that security-related fixes like this should happen somewhat orthogonally with feature/tooling/infrastructure development. Concretely speaking, this pull request should not block 0.16 release, but when it is merged, it can go into 0.16.0-rc2, or maybe even a 0.16.1 patch.

cc: @dstebila

baentsch:
Thanks @xuganyu96 for revisiting this and your thoughts, which I fully support. Particularly this statement

> nor kem_fo_cache_oracle.py, which is AI-generated without my scrutiny

is one that gives me "the creeps": I also heavily use AI by now but at no point in time should this be introduced to a code base, particularly one pertaining to security functionality, without serious review. fwiw, the OpenSSL project is converging on a [declare-and-stand-behind-the-code AI policy](https://github.com/openssl/general-policies/pull/85) which I find a very sensible approach.

As far as (postponing or not) 0.16. is concerned, can I ask whether there is a specific need for specific features (like this one) in 0.16? Why not push this whole topic to 0.17 instead of already now talking about a patch/0.16.1.?

src/common/common.h
```diff
@@ -19,6 +19,61 @@
...
+#define OQS_MEM_BLACK_BOX(v)                                                   \
+    do {                                                                       \
+        __asm__ volatile("" : "+r"(v) :);                                      \
+    } while (0)
+
+#else
+
+#pragma message("WARNING: non-GNUC/Clang toolchain detected. liboqs cannot     \
+                 guarantee optimization barrier on unsupported platforms.      \
+                 Compiler may introduce cache timing side channels. Please     \
```
bhess: nit: 
> Compiler may introduce cache timing side channels.

Not only cache-timing side channels, perhaps put "cache" in parenthesis.

bhess (commented):
Thanks for updating and trimming the PR @xuganyu96, it looks much cleaner/simpler now without the FO test case.

PR #2433, opened 2026-05-14
Fall back when EVP_DigestSqueeze is unavailable or unsupported

### LINKED ISSUE #2042: liboqs Fails Depending on the OpenSSL Provider Version Used
An active OpenSSL FIPS Provider can be based on a lower version of OpenSSL than the version of libcrypto currently loaded.
e.g. OpenSSL version could be 3.3 while the FIPS Provider could based on 3.0.

In this scenario, I came across the case where liboqs use of OpenSSL's SHA3 APIs failed. 
SHA3_shake256_x4_inc_squeeze(), SHA3_shake128_x4_inc_squeeze failed because EVP_DigestSqueeze is not supported in OpenSSL 3.0.

The question therefore is, how can this be handled in general? I think it would be good to keep liboqs provider agnostic if possible.

dstebila:
_PR description written by Claude._

## Summary

Fixes #2042 and the root cause of open-quantum-safe/liboqs-python#122.

Both issues trace back to the same code: `EVP_DigestSqueeze` was called
unconditionally on OpenSSL 3.3+ headers, with no handling for the cases
where the symbol or the operation isn't actually available at runtime.

- **#2042**: with an active FIPS 3.0 provider under libcrypto 3.3+,
  `EVP_DigestSqueeze` returns `EVP_R_METHOD_NOT_SUPPORTED` and SHAKE
  incremental squeeze produces no output.
- **liboqs-python#122**: every other OpenSSL EVP call in
  [src/common/ossl_functions.h](src/common/ossl_functions.h) goes
  through the dlopen-wrapper indirection, but `EVP_DigestSqueeze` was
  the lone exception. `liboqs.so` therefore carried
  `U EVP_DigestSqueeze` and `dlopen` failed on libcrypto < 3.3.

## Approach

A single mechanism resolves both:

1. **`OPTIONAL_FUNC` macro** alongside `FUNC` / `VOID_FUNC` in the
   dlopen wrapper machinery. Optional symbols use a non-fatal `dlsym`
   (the default `ensure_symbol` calls `exit(EXIT_FAILURE)` on missing
   symbols — desired behavior for required APIs). The wrapper returns
   0 when the pointer is NULL; an `_oqs_ossl_has_<name>()` predicate
   lets callers probe.
2. **`OSSL_FUNC_AVAILABLE(name)`** — the probe in DLOPEN builds, a
   constant `1` in non-DLOPEN builds (the linker has already bound the
   symbol).
3. **`EVP_DigestSqueeze`** added to
   [ossl_functions.h](src/common/ossl_functions.h) as an
   `OPTIONAL_FUNC`, gated on `OPENSSL_VERSION_NUMBER >= 0x30300000L`.
4. **Process-wide one-shot probe**
   ([oqs_ossl_can_digest_squeeze](src/common/ossl_helpers.c)): on first
   call, allocate a throwaway SHAKE128 context, attempt a 1-byte
   `EVP_DigestSqueeze`, cache the boolean result.
   `pthread_once`-guarded under `OQS_USE_PTHREADS`. Provider capability
   does not change across the process lifetime, so the result is
   reusable everywhere.
5. **Squeeze call sites** in
   [ossl_sha3.c](src/common/sha3/ossl_sha3.c) and
   [ossl_sha3x4.c](src/common/sha3/ossl_sha3x4.c) become:
   ```c
   if (oqs_ossl_can_digest_squeeze()) {
       OQS_OPENSSL_GUARD(OSSL_FUNC(EVP_DigestSqueeze)(s->mdctx, output, outlen));
       return;
   }
   /* existing clone + EVP_DigestFinalXOF fallback, unchanged */
   ```
   `OQS_OPENSSL_GUARD` turns a post-probe failure into a loud
   `exit(EXIT_FAILURE)` rather than silently emitting wrong
   cryptographic output.

## Test plan

Verified in Docker on `ubuntu:25.04` (OpenSSL 3.4.1, aarch64):

- [x] On `main`, `nm -uD build/lib/liboqs.so | grep EVP_Digest` shows
      `U EVP_DigestSqueeze`. On this branch it shows nothing. Closes
      the dlopen symbol error behind liboqs-python#122.
- [x] `test_sha3` (SHA3-256/384/512, SHAKE-128/256, four-way parallel
      SHAKE-128/256 KATs) passes with `OQS_DLOPEN_OPENSSL=OFF` and
      `OQS_USE_SHA3_OPENSSL=ON`.
- [x] Same `test_sha3` passes with `OQS_DLOPEN_OPENSSL=ON`.
- [x] `kat_kem ML-KEM-768` and `kat_sig ML-DSA-65` (both SHAKE-heavy)
      pass under both DLOPEN configurations.
- [ ] **Not yet exercised in CI:** the actual FIPS-3.0-under-libcrypto-3.3
      runtime trigger (#2042 scenario). The fallback path is the same
      clone + `EVP_DigestFinalXOF` sequence used on OpenSSL < 3.3
      today, which is exercised by every pre-3.3 build. A CI job that
      forces the probe to return 0 (or runs against a FIPS-only
      provider) would close this gap.

## Notes for reviewers

- No benchmark of fast-path vs. fallback-path is included. If the
  performance delta is small, deleting the `EVP_DigestSqueeze` path
  entirely and always using the clone fallback would simplify this
  further. Open to going that way if measurements support it.
- `OQS_OPENSSL_GUARD` is reused from `aes/aes_ossl.c`. It uses
  `exit(EXIT_FAILURE)`, which is consistent with existing liboqs
  practice but is a more aggressive failure mode than main's
  silent-wrong-output. The tradeoff is: noisy crash vs. silently broken
  cryptography. Happy to discuss.
- Probe is lazy (first squeeze call) rather than eager (library init).
  Trivial latency on cold start; could move to `OQS_init` if eager
  initialization is preferred.

---

### DIFF
--- a/src/common/ossl_functions.h
+++ b/src/common/ossl_functions.h
@@ -18,6 +18,14 @@ FUNC(int, EVP_DigestInit_ex,
      (EVP_MD_CTX *ctx, const EVP_MD *type, ENGINE *impl), (ctx, type, impl))
 FUNC(int, EVP_DigestUpdate, (EVP_MD_CTX *ctx, const void *d, size_t cnt),
      (ctx, d, cnt))
+#if OPENSSL_VERSION_NUMBER >= 0x30300000L
+/* Optional: may be absent from the libcrypto loaded at runtime, or rejected by
+ * the active provider (e.g. a FIPS 3.0 provider under libcrypto 3.3). Callers
+ * must check OSSL_FUNC_AVAILABLE and the return value. */
+OPTIONAL_FUNC(int, EVP_DigestSqueeze,
+              (EVP_MD_CTX *ctx, unsigned char *out, size_t outlen),
+              (ctx, out, outlen))
+#endif
 FUNC(int, EVP_EncryptFinal_ex,
      (EVP_CIPHER_CTX *ctx, unsigned char *out, int *outl), (ctx, out, outl))
 FUNC(int, EVP_EncryptInit_ex,
--- a/src/common/ossl_helpers.c
+++ b/src/common/ossl_helpers.c
@@ -14,6 +14,9 @@
 
 static pthread_once_t init_once_control = PTHREAD_ONCE_INIT;
 static pthread_once_t free_once_control = PTHREAD_ONCE_INIT;
+#if OPENSSL_VERSION_NUMBER >= 0x30300000L
+static pthread_once_t squeeze_probe_once_control = PTHREAD_ONCE_INIT;
+#endif
 #endif
 
 static EVP_MD *sha256_ptr, *sha384_ptr, *sha512_ptr,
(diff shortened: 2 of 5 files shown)


ashman-p (commented):
I believe this check is suitable for static builds. However, since liboqs can be compiled as a shared library, should the check include a runtime component? For instance, a runtime check that verifies the presence of the squeeze function in the vector table?

dstebila:
> I believe this check is suitable for static builds. However, since liboqs can be compiled as a shared library, should the check include a runtime component? For instance, a runtime check that verifies the presence of the squeeze function in the vector table?

I think it covers things at runtime: `oqs_ossl_can_digest_squeeze` at runtime does a test squeeze to see if the function is available, and records the result in memory so that subsequent calls make use of that information. https://github.com/open-quantum-safe/liboqs/pull/2433/changes#diff-806e662bc37238f39f9dc231d698a4f50a88df14f68df44093efce82f2d23b6cR115-R146

But I don't know a lot about shared library build/runtime situations, so maybe I'm misunderstanding.

ashman-p:
> > I believe this check is suitable for static builds. However, since liboqs can be compiled as a shared library, should the check include a runtime component? For instance, a runtime check that verifies the presence of the squeeze function in the vector table?
> 
> I think it covers things at runtime: `oqs_ossl_can_digest_squeeze` at runtime does a test squeeze to see if the function is available, and records the result in memory so that subsequent calls make use of that information. https://github.com/open-quantum-safe/liboqs/pull/2433/changes#diff-806e662bc37238f39f9dc231d698a4f50a88df14f68df44093efce82f2d23b6cR115-R146
> 
> But I don't know a lot about shared library build/runtime situations, so maybe I'm misunderstanding.

The case I was considering is that liboqs is built as a shared library with a version of OpenSSL that supports the feature. However, at runtime, it is loaded on a system that has a downgraded version of OpenSSL.

I can try some experiments to see?

dstebila:
> The case I was considering is that liboqs is built as a shared library with a version of OpenSSL that supports the feature. However, at runtime, it is loaded on a system that has a downgraded version of OpenSSL.
> 
> I can try some experiments to see?

That would be great!

ashman-p:
> > The case I was considering is that liboqs is built as a shared library with a version of OpenSSL that supports the feature. However, at runtime, it is loaded on a system that has a downgraded version of OpenSSL.
> > I can try some experiments to see?
> 
> That would be great!

Thinking it through... I guess the application would fail to load for the scenario I described. Thanks for taking care of this issue.

ashman-p (approved):

PR #2435, opened 2026-05-16
Add NIST LMS Signature verify KATs

ashman-p:
Adds additional NIST Signature Verification KATs from https://github.com/post-quantum-cryptography/KAT/blob/main/LMS/

Theses LMS KATs assume a layer value of 1, which is not explicitly included as with HSS public keys.
Utility functions were added to improve parsing the public key buffer passed in.


src/sig_stfl/lms/external/hss_verify_inc.c
```diff
@@ -16,6 +16,62 @@
...
+ *                    LMS format: lm_type || lm_ots || ...
+ * big-endian 32-bit word (bytes 8..11) is a supported LM-OTS parameter set
+ * (see OQS_LMOTS_TYPE_LIST). In that layout, bytes 0..3 are the level count,
+ * bytes 4..7 are the top-level LMS type, and bytes 8..11 are lm_ots. For a
+ * bare LMS public key, bytes 8..11 are the start of the I field and are not
+ * generally a valid LM-OTS id, so this function returns false.
+ *
+
+ * Returns true if the third word matches a known LM-OTS code; false otherwise.
+ */
```
dstebila, line 35: Is it the case that bare LMS keys *definitely will not* contain one of the target values in bytes 8..11, or only that such a value will occur with small probability?  If the I value is uniformly random (I'm not sure if it is), then 4 bytes take on a target value with probability 2^(-32), which is small but not negligible.

ashman-p: @dstebila, the function has been updated to make it much more reliable. Thanks for pointing this out.

dstebila (approved):

PR #2438, opened 2026-05-19
Fix comments in SHA3 AVX512VL assembly file

dstebila:
Reported by Anthony Plank

Comments in SHA3 AVX512VL assembly file refer to the wrong data type, but the the code is accurage.

### DIFF
--- a/src/common/sha3/avx512vl_low/SHA3-times4-AVX512VL.S
+++ b/src/common/sha3/avx512vl_low/SHA3-times4-AVX512VL.S
@@ -347,7 +347,7 @@
 
 # -----------------------------------------------------------------------------
 #
-# void SHA3_shake128_x4_inc_ctx_reset_avx512vl(OQS_SHA3_shake128_inc_ctx *state);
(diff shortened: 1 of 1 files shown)


baentsch (approved):

baentsch:
Side note: suggest adding [skip ci] to such PRs again to save compute cycles...

OK, ridiculous comment given the use of compute-cycle-gobbling AIs for reviews etc. :-/

bhess (approved):
LGTM

dstebila:
> Side note: suggest adding [skip ci] to such PRs again to save compute cycles...

Definitely... I think I am a bit confused on whether [skip ci] actually works. I don't see any "skip ci" phrase in our GitHub workflows files, so I was wondering if it still has an effect, but checking the [documentation](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs) now it looks like it's built into GitHub Actions and we don't need to enable it manually.  Will do this in the future, thanks for commenting about it.

PR #2442, opened 2026-05-20
fix: incorrect incremental absorption in AVX512VL impl of SHA3-512

xuganyu96:
# Summary
This pull request fixes some incorrect logic in the AVX512VL implementation of SHA3-512 and provides a regression test that would have failed without the fix.

# AI/LLM disclosure
The assembly fix was generated by Claude Opus 4.7 using Claude Code.

* [**NO**] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
* [ **NO** ] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged. Also, make sure to update the list of algorithms in the continuous benchmarking files: .github/workflows/kem-bench.yml and sig-bench.yml)

### DIFF
--- a/tests/test_sha3.c
+++ b/tests/test_sha3.c
@@ -458,9 +458,26 @@ int sha3_512_kat_test(void) {
 	OQS_SHA3_sha3_512_inc_ctx_reset(&state);
 	OQS_SHA3_sha3_512_inc_absorb(&state, msg1600, 200);
 	OQS_SHA3_sha3_512_inc_finalize(hash, &state);
+
+	if (are_equal8(hash, exp1600, 64) == EXIT_FAILURE) {
+		status = EXIT_FAILURE;
+	}
+
+    /* Special test case: absorb chunks of nasty sizes */
+	clear8(hash, 200);
+    size_t absorbed = 0;
+    size_t max_seglen = OQS_SHA3_SHA3_512_RATE - 1;
+	OQS_SHA3_sha3_512_inc_ctx_reset(&state);
+    while (absorbed < 200) {
+        size_t seglen = 200 - absorbed < max_seglen ? 200 - absorbed : max_seglen;
+        OQS_SHA3_sha3_512_inc_absorb(&state, msg1600, seglen);
+        absorbed += seglen;
+    }
+	OQS_SHA3_sha3_512_inc_finalize(hash, &state);
 	OQS_SHA3_sha3_512_inc_ctx_release(&state);
 
 	if (are_equal8(hash, exp1600, 64) == EXIT_FAILURE) {
+        printf("ERROR: SHA3-512 non-rate-multiple absorption incorrect output\n");
 		status = EXIT_FAILURE;
 	}
 
--- a/tests/test_sha3.c
+++ b/tests/test_sha3.c
@@ -463,21 +463,21 @@ int sha3_512_kat_test(void) {
 		status = EXIT_FAILURE;
 	}
 
-    /* Special test case: absorb chunks of nasty sizes */
+	/* Special test case: absorb chunks of nasty sizes */
 	clear8(hash, 200);
-    size_t absorbed = 0;
-    size_t max_seglen = OQS_SHA3_SHA3_512_RATE - 1;
+	size_t absorbed = 0;
+	size_t max_seglen = OQS_SHA3_SHA3_512_RATE - 1;
 	OQS_SHA3_sha3_512_inc_ctx_reset(&state);
-    while (absorbed < 200) {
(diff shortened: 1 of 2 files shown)


xuganyu96:
@dstebila where did liboqs acquire this AVX512VL implementation? I found [this Intel white paper](https://builders.intel.com/docs/networkbuilders/post-quantum-cryptography-accelerating-open-quantum-safe-library-with-intel-avx-512-keccak-1600-implementation-1762253806.pdf) which seems to suggest an one-off collaboration. Is it possible to reach the original collaborators?

I don't know enough x86 assembly to vet the correctness of the proposed fix from Claude. Even if it passes all the unit tests, including the new ones I added, I still feel hesitant about committing unfamiliar AVX512 assembly into the code base. To me a more conservative approach would be to turn off `OQS_USE_SHA3_AVX512VL` by default and issue a warning in the same way old implementation of HQC from PQClean was turned off.

dstebila:
> @dstebila where did liboqs acquire this AVX512VL implementation? I found [this Intel white paper](https://builders.intel.com/docs/networkbuilders/post-quantum-cryptography-accelerating-open-quantum-safe-library-with-intel-avx-512-keccak-1600-implementation-1762253806.pdf) which seems to suggest an one-off collaboration. Is it possible to reach the original collaborators?
> 
> I don't know enough x86 assembly to vet the correctness of the proposed fix from Claude. Even if it passes all the unit tests, including the new ones I added, I still feel hesitant about committing unfamiliar AVX512 assembly into the code base. To me a more conservative approach would be to turn off `OQS_USE_SHA3_AVX512VL` by default and issue a warning in the same way old implementation of HQC from PQClean was turned off.

The main contributors from Intel were @mdcornu @tkanteck and @erdincoz. Maybe one of them could take a look at this?

tests/test_sha3.c
```diff
@@ -458,9 +458,26 @@ int sha3_512_kat_test(void) {
...
+	}
+
+	/* Special test case: absorb chunks of nasty sizes */
+	clear8(hash, 200);
+	size_t absorbed = 0;
+	size_t max_seglen = OQS_SHA3_SHA3_512_RATE - 1;
+	OQS_SHA3_sha3_512_inc_ctx_reset(&state);
+	while (absorbed < 200) {
+		size_t seglen = 200 - absorbed < max_seglen ? 200 - absorbed : max_seglen;
+		OQS_SHA3_sha3_512_inc_absorb(&state, msg1600, seglen);
```
mdcornu: this should be absorbing `msg1600 + absorbed` right?

mdcornu (commented):
Nice find! The fix looks good but the new test case seems to be passing only because msg1600 repeats the same byte value for the entire message. It might be a good idea to add another vector to validate the incremental calls.

tests/test_sha3.c
dstebila: You're right, I've now fixed that in [19fd5b7](https://github.com/open-quantum-safe/liboqs/pull/2442/commits/19fd5b783412aac835451cd79bab837996c52cc3).

mdcornu (approved):
Looks good 👍

tkanteck:
Thanks for the fix and for reviewing this area.
We remain fully committed to keeping the AVX512VL implementation secure and fully functional. We’d be happy to help strengthen the existing test coverage, in particular by adding more comprehensive and targeted vectors.

dstebila:
> Nice find! The fix looks good but the new test case seems to be passing only because msg1600 repeats the same byte value for the entire message. It might be a good idea to add another vector to validate the incremental calls.

Yes, I thought about a new test vector, but then apparently the repeated A3 test vector is part of some NIST test vector setup, so I didn't want to go changing it. But if you think it's worth adding another vector, that would be great.

mdcornu:
> > Nice find! The fix looks good but the new test case seems to be passing only because msg1600 repeats the same byte value for the entire message. It might be a good idea to add another vector to validate the incremental calls.
> 
> Yes, I thought about a new test vector, but then apparently the repeated A3 test vector is part of some NIST test vector setup, so I didn't want to go changing it. But if you think it's worth adding another vector, that would be great.

Sure. I can do that once this PR is merged 👍

dstebila:
@xuganyu96 I think this is okay to merge.

mdcornu:
@dstebila I have created a PR with the SHA3 test updates here: https://github.com/open-quantum-safe/liboqs/pull/2450

PR #2445, opened 2026-05-23
Update mldsa-native to v1.0.0-beta2

mkannwischer:
See the full release notes at
https://github.com/pq-code-package/mldsa-native/releases/tag/v1.0.0-beta2

The most noteworthy change is that the reduced RAM mode `MLD_CONFIG_REDUCE_RAM` now uses much less memory is no longer experimental, and fully covered by CBMC proofs. liboqs could consider offering this configuration option to users, but that is not yet done in this PR.

The large diff is primarily due to renaming of files and switching to Doxygen comment style.


xuganyu96 (approved):
For other reviewers: this pull request is smaller than it seems. Everything under `src/sig/ml_dsa/mldsa-native_xxx` is copied from upstream repository and does not require review. There are only 6 files that need attention:

- 2 files under `docs`
- `copy_from_upstream.yml`
- `src/sig/ml_dsa/CMakeLists.txt`
- `ALGORITHMS.md`
- `README.md`

bhess (approved):
LGTM, thanks @mkannwischer for the update!

PR #2449, opened 2026-05-29
Include constant-time analysis framework

pablo-gf:
This PR includes the constant-time testing framework developed throughout the course of the [LFX Mentorship project](https://github.com/orgs/open-quantum-safe/projects/10) supervised by @bhess. These are the main features of this framework:

- It is based on two dynamic analysis tools: Valgrind's memcheck with the Kyberslash patch (renamed to Valgrind-Varlat), and Clang's MemSanitizer (MemSan) to detect possible sources (warnigns) of non-constant-time behavior.
- Provides unified scripts for running constant-time checks both in the CI pipeline and locally.
- Includes a script that aggregates analysis outputs and generates visual graphs to identify and track constant-time warning trends.
- Includes several documentation files outlining the installation, setup, and usage process.

I would be happy to walk through the details of any of these features either here or during a status call. You can also find detailed guides in the documentation files I’ve added under `tests/ct_tooling` and its subdirectories. Feel free to throw any feedback or comments so that we can move this from draft to "ready for review"!

AI assistance was used to generate documentation and minor editing. The resulting code was reviewed, edited and verified for accuracy.


.github/workflows/ct-tooling.yml
```diff
@@ -0,0 +1,230 @@
+name: ct-tooling
+
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
+
+jobs:
+  interactive-inputs:
```
bhess: My understanding is that this job requires interactive input to select the algorithms. Is that correct? If so, what is the intended usage, and can it be configured to run non‑interactively (e.g., for inclusion in the weekly CI runs)?

.github/workflows/ct-tooling.yml
```diff
@@ -0,0 +1,230 @@
...
+          title: Select algorithm(s) Valgrind-Varlat CT Testing'
+          interactive: |
+            fields:
+              - label: runtime-options
+                properties:
+                  description: Choose one or more algorithms to execute valgrind-varlat constant-time testing on
+                  display: Select the algorithm(s) to execute valgrind-varlat constant-time testing on
+                  type: multiselect
+                  choices:
+                    - "BIKE-L1"
```
bhess: My concern with this list is maintainability: if the set of algorithms in liboqs changes, the list would need to be updated manually. Is there a way to generate this list dynamically based on the algorithms currently available?

tests/ct_tooling/tools/memsan/test_sig.c
```diff
@@ -0,0 +1,396 @@
+// SPDX-License-Identifier: MIT
```
bhess: The test_sig.c / test_kem.c files and the corresponding CMakeLists.txt appear to be largely copied from the versions in the /tests directory. Could we avoid duplicating these files and instead reuse the originals with additional macro definitions (e.g., to disable signature verification during ct-testing)? Or is there a specific reason why maintaining separate copies is necessary here?

tests/ct_tooling/tools/memsan/false_positives/ml_kem
```diff
@@ -0,0 +1,15 @@
+fun:PQCP_MLKEM_NATIVE_MLKEM*_C_poly_decompress_d*
```
bhess, line 1: I really like this approach, having an allowlist mechanism similar to the Valgrind one makes a lot of sense.

bhess (commented):
Thank you @pablo-gf for this work on improving the constant‑time tooling in liboqs, this is an important step towards more exhaustive ct-testing in liboqs.

I’ve left a few initial comments inline.

> I would be happy to walk through the details of any of these features either here or during a status call.

That would be excellent. @dstebila / @RodriM11: could we allocate some time in an upcoming status call for @pablo-gf to walk us through this contribution? I think a walkthrough would be very valuable.

.github/workflows/ct-tooling.yml
```diff
@@ -0,0 +1,230 @@
...
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # pin@v4
+
+      - name: Build valgrind_varlat
+        shell: bash
+        run: |
+          set -eu -o pipefail
+          INSTALL_PREFIX="$PWD/valgrind_varlat"
+
+          echo "Cloning Valgrind's source code"
+          git clone git://sourceware.org/git/valgrind.git valgrind_varlat_src> /dev/null 2>&1 || true
```
bhess: The statements with `|| true` would silently fail.

RodriM11:
> That would be excellent. @dstebila / @RodriM11: could we allocate some time in an upcoming status call for @pablo-gf to walk us through this contribution? I think a walkthrough would be very valuable.

For sure. Next week is TSC meeting, so we could do it the 9th of June, if @pablo-gf is available.

tests/ct_tooling/ct_test.sh
```diff
@@ -0,0 +1,429 @@
...
+            # Generate suppression flags for all suppression files containing false positives
+            SUP_DIR="$SCRIPT_DIR/tools/memsan/false_positives"
+            SUP_FLAGS=()
+            for f in "$SUP_DIR"/*.txt; do
+                [ -f "$f" ] || continue
+                SUP_FLAGS+=( "-fsanitize-ignorelist=$f" )
+            done
+
+            # Create backup files of the original tests files
+            mv "$LIBOQS_DIR/tests/CMakeLists.txt" "$LIBOQS_DIR/tests/CMakeLists.txt.bak"
```
bhess: This looks brittle since it makes assumptions about the liboqs file structure. IMO avoiding replacing the original files and instead work CT-specific macros in the original files would be more robust.

.github/workflows/ct-tooling.yml
pablo-gf: Yes, that is correct. The original idea was to execute these tests selectively—like when introducing a new algorithm or modifying the implementation of one. In a previous discussion with @dstebila, we noted the importance of filtering out false-positive warnings to prevent spreading false alarms. I went for the interactive-input setup so that we can execute CT tests only on those algorithms that have gone through false-positive exclusion. If we want to include this testing in CI, we can make it non-interactive, but that would require hard-coding the algorithms that went through false positive exclusion. Another option is simply running the tests on all algorithms even if that means retrieving many warnings that may be false-positives. However, this may require too many resources due to the wide range of variables handled by the framework, which was another reason why I went for the interactive-input setup. I think it comes down to what we want the ultimate purpose of this framework to be in liboqs. Are we looking for a standard CI test that runs automatically, or a specialized tool for specific occasions?

.github/workflows/ct-tooling.yml
pablo-gf: Thank you for pointing this out. We could fix this by retriving all available algorithms using the `available_kems_by_name()` and `available_sigs_by_name()` from `helpers.py`. I have included it in the latest commit, and the tests are working on my fork.

pablo-gf: Now that I think about it, I could also create a separate PR to fix this same issue for the [kem-bench.yaml](https://github.com/open-quantum-safe/liboqs/blob/f986aea60a9f3cb4055474aa212538bb0b14f1fe/.github/workflows/kem-bench.yml#L54) and [sig-bench.yaml](https://github.com/open-quantum-safe/liboqs/blob/f986aea60a9f3cb4055474aa212538bb0b14f1fe/.github/workflows/sig-bench.yml#L54) workflows with the same solution, if you find that appropiate.

tests/ct_tooling/tools/memsan/test_sig.c
pablo-gf: Yes, I think that sounds like a better approach over the preliminary copy-pasting approach. I have updated it in the last commit and the tests are passing in my fork.

.github/workflows/ct-tooling.yml
pablo-gf: Thanks! Forgot to remove these after testing. Should I also remove `> /dev/null 2>&1`? I initially kept them so that the output in GitHub Actions looks cleaner and it is easier to spot the pass/fail of tests, but we can keep the output if preferred.

tests/ct_tooling/ct_test.sh
pablo-gf: Fixed per the previous comment.

pablo-gf:
> I’ve left a few initial comments inline.

Thank you for your comments, @bhess. Let me know if you have any further ideas/suggestions.

> For sure. Next week is TSC meeting, so we could do it the 9th of June, if @pablo-gf is available.

Sounds good @RodriM11. See you next week then!

pablo-gf:
A couple of questions, @bhess:

With respect to the bot warnings: Is the "verified creator" an issue in this case? And would the pip command also require a pin by hash here?

Also, the failing job in CI seems to be a network issue during the docker pull. Any chance that could be re-executed?

bhess:
> With respect to the bot warnings: Is the "verified creator" an issue in this case? And would the pip command also require a pin by hash here?

I’m not entirely sure what to make of the “verified creator” warning, but it may have security implications. From what I can tell, this action is mainly meant for interactive use.
As you noted in the inline comment, this shouldn’t be necessary for regular CI runs if we run it non‑interactively, does that sound right?
For manual runs, could we instead trigger the action manually and pass the required parameters (the algorithm to test) explicitly?

> Also, the failing job in CI seems to be a network issue during the docker pull. Any chance that could be re-executed?

Done

pablo-gf:
> I’m not entirely sure what to make of the “verified creator” warning, but it may have security implications. From what I can tell, this action is mainly meant for interactive use. As you noted in the inline comment, this shouldn’t be necessary for regular CI runs if we run it non‑interactively, does that sound right? For manual runs, could we instead trigger the action manually and pass the required parameters (the algorithm to test) explicitly?

Sounds good, @bhess. I looked into it and we could also implement it using the functionality provided by Github Actions inputs, rather than this external party. Maybe this could simplify the whole thing for both CI tests and manual tests. 
How would you like it to be included (weekly tests, release tests...)?
What algorithms should be included? I set the ML-KEM variants as default for now.

.github/workflows/ct-tooling.yml
bhess: IMO if the output is helpful for debugging it should be kept, if not it could be sent to dev/null

bhess:
> Sounds good, @bhess. I looked into it and we could also implement it using the functionality provided by Github Actions inputs, rather than this external party. Maybe this could simplify the whole thing for both CI tests and manual tests.
> How would you like it to be included (weekly tests, release tests...)?

Thanks for the updates @pablo-gf and for looking into this. I think it would be great to rely on the [CI usage guidelines](https://github.com/open-quantum-safe/tsc/blob/main/guidelines/ci_usage.md) for this. I assume the CI tests will fall here:

> If the limits (1 hour) are exceeded, consider:
> Adding parallel jobs.
> Moving longer-running jobs to weekly workflows (up to 2 hours).
> Moving jobs running longer than 2 hours to tests only triggered manually.
> All workflows, including those only triggered manually, should be run before a release.

Ideally, we would first try to add parallel jobs so that each job stays within the 2-hour limit and can be included in the weekly workflow.
If that's not feasible, then running the tests manually (on demand or before a release) would be the fallback.
It would be great to support both use cases (manual runs and regular CI), for example by using GitHub Actions inputs as you suggested.

> What algorithms should be included? I set the ML-KEM variants as default for now.

I'd recommend adding those that are triaged.

pablo-gf:
Thanks for the comment @bhess. I have included the framework in the weekly tests, since the [latest version executes in around 1 hour](https://github.com/pablo-gf/liboqs/actions/runs/27123205117). Obviously, this would increase as more algorithms are triaged and included in the testing framework, so this is something to keep in mind for the future. At the moment, it is executing 5 jobs in parallel on all ML-KEM variants, which are those containing suppression files.

Any suggestions on the failing CI tests?

.github/workflows/ct-tooling.yml
```diff
@@ -0,0 +1,154 @@
...
+          - compiler: gcc
+            opt_flag: "-O2 -fno-vectorize"
+          - compiler: gcc
+            opt_flag: "-O3 -fno-vectorize"
+      max-parallel: 5
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # pin@v4
+
+      - name: Build valgrind_varlat
```
xuganyu96: Following up on Jun 9's status call: this can be baked into the [CI image](https://github.com/open-quantum-safe/ci-containers).

pablo-gf: Here's the PR: https://github.com/open-quantum-safe/ci-containers/pull/106

pablo-gf:
I have moved the Kyberslash patch to ci-containers so CI test should be much faster now. I also updated the documentation for false-positive suppression block handling, and merged the previous tooling with this new framework. I essentially combined everything into one single ct_tooling directory, moving the "passes" and "issues" that were already logged with the previous tooling into the new directory structure.

After talking with @bhess, we noticed that the previous tooling used a python script used for testing that is not currently used by the new framework. As a result, I have removed it. If anyone thinks differently, please let us know.

With this, I think it should be ready for review. Let me know if you have any comments or suggestions!

tests/ct_tooling/analyze_results.py
```diff
@@ -0,0 +1,258 @@
...
+        content = f.read()
+    
+    header_re = re.compile(r'^Testing (?:KEM:|SIG:|)\s*(\S+) \.\.\. (?:PASS|FAIL)', re.MULTILINE)
+    count_re = re.compile(r'Found (\d+) (?:Valgrind|MemSan|)?\s*warnings')
+
+    matches = list(header_re.finditer(content))
+    for idx, m in enumerate(matches):
+        alg_name = m.group(1)
+        # Skip SPHINCS and SLH_DSA algorithms
+        if 'SPHINCS' in alg_name or 'SLH_DSA' in alg_name:
```
baentsch: I thought Sphincs has been removed?

tests/ct_tooling/README.md
```diff
@@ -0,0 +1,171 @@
...
+./ct_test.sh valgrind-varlat clang generic -O2 all
+./ct_test.sh valgrind-varlat clang-20 auto -O3 -fno-tree-vectorize kems
+./ct_test.sh valgrind-varlat gcc-14 generic -O2 -fno-tree-vectorize Kyber768
+./ct_test.sh memsan clang generic -O1 ML-DSA-44
+./ct_test.sh memsan clang-20 auto -O3 all
+```
+
+The `local_testing_example.sh` script demonstrates how to use `ct_test.sh` to run CT tests locally across a variety of compilers, compiler versions, liboqs target builds, and optimization flags.
+
+The `ct-tooling-valgrind-varlat.yml` and `ct-tooling-memsan.yml` workflows also use `ct_test.sh` to execute CT tests in CI on user-selected algorithms (using [interactive-inputs](https://github.com/marketplace/actions/interactive-inputs)) when the workflows are manually triggered on GitHub Actions.
```
baentsch, line 55: Hmm - that begs the question which algorithms are executed when done automatically in CI. Worthwhile stating here?

.github/workflows/ct-tooling.yml
```diff
@@ -0,0 +1,125 @@
...
+  contents: read
+
+on:
+  workflow_dispatch:
+    inputs:
+      algorithms:
+        description: 'Algorithms to execute CT testing on (comma-separated)'
+        required: true
+        type: string
+        default: 'ML-KEM-512,ML-KEM-768,ML-KEM-1024'
```
baentsch, line 13: Hmm, does this mean only these algs are CT-tested now?

tests/ct_tooling/README.md
```diff
@@ -0,0 +1,171 @@
...
+Both tools are driven by the single unified `ct_test.sh` script located at the root of `tests/ct_tooling/`. The tool to execute is selected via the first argument:
+
+```bash
+./ct_test.sh <tool> <compiler_version> <liboqs_build> <opt_flags...> <input>
+```
+
+- `tool`: `valgrind-varlat` or `memsan`
+- `compiler_version`: clang, clang-20, gcc, gcc-14, ...
+- `liboqs_build`: `generic` or `auto`
+- `input`: `all`, `kems`, `sigs`, or a specific enabled algorithm variant
```
baentsch: That's perfect -- is "all" enabled by default? But then how does "all" relate to the exclusion of SLH-DSA?

tests/ct_tooling/README.md
```diff
@@ -0,0 +1,171 @@
...
+    </tr>
+  </tbody>
+</table>
+
+Once the script builds each configuration into a build folder, it calls the test execution function (`test()`) on the build folder generated.
+
+2. **Test execution**
+Then, `test()` is tasked with executing the tool's test on selected liboqs algorithms. Each tool has a different process through which it parses the tool's output to keep unique instances of the warnings, which are further detailed in their respective README files: [Valgrind-Varlat's README](tools/valgrind_varlat/README.md) and [MemSan's README](tools/memsan/README.md).
+
+Both tools enforce a warning cap of 100,000 unique warnings per algorithm run. Once the cap is reached, further warnings are suppressed. All SPHINCS and SLH-DSA signature variants are currently skipped during SIG tests due to the excessive time they require to execute.
```
baentsch: Again, does Sphincs warrant mentioning? What about stfl-sigs? They're even worse from a runtime perspective,no?

tests/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: MIT
 
 option(OQS_ENABLE_TEST_CONSTANT_TIME "Build test suite with support for Valgrind-based detection of non-constant time behaviour." OFF)
+option(OQS_ENABLE_TEST_CONSTANT_TIME_MEMSAN "Build test suite with support for MemorySanitizer-based detection of non-constant time behaviour." OFF)
```
baentsch, line 4: If we introduce a variable to enable memsan-only testing and define the previous variable to be Valgrind only testing, users get confused: Does setting "OQS_ENABLE_TEST_CONSTANT_TIME" enable _all_ CT testing or just Valgrind-based testing? Wouldn't it be more sensible to add also a OQS_ENABLE_TEST_CONSTANT_TIME_VALGRIND option and re-define OQS_ENABLE_TEST_CONSTANT_TIME to enable both OQS_ENABLE_TEST_CONSTANT_TIME_VALGRIND and OQS_ENABLE_TEST_CONSTANT_TIME_MEMSAN?

tests/test_kem.c
```diff
@@ -296,7 +300,7 @@ static OQS_STATUS kem_test_correctness(const char *method_name, bool derand) {
 		goto err;
 	}
 
-#ifndef OQS_ENABLE_TEST_CONSTANT_TIME
+#if defined(OQS_ENABLE_TEST_CONSTANT_TIME) || defined(OQS_ENABLE_TEST_CONSTANT_TIME_MEMSAN)
```
baentsch: This shows the problem of bad option definitions. IMO there should only one define be necessary here...

tests/ct_tooling/tools/valgrind_varlat/README.md
```diff
@@ -0,0 +1,169 @@
+# Valgrind-Varlat
+This directory contains the necessary files to execute Valgrind's memcheck tool on liboqs with [Daniel Bernstein's Kyberslash patches](https://kyberslash.cr.yp.to/papers.html) (valgrind-try-patch-20250805.txt and valgrind-varlat-patch-20250805.txt) and another patch including variable latency warnings in the suppression block (valgrind_varlat_sup_block.txt).
```
baentsch: Why still this reference to a patch we now have in CI? If just for completeness, what about mentioning that? If no longer needed, why not remove this text?

.github/workflows/ct-tooling.yml
baentsch: > Are we looking for a standard CI test that runs automatically, or a specialized tool for specific occasions?

IMO both - properly configurable.

baentsch (commented):
Thanks again for the work @pablo-gf ! It seems this needs a do-over to catch up with "latest" developments (alg removal, etc.). Please also see my single comments. I did not look at the underlying logic, but just how this presents itself to users. A more general question I have is how this matches up to what is stated in https://github.com/open-quantum-safe/liboqs/blob/main/SECURITY.md? In particular, how are the results represented in the algorithm datasheets as claimed?

tests/ct_tooling/analyze_results.py
pablo-gf: True,  I also removed its mention from `ct-test.sh` and the README files.

tests/ct_tooling/README.md
pablo-gf: CI currently defaults to ML-KEM variants only because they have been thoroughly examined to exclude false-positives. I did not include it in the README because I thought the default algorithms used used for CI could potentially change over time, but let me know if you think we should include it anyways.

Something to keep in mind as we go over the warnings found is that while we can expand these defaults as we want, including a large number of default algorithms would also signifcantly increase the time and resources used during testing.

.github/workflows/ct-tooling.yml
pablo-gf: For CI, yes. That is correct.

tests/ct_tooling/README.md
pablo-gf: For CI, the default option tests all ML-KEM variants as I mentioned earlier. The "all" option excludes SLH-DSA variants. I forgot to state that in the README. Thanks for noticing.

tests/ct_tooling/README.md
pablo-gf: Thanks for noticing, I removed SPHINCS mention. Currently, no tests are executed for stfl-sigs. That is something we would like to enable in the future.

tests/CMakeLists.txt
pablo-gf: Yes, I think this is a better approach too. In this case, each tool requires a mutually exclusive testing environment since they rely on different compilation and execution frameworks to flag uninitialized secret data, so we would stay with OQS_ENABLE_TEST_CONSTANT_TIME_VALGRIND and OQS_ENABLE_TEST_CONSTANT_TIME_MEMSAN only. I have updated this in the latest version and documented it in `CONFIGURE.md` too.

tests/ct_tooling/tools/valgrind_varlat/README.md
pablo-gf: Oops. Thanks for noticing! I rephrased it to simply mention that this repository uses such patches instead. I also included a link to the CI repo in case somebody wants to replicate it locally.

pablo-gf:
Thank you for the comments and suggestions @baentsch. I have fixed and answered all, but let me know if you have any more.

> A more general question I have is how this matches up to what is stated in https://github.com/open-quantum-safe/liboqs/blob/main/SECURITY.md? In particular, how are the results represented in the algorithm datasheets as claimed?

Regarding this, we could extend them to include a column for MemSan tests as well. What are your thoughts about this @bhess? How is this constant-time information reflected on the datasheet usually pipelined?

bhess:
> Regarding this, we could extend them to include a column for MemSan tests as well. What are your thoughts about this @bhess? How is this constant-time information reflected on the datasheet usually pipelined?

IMO, the column in the datasheet is mainly there to tell the reader whether the constant-time claim was verified by our tooling. Whether that verification was done using Valgrind or MemSan seems secondary, so I'm not sure there's a need for an additional column. The sync between test results and datasheets is currently done manually, if I'm not mistaken.

pablo-gf:
Sorry for the wait. Thanks for the comments @bhess. I guess in that case we wouldn't need to change anything on those datasheets, but rather manually update them as we analyze the warnings and categorize them as true positives. Let me know if you have any further comments after the issues I fixed last time @baentsch.

tests/CMakeLists.txt
bhess: I like this too. However, as far as I can see there are still `#cmakedefine` entries missing in oqsconfig.h.cmake in order for the new switches to be effective.

If I search for `OQS_ENABLE_TEST_CONSTANT_TIME` in the code base there are still multiple occurrences (e.g., in workflows and also in upstream glue code like mlkem-native), but I don't see a mechanism currently that sets this option except a remaining `#cmakedefine`.

tests/test_kem.c
```diff
@@ -296,7 +300,7 @@ static OQS_STATUS kem_test_correctness(const char *method_name, bool derand) {
 		goto err;
 	}
 
-#ifndef OQS_ENABLE_TEST_CONSTANT_TIME
+#if defined(OQS_ENABLE_TEST_CONSTANT_TIME_VALGRIND) || defined(OQS_ENABLE_TEST_CONSTANT_TIME_MEMSAN)
```
bhess, line 303: I thinks this inverts the intended semantics: on main this is skipped under CT testing, now it is enabling it. Wrapping it around a "!" should fix it.

tests/test_kem.c
```diff
@@ -352,7 +356,7 @@ static OQS_STATUS kem_test_correctness(const char *method_name, bool derand) {
 	return ret;
 }
 
-#ifdef OQS_ENABLE_TEST_CONSTANT_TIME
+#if defined(OQS_ENABLE_TEST_CONSTANT_TIME_VALGRIND) || defined(OQS_ENABLE_TEST_CONSTANT_TIME_MEMSAN)
```
bhess, line 359: same here

tests/test_kem.c
```diff
@@ -414,7 +418,7 @@ int main(int argc, char **argv) {
 		return EXIT_FAILURE;
 	}
 
-#ifdef OQS_ENABLE_TEST_CONSTANT_TIME
+#if defined(OQS_ENABLE_TEST_CONSTANT_TIME_VALGRIND) || defined(OQS_ENABLE_TEST_CONSTANT_TIME_MEMSAN)
```
bhess, line 421: same here

tests/test_kem.c
```diff
@@ -23,6 +23,10 @@
 #include "system_info.c"
 #include "test_helpers.h"
 
+#if defined(OQS_ENABLE_TEST_CONSTANT_TIME_MEMSAN)
+#include <sanitizer/msan_interface.h>
```
bhess, line 27: I think is is already included via test_helpers.h

tests/ct_tooling/ct_test.sh
```diff
@@ -0,0 +1,404 @@
...
+
+    cd "$LIBOQS_DIR"
+
+    # Execute CT tests based on the tool selected
+    case "$TOOL" in
+        valgrind-varlat)
+            # Generate suppression flags for all suppression files containing false positives
+            SUP_DIR="$SCRIPT_DIR/tools/valgrind_varlat/false_positives"
+            SUP_FLAGS=()
+            for f in "$SUP_DIR"/*.txt; do
```
bhess, line 132: The committed files seem to have no *.txt extension do they won't match here.

tests/ct_tooling/ct_test.sh
```diff
@@ -0,0 +1,404 @@
...
+                # Count and store the number of unique warnings obtained and store it
+                END {
+                    if (!terminated && warnings > 0) {
+                        print warnings > log_file ".count"
+                    } else if (!terminated) {
+                        print 0 > log_file ".count"
+                    }
+                }
+            '
+            EXIT_CODE=$?
```
bhess, line 264: Raised by AI agent:

`$?` is awk's exit status — use ${PIPESTATUS[0]} like the valgrind path; as written a functional test failure with no MSan report records as PASS. Also set MSAN_OPTIONS=halt_on_error=0 when invoking the binary (required for -fsanitize-recover to actually continue), and note awk's exit 1 at the warning cap will kill the whole script via set -e — return the count through the .count file instead of a non-zero exit.

tests/ct_tooling/ct_test.sh
```diff
@@ -0,0 +1,404 @@
...
+            echo "Architecture: ${ARCH}"
+            echo "========================================"
+            echo ""
+        } > "$SUMMARY_FILE"
+    else
+        echo "" >> "$SUMMARY_FILE"
+    fi
+
+    PASS_COUNT=0
+    FAIL_COUNT=0
```
bhess, line 120: The PASS_COUNT and FAIL_COUNT variables seem to be updated but never used.

tests/ct_tooling/README.md
```diff
@@ -0,0 +1,169 @@
...
+./ct_test.sh valgrind-varlat clang generic -O2 all
+./ct_test.sh valgrind-varlat clang-20 auto -O3 -fno-tree-vectorize kems
+./ct_test.sh valgrind-varlat gcc-14 generic -O2 -fno-tree-vectorize Kyber768
+./ct_test.sh memsan clang generic -O1 ML-DSA-44
+./ct_test.sh memsan clang-20 auto -O3 all
+```
+
+The `local_testing_example.sh` script demonstrates how to use `ct_test.sh` to run CT tests locally across a variety of compilers, compiler versions, liboqs target builds, and optimization flags.
+
+The `ct-tooling-valgrind-varlat.yml` and `ct-tooling-memsan.yml` workflows also use `ct_test.sh` to execute CT tests in CI on user-selected algorithms (using [interactive-inputs](https://github.com/marketplace/actions/interactive-inputs)) when the workflows are manually triggered on GitHub Actions.
```
bhess, line 55: The README mentions `ct-tooling-valgrind-varlat.yml` and `ct-tooling-memsan.yml`, but I see a single `ct-tooling.yml` file.

tests/ct_tooling/README.md
```diff
@@ -0,0 +1,169 @@
...
+│   └── ...
+├── <compiler>_<build_config>/
+│   ├── O0/
+│   └── ...
+```
+
+The summary file for each run includes the compiler path, compiler version, target architecture, and compilation flags used, followed by a pass/fail line for each algorithm tested.
+
+### Cathegorizing warnings and false positive handling
+This framework does not directly differentiate between false positives and issues directly. It must be manually maintained by contributors, who carefully analyze the framework's output and label them.  The label is for auditors. We assign "false positive" to a warning that is known not to be a security threat, and we store its suppression file in the corresponding file and subdirectory. We "raise an issue" about any other error, and we store the corresponding suppression file in the "issues" subdirectory. These can be found in `liboqs/tests/ct_tooling/tools/{valgrind_varlat, memsan}/{false_positives, issues}`. If you are unsure where your suppression file belongs, then save it to the "issues" subdirectory.
```
bhess, line 144: Just to double-check, as I haven't seen the "issues" subdirectories consumed anywhere in the scripts.

tests/ct_tooling/local_testing_example.sh
```diff
@@ -0,0 +1,17 @@
...
+# Iterate through the compilation options provided by the framework to execute constant time tests
+# Adjust valgrind-varlat/memsan accordingly (note that gcc uses -fno-tree-vectorize while clang uses -fno-vectorize)
+
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+for compiler_version in gcc gcc-14 clang clang-20; do
+    for liboqs_build in generic auto; do
+        for opt_flag in -O0 -O1 -O2 -O3 -Os -Ofast "-O2 -fno-tree-vectorize" "-O3 -fno-tree-vectorize"; do
```
bhess, line 13: The comment above says clang needs `-fno-vectorize` but here it feeds `-fno-tree-vectorize`.

bhess (commented):
Thanks for the iteration @pablo-gf. I did one more pass, please see the comments inline.

Additionally, I've seen that the `extended.yml` still tries to use the old `test_constant_time` tests.

tests/test_kem.c
pablo-gf: Ops, sorry about that. I fixed this using the `OQS_ENABLE_TEST_CONSTANT_TIME` variable to together with "#ifndef" as it was before, making it less wordy. This variable is set based on whether any of the `OQS_ENABLE_TEST_CONSTANT_TIME_VALGRIND` or `OQS_ENABLE_TEST_CONSTANT_TIME_MEMSAN` is set during compilation.

tests/ct_tooling/ct_test.sh
pablo-gf: I added the MemSan suggestions and they seem to work fine. I also deleted the max warning cap, since that is not a scenario that takes place anymore after only storing unique warnings (that was from early testing). Thanks for locating the pass and fail counts issue, which were used in a preliminary version of the tests, but not anymore.

tests/ct_tooling/README.md
pablo-gf: I fixed the README's discrepancies due to different versions being updated. With regards to the "issues" directory not being called, I also included it so "owned" issues are not reported either.

pablo-gf:
Sorry for the wait, finally found some time to review your comments, @bhess. Here I leave you some comments and thoughts on the changes I made besides the general typos/smaller fixes I answered to in your previous comments:

- Included the ct-tooling workflow within the extended tests, running the same way as in the weekly tests. It currently tests ML-KEM only because those have been analyzed and all false positives have been integrated into the framework. This can be extended to other families once this analysis is performed on the other algorithms.

- External glue code (like the code coming from mlkem-native) uses OQS_ENABLE_TEST_CONSTANT_TIME, so we still need to support it. The same goes for test_kem.c and test_sig.c, where certain functionalities are toggled based on whether we are in "constant-time mode." We must therefore distinguish between functionalities checking if constant-time testing is enabled in general (mlkem-native, test_kem.c, test_sig.c) and functionalities that care about how memory is poisoned (only tests/test_helpers.h). To handle this, we have two mutually exclusive user-facing variants used by tests/test_helpers.h (as previously discussed, it is not possible to poison memory and run both tools simultaneously, since running a MemSan-instrumented binary inside Valgrind is invalid). In CMake, we automatically set OQS_ENABLE_TEST_CONSTANT_TIME to ON if either of the two user variables is enabled, allowing glue code and other features to run regardless of the chosen option. I updated oqsconfig.h.cmake so that it exposes all three variables.

- Moved the constant-time options from tests/CMakeLists.txt to the root CMakeLists.txt. This ensures the OQS_ENABLE_TEST_CONSTANT_TIME flag is set when oqsconfig.h.cmake is populated, since the tests subdirectory is added after oqsconfig.h.cmake is configured in the root file.

- Removed OQS_ENABLE_TEST_CONSTANT_TIME_OPTIMIZED mode from compiler_opts.cmake, as optimization level selection is now handled directly by the user when using the new framework.

- I noticed that suppression files must exclusively use function-based rules (fun:) rather than source-line rules (src:). During testing, I observed that compiler optimizations (like -O2) often strip or shift line-level debug metadata, causing valid suppressions to fail and allowing warnings to slip through. As a thought, we could maybe simply rely on function names and wildcards (...).

tests/ct_tooling/ct_test.sh
```diff
@@ -0,0 +1,349 @@
...
+            '
+            EXIT_CODE=${PIPESTATUS[0]}
+            ERROR_COUNT=$(cat "$LOG_FILE.count" 2>/dev/null); ERROR_COUNT=${ERROR_COUNT:-0}
+            ;;
+        *)
+            echo "Unknown tool: $TOOL"; return 1
+            ;;
+    esac
+
+    if [ "$ERROR_COUNT" -gt 0 ]; then
```
bhess, line 222: Here it prints "FAIL" but CI still passes, see for example https://github.com/pablo-gf/liboqs/actions/runs/34855359149/job/104338513168.
I think it should return an error code so that the CI workflow fails.

tests/ct_tooling/ct_test.sh
```diff
@@ -0,0 +1,349 @@
...
+    echo -n "Testing $ALGORITHM ... " | tee -a "$SUMMARY_FILE"
+
+    cd "$LIBOQS_DIR"
+
+    # Execute CT tests based on the tool selected
+    case "$TOOL" in
+        valgrind-varlat)
+            # Generate suppression flags for all suppression files containing false positives
+            SUP_DIR="$SCRIPT_DIR/tools/valgrind_varlat/false_positives"
+            ISSUES_DIR="SCRIPT_DIR/tools/valgrind_varlat/issues"
```
bhess, line 128: This needs to be `$SCRIPT_DIR`, otherwise it won't be resolved correctly.

tests/ct_tooling/ct_test.sh
```diff
@@ -0,0 +1,349 @@
...
+            SUP_DIR="$SCRIPT_DIR/tools/memsan/false_positives"
+            ISSUES_DIR="$SCRIPT_DIR/tools/memsan/issues"
+            SUP_FLAGS=()
+            for f in "$SUP_DIR"/*; do
+                [ -f "$f" ] || continue
+                SUP_FLAGS+=( "-fsanitize-ignorelist=$f" )
+            done
+            for f in "$ISSUES_DIR"/*; do
+                [ -f "$f" ] || continue
+                SUP_FLAGS+=( "--suppressions=$f" )
```
bhess, line 56: Is the "suppressions" flag here correct, as this is under the memsan-case?

.github/workflows/extended.yml
```diff
@@ -7,35 +7,12 @@ on: [workflow_call, workflow_dispatch]
...
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # pin@v7.0.1
-      - name: Configure
-        run: mkdir build && cd build && cmake -GNinja ${{ matrix.CMAKE_ARGS }} .. && cmake -LA -N ..
-      - name: Build
-        run: ninja
-        working-directory: build
-      - name: Run tests
-        timeout-minutes: 360
-        run: mkdir -p tmp && SKIP_ALGS='${{ matrix.SKIP_ALGS }}' python3 -m pytest --verbose ${{ matrix.PYTEST_ARGS }}
+  ct-tooling:
```
bhess, line 10: This runs only the ML-KEM variants, whereas the previous extended tests covered other algorithms too. As discussed, gradually enabling algorithms in the new testing matrix makes sense, but I think we should preserve the existing CT coverage during that transition.

Could we either keep the previous CT job and its test runner alongside the new workflow until that coverage has been migrated, or use the new framework to test the previously covered algorithms with just the configurations we already ran? That would let us expand the new matrix gradually without losing existing coverage.

.github/workflows/ct-tooling.yml
```diff
@@ -0,0 +1,125 @@
...
+        run: |
+          set -eu -o pipefail
+          cd "$GITHUB_WORKSPACE/tests/ct_tooling"
+          chmod +x ct_test.sh
+          ./ct_test.sh valgrind-varlat ${{ matrix.compiler }} ${{ matrix.liboqs_build }} "$OPT_FLAG" "$ALGORITHM"
+
+      - name: Upload valgrind_varlat logs
+        uses: actions/upload-artifact@1746f4ab65b179e0ea60a494b83293b640dd5bba # pin@v4
+        with:
+          name: valgrind_varlat_${{ matrix.compiler }}_${{ matrix.liboqs_build }}_${{ matrix.opt_flag }}_logs
```
bhess, line 89: If I'm understanding this correctly the uploads are per algorithm. If so, does it make sense to encode the algorithm name in the log file?

.github/workflows/ct-tooling.yml
```diff
@@ -0,0 +1,125 @@
...
+        run: |
+          set -eu -o pipefail
+          cd "$GITHUB_WORKSPACE/tests/ct_tooling"
+          chmod +x ct_test.sh
+          ./ct_test.sh memsan ${{ matrix.compiler }} ${{ matrix.liboqs_build }} "$OPT_FLAG" "$ALGORITHM"
+
+      - name: Upload memsan logs
+        uses: actions/upload-artifact@1746f4ab65b179e0ea60a494b83293b640dd5bba # pin@v4
+        with:
+          name: memsan_${{ matrix.compiler }}_${{matrix.liboqs_build}}_${{ matrix.opt_flag }}_logs
```
bhess, line 123: same comment as with the valgrind-logs

bhess (commented):
Thanks for the update, Pablo, and for working through the earlier feedback. I think we're very close now. I've left a few follow-up comments on CI failure reporting and preserving existing test coverage that I think should be addressed before we can proceed.

tests/ct_tooling/ct_test.sh
pablo-gf: Perfect, I updated the workflow so that CI tests now fail if CT warnings are encountered without stopping the rest of the matrix execution and still uploading artifacts.

.github/workflows/extended.yml
pablo-gf: Thanks for noticing this. I created a new job with the pevious configurations (valgrind-varlat, gcc, generic/hashwel, O0) in the `extended.yml` workflow. This way, the extended tests maintain the previous coverage but using the new tooling. If more specific tests (memsan-based, other optimization flags, other builds, etc) are desired, users can directly use the ct-tooling.yml workflow instead.
 While testing it on CI, I noticed an exit code 1 failure on Falcon during the `haswell` extended tests. The issue is caused by the custom Kyberslash (`valgrind_varlat`) patch on Valgrind, specifically inside `mkPCastTo(1)`. Falcon on `haswell` uses 256-bit AVX2 floating-point intrinsics (`_mm256_mul_pd` / `_mm256_div_pd`), which produce `V256` vector operations. The variable-latency check attempts to cast these 256-bit inputs using `mkPCastTo(1)`, but Valgrind’s cast helper only supports scalar sizes (up to 64-bit). As a result, Valgrind seems to hit an unhandled type case, and panics with `Memcheck: the 'impossible' happened: mkPCastTo(1)`. Unfortunately, this is a problem of mixing the Kyberslash patch with Fallcon's internal functioning. I bypassed this issue by executing Falcon tests with plain valgrind in a separate job, if you see that appropiate. I included this behavior in a comment and enabled plain valgrind execution in the framework as a result.

.github/workflows/ct-tooling.yml
pablo-gf: Yes, that could be helpful.

pablo-gf:
Thanks again @bhess! I went through your suggestions. Let me know if you or anybody else have any more feedback before this can be merged.

PR #2462, opened 2026-06-09
validate signature length to prevent out of bounds read

loganaden:


RodriM11:
Hi @loganaden ! Thanks for the contribution. I think it would be best for every PR created to have a brief description of the purpose of it, just so everybody can have an initial understanding of what they are about to review, as well as the need for it.

In this case, I believe any proposed changes to the SLH-DSA implementation should be done in the original repository. You can see the information about the current implementation source in [here](https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/sig/slh_dsa.md).

loganaden:
Thanks @RodriM11. PR opened in upstream implementation.

RodriM11:
I will close this PR then.

PR #2463, opened 2026-06-10
Switch linux_x86_emulated to download Intel SDE from alternative source

### LINKED ISSUE #2461: Download Intel SDE from alternative source
A number of CI jobs failed due to <https://downloadmirror.intel.com/850782/sde-external-9.53.0-2025-03-16-lin.tar.xz> returning an empty file. This is likely because Intel's website recently deployed some kind of bot defense.

Bun chose to host a copy of SDE on their own infrastructure. I've raised a ~~[pull request](https://github.com/open-quantum-safe/www/pull/330)~~ [another pull request](https://github.com/open-quantum-safe/ci-containers/pull/103) to do the same. When it is merged, ~~I will raise another pull request here to switch SDE download to <https://openquantumsafe.org/assets/sde-external-9.53.0-2025-03-16-lin.tar.xz>~~ I will switch liboqs' GitHub Actions workflow to download from the alternative source.

xuganyu96:
This pull request fixes #2461

## Summary

A copy of Intel SDE 9.53.0 will be checked into open-quantum-safe/ci-containers. After that, `linux_x86_emulated` will switch to download the SDE from the aforementioned repository instead of <https://mirrordownload.intel.com>.

I also added a step in the job to facilitate visually checking the SHA256 checksum of the SDE.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -277,6 +277,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # pin@v4
       - name: Setup Intel SDE
         # TODO: one day we will automate the checksum. For now, just inspect visually
+        #   source: https://www.intel.com/content/www/us/en/download/684897/850782/intel-software-development-emulator.html
         run: |
           wget -O sde-external-9.53.0-2025-03-16-lin.tar.xz "$SDE_URL" && \
             sha256sum sde-external-9.53.0-2025-03-16-lin.tar.xz && \
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -271,8 +271,7 @@ jobs:
             PYTEST_ARGS: tests/test_hash.py::test_sha3 tests/test_kat.py tests/test_acvp_vectors.py
     env:
       # NOTE: https://downloadmirror.intel.com returns 202 and an empty file
-      # SDE_URL: https://downloadmirror.intel.com/850782/sde-external-9.53.0-2025-03-16-lin.tar.xz
-      SDE_URL: https://raw.githubusercontent.com/open-quantum-safe/ci-containers/refs/heads/main/intel-sde/sde-external-9.53.0-2025-03-16-lin.tar.xz
+      SDE_URL: https://api.github.com/repos/open-quantum-safe/ci-containers/contents/intel-sde/sde-external-9.53.0-2025-03-16-lin.tar.xz
     steps:
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # pin@v4
@@ -281,7 +280,7 @@ jobs:
       #   https://www.intel.com/content/www/us/en/download/684897/850782/intel-software-development-emulator.html
         run: |
           set -e && \
-          wget -O sde-external-9.53.0-2025-03-16-lin.tar.xz "$SDE_URL" && \
+          wget --header="Accept: application/vnd.github.raw+json" "$SDE_URL" && \
             sha256sum sde-external-9.53.0-2025-03-16-lin.tar.xz > lhs && \
             echo "f55138df53378198e8c0a89598351cdb3c5e7f8819e63e472b0bc179afaad34c  sde-external-9.53.0-2025-03-16-lin.tar.xz" > rhs && \
             diff lhs rhs && \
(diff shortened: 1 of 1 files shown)


RodriM11:
Thanks for the prompt response to this @xuganyu96 !

.github/workflows/linux.yml
```diff
@@ -270,14 +270,22 @@ jobs:
...
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # pin@v4
       - name: Setup Intel SDE
+      # NOTE: source of checksum:
+      #   https://www.intel.com/content/www/us/en/download/684897/850782/intel-software-development-emulator.html
         run: |
-          wget -O sde.tar.xz "$SDE_URL" && \
-            mkdir sde && tar -xf sde.tar.xz -C sde --strip-components=1 && \
+          set -e && \
+          wget -O sde-external-9.53.0-2025-03-16-lin.tar.xz "$SDE_URL" && \
```
baentsch: That's a bit surprising: Is there no way you can simply reference the file (by local path) within the container? Why `wget` ?

xuganyu96: I chose to not bake the SDE into the CI image since it is only used in one CI job.

baentsch: This approach imo has two disadvantages: 1) People cannot easily run this in a purely local CI run (with an image containing this file downloaded) and 2) This seems brittle in the sense that a GH download path is baked into the CI script that might change based on how GH stores things.

xuganyu96: I want to push back against baking the Intel SDE into the CI image.

> 1) People cannot easily run this in a purely local CI run 

Please clarify the context in which a user/developer would run this test locally. I don't think there is a "purely local" CI run, since neither liboqs source code nor some test dependencies (e.g.
[ACVP test vectors](https://github.com/open-quantum-safe/liboqs/pull/2303)) were baked into the CI image; they were fetched at container runtime instead.

> 2) This seems brittle in the sense that a GH download path is baked into the CI script

Fair point. I've moved the download URL from <https://raw.githubusercontent.com> to <https://api.github.com>, which should be more stable.

The test binary itself is relatively small (~28 MB) but still non-trivial. Baking it into the CI image means every CI job will download an extra 28MB piece of data when pulling the image, whereas fetching it at container run time means only one job needs to download this extra dependency. I guess we don't really pay extra for network bandwidth, so this is less of a practical concern than "just following best practice".

mdcornu:
@xuganyu96 @baentsch Obviously this is merged already, and is probably a more robust solution, but I just wanted to let you know that the issue with the SDE download being blocked has now been resolved. Apologies for the inconvenience.

baentsch:
> @xuganyu96 @baentsch Obviously this is merged already, and is probably a more robust solution, but I just wanted to let you know that the issue with the SDE download being blocked has now been resolved. Apologies for the inconvenience.

Thanks for letting us know @mdcornu - conceptually, at least I prefer to obtain dependencies from as close as possible to their origin -- so let me ask you whether you'd know of a reason that we should revert to pulling straight from Intel? Is it possible that updates will be done that we'd now miss with this "local copy" approach?

mdcornu:
> Thanks for letting us know @mdcornu - conceptually, at least I prefer to obtain dependencies from as close as possible to their origin -- so let me ask you whether you'd know of a reason that we should revert to pulling straight from Intel? Is it possible that updates will be done that we'd now miss with this "local copy" approach?

@baentsch 
Unfortunately I don't believe there is a "latest SDE version" URL that will always provide the most up-to-date version of SDE.
Whatever approach is used, updating to the latest version will, at minimum, require updating the download link to point to a new SDE version (latest is 10.8.0).

Realistically, updating to newer versions may only be necessary if you need to emulate ISA on newer Intel processors not supported by your current SDE version, or if you encountered a bug etc.

xuganyu96:
> Is it possible that updates will be done that we'd now miss with this "local copy" approach?

My personal preferences are to treat test dependencies in the same way as build dependencies. It might actually be a good idea to pin the version of Intel's emulator in the same way we pin compiler toolchain (e.g. Pin Clang to version 18 since some side-channel regressions can only be done with this version or newer). Admittedly Intel emulator is a small piece of the test suite and probably not worth the fuzz, but if necessary we can even implement some kind of fallback where the test reaches for Intel's website first then fall back to self-hosted copies. I think this is engineering overkill. It is probably better to just leave it be until updates are explicitly justified.

Issue #2437, opened 2026-05-19
API hygiene: OQS_STATUS enum vs mlkem-native return values + accel branches missing on *_derand variants


CrunchyJohnHaven:
**Context:** Static review pass against `main @ f28cef1` (LLM-assisted), focused on the ML-KEM dispatcher and integration layer ahead of broader PQC deployment. Filing two related observations as questions; happy to PR fixes if maintainers agree they're worth changing.

---

### 1. `OQS_STATUS` enum is not closed; mlkem-native errors `-2` / `-3` flow through unchanged

`OQS_STATUS` ([common.h:116-123](https://github.com/open-quantum-safe/liboqs/blob/main/src/common/common.h#L116-L123)) is defined as:

```c
typedef enum {
    OQS_ERROR = -1,
    OQS_SUCCESS = 0,
    OQS_EXTERNAL_LIB_ERROR_OPENSSL = 50,
} OQS_STATUS;
```

The ML-KEM dispatcher ([kem_ml_kem_768.c:81 et al.](https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/ml_kem/kem_ml_kem_768.c#L81)) does:

```c
return (OQS_STATUS) PQCP_MLKEM_NATIVE_MLKEM768_C_keypair(public_key, secret_key);
```

`PQCP_MLKEM_NATIVE_MLKEM768_C_keypair` is the preprocessor-renamed `mlk_kem_keypair` ([config_c.h:61-65](https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/ml_kem/mlkem-native_ml-kem-768_ref/integration/liboqs/config_c.h#L60-L65) namespace prefix + [kem.h:51](https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/ml_kem/mlkem-native_ml-kem-768_ref/mlkem/src/kem.h#L51) macro). That function can return `MLK_ERR_OUT_OF_MEMORY = -2` or `MLK_ERR_RNG_FAIL = -3` ([mlkem-native common.h:267,270](https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/ml_kem/mlkem-native_ml-kem-768_ref/mlkem/src/common.h#L267) and uses in [kem.c:263,270](https://github.com/open-quantum-safe/liboqs/blob/main/src/kem/ml_kem/mlkem-native_ml-kem-768_ref/mlkem/src/kem.c#L263)).

[...]

PR #2469, opened 2026-06-21
Document imported implementation status contract

Ap4sh:
## summary

Documents the algorithm-independent return-value contract for imported KEM and signature implementations

The generated dispatchers expose implementation results as `OQS_STATUS`. Imported entry points now explicitly need to return a documented status, using `OQS_SUCCESS` for success and `OQS_ERROR` for generic failures. Upstreams with native status codes need to translate them in their integration layer before those values reach liboqs

This replaces the earlier ML-KEM-specific template change with the general contract discussed in review. It gives mlkem-native a documented integration target without adding source-specific logic to the generic generators

Refs #2437

## testing

- `git diff --check`
- `python3 -m pytest --verbose tests/test_code_conventions.py` in `openquantumsafe/ci-ubuntu-latest:latest` (`265 passed, 3 skipped`)
- `cmake -B /build -GNinja -DOQS_STRICT_WARNINGS=ON -DOQS_MINIMAL_BUILD="KEM_ml_kem_768;SIG_ml_dsa_65" --warn-uninitialized .`
- `ninja -j2`
- `ninja -j2 gen_docs`
- `./scripts/run_doxygen.sh /usr/bin/doxygen ./docs/.Doxyfile /build`
- `./scripts/validate_cbom.sh` in `openquantumsafe/ci-ubuntu-latest:latest`

## checklist

* [ ] Does this PR change the input/output behaviour of a cryptographic algorithm, i.e. known answer test values? No
* [ ] Does this PR change the list of algorithms available, or otherwise change an API? No

## AI use

the idea and final direction were mine. i used AI assistance mostly for autocomplete-style edits, checking a few local paths, and helping run local checks. i reviewed the final diff and test results myself before submitting this

### DIFF
--- a/scripts/copy_from_upstream/README.md
+++ b/scripts/copy_from_upstream/README.md
@@ -92,6 +92,19 @@ implementations:
         required_flags: [avx2]
 ```
 
+### Return Value Contract
+
+The generated KEM and signature dispatchers cast implementation return values
+to `OQS_STATUS`. Every implementation entry point invoked by a generated
+dispatcher, whether named explicitly in `META.yml` or selected by convention,
+must therefore return a documented `OQS_STATUS` value. In particular,
+successful operations return `OQS_SUCCESS` and generic failures return
+`OQS_ERROR`. An upstream whose native API uses different status codes must
+translate them in the upstream integration layer before exposing the entry
+point to liboqs. An integration-specific wrapper referenced by `META.yml` is one
+way to do so. Native status values not defined by `OQS_STATUS` must not reach a
+generated dispatcher.
+
 ### Common Dependencies
 
 Shared code used by multiple implementations is declared in a commons YAML file (referenced by `common_meta_path` in the upstream entry):
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -157,6 +157,12 @@ extern "C" {
 /**
  * Represents return values from functions.
  *
+ * Functions declared to return OQS_STATUS must return one of the values below.
+ * Algorithm implementations imported into liboqs must return OQS_SUCCESS on
+ * success and OQS_ERROR for generic failures. Implementations with their own
+ * status codes must translate them in the upstream integration layer before
+ * exposing the function to liboqs.
+ *
  * Callers should compare with the symbol rather than the individual value.
  * For example,
  *


src/kem/ml_kem/kem_ml_kem_512.c
```diff
@@ -78,54 +80,54 @@ OQS_API OQS_STATUS OQS_KEM_ml_kem_512_keypair_derand(uint8_t *public_key, uint8_
...
 #endif
 }
 
 OQS_API OQS_STATUS OQS_KEM_ml_kem_512_keypair(uint8_t *public_key, uint8_t *secret_key) {
 #if defined(OQS_USE_CUPQC) && defined(OQS_ENABLE_KEM_ml_kem_512_cuda)
-	return (OQS_STATUS) cupqc_ml_kem_512_keypair(public_key, secret_key);
+	return oqs_ml_kem_native_status_to_oqs(cupqc_ml_kem_512_keypair(public_key, secret_key));
 #elif defined(OQS_USE_ICICLE) && defined(OQS_ENABLE_KEM_ml_kem_512_icicle_cuda)
-	return (OQS_STATUS) icicle_ml_kem_512_keypair(public_key, secret_key);
+	return oqs_ml_kem_native_status_to_oqs(icicle_ml_kem_512_keypair(public_key, secret_key));
```
mkannwischer: cupqc/icicle should not be using changed, right?

mkannwischer (changes requested):
In general, I agree with this change, but it should only affect mlkem-native, not the other ML-KEM implementations.

src/kem/ml_kem/kem_ml_kem_512.c
Ap4sh: good catch, updated this to only normalize the mlkem-native return paths and leave cupqc/icicle unchanged

tests/test_ml_kem_status.c
```diff
@@ -0,0 +1,34 @@
...
+		fprintf(stderr, "native status %d mapped to %d, expected %d\n",
+		        native_status, actual_status, expected_status);
+		return EXIT_FAILURE;
+	}
+
+	return EXIT_SUCCESS;
+}
+
+int main(void) {
+	int failing_statuses[] = {-1, -2, -3, 1};
```
baentsch: what are these "magic constants"? Please document (defines?).

Ap4sh: Good point, thanks. Those values come from the mlkem-native status convention, but leaving them as raw literals made the test harder to read. I pushed a small follow-up that names them locally and documents `0` as success, the named mlkem-native failures `MLK_ERR_FAIL`, `MLK_ERR_OUT_OF_MEMORY`, and `MLK_ERR_RNG_FAIL`, plus one synthetic non-zero failure case to keep the fallback mapping covered

Ap4sh:
quick ping on this one: the branch has been updated so only the `PQCP_MLKEM_NATIVE_*` paths go through the status mapper; cupqc/icicle still return with the plain `OQS_STATUS` cast. the status test constants are named and documented now too

could you take another look when you have a chance?

xuganyu96 (changes requested):
Hi @Ap4sh, thank you for identifying the issue and proposing a fix!

I have some concerns with the approach taken in this pull request. The Jinja template used in kem_scheme.c is used to generate many KEM schemes. Having algorithm-specific logic (if algorithm is ML-KEM then do something special) and/or implementation-specific logic (if implementation comes from PQCP then do something special) means having more implicit assumptions scattered in more places in this codebase, which is exactly the kind of scattershot modification I want to push back on, since it adds more places future contributors need to check when maintaining this feature.

I think the core of the problem is that there is no globally documented standard describing the meaning of non-zero return codes from `keypair|encap|decap|sign|verify`. The next best thing would be a liboqs-defined standard that all implementations within liboqs would conform to.
For this specific case, I'd suggest one of two alternatives:

- Have mlkem-native own this normalization itself, for example via a header file in `mlkem-native/integration/liboqs/`
- If that's not workable, apply a git patch that lives under liboqs but is scoped only to mlkem-native, rather than modifying the shared Jinja template

cc: @mkannwischer

mkannwischer:
> - Have mlkem-native own this normalization itself, for example via a header file in `mlkem-native/integration/liboqs/`
> 
> - If that's not workable, apply a git patch that lives under liboqs but is scoped only to mlkem-native, rather than modifying the shared Jinja template

I'm opposing to having local patches in liboqs as it defeats the purpose of maintaining the integration files upstream. 
Should OQS define some standard for the meaning of return values somewhere (ideally define some symbols for it), I have no issue with maintaining a wrapper upstream that does the normalization. 

If you point me to the documentation/header that defines it I can get that added.

> 
> 
> 
> cc: @mkannwischer

xuganyu96:
> Should OQS define some standard for the meaning of return values somewhere (ideally define some symbols for it)

I prefer this approach, too. @Ap4sh Would you be able to take on this?

Ap4sh:
> I prefer this approach, too. @Ap4sh Would you be able to take on this?

Yep. Checking this!

Ap4sh:
reworked this around the general contract instead. the generic templates are unchanged now, and the contract is documented in `src/common/common.h` plus the importer README. both META-selected and convention-selected entry points need to return a documented `OQS_STATUS` value, while native-only status values stay in the upstream integration layer

mkannwischer:
I have opened upstream issues: 
 - https://github.com/pq-code-package/mlkem-native/issues/1802
 - https://github.com/pq-code-package/mldsa-native/issues/1296
 
 I'll try to get this implement for the next releases.

baentsch (approved):
This to me now looks like a pure --and IMO very sensible-- documentation update which I'd suggest to merge as-is. Objections @mkannwischer @xuganyu96 ?

mkannwischer (approved):

xuganyu96 (approved):
I am so sorry for the late review. Will merge as soon as possible.

PR #2470, opened 2026-06-22
Update mlkem-native to v1.2.0

mkannwischer:
Bump the mlkem-native upstream from v1.1.0 to v1.2.0.

v1.2.0 adds a ppc64le (POWER8 vector) backend, but that will be added to liboqs in a follow-up PR. It is, hence, patched out here.

See the full release notes at
https://github.com/pq-code-package/mlkem-native/releases/tag/v1.2.0


bhess (approved):
Thanks for the update with the new mlkem-native release @mkannwischer .

LGTM.

After this PR lands, I will follow up with a PR fully integrating the ppc64le backend (continuing https://github.com/open-quantum-safe/liboqs/pull/2464)

xuganyu96 (approved):

PR #2471, opened 2026-06-24
Add parantheses around the use of sizeof operator

Hussain1811:
**Purpose**
This PR fixes a formatting issue with the use of the sizeof operator.
Basically surrounding the use of sizeof operator with parantheses.

**Checklist**
Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?
No. This PR just addresses a code formatting issue.

Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API?
No. This PR just addresses a code formatting issue.

### DIFF
--- a/tests/test_helpers.c
+++ b/tests/test_helpers.c
@@ -203,7 +203,7 @@ OQS_STATUS test_sig_bitflip(OQS_SIG *sig, uint8_t *message, size_t message_len, 
 			} else {
 				rc = OQS_SIG_verify(sig, message, message_len, signature, signature_len, public_key);
 			}
-			OQS_TEST_CT_DECLASSIFY(&rc, sizeof rc);
+			OQS_TEST_CT_DECLASSIFY(&rc, sizeof(rc));
 			if (rc != OQS_ERROR) {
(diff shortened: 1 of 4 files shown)


bhess (commented):
AFAIK parentheses after sizeof are optional in C if the argument is not a type. Did you encounter any formatting issues in practice because of this?

xuganyu96:
> AFAIK parentheses after sizeof are optional in C if the argument is not a type. Did you encounter any formatting issues in practice because of this?

I think it's cleaner to have parenthesis everywhere.

xuganyu96 (approved):

RodriM11:
> I think it's cleaner to have parenthesis everywhere.

However, we won't have it _everywhere_ everywhere as some native implementation code also uses sizeof in that format, right?

Based on a quick search, there are a number of algorithms that also use it that way (e.g. Classic McEliece, HQC or Falcon). There is also one use in liboqs common codebase [here](https://github.com/open-quantum-safe/liboqs/blob/main/src%2Fcommon%2Faes%2Faes_c.c).

What I mean is that certain "differences" are to be expected in between implementations, and therefore focus on these adaptations seem a bit futile, given the amount of work ahead.

PR #2473, opened 2026-06-25
tests: Mark rc as public via declassification

Hussain1811:
**Purpose**
Without the declassify call, Valgrind would report a false positive — flagging the if (rc != OQS_SUCCESS) check as secret-dependent control flow, when in reality the return code is intentionally a public value.

**Checklist**
Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  
No 
Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API?
No


bhess (approved):
Thanks for the improvements @Hussain1811, LGTM.

It seems that since we currently don't CT-test the stfl signatures and the extended tests didn't pop up as a false positive.

xuganyu96 (approved):

PR #2477, opened 2026-06-26
Add security research guidance and threat model

tob-scott-a:
This adds two Markdown files under `.github` to instruct AI agents how to more effectively work with liboqs. The intention is to improve the signal-to-noise ratio and reduce the rate of invalid LLM-generated security reports.

   * No.
   * No.

These documents were generated by Codex-5.5-Cyber as part of the Patch the Planet program.


baentsch:
Why in the world did this contribution not get reviewed & merged? IMO it'd save us from lots of AI crap. Why is it not in Ready for Review state? @tob-scott-a Thanks a million for this proposal! Would you be OK with me/@claude adapting it for oqs-provider?

baentsch (approved):

RodriM11:
> Why is it not in Ready for Review state?

It is not just because that is the way all PRs from the TrailOfBits collaboration were created. But, it does not mean that more work on the PRs is expected (i.e. we could merge it as is).

I concur that this is a useful addition.

RodriM11 (approved):

RodriM11:
thanks @xuganyu96 !

RodriM11:
> Would you be OK with me/@claude adapting it for oqs-provider?

That would be a great addition @baentsch ! Thanks for staying on top of this.

PR #2480, opened 2026-06-26
Add bounded CBMC proofs for liboqs-owned code

tob-scott-a:
This PR adds a bounded CBMC proof suite under `tests/cbmc/` for small utility and wrapper functions that are implemented directly by liboqs.

The proof set intentionally excludes imported cryptographic implementations and focuses on common API contracts at the liboqs-owned boundary. It adds:

- a reproducible `tests/cbmc/run.sh` runner
- four harnesses
- a narrow model layer for public headers and free helpers
- checked-in raw CBMC outputs and assertion coverage runs
- a manifest, per-proof summaries, a consolidated report, and an explicit list of unproven/out-of-scope segments

The included proofs cover:

- `OQS_MEM_secure_bcmp`
  - bounded functional correctness for inputs up to four bytes
  - return-value normalization to `0` or `1`
  - non-mutation of inputs

- generic signature dispatchers in `src/sig/sig.c`
  - NULL object/callback rejection
  - argument forwarding
  - one-callback invocation behavior
  - normalization of non-success callback statuses to `OQS_ERROR`

- stateful signature dispatchers in `src/sig_stfl/sig_stfl.c`
  - NULL object/callback handling
  - argument forwarding
  - wrapper-owned NULL rejection for serialize/deserialize
  - callback status propagation rules

- stateful lock/unlock helper contracts
  - NULL key rejection
  - no-op behavior when callbacks are unset
  - mutex validation
  - single callback invocation with the configured mutex

All four proofs are marked `PROVED_BOUNDED` and the checked-in runs were generated with CBMC 6.9.0. The proof suite uses bounds, pointer, pointer-overflow, division-by-zero, signed-overflow, and conversion checks with loop unwinding set to 5 and unwinding assertions enabled.

This PR does not change algorithm implementations, serialization formats, or public API behavior.

This PR does not currently add a GitHub Actions workflow. The proof suite is fully reproducible via `tests/cbmc/run.sh`, but integrating CBMC into CI requires a separate decision about how to install and pin the CBMC toolchain in the project's CI images.

No issue is resolved by this PR.

This contribution was prepared with the assistance of generative AI for drafting proof harnesses and documentation. I reviewed the resulting code, ran the proof suite, and verified the checked-in results and claims before submission.

As with the previous pull requests, this was generated with Codex-5.5-Cyber as part of the Patch The Planet program.

### DIFF
--- a/tests/cbmc/README.md
+++ b/tests/cbmc/README.md
@@ -1,0 +1,32 @@
+# CBMC Proofs For Liboqs-Owned Code
+
+This directory contains bounded proofs for small API and utility functions that
+are implemented directly by liboqs. The proof set intentionally excludes copied
+or imported cryptographic implementations.
+
+Run the suite from the repository root:
+
+```sh
+tests/cbmc/run.sh
+```
+
+The runner writes safety results and companion assertion-coverage runs to
+`tests/cbmc/raw-runs/`. The checked-in results were produced with CBMC 6.9.0 at
+commit `d95fc734f5de9bdfc80c8c65d8166b30e9cff5cd`.
+
+| Proof | Owned source | Status |
+| --- | --- | --- |
+| `mem_secure_bcmp` | `src/common/common.c` | `PROVED_BOUNDED` |
+| `sig_status_normalization` | `src/sig/sig.c` | `PROVED_BOUNDED` |
+| `sig_stfl_dispatch_contract` | `src/sig_stfl/sig_stfl.c` | `PROVED_BOUNDED` |
+| `sig_stfl_lock_contract` | `src/sig_stfl/sig_stfl.c` | `PROVED_BOUNDED` |
+
+The model header disables every imported algorithm implementation while keeping
+the public `OQS_SIG` and full `OQS_SIG_STFL` layouts visible. This lets CBMC
+compile the real liboqs-owned translation units and reason about their wrapper
+control flow without claiming anything about algorithm internals.
+
+Every safety run uses bounds, pointer, pointer-overflow, division-by-zero,
+signed-overflow, and conversion checks with loop unwinding set to 5 and
+unwinding assertions enabled. The companion coverage runs are not proof
+statuses; they are kept to show that the harness assertions are reachable.
--- a/tests/cbmc/harnesses/mem_secure_bcmp_harness.c
+++ b/tests/cbmc/harnesses/mem_secure_bcmp_harness.c
@@ -1,0 +1,51 @@
+// SPDX-License-Identifier: MIT
+
+#include <stddef.h>
+#include <stdint.h>
+
+#include <oqs/oqs.h>
+
+#define OQS_CBMC_BCMP_MAX_LEN 4
+
+extern unsigned char nondet_uchar(void);
+extern size_t nondet_size_t(void);
+
+int main(void) {
+	uint8_t a[OQS_CBMC_BCMP_MAX_LEN];
+	uint8_t b[OQS_CBMC_BCMP_MAX_LEN];
+	uint8_t a_before[OQS_CBMC_BCMP_MAX_LEN];
+	uint8_t b_before[OQS_CBMC_BCMP_MAX_LEN];
+	size_t len = nondet_size_t();
+	int expected = 0;
+
+	__CPROVER_assume(len <= OQS_CBMC_BCMP_MAX_LEN);
+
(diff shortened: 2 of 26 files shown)


hanno-becker:
Seeing that @baentsch tagged this with `Help Wanted`, allowing myself to chime in: I would advise to have CBMC in CI from the start if you are serious about it, and make sure that you have an automated setup flow that works both in CI and for the maintainers (at least). I would otherwise expect maintenance and reproducibility issues. In mlx-native, we use `nix`; you find everything online. 

I would also suggest that you think about what you want to achieve with CBMC: What properties are you hoping to prove? What scope are you aiming for?

baentsch:
Thanks @hanno-becker for your feedback and thoughts. 

Unfortunately, time to think about new things doesn't really seem to be available any more: The OQS team is losing more and more active participation and the previously active contributors divert more and more of their attention to non-OQS stuff: The net in this case was that the OQS TSC couldn't decide to even discuss what to do with this proposal (and its siblings #2478, #2479), let alone even to decide to do any work on it -- i.e., I'm afraid OQS is very far away indeed from discussing your concrete --imo sensible-- questions.

Our prime focus is on integrating the round 3 sig algs to return OQS to the research utility it once had. Unfortunately, there's also lots of effort still spent on non-research topics (following APIs for done-and-dealt with already standardized algs, for example or wondering why the OQS CI is unwieldy, or re-factoring stuff) causing work and imo distractions from this focus. And I'm afraid I currently have to count CBMC into that: If someone new were looking into that, that were great and welcome (hence the tag), but as long as we don't even get reviews for the many open code-carrying and un-attended PRs done, I don't see us taking on things that require additional thinking and code writing.

hanno-becker:
In that case, I don't think it makes sense to integrate CBMC. It adds value, but it is also a maintenance burden. You need someone to be willing and available to dig into (sometimes random) issues when they arise. If that is unrealistic -- which seems to be the case -- my own advise would be to hold off (for now).

dstebila:
> In that case, I don't think it makes sense to integrate CBMC. It adds value, but it is also a maintenance burden. You need someone to be willing and available to dig into (sometimes random) issues when they arise. If that is unrealistic -- which seems to be the case -- my own advise would be to hold off (for now).

Thanks for the thoughts Hanno. In general having high assurance is a great goal, but I think the concerns about maintenance are valid here. I don't think anyone involved in OQS on an ongoing basis at the moment has CBMC experience, so we'd have to rely on occasional contributors or LLMs to maintain this code.

RodriM11:
Thank you @hanno-becker and @baentsch for chiming in. I concur with @dstebila 's opinion: I believe CBMC is a highly interesting goal and definitively something that we should pursue.

> The net in this case was that the OQS TSC couldn't decide to even discuss what to do with this proposal

We did [agree](https://github.com/open-quantum-safe/tsc/blob/main/meetings/2026-07-07/minutes.md#security-response-process) that the TrailOfBits PRs were to have less priority right now that the round 3 algorithm inclusion. It is true that our current bandwidth is limited and therefore prioritizations are expected. 

However, these are inclusion that IMO I would expect to be able to start considering once we are able to advance with our currently planned efforts, or if someone new is willing to help in this endeavor.

baentsch:
After this general agreement that OQS doesn't have the bandwidth for this right now, can I suggest to close this PR and change it into an issue (as per [suggestion](https://github.com/open-quantum-safe/tsc/discussions/321))? Benefit: We don't rehash things repeatedly when regularly reviewing the PR backlog.

dstebila:
Thanks very much for this contribution @tob-scott-a, and thanks @hanno-becker for the expert input.

Following the discussion above, I'm closing this PR and have opened #2504 to track the idea, per the approach proposed in [tsc#321](https://github.com/open-quantum-safe/tsc/discussions/321) for `help wanted` PRs that don't yet have a reviewer or implementer to take them on. This keeps the proposal and the open design questions visible without the PR backlog needing to re-examine it at every review.

To be clear, this is a "not now" rather than a "no": the concern is maintenance ownership, not the value of the work. Nobody currently active in OQS has CBMC experience, and our near-term focus is the round 3 signature algorithms.

Nothing is lost by closing — the branch remains available via `git fetch origin refs/pull/2480/head` or the [patch](https://github.com/open-quantum-safe/liboqs/pull/2480.patch), and #2504 records what was proposed and what would need to be settled before picking it up again.

PR #2485, opened 2026-07-01
Update Kyber algorithm status from Community to Supported

loganaden:


RodriM11:
Hi @loganaden ! The PR is not correct, as there are other elements to be updated to document algorithm support. I am waiting on some answers from the remaining teams and will create a PR to update them together. But thanks for contributing. On your behalf, it would be greatly appreciated a statement of support for the "Actively Maintained" status you committed to. I don't know exactly where that statement would make the most logic (maybe just a document under a repository in your GitHub account or something like that would suffice)

baentsch:
Thanks for the stated willingness to support this algorithm @loganaden! This should be very clear-cut, though, e.g., along the lines done in https://github.com/open-quantum-safe/oqs-demos/ where there are specific names/GH handles. In oqs-demos the obligation is comparatively "light" (along the lines of "will look at a problem when you have time") but for `liboqs` I'd argue a change from "Community" (i.e., clearly "unlikely someone will quickly handle") to "Supported" must be accompanied by a (documented) name (and commitment): See https://github.com/open-quantum-safe/liboqs/blob/main/ALGORITHMS.md#upstream-maintenance

> I don't know exactly where that statement would make the most logic (maybe just a document under a repository in your GitHub account or something like that would suffice)

fwiw, I'd suggest this statement to be within the `liboqs` algorithm support page (if too long, maybe by link) -- but in any case, it should be a (new) line item in ALGORITHMS.md: Contact and statement of support (level/commitment) -- for each alg: That's where a user should be able to judge _in one place_ the state of code (incl. support/maintenance) for each algorithm.

xuganyu96 (changes requested):
In addition to @RodriM11 and @baentsch 's input, I also want to add that this table is generated from `docs/algorithms/kem/kyber.yml` under the key `oqs-support-tier`. You need to modify the source of truth there, then run `copy_from_upstream` to re-generate the markdown.

I am also not clear on how you plan to answer maintenance request. Will there be a fork of `pq-crystals/kyber`? If yes, then I think a integration datasheet under that fork `cyberstorm/pq-crystals-kyber/integration/liboqs/kyber_<512|768|1024>.yml` could be an appropriate place to declare the commitment to support. What do you think @RodriM11?

RodriM11:
> I am also not clear on how you plan to answer maintenance request. Will there be a fork of pq-crystals/kyber? If yes, then I think a integration datasheet under that fork cyberstorm/pq-crystals-kyber/integration/liboqs/kyber_<512|768|1024>.yml could be an appropriate place to declare the commitment to support. What do you think @RodriM11?

I completely agree, that would be ideal. 

In any case, I have to draft another PR to update the commitment status from (some of) the algorithms remaining, and in there we will reference this commitment from @loganaden , including a statement of support that should not be a part of the `liboqs` own repository. 

But, the Tier of the algorithm is not an object of modification, so I think we could proceed to close this PR.

xuganyu96:
> In any case, I have to draft another PR to update the commitment status from (some of) the algorithms remaining, and in there we will reference this commitment from @loganaden , including a statement of support that should not be a part of the `liboqs` own repository.

@loganaden feel free to close this pull request if you agree. Thank you.

loganaden:
closing.

Issue #2179, opened 2025-06-25
Check for memset in test_code_conventions.py


SWilson4:
All existing uses of memset would need to be either (1) allow-listed with `// IGNORE memory-check` or (2) replaced with OQS_MEM_cleanse.

PR #2488, opened 2026-07-04
Fix comment and string handling in the memory-function check

mohamadm0meni:
`test_memory_functions` tracked comments line by line, which had holes: a line containing both `/*` and `*/` was skipped entirely (`free(p); /* cleanup */` passed the check), a `/*` inside a string literal flipped the tracker into comment state and hid everything up to the next `*/`, and `free (p)` with a space before the parenthesis never matched. It also flagged names that only appear in comments or strings, which would force pointless `IGNORE` annotations.

Replaced that with a single pass that blanks out comments and string/character literals (newlines kept, so line numbers stay valid) and searches the result with one combined pattern. `// IGNORE memory-check` behaves as before. The `\b_<func>\(` exclusion is dropped: `\b` never matches between two word characters, so it had no effect. The checker is factored into its own function with parametrized unit tests; those run in the existing code conventions CI job. `src/` is still clean under the stricter parser, no new annotations needed.

Doesn't close an issue, but related to #2179: adding `memset` to this check needs a parser that doesn't miss code next to comments or flag names inside them.


mohamadm0meni:
Hello,
As this is my first PR here, the CI workflows are pending manual approval. Could a maintainer please trigger them?
I will mark the PR as ready for review once the tests pass.
Thank you.

xuganyu96:
Hi @mohamadm0meni! Thank you for the contribution.

I appreciate your efforts to make the existing checker more robust. However, the proposed regular expressions still miss cases that can occur in practice, such as macros that expand to non-OQS memory functions or calls through function pointers. They can also produce false positives for struct members named free or malloc, even when those members point to approved OQS functions. Regular expressions are useful for a lightweight convention check, but they cannot
reliably determine which function is being called.

Please consider syntax-aware tools such as Tree-sitter, or compiler-based tools such as clang-query, clang-tidy, or CodeQL. These offer different levels of analysis; resolving function-pointer targets may require additional work. We could use the results to build an explicit inventory of memory-function calls, classifying reviewed calls as “OQS-approved API,” “explicitly allowed exception,” or “non-OQS-approved API,” while tracking unresolved calls separately. This would give us a clearer basis for detecting regressions, provided we also account for code excluded by particular build configurations. You’re welcome to open a follow-up pull request incorporating these suggestions. Thank you!

Issue #2474, opened 2026-06-25
[basic-downstream / trigger-downstream-ci] CI job does not appear to handle failures correctly


jplomas:
### Description

A recent push to liboqs/main caused oqs-provider CI to start failing.

The basic-downstream / trigger-downstream-ci workflow seems designed to prevent this, but appears to handle errors incorrectly, resulting in a successful run regardless as to whether the triggered CI passes or fails.

For example:

https://github.com/open-quantum-safe/liboqs/actions/runs/27785308182/job/82220378025

The `Trigger oqs-provider job` reported as successful despite reporting an error state / 201
```
{
  "number" : 2205,
  "state" : "errored",
  "id" : "2a5b2378-8f6f-4302-98dd-bbbf5a916ff0",
  "created_at" : "2026-06-18T19:50:56.723Z"
}
201
```
c.f. the other jobs which exited with 204 and no reported errors

See also discussion on `oqs-provider`: https://github.com/open-quantum-safe/oqs-provider/issues/772#issuecomment-4787921496

### Expected behaviour

It looks like this job is set to guard against triggering downstream CI failures

### liboqs version

main

jplomas:
Actually, perhaps just clarification in CI.md, skipping the downstream-basic.yml checks on forks and deciding where liboqs and provider incompatibilities surface is more pragmatic.

baentsch:
> Actually, perhaps just clarification in CI.md, skipping the downstream-basic.yml checks on forks and deciding where liboqs and provider incompatibilities surface is more pragmatic.

For forks, this is a pragmatic and OKish solution. For this project it arguably is not: `oqsprovider` now begins to collect PRs being blocked on a `liboqs` change that has been labelled as breaking the `oqsprovider` on [April 14](https://github.com/open-quantum-safe/liboqs/pull/2407). The problem surfaced two months later with #2407 merging without the update having been done. IMO at latest at that stage, CI should have stopped that merge, so I completely agree with this issue needing a fix -- and I think the fix needs to be code, not reliance on a person scrambling to fix something where the problem was obvious for many weeks and months.

jplomas:
Thinking through a few options:
(1) If the downstream builds fail, this could file an issue on the relevant downstream repo: better surfacing of issues than the present setup, but this is not a gate to downstream-breaking code making it into liboqs.
[...]

PR #2489, opened 2026-07-05
ci: fix downstream-basic trigger logic (#2474)

mohamadm0meni:
Partially addresses #2474.
Replaces the CircleCI-based check for oqs-provider (decommissioned) and the substring/status-code check for the rest with a shared script, scripts/trigger_downstream_ci.py, that dispatches each downstream CI and polls the actual run to a terminal state before reporting success.
oqs-provider is temporarily excluded: none of its current workflows expose a repository_dispatch or workflow_dispatch trigger, so it cannot be wired in from this repo alone. A follow-up PR to oqs-provider is needed first.
Whether downstream compatibility should block liboqs PRs (vs. today's post-merge/manual checking) is a separate policy question, out of scope here.


mohamadm0meni:
Hello,

The CI workflows are currently pending manual approval. Could a maintainer please trigger them? 

I will mark the PR as ready for review once the tests pass.

Thank you in advance.

baentsch:
> The CI workflows are currently pending manual approval. Could a maintainer please trigger them?

Done & passed, but I didn't see where the new code ran.

Also tagging @jplomas to take a look at this PR as this pertains to #2474

jplomas:
If this works as it appears, it should be a small improvement over the existing setup but not a full fix for #2474.

I say "if" because the current CI will not confirm this until `commit-to-main` runs. `downstream-basic` is not triggered on PRs, which perhaps is part of the broader issue.

For liboqs/oqs-provider compatibility, I think we still need an explicit policy decision:

1. If downstream compatibility should block liboqs PRs, then CI needs to build/test oqs-provider against the liboqs PR merge SHA, either directly inside liboqs CI or by triggering downstream with a liboqs ref and waiting/polling for the downstream result.

2. If downstream compatibility is advisory/post-merge only, which I think is the implication in the current setup and docs, failures should create a more visible notification/issue.

I would suggest that the policy decision be made first, then a full fix be proposed.

xuganyu96 (changes requested):
Hi @mohamadm0meni, thank you for your contribution!

The intention behind these jobs is to check that changes in the upstream project (i.e. liboq) do not break downstream projects (i.e. oqs-provider, openssh, boringssl, python/cpp/go/rust wrappers). Based on my reading, the problem with the current setup is that it only triggers the downstream projects' GitHub Actions workflows but do not check their results. Instead, the current checking logic only "intends to ensure" that the workflows have been triggered. Even that is not implemented correctly. For example, the URL for triggering oqs-provider's CI pipeline is simply wrong: we don't use Circle CI anymore and have moved everything onto GitHub Actions.

Please consider the following points when writing a complete solution:

- Do we want to interact with GitHub API (`https://api.github.com`) or use some wrappers like `gh` CLI? (I personally prefer interacting with the HTTP API directly)
- Do we want to isolate the logic into a separate script file living under `scripts/` instead of writing inline bash scripts in workflow YAML? The former carries an additional advantage of making local runs easier.
- How to check the status of the triggered workflow run? Check GitHub API's documentation: what are some terminal states? Also consider setting a timeout so this job doesn't run indefinitely. Given these requirements, we might even want to consider using a richer language like Python instead of bash scripting

References:

-  [Get a workflow run](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2026-03-10#get-a-workflow-run)

cc: @baentsch @RodriM11 @dstebila @jplomas

mohamadm0meni:
@xuganyu96  Thanks for the review. Updated to use the HTTP API directly via a stdlib-only Python script in scripts/, testable locally via --api-base. It now dispatches, locates the spawned run(s), and polls to a terminal status (with a 45-minute timeout), failing on anything other than success/neutral/skipped. CircleCI is removed entirely.
One open item: oqs-provider has no remote trigger today, so it's excluded pending a companion PR there — happy to send that next if you agree with the approach.

xuganyu96 (changes requested):
Sorry for the late review. I appreciate the delegation of non-trivial logic to
a Python script, but after reading the script I still feel that it is a lot to
maintain. I am aware that I previously expressed a personal preference for
interacting with the raw HTTP REST API, but this opinion has changed as I became
aware that GitHub CLI `gh` is indeed a first-party tool and that it is already installed
on all GitHub-hosted runners. Could you explore implementing the same "dispatch
and poll" logic with `gh` and see if it is indeed easier to work with?

We now have a oqs-provider release test living under `downstream-release.yml`.
You can skip this test in your work.

Thank you.

PR #2492, opened 2026-07-14
Fix integer overflow in OQS_MEM_aligned_alloc size calculation

Vishnu2707:
Both the OpenSSL and fallback branches of OQS_MEM_aligned_alloc compute size + offset with no overflow check before passing the result to the underlying allocator. When size is near SIZE_MAX, this wraps to a small value, so the allocator returns a real but far too small buffer while the caller still treats the returned pointer as valid for the full requested size. Add a guard before each allocation call that rejects the request instead of letting the addition wrap.

Confirmed with ASan: before this change, an unaligned near-SIZE_MAX size drives CRYPTO_malloc/malloc down to an 8-byte (OpenSSL branch) or 1-byte (fallback branch) allocation, and the pointer-alignment bookkeeping inside OQS_MEM_aligned_alloc itself immediately writes past it (heap-buffer-overflow at common.c:329 and common.c:386 respectively). After the fix both cases return NULL cleanly.

Also adds the alignment/size validation that the fallback branch already performs (power-of-two alignment >= sizeof(void*), size a multiple of alignment) to the OpenSSL branch, which was missing it entirely.

See GHSA-54w9-9rvw-3fg4 for the original report.

### DIFF
--- a/src/common/common.c
+++ b/src/common/common.c
@@ -312,7 +312,15 @@ void *OQS_MEM_aligned_alloc(size_t alignment, size_t size) {
 	if (!size) {
 		return NULL;
 	}
+	// Check alignment (power of 2, and >= sizeof(void*)) and size (multiple of alignment)
+	if (alignment & (alignment - 1) || size & (alignment - 1) || alignment < sizeof(void *)) {
+		errno = EINVAL;
+		return NULL;
+	}
 	const size_t offset = alignment - 1 + sizeof(uint8_t);
+	if (offset > SIZE_MAX - size) {
+		return NULL;
+	}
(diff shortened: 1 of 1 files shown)


dstebila (approved):

dstebila:
Thanks @Vishnu2707!

Vishnu2707:
@baentsch , this PR basically addresses the point mentioned in the [GHSA-54w9-9rvw-3fg4](https://github.com/open-quantum-safe/liboqs/security/advisories/GHSA-54w9-9rvw-3fg4).

baentsch:
> @baentsch , this PR basically addresses the point mentioned in the [GHSA-54w9-9rvw-3fg4](https://github.com/open-quantum-safe/liboqs/security/advisories/GHSA-54w9-9rvw-3fg4).

Much appreciated @Vishnu2707 : Thanks for the contribution!

PR #2493, opened 2026-07-15
Add CI runtime report generator

baentsch:
This adds a scheduled (weekly) workflow that reports how much CI compute
time is spent per algorithm family, to give maintainers ongoing
visibility into CI cost distribution as new PQC families are added, in line and
with the goal to provide tool support for https://github.com/open-quantum-safe/tsc/blob/main/guidelines/ci_usage.md.

Testing performed

Every subcommand of ci_family_report.py, and the full three-job
workflow (generate-family-matrix -> profile-family matrix ->
aggregate-and-report, including real artifact upload/download between
jobs), have been run successfully end to end locally against the real
liboqs source and the real GitHub Actions API, using
[act](https://github.com/nektos/act) (see scripts/README.md for how).
This wasn't run inside an actual GitHub Actions runner, since
workflow_dispatch only becomes available once a workflow file is merged
to the default branch.

Related follow-up

While building this, the report surfaced that SLH-DSA's dedicated CI jobs
take disproportionately long (70-90 min vs. ~15 min baseline on the same
platform, for identical test scope). Analysis and possible fix will be filed separately.

* [ no Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged. Also, make sure to update the list of algorithms in the continuous benchmarking files: .github/workflows/kem-bench.yml and sig-bench.yml)

This has been created with substantial AI assistance (@claude) with sanity, though not line-by-line, checking by the main author. Risk of errors is perceived as low as this is a report-only generator, separate from any user-level/runtime code.

### DIFF
--- a/.github/workflows/ci-family-usage-report.yml
+++ b/.github/workflows/ci-family-usage-report.yml
@@ -1,0 +1,146 @@
+name: CI Family Usage Report
+
+# Periodically (default: weekly) generates a report showing how much
+# GitHub Actions CI compute time is spent per liboqs algorithm family.
+# All logic lives in a single script: scripts/ci_family_report.py
+# (subcommands: generate-definitions, build-matrix, measure-family,
+# merge, fetch-ci-data, report). See that script's docstring and
+# scripts/README.md for details.
+
+on:
+  schedule:
+    - cron: '0 3 * * 1'   # every Monday at 03:00 UTC
+  workflow_dispatch:
+    inputs:
+      n_ci_runs:
+        description: 'Number of real CI runs to analyze for timing data'
+        default: '5'
+        required: false
+
+permissions:
+  contents: read
+  actions: read
+
+env:
+  N_CI_RUNS: ${{ github.event.inputs.n_ci_runs || '5' }}
+
+jobs:
+  generate-family-matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.gen.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Generate family definitions and build matrix
+        id: gen
+        run: |
+          python3 scripts/ci_family_report.py generate-definitions . --out family_definitions.json
+          python3 scripts/ci_family_report.py build-matrix family_definitions.json --github-output
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: family-definitions
+          path: family_definitions.json
+          retention-days: 14
+
+  profile-family:
+    needs: generate-family-matrix
+    runs-on: ubuntu-latest
+    # SLH-DSA (156 parameter variants including tests) can take noticeably
+    # longer than the other families -- hence the generous timeout.
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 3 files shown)


dstebila:
Thanks for taking this on -- a good step towards taming our CI usage, a long-standing goal. If I understand the PR correctly, it adds the report to the repo as a file. If so that seems a little suboptimal, as most people checking out the repo wouldn't have any interest in that. Might it be possible to add the report as an issue or a comment on a discussion? Or if it's going to be archived as a file in a repo, would putting it in e.g. the ci-containers repo be more out of the way?

baentsch:
> Might it be possible to add the report as an issue or a comment on a discussion? 

Agree, the optional persistent repo store is overkill; PR changed to keep track in an issue. But I assume the weekly run report is going to be the primary "go-to" place to see where CI "goes overboard".

dstebila (dismissed):

xuganyu96 (dismissed):

xuganyu96 (commented):
I would like to propose an alternative architecture for aggregating test statistics using two features that are already built into pytest:

- Pytest can output JUnitXML file, which includes each test's duration
- Pytest can record property using `record_property` fixture, which we can use to record the algorithm family and algorithm name of each test case

Going this route means we need some small modifications to existing pytest test cases (i.e. adding `record_property`) and an one-line change to how we invoke pytest. We will also need to find appropriate local tools and GitHub Actions workflow to process JUnitXML. I feel optimistic about finding some off-the-shelf things that are just ready to go and don't require maintaining another 600-line Python script and 160-line custom-built GitHub Action workflow.

*References*:

- [Managing pytest output](https://docs.pytest.org/en/stable/how-to/output.html)
- [action-junit-report](https://github.com/marketplace/actions/junit-report-action)
- [Junit test dashborad](https://github.com/marketplace/actions/junit-test-dashboard)

baentsch:
> I feel optimistic about finding some off-the-shelf things that are just ready to go and don't require maintaining another 600-line Python script and 160-line custom-built GitHub Action workflow.

You're absolutely right @xuganyu96 that more elegant and off-the-shelf-components solutions can be created, but designing and developing such solution takes (more) time (and if only to learn about the COTS pieces). This time has not been found so far and in consequence, CI did grow unchecked, becoming an impediment to the dev cycle and violating [our own guidelines](https://github.com/open-quantum-safe/tsc/blob/main/guidelines/ci_usage.md) by factors/>>100%. 

Regarding maintenance concerns: These are conceptually also agreed, but apply to a problem (monitoring CI use) that has not been tended to at all so far: Thus I'd ask whether an unmaintained solution to solve a problem is better than no(t finding time to develop a perfect) solution to that problem?

We can further always revert this commit without loss to the project as it has no interaction points with any other component, e.g., at the moment that an off-the-shelf-component solution, maybe with a nice UI, has been developed.

dstebila (dismissed):

xuganyu96 (approved):

PR #2494, opened 2026-07-15
release triggering for supported and ready downstreams and fix CI on forks

jplomas:
Proof-in-doing of the concept discussed in the liboqs status meeting Tue 14 Jul 2026: the uses liboqs-go as an example.

A release of `liboqs` triggers release(s) of [tested and ready] downstream language wrappers.

This PR is in parallel with the PR on `liboqs-go`, which contains the release job triggered: https://github.com/open-quantum-safe/liboqs-go/pull/55

A release job is only _dispatched_ on `liboqs` release, so there are no release-blocking failures if the downstream repo is not ready or is breaking. However, if any necessary changes are in situ, it will allow the language wrapper release to be made at the same time as the main liboqs release.

The `downstream-release` workflow is renamed `downstream-release-tests` to better reflect the “readiness” functionality, while `downstream-release` now handles the actual downstream release, as its name implies.

Also fixes a longstanding issue preventing all CI passing on forks: there is now a repository check to skip downstream dispatches where repository secrets are absent.

Claude assistance with greps and hash pins; code review by corerabbit.ai 

Motivation:

liboqs-go wrapper releases stalled between 0.10 and 0.15, though only the latter release contained any breaking changes.  Similar CI changes could rapidly speed up the deployment of other language wrappers.

### DIFF
--- a/.github/workflows/downstream-basic.yml
+++ b/.github/workflows/downstream-basic.yml
@@ -8,6 +8,8 @@ on: [workflow_call, workflow_dispatch]
 jobs:
 
   trigger-downstream-ci:
+    # Dispatch secrets only exist upstream; skip in forks.
+    if: github.repository == 'open-quantum-safe/liboqs'
     runs-on: ubuntu-latest
     steps:
       - name: Trigger OQS-BoringSSL CI
--- a/.github/workflows/downstream-release-tests.yml
+++ b/.github/workflows/downstream-release-tests.yml
@@ -1,0 +1,30 @@
+name: Downstream release tests
+
+permissions:
+  contents: read
+
+on: [workflow_call, workflow_dispatch]
+
+# Trigger oqs-provider release tests.
+# When triggered by a release (see release.yml), the liboqs release tag and the provider "<release tag>-tracker" branch are used.
+# When triggered by a commit message (see filter.yml), the triggering liboqs branch and the provider "<liboqs branch>-tracker" branch are used.
+# If the tracker branch does not exist, the downstream pipeline should detect it and run on the main branch instead.
+
+jobs:
+  oqs-provider-release-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout release tests script
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # pin@v4
+        with:
+          sparse-checkout: |
+            scripts/provider-test-trigger.sh
+          sparse-checkout-cone-mode: false
+      - name: Trigger oqs-provider release tests
+        run: |
+          CURL_FLAGS="--silent --write-out \n%{response_code}\n" \
+          ACCESS_TOKEN="${{ secrets.OQSBOT_GITHUB_ACTIONS }}" \
+          LIBOQS_REF="${{ github.ref_name }}" \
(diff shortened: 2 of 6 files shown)


baentsch:
The job name renaming makes perfect sense: One to test readiness (1), one to do release(s) (2). I wonder though, whether it wouldn't be sensible to first have automated readiness tests for all downstreams before having automated releases -- or do I misunderstand the logic of this PR to be (1) for `oqsprovider` (as before) and (2) for `liboqs-go` (new) only, i.e., no (1) for `liboqs-go`?

jplomas:
The status quo of downstream readiness is relatively maintained: (1) is for PR authors to see if their changes will break downstream, (2) is for release managers to automatically release downstream which are ready or those which do not break.  The errors for both (1) and (2) surface in the affected downstream repository.

This PR is minimal -- (1) for `liboqs-go` and all the language wrappers would be very useful imo -- as the overall approach has yet to be green-lit.  

From the status meeting, a summary (which might be my misinterpretation) is that downstream readiness cannot slow down a liboqs release.

The work regarding #2474 and #2489 aligns more with the:
> automated readiness tests for all downstreams before having automated releases
design, but this is contrary to what I understand to be desirable.

From a "fresh look" perspective, I think I would favour defining critical (supported) downstreams, and building their `main` as a gating CI job on all liboqs PRs and pushes, but I appreciate there does not appear to be the appetite for this approach.

baentsch:
> From a "fresh look" perspective, I think I would favour defining critical (supported) downstreams, and building their main as a gating CI job on all liboqs PRs and pushes, but I appreciate there does not appear to be the appetite for this approach.

I'd personally also favour that from the perspective of being able to more easily/quickly do releases. But this would mean _very_ (re)active downstream maintenance (which doesn't exist) and substantially higher CI costs (contradicting https://github.com/open-quantum-safe/tsc/blob/main/guidelines/ci_usage.md).

So what about this compromise: Run downstream release readiness tests on all "critical" downstreams not at push but only at a merge highlighting problems thus possibly unearthed (and not discovered in PR review) in a (newly, automatically generated) issue (instead of blocking/failing the PR)?

jplomas:
That would seem a reasonable compromise.

Notably, the "pull from downstream and test in liboqs CI" is different from the approach in #2474

xuganyu96 (changes requested):
The fix for running CI on forks looks good to me.

The merge conflict is due to the existing `oqs-provider-release-test`. Does it duplicate what we have here?

I don't feel confident about coupling downstream release tests with liboqs development. In practice, I review pull requests in batches, so "commit to main" usually have bursts of activities that might flood the downstream projects with unnecessary runs. I would vote for putting the release tests as a "manual dispatch only" job or putting it into weekly tests.

Last but not least, consider using GitHub CLI for interacting with GitHub Actions, since it is a first-party tool and is installed by default on all GitHub-hosted runners. Using `gh` is probably easier than maintaining raw logic with bash and `curl`.

jplomas:
Understand.  As I recall, this was to prevent liboqs merges to main silently break provider but as there are a multitude of approaches and options, it is probably best to close this PR until sequencing is defined by process, rather than having the sequence defined by the code.  Apologies, I should have done this earlier.

PR #2496, opened 2026-07-17
Additional CI profiling and sig_stfl CI bug fixes

baentsch:
This PR generates additional output that can be used to analyze CI runtime problems.

It unearthed an incorrect Windows CI test setup (Debug instead of Release) that is also fixed in this PR, leading to massive CI runtime improvements (approx. 4x for the slowest jobs, SLH-DSA on Windows).

That fix in turn exposed a latent bug in the stateful hash based signature test setup on Windows (a missing CMake output-path override for SIG_STFL_TESTS, previously masked by the Debug build's incidental fallback path). This is also fixed here.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -20,7 +20,7 @@ jobs:
           - name: arm64-slhdsa
             runner: ubuntu-24.04-arm
             container: openquantumsafe/ci-ubuntu-latest:latest
-            PYTEST_ARGS: --maxprocesses=10 --ignore=tests/test_kat_all.py
+            PYTEST_ARGS: --maxprocesses=10 --ignore=tests/test_kat_all.py --durations=0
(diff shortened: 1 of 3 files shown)


ashman-p (approved):

xuganyu96 (changes requested):
The bug fix for Windows platform tests looks good to me.

I have some reservations about adding `--durations=0` to `pytest` as it adds clutter to logging, which is already enormous and difficult to navigate. A cleaner alternative would be to output detailed test reports to JUnit XML using the [`--junit-xml=/path`](https://docs.pytest.org/en/stable/how-to/output.html#creating-junitxml-format-files) option, then use `upload-artifact` to preserve the XML report. JUnit is a widely used unit test framework, so there should be a plethora of mature tools for analyzing them.

I would suggest simply ripping out the `--durations=0` from this pull request so it is narrowly focused on bug fix. The test report analysis component can be rolled into other CI enhancements you are already proposing at a later time.

PR #2500, opened 2026-07-21
CI Workflow optimizer

baentsch:
Proposal to fix #2497 & #2498. ~NOT ready for merge (private CI image). Goal of PR: Test & time new CI logic.~

AI has been used, but code has been human reviewed.

PR #2501, opened 2026-07-22
Add algorithm-integrator guidance to copy_from_upstream README

bhess:
Extends the copy_from_upstream README with more useful information regarding algorithm integraton:

- Pointers to the OQS KEM/SIG and common APIs + the OQS_SUCCESS/OQS_ERROR convention
- Integration glue: routing upstream RNG/hash/AES onto OQS primitives, and symbol namespacing
- Registering constant-time suppressions (passes/issues)
- Generated vs. hand-written files

Text assisted by AI


RodriM11:
Thanks for this @bhess !

RodriM11 (dismissed):

scripts/copy_from_upstream/README.md
```diff
@@ -134,10 +134,76 @@ When `include_only: false` (default), common sources are compiled once as a CMak
...
+The header of [`tests/test_constant_time.py`](../../tests/test_constant_time.py) explains how to turn Valgrind's output into a suppression file.
+
 ## Adding a New Upstream Algorithm
 
 1. Add the upstream repo to the `upstreams` section in `copy_from_upstream.yml`
 2. Add the family and its schemes to `kems` or `sigs`, referencing the upstream
-3. Ensure the upstream provides a `META.yml` per scheme with source lists and metadata
-4. If the algorithm has shared code across implementations, set up common dependencies
-5. Run `python3 copy_from_upstream.py copy` and verify the generated output
+3. In the upstream repo, provide a per-scheme `META.yml` (lengths, NIST level, `nistkat-sha256`, and `implementations` with `sources`, `compile_opts`, and the `signature_*` entrypoint symbol names — see [The OQS API](#the-oqs-api))
```
ashman-p, line 205: This needs coordination with the upstream authors, correct? If so, could there be expanded notes on this point (suggestions on how and who to reach out to)?

baentsch: Conceptually completely right @ashman-p . I'd love to see that, too.

But if we demand this here, we need to also have it for the algs already integrated -- and most notably, a field to fill in in the META.yml (sth like "primary contact") to point to here. 

If I'm not mistaken we still don't have that but still always have to "manually" contact the upstreams and hope and pray for their attention in case sth goes really wrong, right? 

Now, when checking things for the supposedly maintained upstreams OQS lists as such, I even fail to see a clear-cut responsibility statement or contact, e.g. [here for ML-KEM](https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/kem/ml_kem.md#support) or [here for XMSS](https://github.com/open-quantum-safe/liboqs/blob/main/docs/algorithms/sig_stfl/xmss.md), so I guess the point is moot: No-one wants to stick their head out (be someone "to reach out to").

--> Worthwhile a separate issue to improve this, I'd say. I thought we had one on this since years but maybe I'm wrong.

But not an issue in this PR. Thanks @bhess for improving things here and preparing this tooling for easier understanding!

baentsch (dismissed):

scripts/copy_from_upstream/README.md
bhess: Good point. I updated the README pointing to: the datasheet's `upstream-maintenance` and `upstream-statement-url` fields (which render into the support tables in README/ALGORITHMS.md), plus a note that submitters are welcome to add a [CODEOWNERS](https://github.com/open-quantum-safe/liboqs/blob/main/.github/CODEOWNERS) entry for their algorithm. As far as I can see there we already have the direct contacts for those algorithms with a claimed maintenance status, including ML-KEM and XMSS.

baentsch: Ahh - the CODEOWNERS page: That's a place I didn't check. Very good, thanks @bhess. Worthwhile pointing there (from the main README/ALGORITHMS.md), too? Would resolve @ashman-p 's original suggestion "automatically".

Regarding your suggested text, what about this formulation: "... are invited to add a code-owner entry for their algorithm in [`.github/CODEOWNERS`](../../.github/CODEOWNERS) (e.g. `/src/<kem|sig>/<family> @your-handle`) to stay in the loop; if the algorithm should be marked "Tier 1/Maintained" such entry is required."

bhess: Thanks for the suggestion @baentsch, I adopted the wording and a link to the CODEOWNERS file in ALGORITHMS.md.

baentsch (approved):
Thanks @bhess! Makes expectations & use of the tooling much more self-explanatory!

RodriM11 (approved):

baentsch:
fwiw, a [skip ci] would have been nice to the environment (particularly as #2500 has not landed): Another useless 2000 CPUmins wasted....

Issue #2498, opened 2026-07-20
CI re-design & review responsibility proposal


baentsch:
In extension of #2497 this is to propose a complete re-design of CI along a strict principle:

**Make contribution feedback fast again**: Strict adherence, ideally substantial improvement over the times mandated in https://github.com/open-quantum-safe/tsc/blob/main/guidelines/ci_usage.md.

Rationale: Simple pushes take nearly 10 minutes to come back with a result; PR feedback can take (way) more than an hour. Both break a normal development flow, leading to contributor multitasking or work on other projects when contributors wait for CI feedback. Reviewers are also not exactly well-served by this. 

All of this is bad for the project and the environment, and anathema to OQS wanting people to quickly and easily integrate new algorithms or enjoy working on new features.

 **Proposal**:

The way to achieve this will follow several approaches that I want to hear comments on below before putting in more work, particularly if there were ideological reasons to object or plainly no time or interest to review this work.

1) All ACVP tests will be removed from the automated push-PR-merge line and only be done when [extended tests] is explicitly requested, i.e., run in the worst case on the weekend. There simply is not enough active development any more in the core logic of those algorithms covered by ACVP testing to warrant this testing as often as is done currently.

2) [code coverage] testing will be streamlined: There is no CI automation stopping merges on code coverage regression and no known instance where code coverage reports influenced a PR. Moreover, the combination with ACVP testing has a miniscule effect on coverage (<0.1%) and a huge impact on runtime: Below the most extreme recent example: More than a day runtime: 

<img width="1535" height="441" alt="Image" src="https://github.com/user-attachments/assets/0c8bea3d-a54a-4484-a2c5-5ce80b792ab8" />

In some instances, SLH-DSA testing alone takes 98.5% of all CI cycles. This is hilarious -- for an algorithm family that is not exactly on the forefront of research nor actual deployments.

3) Generally, any algorithm-specific support infrastructure testing, e.g., `libjade`, will also be moved to [extended testing] only: That alone currently incurs a 100% push-wait time penalty.

4) Algorithm-independent, but architecture/infrastructure specific testing will be streamlined for push testing and moved to [full tests] only for the current test set; the (bad) example is nix, currently causing a >400% penalty on push testing.

5) The CI image will get streamlined: It's grown way too large, penalizing every single task with unnecessary startup-overhead (WIP at https://github.com/baentsch/ci-containers/blob/main/ubuntu-latest/Dockerfile.slim).

[...]

PR #2509, opened 2026-07-27
further CI optimizations

baentsch:
This extends #2500 by moving further long-running, highly unlikely to fail tests to extended/weekly testing.

* [no] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged. Also, make sure to update the list of algorithms in the continuous benchmarking files: .github/workflows/kem-bench.yml and sig-bench.yml)

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -18,12 +18,6 @@ jobs:
             PYTEST_ARGS: --maxprocesses=10 --ignore=tests/test_kat_all.py
             CMAKE_ARGS: -DOQS_ENABLE_SIG_SLH_DSA=OFF -DOQS_ENABLE_SIG_STFL_LMS=ON -DOQS_ENABLE_SIG_STFL_XMSS=ON -DOQS_HAZARDOUS_EXPERIMENTAL_ENABLE_SIG_STFL_KEY_SIG_GEN=ON
 
-          - name: alpine
-            runner: ubuntu-latest
-            container: openquantumsafe/ci-alpine-amd64:latest
-            CMAKE_ARGS: -DOQS_STRICT_WARNINGS=ON -DOQS_USE_OPENSSL=ON -DBUILD_SHARED_LIBS=ON -DOQS_ENABLE_SIG_SLH_DSA=OFF -DOQS_HAZARDOUS_EXPERIMENTAL_ENABLE_SIG_STFL_KEY_SIG_GEN=ON -DOQS_ENABLE_SIG_STFL_XMSS=ON -DOQS_ENABLE_SIG_STFL_LMS=ON
-            PYTEST_ARGS: --ignore=tests/test_alg_info.py --ignore=tests/test_kat_all.py
-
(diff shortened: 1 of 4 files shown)


.github/workflows/linux.yml
```diff
@@ -118,7 +112,7 @@ jobs:
         timeout-minutes: 60
         env:
           OQS_TEST_SIG_CTX_STEP: "255"
-        run: mkdir -p tmp && python3 -m pytest --verbose --ignore=tests/test_code_conventions.py --ignore=tests/test_acvp_vectors.py --numprocesses=auto ${{ matrix.PYTEST_ARGS }}
+        run: mkdir -p tmp && python3 -m pytest --verbose --ignore=tests/test_code_conventions.py --ignore=tests/test_acvp_vectors.py --ignore=tests/test_wycheproof_vectors.py --numprocesses=auto ${{ matrix.PYTEST_ARGS }}
```
dstebila, line 115: Do wycheproof tests really cost that much?  As far as I can tell it's about a 5-10 second cost.

baentsch: Agreed, they don't. They just annoyingly keep failing irregularly with network failures when pulling the data, requiring unnecessary manual CI re-runs:  The last time happened just today in the merge of #2500 and I personally clicked pretty often in the past few months on "CI re-run" to get CI to green on PRs otherwise stalled in "failed" CI.

As reduction (in the friction of) CI is the main purpose of this work and shift-to-extended is in line of the spirit of item 1 in https://github.com/open-quantum-safe/liboqs/issues/2498 (not unnecessarily often repeating tests on standardized, i.e., by-definition hardly changing, code) I still think this is reasonable.

A lot of text again, I know, Sorry and I hope you're still reading... The answer by @claude would have been _much_ more verbose :-)

dstebila: > Agreed, they don't. They just annoyingly keep failing irregularly with network failures when pulling the data, requiring unnecessary manual CI re-runs: The last time happened just today in the merge of #2500 and I personally clicked pretty often in the past few months on "CI re-run" to get CI to green on PRs otherwise stalled in "failed" CI.
> 
> As reduction (in the friction of) CI is the main purpose of this work and shift-to-extended is in line of the spirit of item 1 in #2498 (not unnecessarily often repeating tests on standardized, i.e., by-definition hardly changing, code) I still think this is reasonable.
> 
> A lot of text again, I know, Sorry and I hope you're still reading... The answer by @claude would have been _much_ more verbose :-)

Ah, okay, that makes sense. My Claude when reviewing this noticed that the caching mechanism was vacuous, but I didn't want to pull that into my comment. I'm okay with keeping things simple and moving on with this.

dstebila (approved):

xuganyu96 (approved):

PR #2510, opened 2026-07-28
Add requirements.txt for Python dependencies

charishma1407:
**Purpose:**
 This PR adds requirements.txt file for installing python dependencies.
Updated README.md as per the requirements in Ubuntu and MacOs.

**Checklist**
1. Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  (If so, a version bump will be required from *x.y.z* to *x.(y+1).0*.)
**No This PR is a document and dependency change only**
2. Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider]
**No**

**Generative AI**
This contribution is not made with the help of a generative AI.


dstebila:
Hi @charishma1407, thanks for looking at this. As far as I can tell, on Ubuntu the preferred way to install packages is via apt, rather than pip3. Is there a reason to do it the way you've proposed?

charishma1407:
You're right — apt is the preferred and safer way to install packages on Ubuntu as it avoids conflicts with system-managed Python. I have reverted the Ubuntu install step back to use apt and keep requirements.txt only for macOS and other platforms where pip3 is the standard approach.
I thought of making it to generic so I made pip3 for both macOS and ubuntu.

xuganyu96:
> Hi @charishma1407, thanks for looking at this. As far as I can tell, on Ubuntu the preferred way to install packages is via apt, rather than pip3. Is there a reason to do it the way you've proposed?

With `pip` we can pin exact version and verify sha-256 checksum. I would actually argue against installing pytest (and its plugins) and other script dependencies via system package manager precisely because `apt` versions might drift way from `pip` versions, which then makes CI image dependencies drift from local development setup. Especially for local setup, a virtual environment (i.e. venv, virtualenv, conda) should be the strongly recommended default.

requirements.txt
xuganyu96, line 1: The test suite also requires `requests`. Please also pin exact version, provide sha-256 hashes. It is a good idea to run `pip` with `--require-hashes`.

xuganyu96 (changes requested):

xuganyu96 (approved):

dstebila:
@xuganyu96 is this ready to merge?

xuganyu96:
> @xuganyu96 is this ready to merge?

Yes

PR #2511, opened 2026-07-29
Fix out-of-bounds access in xmss_core_sign on secret key OID mismatch

Ramya-9353:
Repro: deserialise an XMSS secret key blob of exactly `sig->length_secret_key` bytes whose first four bytes name a different parameter set, then call `OQS_SIG_STFL_sign`. For `XMSS-SHA2_10_192` (1057-byte key) carrying the OID of `XMSS-SHA2_20_512` (`sk_bytes` 4973), ASAN reports `heap-buffer-overflow READ of size 1 ... 1216 bytes after 1057-byte region` in `bytes_to_ull` (`utils.c:28`), reached via `xmssmt_deserialize_state` (`xmss_core_fast.c:111`) <- `xmss_deserialize_state` <- `xmss_core_sign` (`xmss_core_fast.c:663`) <- `xmss_sign` (`xmss.c:95`) <- `OQS_SIG_STFL_sign` (`sig_stfl.c:1035`). Three calls from the documented public API, with the blob fully attacker-controlled.

Cause: the OID is part of the secret key material and selects `n`, `tree_height`, `d`, `index_bytes`, `sk_bytes`, `sig_bytes` and `wots_sig_bytes`. `OQS_SECRET_KEY_XMSS_deserialize_key` checks only that the blob length equals `sk->length_secret_key`, and `xmss_sign` re-parses the OID out of the loaded bytes, so the BDS state is walked with the geometry of the named parameter set rather than the allocated one. The signature is written past the end of the caller's buffer for the same reason.

Fix: decode `secret_key_data[0..3]` as a big-endian OID in the per-variant sign wrapper and reject anything other than the variant's compile-time OID constant, before any offset is derived from it. The check sits in the `XMSS_ALG` macro, so it reaches all 37 XMSS and XMSS^MT variants from one place. `xmss_keypair` writes the variant's own OID into the secret key, so a legitimately serialised key always carries the matching value and valid signing is untouched; the XMSS and XMSS^MT KATs are unchanged.

The public-key side of this check already lives in the same macro's verify wrapper (GHSA-2wxh-55qf-c7wg); the secret key side was left open. `OQS_SECRET_KEY_XMSS_deserialize_key` cannot make the equivalent check itself, since it is shared by every variant and the key object carries only a length, leaving it unable to tell an XMSS OID from an XMSS^MT OID with the same numeric value. `sigs_remaining` and `sigs_total` read only the OID and index prefix, so `sign` is the sole consumer that derives buffer offsets from it. The regression test adds a third sub-case to `test_invalid_sig`: unpatched, 33 of the 37 variants abort under ASAN and the other 4 accept the mismatched key and sign with it, so every variant is covered.

### DIFF
--- a/src/sig_stfl/xmss/sig_stfl_xmss_xmssmt.c
+++ b/src/sig_stfl/xmss/sig_stfl_xmss_xmssmt.c
@@ -69,6 +69,19 @@ OQS_API OQS_STATUS OQS_SIG_STFL_alg_xmss##xmss_v##_keypair(XMSS_UNUSED_ATT uint8
 }\
 \
 OQS_API OQS_STATUS OQS_SIG_STFL_alg_xmss##xmss_v##_sign(uint8_t *signature, size_t *signature_len, const uint8_t *message, size_t message_len, OQS_SIG_STFL_SECRET_KEY *secret_key) {\
+        if (secret_key == NULL || secret_key->secret_key_data == NULL) {\
+                return OQS_ERROR;\
+        }\
+        /* Reject secret keys whose OID disagrees with the variant the caller invoked. */\
+        /* The OID is part of the key material and selects every buffer offset used by */\
+        /* xmss_core_sign, so a deserialized key naming a larger parameter set reads and */\
+        /* writes past both secret_key_data and the caller's signature buffer. */\
+        const uint8_t *sk_oid_bytes = (const uint8_t *)secret_key->secret_key_data;\
+        uint32_t sk_oid = ((uint32_t)sk_oid_bytes[0] << 24) | ((uint32_t)sk_oid_bytes[1] << 16)\
+                        | ((uint32_t)sk_oid_bytes[2] <<  8) | ((uint32_t)sk_oid_bytes[3]);\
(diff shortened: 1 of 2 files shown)


Ramya-9353:
Amended the commit with the missing sign-off; DCO check is green now.

Ramya-9353:
any update?

ashman-p:
Thanks for this. What do you think about adding this patch for completion?
[stfl_sig.patch](https://github.com/user-attachments/files/31487973/stfl_sig.patch)

Ramya-9353:
Good call, folded it in. sigs_remaining and sigs_total both re-read the OID and index the same way sign does, so a foreign OID could walk past secret_key_data there too. Pushed the patch with one small change: in the test the sigs_* acceptance branch now sets status to OQS_SUCCESS rather than OQS_ERROR, otherwise the assertion below never fires and the sub-case wouldn't catch a regression on those two paths. Verified by stubbing out the three OID guards, which makes the test fail as expected, and all the XMSS/XMSSMT KATs still pass with the guards in place.

ashman-p:
Thanks for this @Ramya-9353

ashman-p (approved):
Thanks for the contribution.

xuganyu96 (approved):

PR #2520, opened 2026-08-08
Fix oqs_sig_stfl_lms_sign overflow on cross-parameter-set secret key

Ramya-9353:
Repro: with `-DOQS_HAZARDOUS_EXPERIMENTAL_ENABLE_SIG_STFL_KEY_SIG_GEN=ON`, deserialise an HSS `LMS_SHA256_H10_W2_H10_W2` key (9300-byte signature) into an `LMS_SHA256_H5_W8` secret-key object (`length_signature` 1296) via `OQS_SIG_STFL_SECRET_KEY_deserialize`, then `OQS_SIG_STFL_sign`. Every LMS variant shares the same 64-byte private key, so the key loads, and the sign writes 9300 bytes into the caller's 1296-byte buffer:

```
==ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 9300 ... 0 bytes after 1296-byte region
    #1 oqs_sig_stfl_lms_sign      sig_stfl_lms_functions.c:617
    #2 OQS_SIG_STFL_alg_lms_sign
    #3 OQS_SIG_STFL_sign
```

Cause: `oqs_deserialize_lms_key` checks only the total blob length against the constant 64-byte private key, so it accepts a key for any parameter set. `oqs_sig_stfl_lms_sign` then takes `sig_len` from the loaded key's own parameter set and copies it into the caller's buffer, which the API sizes to the object's `length_signature`.

Fix: reject a deserialised key whose parameter set yields a different signature length than the variant the object was created for. Each variant's `length_signature` is threaded into `oqs_deserialize_lms_key` and compared against `hss_get_signature_len` of the parsed set. XMSS is not affected: its per-variant secret keys have distinct lengths, so the existing exact-length check already rejects a mismatch. The added regression test fails on the current tree and passes with the fix.

* [ ] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API?


Ramya-9353:
Amended the commit with the missing sign-off and force-pushed; DCO is green now. No code changes.

ashman-p:
Hi @Ramya-9353, thanks for the updates. Also, were any of the code changes AI-generated? If so, please cite resources used.

Ramya-9353:
The code changes themselves I wrote by hand; I do use an assistant a little for research and sanity-checking along the way. Resources: RFC 8554 for the LMS/HSS parameter sets and signature sizes, the vendored hash-sigs sources (hss.h, hss_get_signature_len, the 64-byte private-key format) for the deserialise path, and AddressSanitizer to confirm the overflow at sig_stfl_lms_functions.c:617 and that the fix rejects the mismatched key before sign.

ashman-p (dismissed):
@Ramya-9353, Thanks for the contribution.

tests/test_sig_stfl.c
```diff
@@ -482,6 +482,60 @@ static OQS_STATUS test_invalid_sig(const char *method_name) {
...
+	const char *big = "LMS_SHA256_H5_W1";   /* length_signature 8688 */
+	const char *small = "LMS_SHA256_H5_W8";  /* length_signature 1296 */
+	OQS_STATUS ret = OQS_ERROR;
+	uint8_t *pk = NULL, *blob = NULL;
+	size_t bloblen = 0;
+
+	OQS_SIG_STFL *sig_big = OQS_SIG_STFL_new(big);
+	OQS_SIG_STFL_SECRET_KEY *sk_big = OQS_SIG_STFL_SECRET_KEY_new(big);
+	OQS_SIG_STFL_SECRET_KEY *sk_small = OQS_SIG_STFL_SECRET_KEY_new(small);
+	if (sig_big == NULL || sk_big == NULL || sk_small == NULL) {
```
ashman-p: Looks like 'small' is already defined in Windows API. Rename this variable here.

```suggestion
	const char *lms_big = "LMS_SHA256_H5_W1";   /* length_signature 8688 */
	const char *lms_small = "LMS_SHA256_H5_W8";  /* length_signature 1296 */
	OQS_STATUS ret = OQS_ERROR;
	uint8_t *pk = NULL, *blob = NULL;
	size_t bloblen = 0;

	OQS_SIG_STFL *sig_big = OQS_SIG_STFL_new(lms_big);
	OQS_SIG_STFL_SECRET_KEY *sk_big = OQS_SIG_STFL_SECRET_KEY_new(lms_big);
	OQS_SIG_STFL_SECRET_KEY *sk_small = OQS_SIG_STFL_SECRET_KEY_new(lms_small);
	if (sig_big == NULL || sk_big == NULL || sk_small == NULL) {
```

ashman-p (approved):
The Windows build issue has bee resolved now.

PR #2522, opened 2026-08-10
Update mldsa-native to v2.0.0 [full tests] [extended tests]

### LINKED ISSUE #2518: Update mldsa-native to v2.0.0
We have just shipped the first stable release of mldsa-native: https://github.com/pq-code-package/mldsa-native/releases/tag/v2.0.0

(called v2 because conventions and configuration is aligned with the upcoming mlkem-native v2).

mkannwischer:
Bump the mldsa-native upstream from v1.0.0-beta2 to v2.0.0, the first stable release of mldsa-native.

Notable changes:
- All x86_64 and AArch64 assembly is not proven to be functionally correct and memory-safe using HOL-Light. All x86_64 and AArch64 assembly except the rejection sampling routines are also proven to be constant-time.
- signature/verification no longer carry a siglen argument, since ML-DSA signatures have a fixed length. The generated OQS wrappers are insulated from this by upstream's new integration/liboqs/sig_glue.c, which re-exposes the length-carrying API.
- Entry points, including the external-mu ones, gain an `_oqs` suffix as they are wrappers collapsing mldsa-native's more fine grained error codes to OQS_SUCCESS/OQS_ERROR.

tests/vectors_sig.c calls the internal signing and verification entry points directly, so it has to drop the siglen argument as well.

See the full release notes at
https://github.com/pq-code-package/mldsa-native/releases/tag/v2.0.0

- Resolves #2518 

NB: I'm not thrilled that `tests/vectors_sig.c` contains mldsa-native specific code . I'd very much prefer if only functions included in the META.yml are called. If additional functions are needed, I'm happy to add them to the integration glue if there is a well defined API. I'd argue that tests within OQS should be testing the OQS APIs only and not test the mldsa-native APIs directly.


baentsch:
Thanks for the update @mkannwischer .

> I'm not thrilled that tests/vectors_sig.c contains mldsa-native specific code . 

That's an understatement whose core statement I whole-heartedly support: `liboqs` has taken on so many algorithm-specific twists and turns that I lost understanding of this "bespoke machinery" and therefore had to cease reviewing anything non-generic.

> I'd very much prefer if only functions included in the META.yml are called. If additional functions are needed, I'm happy to add them to the integration glue if there is a well defined API. 

Thanks for that willingness. But for that there would need to be a concerted effort and willingness to bring back complete API "algorithm-agnosticity" (I know the term doesn't exist in English, but I can't help my Latin :) and I'm not sure there's bandwidth/appetite for that.

> I'd argue that tests within OQS should be testing the OQS APIs only and not test the mldsa-native APIs directly.

Again, totally agreed. Unfortunately I think I fought a lost battle for generic APIs against limited time for OQS by humans; and AIs (reviewing and/or coding) also just do "the bare minimum to not get fired" (if you know the film "Office Space" :).

I sum, to not have this PR be held up/lengthen the backlog, I'd tend to suggest just merging this without further checks as it obviously used the standard copy-from-upstream machinery without changes and it passes your and OQS' CI.

mkannwischer:
> Thanks for the update @mkannwischer .
> 
> > I'm not thrilled that tests/vectors_sig.c contains mldsa-native specific code .
> 
> That's an understatement whose core statement I whole-heartedly support: `liboqs` has taken on so many algorithm-specific twists and turns that I lost understanding of this "bespoke machinery" and therefore had to cease reviewing anything non-generic.
> 
> > I'd very much prefer if only functions included in the META.yml are called. If additional functions are needed, I'm happy to add them to the integration glue if there is a well defined API.
> 
> Thanks for that willingness. But for that there would need to be a concerted effort and willingness to bring back complete API "algorithm-agnosticity" (I know the term doesn't exist in English, but I can't help my Latin :) and I'm not sure there's bandwidth/appetite for that.
> 
> > I'd argue that tests within OQS should be testing the OQS APIs only and not test the mldsa-native APIs directly.
> 
> Again, totally agreed. Unfortunately I think I fought a lost battle for generic APIs against limited time for OQS by humans; and AIs (reviewing and/or coding) also just do "the bare minimum to not get fired" (if you know the film "Office Space" :).
> 
> I sum, to not have this PR be held up/lengthen the backlog, I'd tend to suggest just merging this without further checks as it obviously used the standard copy-from-upstream machinery without changes and it passes your and OQS' CI.

Thanks! I think we are in agreement.
I didn't mean to suggest that this should be fixed before we merge this PR, but rather consider it in the future (or at least not make it worse by adding more implementation-specific tests). If there is consensus that the mldsa-native-specific tests in tests/vectors_sig.c should be removed, I can open a follow-up PR to drop them.

baentsch:
> Thanks! I think we are in agreement.
I didn't mean to suggest that this should be fixed before we merge this PR, but rather consider it in the future (or at least not make it worse by adding more implementation-specific tests). If there is consensus that the mldsa-native-specific tests in tests/vectors_sig.c should be removed, I can open a follow-up PR to drop them.

Thanks for that offer, too. This I think is worth while doing only when there'd be consensus forming that all alg-specific things should be streamlined. If that doesn't exist, an improvement in just one corner doesn't really move the needle.

RodriM11:
Thank you for the offer @mkannwischer, and for the PR in general!

> This I think is worth while doing only when there'd be consensus forming that all alg-specific things should be streamlined. If that doesn't exist, an improvement in just one corner doesn't really move the needle.

We all want code to be employed as general as possible, without resorting to code-specific of an algorithm when there is no need to. In order to correct those situations, we have to start somewhere, so any efforts are more than welcome.

mkannwischer:
I just rebased to resolve the conflicts. @xuganyu96 @bhess can we get this merged?

bhess (approved):
Thanks for the update @mkannwischer. All looks good to me, ready to merge from my side, pending a second approval.

xuganyu96 (approved):

PR #1123, opened 2021-11-16
Adding multiple upstream support to doc generation

Martyrshot:
This PR fixes an issue discovered by  Michael (https://github.com/open-quantum-safe/liboqs/pull/1117#issuecomment-969116802).

This will update legacy yaml files so that the new format is followed, and when a yaml file is updated the script it indicate it has done so. When this happens I recommend double checking the yaml file to make sure the ordering "makes sense" to human eyes.

I also manually patched the yaml files of schemes that the copy_from_upstream.py script doesn't manage.

PR #2547, opened 2026-09-04
Fix stale loop variable in copy_from_upstream arch-specific ignore check

mohamadm0meni:
Two expressions in `load_instructions()` read a loop variable belonging to a
different loop. Two characters each: `impl['name']` to `imp['name']`.

```python
# scripts/copy_from_upstream/copy_from_upstream.py
# :304 — the same key, built correctly, a few lines earlier in the same function
if 'ignore' in upstream and "{}_{}_{}".format(upstream['name'], scheme['pqclean_scheme'], imp['name']) in upstream['ignore']:

# :317-321 — the loop variable is `imp`, but :320 formats the key from `impl`
for imp in implementations:
    upstream = upstreams[family['arch_specific_upstream_locations'][arch]]
    if (arch in family['arch_specific_implementations'] and imp['name'] in family['arch_specific_implementations']) \
            and ('ignore' not in upstream or ('ignore' in upstream and "{}_{}_{}".format(upstream['name'], scheme['pqclean_scheme'], impl['name']) \
                    not in upstream['ignore'])):
```

`impl` is never bound in that loop, only at `:358` and `:469` — later, over a
different list, for the previous scheme. It does not raise because `hqc`
(`copy_from_upstream.yml:159`) is processed before `kyber` (`:176`) and has
already run its `:358` loop. So for `kyber512`, `:320` builds
`oldpqclean_kyber512_ref` where `:304` would build
`oldpqclean_kyber512_aarch64`. Nothing changes as a result: the only upstream
reaching this expression with an `ignore` key is `oldpqclean`, which declares
`ignore: ""` (`copy_from_upstream.yml:21`), and any string is `not in ""`.

`kyber768` and `kyber1024` agree by construction rather than by correctness:
`:323` appends the aarch64 implementation to `metadata['implementations']`, so
the `:358` loop ends on it and the next scheme reads `aarch64` — the right
string by the wrong route. Only the first scheme of the first family to reach
this code diverges.

`:320` is live: three evaluations per `-d copy`, one of which reads a name from
another scheme, and that was measured. `:427`, with `:411` and `:424-428` the
signature-side copies of the same pair, is a defect by inspection only — no
signature family declares `arch_specific_upstream_locations`, so `:399` never
creates the `extras` dict `:421` guards on. It is fixed here because if one
were to declare it the expression would become live in the state the KEM side
is in now; I make no prediction about whether that will happen.

I ran the real `load_instructions()` against the real, unmodified
`copy_from_upstream.yml` with tracing on these lines, before and after the
change, for both shipped configurations. The instruction tree that all
downstream processing consumes is byte-identical at the current pins. Beyond
that tree, `basic.yml`'s upstream check regenerates the vendored sources with
the patched script and asserts `git status --porcelain` is empty; it ran on
this commit and passed, again under these configurations at these pins. That
check ran against the pull request's merge ref, so the regeneration was of this
change merged into `main` rather than of the branch alone.

What was run, on my fork: both jobs that execute this script passed. `-d copy`
ran on this commit. `-d libjade` is never triggered by a pull request, so it
will not appear in this PR's checks; I ran it separately on the same two-line
change, and the inputs that job reads — `copy_from_upstream.py` and
`copy_from_libjade.yml` — are byte-identical to the ones here. The other
`extended.yml` failures on that run reproduce identically on an unmodified
`main`; details below.

**Use of generative AI**, per CONTRIBUTING.md. I used Claude throughout. The
starting hypothesis came from a model; I treated it as a lead and tried to
refute it first, and two of its claims proved wrong and were dropped. The model
wrote the tracing harness and a first draft of this text. I verified everything
asserted here by running it — line numbers and scopes re-derived with `ast`,
family ordering and `ignore` values read from the yml and the pinned upstream
METAs, the divergence and the before/after comparison from my own runs. I
affirm this contribution.

<details>
<summary>Measured trace, unpatched main</summary>

`sys.settrace` on the suspect lines, recording what `imp` and `impl` are bound
to at the moment the expression is evaluated:

```
line 320: 3 evaluations, all kyber / aarch64 / oldpqclean
  kyber512   imp['name']=aarch64   impl['name']=ref       -> diverged
  kyber768   imp['name']=aarch64   impl['name']=aarch64
  kyber1024  imp['name']=aarch64   impl['name']=aarch64
line 427: 0 evaluations
`impl` unbound at evaluation time: never
```

`ml_kem` also declares `arch_specific_upstream_locations`, so `:319` is reached
27 times against `:320`'s 6. Its upstreams `cupqc` and `icicle` declare no
`ignore` key at all, so `'ignore' not in upstream` short-circuits and `:320` is
never evaluated for them.

The KEM families in `copy_from_upstream.yml` are `classic_mceliece` (`:114`),
`hqc` (`:159`), `kyber` (`:176`), `ml_kem` (`:207`). The HQC METAs declare
exactly one implementation, `ref`, which is where the stale value comes from.

</details>

<details>
<summary>Fork CI: what ran, and the control comparison</summary>

`extended.yml` is `on: [workflow_call, workflow_dispatch]` and its only caller
gates on `[extended tests]` in the head commit message, so a pull request never
runs it. To exercise `-d libjade` I pushed the same tree twice to my fork.

Those two runs were made before this branch was rebased onto the current
`main`, so they carry the same two-line change over the previous base. What the
`-d libjade` job reads is unaffected: `copy_from_upstream.py:863` shows it loads
`copy_from_libjade.yml` and never `copy_from_upstream.yml`, and both that yml
and this file's base blob are byte-identical across the rebase.

| | jobs | success | failure | skipped |
|---|---|---|---|---|
| this commit, clean message | 12 | 8 | 0 | 4 |
| identical tree + `[extended tests]` | 46 | 33 | 10 | 3 |
| unmodified `main` + `[extended tests]` | 46 | 32 | 11 | 3 |

The job sets on the second and third runs are identical, 1:1. Comparing them
by name: **no job failed on the patched tree that did not also fail on the
unmodified one.** One job failed only on the unmodified tree
(`address-sanitizer-slhdsa`), on a 90-minute wall-clock timeout rather than an
assertion, and it passed on the patched tree.

The ten shared failures, with identical signatures on both:

```
nistkat-x64 x4          FAILED tests/test_kat_all.py::test_sig[ML-DSA-{44,65,87}-extmu]
SLH-DSA alpine x3       FAILED tests/test_code_conventions.py::{test_spdx,test_style}
  + alpine-libjade        [Errno 2] No such file or directory: 'tests/run_astyle.sh'
jammy libjade x2        FAILED tests/test_code_conventions.py::test_style
                          Error: Some files need reformatting.
```

`Check default upstream code is properly integrated` and `Check libjade code is
properly integrated` were `success` on every run above. The skipped jobs are
`push.yml` message-tag gates that neither message triggers; they are skips by
design rather than passes.

</details>

<details>
<summary>Equivalence method, dump hashes, and how the fixed line was checked</summary>

The `instructions` structure was serialized deterministically (sorted keys) and
the dumps compared:

```
771677341748001a…   356874 bytes   pre.copy_from_upstream.json
771677341748001a…   356874 bytes   post.copy_from_upstream.json
ea99d92529287cf7…     9160 bytes   pre.copy_from_libjade.json
ea99d92529287cf7…     9160 bytes   post.copy_from_libjade.json
```

The structure stores the upstream dict back onto the objects that came out of
it, so it holds 471 shared references. A reference to an ancestor is emitted as
a marker; a shared but non-ancestor reference is serialized again in full so no
field is dropped. A self-test runs on a deliberately cyclic structure each time,
and a leaf count is compared against a naive walk (`live=6728 canonical=6728`).
As it turns out the live structure holds no cycle at all — plain `json.dumps`
completes without a circular-reference error.

To check that the fixed line stops reading `impl` rather than merely appearing
to, `impl` was rebound in the live frame immediately before the expression to an
object whose `__getitem__` raises. Unpatched, the run dies on the first
evaluation. Patched, it completes with that object in place 27 times across six
schemes, never read.

</details>

<details>
<summary>Harness fidelity, provenance, and what remains unverified</summary>

The harness executes the real `load_instructions()` with `shell()` stubbed, so
nothing is cloned, and serves META content from memory.

> The harness serves the kyber, HQC and pqcrystals METAs from the real pinned
> upstream content, and synthesizes the remaining 84 META files. This
> establishes that the expression is evaluated, that it reads a name bound by a
> different loop, and that the two names can differ — it does not establish the
> exact stale value a full `copy_from_upstream.py -d copy` would produce, since
> that depends on META content this harness did not fetch.

The particular string `oldpqclean_kyber512_ref` comes from METAs served real at
their pinned commits on both ends of the chain: the HQC METAs recovered from
`patches/pqchqc-liboqs-integration.patch`, and the PQClean and pqcrystals kyber
METAs at the commits `copy_from_upstream.yml` pins.

Provenance: the KEM site arrived in `769412623` (#1123, 2021-11-18) and was
copied to the signature side in `55bc8cb38` (#1194, 2022-03-09). The third
`impl['name']` in the file, at `:924`, is correct and untouched — it is inside
`verify_from_upstream()` (`:883`), where `impl` genuinely is the loop variable
(`:915`).

Not established: the exact stale string a full `-d copy` would produce; the
behaviour under any `ignore` value other than the `""` shipped today; and any
configuration other than the two shipped ymls.

</details>


mohamadm0meni:
Updating this comment to cover all three runs on head `f9d8477a4` rather than leaving the first run's snapshot standing.

| run | job | failing step | error |
|---|---|---|---|
| 33858723668 | `linux_x86_emulated (avx512-ml-kem_ml-dsa, -skx)` | `Setup Intel SDE` | `2026-09-04 11:32:02 ERROR 403: rate limit exceeded.` |
| 33858723668 | `linux (noble-shared-noopenssl)` | `Run tests` | `requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))` |
| 33867071879 | `basic-checks / Check code formatting` | `Validate CBOM` | `GnuTLS: Error in the pull function. Unable to establish SSL connection.` |

In the `Setup Intel SDE` job, `Configure`, `Build` and `Run tests` were all skipped, so nothing in this repository was compiled or executed.

In `noble-shared-noopenssl` the two failures were `tests/test_wycheproof_vectors.py::test_wycheproof_vec_kem_keygen[ML-KEM-512]` and `::test_wycheproof_vec_sig_sign[ML-DSA-87]`, with `2 failed, 1187 passed, 1464 skipped in 844.86s`.

In `Validate CBOM`, `scripts/validate_cbom.sh` fetches three CycloneDX schemas; the first two saved and the third did not, leaving `ajv` with `error: Cannot find schema '.../scripts/jsf-0.82.schema.json'`.

Subsequently, on the same head SHA:

- run 33867071879: `linux_x86_emulated (avx512-ml-kem_ml-dsa, -skx)` and `linux (noble-shared-noopenssl)` both completed with conclusion `success`.
- run 33871116055: `basic-checks / Check code formatting` completed with conclusion `success`.

Run 33871116055 completed with conclusion `success`: 71 jobs `success`, 1 `skipped` (`nix-full-build`), 0 failed, 0 cancelled. Checks: 72 pass, 3 skipping.

The two `platform-tests` jobs above also completed with conclusion `success` on run 33786584890, the Main branch tests run for c27c88b76, the commit this branch is based on.

xuganyu96 (approved):
Nice catch. Thanks!

RodriM11:
Hi @mohamadm0meni ! As per your stated AI contribution, I think it would make sense to add Claude as a contributor of your PR (i.e. signing off with Claude also), do you agree?

mohamadm0meni:
> Hi @mohamadm0meni ! As per your stated AI contribution, I think it would make sense to add Claude as a contributor of your PR (i.e. signing off with Claude also), do you agree?

Hi @RodriM11 Sure, I've added Claude as a co-author.

PR #2557, opened 2026-09-08
Fix: ignore .venv in tests/test_spdx.sh

owaldron:
## Summary

This PR introduces the `spdx-ignore.txt` to list ignore filename patterns for files that do not need an `SPDX-License-Identifier`. It also appends `/.venv/` to the existing ignore patterns allowing the test to pass locally when running `ninja run_tests` following the quick start.

## Versioning Questions


xuganyu96 (approved):
A nice improvement to make `test_spdx.sh` more adaptable to individual setup. LGTM.

PR #2562, opened 2026-09-09
Update SNOVA to Round 3

### LINKED ISSUE #2559: Update SNOVA to the Round 3 version
### Description

SNOVA Round 3 is a major update as described in the Supporting Document which is available on our github repository.

For the integration into liboqs the most important change is that we changed the names of the parameter sets.

_No response_

_No response_

I will create a merge request.

### Use of generative AI

The AVX2 version is based on the optimized version and was further optimized with AI-assisted development (Claude Code).

vacuas:
Update SNOVA to the Round 3 version.

Fixes #2559.

I will update oqs-provider next.

The AVX2 version is based on the optimized version and was further optimized with AI-assisted development (Claude Code).


baentsch:
Thanks for this update @vacuas ! Before reviewing 899 files, can you please document what you actually did to prepare this PR? Was it more than updating `copy_from_upstream.yml` and re-running the script? Where should a (human :) reviewer particularly focus?

vacuas:
> Thanks for this update @vacuas ! Before reviewing 899 files, can you please document what you actually did to prepare this PR? Was it more than updating `copy_from_upstream.yml` and re-running the script? Where should a (human :) reviewer particularly focus?

Essentially it was just updating `copy_from_upstream.yml` and re-running the script. Most of the edits to integrate were in the upstream code to resolve various CI issues (https://github.com/PQCLAB-SNOVA/SNOVA/pull/14). 
But as the names and the parameters have changes, the KATs needed to change along with constant time passes. As also the large memory properties have changed there were a few other files that required an update.

To be precise, I edited the following files to prepare for `copy_from_upstream.py`

```
cp copy_from_upstream.yml ~/liboqs/scripts/copy_from_upstream/
cp snova.yml ~/liboqs/docs/algorithms/sig
cp kats.json ~/liboqs/tests/KATs/sig/
cp passes.json ~/liboqs/tests/constant_time/sig/
cp snova ~/liboqs/tests/constant_time/sig/passes/
cp test_sig.c ~/liboqs/tests/
cp issues.json ~/liboqs/tests/constant_time/sig/
cp CMakeLists.txt ~/liboqs/zephyr/
```

I hope this helps.

vacuas:
In addition, you may want to review the way the memory optimized version was integrated.

RodriM11:
Thanks a lot @vacuas !

xuganyu96 (dismissed):

xuganyu96 (changes requested):
Thank you for the update. The pull request looks mostly good to me except for two concerns:

1. For category I and III there is a `X` variant in addition to the normal K, B, S variants listed in the supporting document's recommended list of parameter. The supporting document describes `X` variant in "additional parameter set" as trading very large public key for very small signature, but the `X` name was not documented. I found it a bit confusing "where did the X variants" come from, and it would be great if it can be documented either here or in the supporting document. Also, there is no non-AES X variant: is this intentional?
2. The algorithm filtering is not quite right. With `-DOQS_ALGS_ENABLED=NIST_SIG_ONRAMP`, SNOVA was not enabled.

vacuas:
Thanks @xuganyu96 for this excellent review!

Regarding concern 1: we are discussing the best way to resolve this within the SNOVA team at the moment.

I have a question on concern 2: For SNOVA this is easy to fix. When testing I noted that MQOM is missing in the list of NIST_SIG_ONRAMP and CROSS is still in. Do you want me to update that as well?

Another point: the zephyr currently excludes some of the SNOVA schemes. This is no longer needed as we now have a reduced memory version that uses less than 120KB even when signing at SL V. I will update this as well.

xuganyu96:
> Regarding concern 1: we are discussing the best way to resolve this within the SNOVA team at the moment.

Sounds good.

> I have a question on concern 2: For SNOVA this is easy to fix. When testing I noted that MQOM is missing in the list of NIST_SIG_ONRAMP and CROSS is still in. Do you want me to update that as well?

Please limit the scope of this pull request to SNOVA only. We can fix build configuration with respect to MQOM and/or CROSS at a later pull request.

> Another point: the zephyr currently excludes some of the SNOVA schemes. This is no longer needed as we now have a reduced memory version that uses less than 120KB even when signing at SL V. I will update this as well.

Thank you.

vacuas:
We have decided to remove the short signature variants for this pull request.
This also affects oqs-provider.

xuganyu96 (approved):
Can confirm that `-DOQS_ALGS_ENABLED=NIST_SIG_ONRAMP` now correctly enable SNOVA. Thank you!

baentsch:
Thanks @xuganyu96 for bringing this over the finishing line.

> We have decided to remove the short signature variants for this pull request. This also affects oqs-provider.

At first glance (CI-retriggered now turning green), this indeed unblocks https://github.com/open-quantum-safe/oqs-provider/pull/830 -- unfortunately, now MQOM doesn't build OK any more in `oqsprovider` for aarch64 in MQOM, still blocking progress there. Is this https://github.com/open-quantum-safe/liboqs/pull/2567 ? Will merge that tomorrow unless hearing howls of protest.

rben-dev:
@baentsch : The issue seems to be another one on top of #2567 related to ARM flags. In the CI, from what I see (to be confirmed), the problem lies in the fact that two `-march` are emitted on the same compiler invocations: `-march=armv8-a+crypto` first, and then `-march=armv8-a+crypto+sha3`. This seems to bring a confusion between the compiler pass and the assembler pass:
- The compiler emits `eor3` (specific to sha3 extensions) instructions because it has the `+sha3` flag, but because of the first flag it seems to produce a `.arch armv8-a+crypto` in the generated assembly file
- The assembler fails because the `.arch armv8-a+crypto` is not compatible with the emitted assembly

Anyways, the root cause of the issue seems to be in the way `liboqs` emits incompatible `-march` flags for the same compilation units:  unfortunately, for some compilers versions, the last emitted flag does not necessarily win (or produces inconsistent assembly).
The first aarch64 flag is emitted here: https://github.com/open-quantum-safe/liboqs/blob/cf096a27a729072c2bb40c2922c7533dcf717795/.CMake/compiler_opts.cmake#L63. Then the second one, target dependent, is emitted in `copy_from_upstream`: https://github.com/open-quantum-safe/liboqs/blob/cf096a27a729072c2bb40c2922c7533dcf717795/scripts/copy_from_upstream/src/sig/family/CMakeLists.txt#L16 and so on, discussed in the issues.

Since this touches sensitive parts of the `liboqs` compilation, the way flags overloading is dealt with should be discussed, maybe with @bhess and @xuganyu96? This is a not so easy topic because sometimes overloading flags is indeed necessary (e.g. for `x86_64` adding some options), and sometimes it can bring inconsistencies as in the current ARMv8 case ...

baentsch:
Thanks very much @rben-dev for looking into this! Can I suggest we start discussing this in a new issue instead of a merged PR? Ideal would be a standalone reproducer within `liboqs` alone: I don't understand why this problem only surfaces with `oqsprovider` and not the CI of this project. Did we prune too much of the CI here?

rben-dev:
@baentsch: should be raised (and the difference between `liboqs` CI and `oqs-provider` CI explained) in #2579

Issue #2556, opened 2026-09-08
Support Aarch64 AES acceleration without SHA-3 acceleration


rben-dev:
### Description

As of now, optimized Aarch64 builds support in the META files the `arm_sha3` flag that activates the `-march=armv8-a+crypto+sha3` flags when compiling, see for instance for the signatures:
 
https://github.com/open-quantum-safe/liboqs/blob/8c7f6592e097736c414b08eea12817b29399734f/scripts/copy_from_upstream/src/sig/family/CMakeLists.txt#L12

Wouldn't be interesting to have a dedicated `arm_aes` flag that would only activate the `crypto` flag: `-march=armv8-a+crypto`? As a matter of fact, Apple Silicon have the SHA-3 extensions but this is not the case for some other targets such as the ones in Raspberry Pi boards, while some of them have the AES acceleration.

### Describe the solution you'd like

Add a dedicated `arm_aes` flag when parsing the META files that activates `-march=armv8-a+crypto` without the SHA-3 extensions.

### Are you willing to help develop the solution?

Yes

### Use of generative AI

No

PR #2567, opened 2026-09-12
Fix AND separator in ARM64 required_flags CMake guard

iiviel:
## Problem

**Status: unreleased `main` only — no released version of liboqs is affected.** The defect
reached the generated file with the MQOM v3 integration (`e12880042`), which is not in any
release tag. See [Why it surfaced only now](#why-it-surfaced-only-now).

`.CMake/alg_support.cmake` contains 18 guards with two variable names run together:

```cmake
if(OQS_DIST_ARM64_V8_BUILD OR (OQS_USE_ARM_NEON_INSTRUCTIONS AND OQS_USE_ARM_NEON_INSTRUCTIONSOQS_USE_ARM_SHA3_INSTRUCTIONS))
```

`OQS_USE_ARM_NEON_INSTRUCTIONSOQS_USE_ARM_SHA3_INSTRUCTIONS` is not a variable. CMake's
`if()` auto-dereferences an unquoted argument that names a variable; an undefined one falls
back to its own name as a string, which is not a truthy constant, so it evaluates false.

The conjunction is therefore always false, and the affected NEON implementations can be
enabled **only** via `OQS_DIST_ARM64_V8_BUILD`. On a non-dist ARM64 build (`OQS_DIST_BUILD=OFF`),
setting `OQS_USE_ARM_NEON_INSTRUCTIONS` and `OQS_USE_ARM_SHA3_INSTRUCTIONS` does not enable
them — the build silently falls back to a portable implementation.

Confirmed by evaluating both forms in CMake directly, with both flags `ON` and
`OQS_DIST_ARM64_V8_BUILD=OFF`:

```
-- RESULT broken-guard  = FALSE
-- RESULT intended-guard = TRUE
```

All 18 affected guards are `mqom` `neon`.

## Root cause

Not a hand-edit — a separator bug in the generator template,
`scripts/copy_from_upstream/.CMake/alg_support.cmake/add_enable_by_alg_conditional.fragment`
(lines 50 and 101):

```jinja
{%- if not loop.last or loop.first %} AND {% endif -%}
```

This emits the separator *before* the first flag and *omits* it before the last — inverted
on both ends. The template already writes a leading `OQS_USE_ARM_NEON_INSTRUCTIONS`, so:

| `required_flags` | rendered |
|---|---|
| 1 flag | `NEON AND NEON` — the two errors cancel; harmless duplicate |
| 2 flags | `NEON AND NEON` + `SHA3` → **concatenated** |
| 3 flags | `NEON AND NEON AND SHA3` + `SHA2` → **concatenated** |

The x86 guard on lines 16/89 uses the conventional `{%- if not loop.last %} AND {% endif -%}`
placed *after* each flag and is unaffected.

### Why it surfaced only now

The condition was introduced in 096109052 ("Fix for alg_support.cmake", #1716, 2024-03-07)
and was latent for two and a half years, because rendering is only wrong when an ARM64_V8
platform declares **two or more** `required_flags`. Every NEON platform in-tree had exactly
one until MQOM v3 arrived, whose metadata declares `required_flags: [asimd, sha3]`
(mapped to `arm_neon` / `arm_sha3` by `copy_from_upstream.py:724-728`).

MQOM v3 (`e12880042`) is the first in-tree platform to declare two ARM flags; it is not in
any release tag, so no released version is affected and the bug is currently main-only.
The MQOM v2 integration that shipped in 0.16.0 carried no NEON implementation at all —
`.CMake/alg_support.cmake` at tag `0.16.0` has 41 ARM64 guards, every one single-flag, no
`mqom` `neon` guard among them, and zero occurrences of the concatenated token.

## Fix

Emit `" AND "` before every flag. Since the leading `OQS_USE_ARM_NEON_INSTRUCTIONS` is
already in the template, this is correct for any flag count:

```jinja
{% for flag in platform['required_flags'] %} AND OQS_USE_{{ flag|upper }}_INSTRUCTIONS{% endfor %}
```

Both the template (2 lines) and the generated `.CMake/alg_support.cmake` (18 lines) are updated.

**Single-flag platforms render byte-for-byte identically**, so the other 44 ARM guards in the
generated file (38 from the signature fragment, 6 from the KEM fragment) are untouched and the
diff is confined to the 18 broken ones. Verified by rendering the old and new templates against
1-, 2- and 3-flag inputs, and by confirming the patched generated lines are exactly what the
patched template produces. `if(`/`endif()` counts in the generated file are unchanged (339/344).

The surviving `NEON AND NEON` duplicate in single-flag guards is pre-existing and harmless;
removing the hardcoded leading token would touch all 62 ARM guards, so I left it out of this
PR. Happy to follow up separately if you'd like it cleaned up.

## Testing

- CMake evaluation of both guard forms (above).
- Jinja render comparison, 1/2/3 flags, old vs new template, both fragments.
- Patched generated lines confirmed identical to patched-template output.
- No `INSTRUCTIONSOQS_USE` tokens remain anywhere in the repo.
- Release-tag check confirming the concatenated token is absent from `0.16.0` and that
  `e12880042` is contained in no tag.

I do not have ARM64 hardware, so I have **not** run an end-to-end non-dist ARM64 build with
`OQS_USE_ARM_NEON_INSTRUCTIONS=ON OQS_USE_ARM_SHA3_INSTRUCTIONS=ON` to observe mqom's NEON
implementation being selected. That is the one check I would ask a reviewer with ARM64 CI to
confirm.

## PR template questions

* [ ] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)? — **No.** The change only affects which implementation a build is permitted to select; the NEON and portable implementations produce identical outputs, and no KAT values change.
* [ ] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? — **No.** No algorithm is added, removed or renamed, and no API changes. No downstream or benchmarking metadata updates are required.

## Generative AI declaration

Per CONTRIBUTING.md, declaring AI assistance: this change was produced with the help of
Claude (Anthropic) acting as a coding assistant. The AI located the defect, identified the
template as the root cause, bisected the introducing commit, wrote the template and
generated-file edits, constructed the CMake and Jinja verification cases, and drafted this
description. I reviewed the diff and the reasoning, re-ran the verification myself, and I
affirm the contribution.


baentsch (approved):
LGTM. Clear coding error in the jinja fragment. Thanks @iiviel and @claude for fixing this.

baentsch:
@iiviel could you please add DCO and re-base so this can land?

rben-dev:
Thanks @iiviel for spotting this. As a matter of fact, MQOM v3 declares the `sha3` flag mostly to have the `crypto` AES acceleration, see #2556 and #2565.

iiviel:
Rebased on main and added DCO sign-off, thanks @baentsch.

xuganyu96 (approved):

Issue #2143, opened 2025-05-25
Missing support for parameter sets that use SHA256/192 hashes from draft-fluhrer-lms-more-parm-sets


ymoyal1:
NIST.SP.800-208 and [draft-fluhrer-lms-more-parm-sets/](https://datatracker.ietf.org/doc/draft-fluhrer-lms-more-parm-sets/) define extra parameter set for LMS and HSS algorithms. 

Support for these parameter sets was added to https://github.com/cisco/hash-sigs/tree/192 also. 

The issue is asking to add the support to liboqs as well.

dstebila:
We'd be happy to have them. Would you consider submitting a pull request with the code?

h2parson:
It seems like this has already been supported in Cisco's branch:
https://github.com/cisco/hash-sigs/tree/shake-support

So we will need to pull these changes into libOQS

ashman-p:
@h2parson began work on this, see https://github.com/open-quantum-safe/liboqs/pull/2207
But the effort is being deferred.

PR #2569, opened 2026-09-13
Bind LMS verify to the variant the caller invoked

### LINKED ISSUE #2568: LMS verify does not bind the declared variant to the public key: every LMS object accepts every other variant's signatures
### Description

`sig->verify` is the same function pointer for all 33 LMS variants, so at verify time nothing knows which variant the caller declared. Every LMS `OQS_SIG_STFL` object therefore accepts a genuine public key and signature from every other LMS variant.

**This is an API-contract defect, not a vulnerability, and I am not asking for a CVE or an advisory.** The contract is that `OQS_SIG_STFL_verify` checks the object it was called on; today, for LMS, it checks whatever the public key happens to name. Why it is not a security issue is spelled out under "What this is not" below.

### Root cause

`src/sig_stfl/lms/sig_stfl_lms.c:89` assigns one shared entry point inside the `LMS_ALG` generator macro:

```c
sig->length_signature = OQS_SIG_STFL_alg_lms_##lms_variant##_length_signature; \
\
sig->verify = OQS_SIG_STFL_alg_lms_verify; \
```

The variant's identity is right there on the line above, where `sig->length_signature` is set from the same macro parameter, and is then discarded. `OQS_SIG_STFL_alg_lms_verify` takes no variant, and `hss_validate_signature_init()` derives the level count, the LMS type and the LM-OTS type from the public key alone, so the declared parameter set never enters the decision.

XMSS is built from the same kind of macro but gives each variant its own wrapper, `OQS_SIG_STFL_alg_xmss##xmss_v##_verify`, which closes over that variant's compile-time constants. That wrapper is exactly where both XMSS fixes landed — `signature_len` against `..._length_signature` in ef70dea (CVE-2026-44518), and `pk[0..3]` against `..._oid` in 077e32a (CVE-2026-46344). LMS has no such wrapper, so neither check had anywhere to go.

### Measurement

For each row, a genuine key and signature were generated with that variant and then verified through every enabled object of the family. `foreign-accepted` counts objects other than the signer's own that returned `OQS_SUCCESS`.

```
LMS  H5_W1 / H5_W2 / H5_W4 / H5_W8 / H5_W8_H5_W8   foreign-accepted 32/32 each
XMSS SHA2_10_256 / SHA2_16_256 / MT-SHA2_20/2_256  foreign-accepted  0/36 each
```

Cross-level confusion is included: a single-tree signature is accepted by a two-level HSS object and the reverse. Concretely, an `LMS_SHA256_H5_W8` object, whose `length_signature` is 1296, returns `OQS_SUCCESS` for a 4464-byte `LMS_SHA256_H5_W2` signature.

### What this is not

Stated up front so it does not get re-litigated:

* **Not memory-unsafe.** A mismatched typecode cannot reach past either buffer. Every offset the parser derives is re-checked against the caller's `signature_len`; all five LMS parameter sets are `n = 32`, so the 56-byte `memcpy` into `ctx->final_public_key` is exactly `lm_get_public_key_len()`; and `lm_verify.c:77` requires `signature_len == n * height`, which rejects over-long input as well as short. The bounds question was settled by #2507 (and upstream cisco/hash-sigs#40); credit there, not here.
* **Not the `(void)` casts.** `oqs_sig_stfl_lms_verify` discards the return values of `hss_validate_signature_init` and `_update`, but that is a readability wart rather than a defect: `ctx->status` is set at `hss_verify_inc.c:112` before any early return and re-set on every failure path, and both `_update` (`:248`) and `_finalize` (`:268`) return early on it, so the context is never partially consumed.
* **Not a downgrade.** All 33 LMS parameter sets are SHA-256 with `n = 32`, differing only in Merkle tree height and Winternitz `w`. There is no weaker set to steer a verifier into, unlike XMSS where the same gap spans 192/256/512-bit sets and SHA-2/SHAKE.
* **Needs an attacker-controlled public key**, which is close to assuming the conclusion: whoever supplies the key supplies the trust anchor. No forgery, crash, or denial of service was demonstrated, and verification work stays bounded by the caller's own `signature_len`.

### Negative result, and the controls that make it meaningful

92,148 verify calls across `H5_W1`, `H5_W2`, `H5_W8` and the two-level `H5_W8_H5_W8` — pk header mutation with a correct-size signature, truncation sweeps, signature header mutation, and a 4-byte probe window walked across an entire genuine signature and public key — produced **zero AddressSanitizer and zero MemorySanitizer reports** on current `main`, with the public key and the signature each heap-allocated at exactly the length passed to verify.

That is only worth reporting because the detectors were shown to fire:

* Reverting #2507 and re-running the same harness reproduces the known over-read immediately:
  ```
  ==123822==ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 1
      #0 OQS_LMS_NAMESPACE_get_bigendian        external/endian.c:20:34
      #1 is_hss_public_key                      external/hss_verify_inc.c:40:48
      #2 public_key_levels                      external/hss_verify_inc.c:69:6
      #3 OQS_LMS_NAMESPACE_hss_validate_signature_init  external/hss_verify_inc.c:119:32
  ```
* Removing the `ctx->status` initialiser *and* the assignment on the truncation path makes MemorySanitizer report `use-of-uninitialized-value` at `hss_verify_inc.c:248` through this exact call path, which is silent on unmodified `main`.

Related: #2528 / #2539 note that the stateful fuzz harnesses pass a fixed signature length, so this class of input is not reachable from them today either.

### Expected behaviour

`OQS_SIG_STFL_verify(sig, ...)` should reject a public key or a `signature_len` that does not belong to the variant `sig` was created for, the way the XMSS wrappers already do.

### Why fix it anyway

The contract is that verify checks the object it was called on, and an application that pins `OQS_SIG_STFL_new("LMS_SHA256_H5_W8")` has no way to learn that the pin was ignored. More concretely, the harmlessness above rests entirely on every enabled parameter set sharing SHA-256 and `n = 32`. The moment a set with a different `n` or a different hash is added — #2143 asks for exactly that, the SHA-256/192 sets from draft-fluhrer-lms-more-parm-sets, where `n = 24` — the 56-byte `final_public_key` copy stops being exact and this turns from theoretical into a real over-read. Fixing the binding now is much cheaper than remembering this when those sets land.

### liboqs version

`main` at adbeba1ef (2026-09-14).

### Environment

x86_64 Linux, clang 21.1.8, liboqs 0.16.0, `-DOQS_ENABLE_SIG_STFL_LMS=ON -DOQS_ENABLE_SIG_STFL_XMSS=ON`, with and without `-DOQS_HAZARDOUS_EXPERIMENTAL_ENABLE_SIG_STFL_KEY_SIG_GEN=ON`, under `-DUSE_SANITIZER=Address` and `-DUSE_SANITIZER=Memory`.

### Use of generative AI

Yes. This was prepared with Claude Code (Anthropic): it ran the measurement, the sanitizer builds and the positive controls, and drafted this text. I reviewed the analysis and the numbers before filing.

### Additional information

Fix in the linked PR. cc @ashman-p as the LMS code owner. It touches `sig_stfl_lms.c` and `tests/test_sig_stfl.c`, which #2520 also edits (a different hunk of the same `LMS_ALG` macro), so one of the two will need a trivial rebase depending on merge order.

iiviel:
## Purpose

Fixes #2568.

`sig->verify` was the single shared `OQS_SIG_STFL_alg_lms_verify` for all 33 LMS variants, so at verify time nothing knew which variant the caller declared. `hss_validate_signature_init()` derives the level count and both parameter-set type codes from the public key alone, so every LMS `OQS_SIG_STFL` object accepted every other LMS variant's public key and signature:

```
LMS  H5_W1 / H5_W2 / H5_W4 / H5_W8 / H5_W8_H5_W8   foreign-accepted 32/32 each
XMSS SHA2_10_256 / SHA2_16_256 / MT-SHA2_20/2_256  foreign-accepted  0/36 each
```

Including across HSS level counts, and across signature lengths: an `LMS_SHA256_H5_W8` object, whose `length_signature` is 1296, returned `OQS_SUCCESS` for a 4464-byte `LMS_SHA256_H5_W2` signature. XMSS rejects all of these because its per-variant wrappers check `signature_len` (ef70dea, CVE-2026-44518) and the public-key OID (077e32a, CVE-2026-46344); LMS had no per-variant wrapper for those checks to live in.

**This is an API-contract defect, not a vulnerability, and I am not asking for a CVE or an advisory.** All 33 LMS parameter sets are SHA-256 with `n = 32`, differing only in tree height and Winternitz `w`, so there is no weaker set to steer a verifier into; it needs an attacker-controlled public key, which is close to assuming the conclusion; and no forgery, crash or denial of service was demonstrated. The bounds question in the parser is separately settled by #2507 — 92,148 verify calls with the public key and the signature each allocated at exactly the length passed in produced no ASan and no MSan report on current `main`. The reasoning, and the positive controls that make that negative result mean something, are in #2568.

It is still worth fixing: the contract is that verify checks the object it was called on, and an application that pins `OQS_SIG_STFL_new("LMS_SHA256_H5_W8")` has no way to learn the pin was ignored. The harmlessness also rests entirely on every enabled set sharing `n = 32` — #2143 asks for the SHA-256/192 sets where `n = 24`, at which point the fixed 56-byte `final_public_key` copy stops being exact.

## Change

`LMS_ALG` now emits a per-variant verify wrapper and `sig->verify` points at it. The wrapper rejects a `signature_len` other than the variant's own `length_signature`, and a public key that does not name the variant, before anything derives offsets from the key — the same placement as the XMSS wrapper check in 077e32a. The three values to compare are already encoded in `OQS_LMS_ID_*`, whose nibbles are the RFC 8554 registry codes (`0x0L H(1)W(1) ...`; H5=5..H25=9, W1=1..W8=4), so no new table is needed.

Two details worth a reviewer's attention:

1. **Both public-key serialisations are accepted.** `public_key_levels()` handles both and both reach this API: the HSS form `u32(L) || u32(lm_type) || u32(lm_ots) || I || T[1]` that `keypair()` emits, and the bare RFC 8554 LMS form without the level prefix. The NIST verify vectors added in #2435 use the bare form for `LMS_SHA256_H5_W1` and `LMS_SHA256_H10_W1` (the other 16 KATs use the HSS form), so a naive three-field compare regresses exactly those two KATs — it did, on the first draft of this patch. The two forms cannot be confused: the HSS form opens with a level count in 1..8, the bare form with an LMS type code in 5..9.

2. **`OQS_SIG_STFL_alg_lms_verify` is unchanged and still exported.** It is the family-generic entry point and has no variant to bind to, so it keeps its current parse-what-the-key-says behaviour. The new wrappers are `static`; only `sig->verify` changes, so this adds no public symbols and no API change.

## Testing

Added the cross-variant case to `test_invalid_sig_lms`, mirroring the sub-case `test_invalid_sig` grew for GHSA-2wxh-55qf-c7wg: sign with `LMS_SHA256_H5_W2`, then require both `LMS_SHA256_H5_W8` (differs in `w` and in `length_signature`) and `LMS_SHA256_H5_W8_H5_W8` (differs in HSS level count) to reject it, with a control asserting the signer's own object still accepts. It needs a genuine signature, so it is a no-op without `OQS_HAZARDOUS_EXPERIMENTAL_ENABLE_SIG_STFL_KEY_SIG_GEN`; a crafted-buffer test cannot distinguish fixed from unfixed here, because unpatched LMS rejects a bogus signature on its own merits rather than crashing.

Against the unpatched source with the test in place:

```
ERROR: LMS_SHA256_H5_W8 object accepted a genuine LMS_SHA256_H5_W2 signature
EXIT=1
```

With the fix, on x86_64 Linux / clang 21.1.8, `-DUSE_SANITIZER=Address`, LMS and XMSS enabled:

* all 18 LMS KATs and all 37 XMSS KATs unchanged;
* `test_sig_stfl` passes for all 18 LMS parameter sets with key generation off, and for 14 with it on (H15–H25 keygen is too slow to run locally; they are covered as verifiers via the KATs);
* the cross-variant matrix goes from 32/32 foreign-accepted to **0/32**, diagonal intact;
* clean build under `-DOQS_STRICT_WARNINGS=ON`;
* formatted with `astyle --options=.astylerc` (reports `Unchanged`).

## Notes

cc @ashman-p as the LMS code owner.

This touches `src/sig_stfl/lms/sig_stfl_lms.c` and `tests/test_sig_stfl.c`, which #2520 also edits — a different hunk of the same `LMS_ALG` macro (`sk->deserialize_key` vs `sig->verify`) and a different part of the test file. Whichever lands second will need a trivial rebase; happy to do that on this one either way. #2520 and #2511 close the secret-key side of the same binding question for LMS and XMSS respectively, so with this the four-way matrix (XMSS/LMS × verify/sign) is complete.

Neither box applies: KAT values are unchanged (all 18 LMS and 37 XMSS KATs pass as-is), and no algorithm is added, removed or renamed. The new wrappers are `static` and no public symbol changes, so oqs-provider is unaffected — it does not reach LMS today in any case (open-quantum-safe/oqs-provider#331).

## Use of generative AI

Yes. Prepared with Claude Code (Anthropic): it ran the cross-variant measurement and the sanitizer builds, wrote the wrapper and the regression test, and drafted this description and the issue. I reviewed the change and the test results before submitting.

https://claude.ai/code/session_01RryAB2JbjNYt6N45sM2dcA


ashman-p:
@iiviel, thanks for the code changes. I need to think about it some more. I think the code change is good for LMS but may not address HSS (multi-tree).  
I will get back on this.

PR #2571, opened 2026-09-15
Add AES 192 support.

### LINKED ISSUE #2570: AES 192 Support
### Description

I am investigating the integration of FAEST in liboqs. 

FAEST however has a variant that uses AES192.

While OpenSSL has support for AES 192, not all users will make use of OpenSSL.

https://faest.info/#:~:text=Variants,or%20AES%2D256.

To begin with add support for AES 192 ECB mode.

_No response_

_No response_

### Use of generative AI

_No response_

ashman-p:
Add AES 192 support similar to AES 128 and AES 256.
AES 192 is needed to support FAEST.

AES 192 was sourced from BearSSL, the same code base used for AES 128/256. AES-NI and ARMv8 were handled in the same manner.

Fixes #2570

Updates AES in the common directory. 
4 additional entries where made in struct OQS_AES_callbacks

AI was used to replicate the AES function patterns. Example, added OQS_AES192_ECB_xxx based on OQS_AES128_ECB_xxx.


sfluhrer (approved):
Looks good.  Including the FIPS 197 test vector means that the algorithm is almost certainly right (I assume you tested it on all three environments).  The only remaining possibility is a memory leak or something similar - that looks unlikely.

ashman-p:
> Looks good. Including the FIPS 197 test vector means that the algorithm is almost certainly right (I assume you tested it on all three environments). The only remaining possibility is a memory leak or something similar - that looks unlikely.



> Looks good. Including the FIPS 197 test vector means that the algorithm is almost certainly right (I assume you tested it on all three environments). The only remaining possibility is a memory leak or something similar - that looks unlikely.

@sfluhrer, thanks for taking a look at this. My dev tests covered C, ARM, Intel (handled via container). No inherent leaks found.

src/common/ossl_functions.h
```diff
@@ -44,6 +44,7 @@ FUNC(EVP_MD_CTX *, EVP_MD_CTX_new, (void), ())
 FUNC(int, EVP_MD_CTX_reset, (EVP_MD_CTX *ctx), (ctx))
 FUNC(const EVP_CIPHER *, EVP_aes_128_ecb, (void), ())
 FUNC(const EVP_CIPHER *, EVP_aes_128_ctr, (void), ())
+FUNC(const EVP_CIPHER *, EVP_aes_192_ecb, (void), ())
 FUNC(const EVP_CIPHER *, EVP_aes_256_ecb, (void), ())
 FUNC(const EVP_CIPHER *, EVP_aes_256_ctr, (void), ())
```
RodriM11, line 49: Question: AES-192-CTR is not needed, right? What uses are the CTR variants having right now?

RodriM11:
Hi @ashman-p ! The PR LGTM, thank you for that. I have left a very naive question, just out of curiosity.

RodriM11 (dismissed):

src/common/ossl_functions.h
ashman-p: @RodriM11, FAEST only needs ECB therefore, I started with that requirement. As other requirements arise, we can implement them as needed. I am not readily aware of which algorithms would use counter mode.

RodriM11:
hi! can we get another pair of eyes here? @open-quantum-safe/liboqs-maintainers so we are able to merge this and start with FAEST inclusion.

dstebila:
I've pushed a few small changes. Otherwise looks okay to me.

dstebila (approved):

xuganyu96 (approved):

dstebila:
I will have a follow-up PR with further performance improvements.

Issue #2453, opened 2026-06-03
Adding SDitH


RodriM11:
### Description

This issue tracks inclusion of NIST on-ramp signature algorithm [SDitH](https://sdith.org/index.html) into liboqs.

RodriM11:
The SDitH team will join the September 8th Status Meeting to start discussions on SDitH inclusion.

RodriM11:
I have been toying a little bit with Claude with regards to SDitH inclusion. What about these kinds of `liboqs` wrappings for the code provided by SDitH?

```
#include <oqs/sig.h>
#include "sdith_signature.h"  /* from the vendored SDitH source */

OQS_STATUS OQS_SIG_sdith_cat1_short_keypair(uint8_t *public_key, uint8_t *secret_key) {
    const signature_parameters *p = &CAT1_SHORT_PARAMETERS;
    uint8_t entropy[/* sdith_keygen_entropy_bytes(p), compile-time-known per scheme */];
    uint8_t tmp[/* sdith_keygen_tmp_bytes(p) */];
    OQS_randombytes(entropy, sizeof(entropy));
    sdith_keygen(p, secret_key, public_key, entropy, tmp);
    OQS_MEM_cleanse(tmp, sizeof(tmp));
    return OQS_SUCCESS;
}
/* ...sign / verify follow the same shape, forwarding to sdith_sign / sdith_verify */
```

Should macros for the lengths be employed or we continue to use the runtime SDitH procedures like `sdith_keygen_tmp_bytes`?

owaldron:
Hey @RodriM11, I've been taking a look at this and have a way forward with this same wrapper approach you suggested. I can keep working through the integration and report my findings here and in the next status call

RodriM11:
thanks @owaldron !

owaldron:
I've had some success with the wrapper approach and shimming FIPS202 and AES. There's still some outstanding work, including tests to evaluate the impact of glue code, but in the meantime I opened a draft PR #2574 for feedback

RodriM11:
wow @owaldron ! Thank you very much for the hyper quick PR!!

owaldron:
Hey @RodriM11, I've opened a PR #2586 ready for review.

[...]

PR #2574, opened 2026-09-17
[WIP] SDitH3 integration

owaldron:
## (WIP) SDitH3 Integration (#2453)

This PR experiments with SDitH3 integration into liboqs. The corresponding WIP upstream branch is at [owaldron:feat/liboqs-integration](https://github.com/owaldron/sdith/tree/feat/liboqs-integration).

For completeness sake, this features FIPS202 and AES shims, although the drawbacks to either performance or simpicity may not justify their inclusion in the final integration. Performance tests are still a WIP, but KAT tests are all passing for Apple arm64 and Linux x86 (including avx2) builds.

Here are notes I kept for different design decisions made:

### Overall design
1. Each scheme wraps the `sdith_signature.h` with the appropriate parameters in `api.h/c`
2. Common sources are put into `sdith_common`. The sources files are all common for each scheme (including the wrapper), and only the compile flags change behaviour
3. SDitH's current `avx2` implementations do **not** have  Darwin x86 support, and only have linux support.
	- This was confirmed by attempting to build the sdith repo on Darwin x86, which fails
	- In the SDitH repo, any x86 arch enables the `X86_LINUX` flag, which toggles the inclusion of some linux-only headers

### FIPS202 Shimming
1. `fips202_glue.h` and `fips202x4_glue.h` provide wrappers for the XOF function calls, which selects the 
	- This is required since SDitH uses a different FIPS202 api than provided by liboqs, which are generic with respect to `shake128/256`
	- These wrappers largely select whether to use the 128/256 liboqs functions
2. Added functions  `xof_ctx_release_shake128/256` to release liboqs' heap context when finished 
	- Added appropriate callsites throughout SDitH as well
	- This expands to a no-op for non-liboqs 
- Ablating these changes is a WIP

### **AES Shimming**
1. The avx2 batched AES calls were not replaced with liboqs calls, since they define their own hardware-enable parallelization that aren't reproducible via the liboqs aes api
	- There is a VAES implementation for batch CTR mode encryption with a custom counter (litte-endien with no carry) and different keys per block
	- This isn't possible with liboqs since:
		- ECB mode does not offer a batched api
		- Although there is batched CTR mode, it does not accept custom counters nor different keys
	- There is also parallelized key schedule expansions, which isn't in liboqs
2. The default reference implementation only runs AES in serial, and this was shimmed to liboqs calls
	1. Since the liboqs key schedules are heap-allocated and opaque (unlike sdith's), they are alloc'd, generated and free'd all at encryption time
		1. `aes128_set_key_ref` stores the raw bytes of the initial key in the round-key buffer, since a liboqs round key cannot be pulled out of the opaque struct
		2. Generating the real round keys is deferred to `aes128_nocarry_ref_impl`, where they are used immediately for encryption
			- This does add alloc/free overhead during each `aes128_nocarry_ref_impl` call
	- Ablating this change WIP

****

* [X] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? (If so, PRs in fully supported downstream projects dependent on these, i.e., [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) will also need to be ready for review and merge by the time this is merged. Also, make sure to update the list of algorithms in the continuous benchmarking files: .github/workflows/kem-bench.yml and sig-bench.yml)

****
**Generative AI Disclosure**: generative AI was used throughout the development process to aid with understanding, drafting, debugging,  and reviewing code. Any code generated by AI was reviewed and possibly modified before inclusion.


owaldron:
Closed until changes are ready for review

RodriM11:
hi @owaldron ! What is left to do before it is ready for review?

owaldron:
Hi @RodriM11, Bruce and I went through the code together and want to tweak how AES is shimmed. It should be ready for review again shortly

PR #2576, opened 2026-09-19
common: include opensslconf.h so OPENSSL_NO_STDIO is actually visible

### LINKED ISSUE #2057: opensslconf.h not used

Johnsonajibi:
Fixes #2057.

`common.h` selects between two `OQS_OPENSSL_GUARD()` implementations based on `#ifdef OPENSSL_NO_STDIO`, but never includes any OpenSSL header itself. `OPENSSL_NO_STDIO` only ever gets defined by `<openssl/opensslconf.h>` (pulled in transitively by `<openssl/crypto.h>` and others).

In every real consumer of this macro (`aes_ossl.c`, `sha2_ossl.c`, `rand_nist.c`, `ossl_sha3.c`, `ossl_sha3x4.c`), `oqs/oqs.h` -- and therefore `common.h` -- is included *before* `../ossl_helpers.h`, which is what actually pulls in `<openssl/crypto.h>`. So by the time the preprocessor reaches this `#ifdef`, `OPENSSL_NO_STDIO` has never been defined yet, regardless of how the linked OpenSSL was actually built. The check is structurally a no-op that always selects the non-no-stdio branch.

I confirmed this with a minimal repro mirroring the real include order (`oqs.h`'s macro check happening before `ossl_helpers.h`'s `<openssl/crypto.h>` include): with a stub OpenSSL that defines `OPENSSL_NO_STDIO`, the "before" version always resolves the wrong branch; the "after" version (this PR's fix) correctly resolves it. Also did a full `-DOQS_USE_OPENSSL=ON` debug build against the system OpenSSL 3.0.13 to confirm nothing else regresses -- builds clean, 3306/3306 objects.

Fix: include `<openssl/opensslconf.h>` directly in `common.h`, guarded on `OQS_USE_OPENSSL`, so `OPENSSL_NO_STDIO` reflects the real build configuration of the linked OpenSSL regardless of what a given translation unit's include order happens to be. This is the fix suggested in #2057 by the issue reporter and endorsed by @baentsch there.

I searched the rest of the tree for the same pattern (`OPENSSL_NO_STDIO`, `OPENSSL_VERSION_NUMBER`, `LIBRESSL_VERSION_NUMBER`) -- all other usages are in `.c` files that already include `ossl_helpers.h`/OpenSSL headers earlier in the same translation unit, so they don't have this ordering problem. This appears to be the only instance.

An earlier attempt at this (#2206) was closed without merging; this PR is scoped to just the `#2057` fix.


baentsch (changes requested):
Thanks for the PR, @Johnsonajibi -- unfortunately, this cannot work as-is as it would create an OpenSSL depenency into a `liboqs` common API that does not exist yet (`liboqs` conceptually can be built without OpenSSL -- not sensible, but possible), so this fix needs to be relocated to an internal header file, e.g., ossl_helpers.h.

Johnsonajibi:
Good point, thanks. Moved OQS_OPENSSL_GUARD (and the OPENSSL_NO_STDIO check) into ossl_helpers.h instead -- that header is already OQS_USE_OPENSSL-only and already includes <openssl/crypto.h> before anything in it references OPENSSL_NO_STDIO, so the ordering bug is fixed by construction without common.h gaining any OpenSSL dependency. common.h no longer touches OpenSSL at all now.

Re-verified with the same before/after repro as before, and did a full OQS_USE_OPENSSL=ON build -- sha2_ossl.c, ossl_sha3.c, ossl_sha3x4.c and ossl_helpers.c all call OQS_OPENSSL_GUARD directly and compile clean from the new location.

baentsch (approved):
Thanks for the update @Johnsonajibi . This now LGTM and ready to merge (pending CI passing).

xuganyu96 (commented):
@baentsch @Johnsonajibi

I am not intimately familiar with `OPENSSL_NO_STDIO`. From its name I assume it means `stdio.h` facilities are not available from within the OpenSSL code base, so it seems odd to me to build OpenSSL without stdio but still use `stderr` in liboqs. Do you think we should remove the `fprintf(stderr, ...)` call altogether under the `OPENSSL_NO_STDIO` branch?

Johnsonajibi:
Good question. The two things are independent: `OPENSSL_NO_STDIO` only describes how *OpenSSL itself* was built (its own internals don't link against stdio, which matters for e.g. embedded/no-libc-stdio targets) -- it says nothing about whether the *calling application* can use stdio. liboqs is a separate translation unit linked against its own libc, so `fprintf(stderr, ...)` here is unaffected by how the linked OpenSSL was configured.

What actually changes between the two branches is `ERR_print_errors_fp(stderr)` -- that's an OpenSSL API that itself takes and uses a `FILE*` internally, so it may not be safe to call (or may not even be linked in) when OpenSSL was built with `no-stdio`. That's the only reason for the branch: skip that one OpenSSL call, keep liboqs's own `fprintf` either way.

baentsch:
Thanks for the explanation @Johnsonajibi . A rebase seems to be in order plus a second approval to get this to "all green".

xuganyu96 (approved):
> it says nothing about whether the _calling application_ can use stdio. liboqs is a separate translation unit linked against its own libc, so `fprintf(stderr, ...)` here is unaffected by how the linked OpenSSL was configured.

If a user is building OpenSSL without stdio, it's reasonable to assume that it might also want to build liboqs without stdio. The current fix calls `fprintf(stderr, ...)` regardless of that, which I think might make it difficult for the user. At this moment I am thinking about gating liboqs' calls to `fprintf(stderr, ...)` behind a compile-time flag if not removing this call altogether, but that's probably way out of scope for this pull request.

Johnsonajibi:
Thanks both. On the rebase: GitHub shows this branch as 0 commits behind main (2 ahead), so it should already be current. If you meant squashing the two commits (the first one added the include to common.h, the second moved the whole thing to ossl_helpers.h), I'm happy to do that -- just say so, since a force-push may dismiss the existing approvals.

@xuganyu96 agreed that gating liboqs's own stderr output behind a compile-time flag is a reasonable follow-up, and I'd keep it out of this PR so the fix stays scoped to making the existing OPENSSL_NO_STDIO check actually take effect.

PR #2578, opened 2026-09-19
Pin charset-normalizer in requirements.txt

### LINKED ISSUE #2545: pip install --require-hashes -r requirements.txt fails: charset_normalizer not pinned
The Quickstart tells new contributors to make a venv and run
`pip install --require-hashes -r requirements.txt` (README.md:117 for Ubuntu,
README.md:124 for macOS, and the same line again in requirements.txt:2). That
command does not complete. `requests==2.34.2` depends on `charset_normalizer`,
which is not pinned or hashed anywhere in the file, so pip refuses the whole
install:

```
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    charset_normalizer<4,>=2 from https://files.pythonhosted.org/packages/e9/40/095ce62fa078483cccc1fa2b36e6bc9580b85422a20ee9f925341c50e44f/charset_normalizer-3.5.1-cp314-cp314-macosx_10_15_universal2.whl (from requests==2.34.2->-r requirements.txt (line 116))
```

Dropping the flag does not help. pip switches hash checking on by itself as
soon as a requirements file contains any hash, so plain
`pip install -r requirements.txt` fails the same way.

```
python3 -m venv .venv && source .venv/bin/activate
pip install --require-hashes -r requirements.txt    # exits 1
```

`charset_normalizer` is the only entry missing, and both sibling requirements
files pin the same `requests` version and do list it.

Nothing in CI installs the root file. Every pip install under
`.github/workflows` reads either `.github/workflows/requirements.txt` or
`requirements-linter.txt`, so the command the README hands to new contributors
is never exercised, which is probably why this has gone unnoticed.

There is also no `requirements.in` at the repository root, even though
requirements.txt:4 says the file was generated from one and
docs/PROCEDURES.md:29 asks for both to be checked in. The two sibling
directories each have theirs.

#2524 noticed this in passing and offered to file it separately; that offer
hasn't been picked up, so filing it here.

How would you like this handled?

MrWater00:
requests==2.34.2 depends on charset-normalizer, but it was never pinned or hashed in the root requirements.txt, so
'pip install --require-hashes -r requirements.txt' — the exact command README.md's Quickstart tells new contributors to run — failed on a fresh install.

Added the missing requirements.in at the repository root (per docs/PROCEDURES.md, which documents this file should exist but didn't), listing the four direct dependencies (pytest, pytest-xdist, pyyaml, requests), then regenerated requirements.txt via 'pip-compile --generate-hashes requirements.in' as documented.

Fixes #2545

- [x] No — this PR only touches Python packaging metadata (requirements.txt/requirements.in), not any cryptographic algorithm implementation.
- [x] No — no algorithm list or API changes.

This fix was developed with the help of Claude (Anthropic), which explained
the root cause and pip-tools' workflow, and reviewed each step; I ran every
command myself and verified the fix on a clean install before submitting.


baentsch (approved):
LGTM. Thanks for bringing documentation and code in line again @MrWater00 !

xuganyu96 (approved):

PR #2582, opened 2026-09-20
Add QR-UOV signature schemes

### LINKED ISSUE #2454: Adding QR-UOV
### Description

This issue tracks inclusion of NIST on-ramp signature algorithm [QR-UOV](https://info.isl.ntt.co.jp/crypt/qruov/) into liboqs.

0rlych1kk4:
Adds QR-UOV Round 3 signature schemes to liboqs.

Closes #2454 .

This integrates all 30 QR-UOV parameter-set / PRG combinations, covering the
15 Round 3 parameter sets with both AES and SHAKE PRGs.

The integration includes:

- ref, opt, and AVX2 implementations
- liboqs primitive integration for SHAKE and AES-CTR
- `OQS_randombytes` integration
- detached signature and verification wrappers
- symbol namespacing for coexisting variants
- AVX2 runtime dispatch gated on AVX2, BMI2, and AES CPU features
- liboqs KAT entries for all 30 schemes

The QR-UOV source is currently pinned to:

`de988768f22f894f5a04bd7773fd2dbd9ba1d10f`

### Validation

Tested from a clean liboqs build on macOS/Apple Silicon:

- 30/30 QR-UOV signature runtime tests passed
- 30/30 liboqs KAT checks passed
- `git diff --check` passes
- importer regeneration from the pinned QR-UOV source is reproducible
- AVX2/BMI2/AES runtime dispatch was verified in the generated wrappers

### Version / API questions

- [ ] Does this PR change the input/output behaviour of an existing cryptographic algorithm (i.e., does it change known answer test values)?

No existing algorithm behaviour or KAT values are changed. This PR adds new
QR-UOV algorithms.

- [x] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API?

Yes. This PR adds the QR-UOV signature schemes to the list of algorithms
available through the liboqs signature API.

### Generative AI disclosure

Generative AI was used as a development aid for documentation drafting. All
changes were reviewed by me, and the resulting implementation was built and
validated locally, including runtime signature tests and KAT verification.


RodriM11:
Hi @0rlych1kk4 ! first of all, let me thank you sincerely for the effort put into this, and for helping us with this endeavor. From a quick look, and based upon the number of new files present, is there a way to group common code in any way? +2k review files look spooky....

I have seen how the repo of the QR-UOV team works and it is true that a lot of files are exact, but they depend on compilation macros which represent different parametrization, so that effort may be be difficult.

0rlych1kk4:
Thanks @RodriM11 . Yes, most of the file count comes from the importer materializing the same QR-UOV sources separately for each parameter set, PRG, and implementation. The upstream implementation selects the parameterization primarily through compile-time macros, so there should be room to reduce duplication.
I’ll look at restructuring the integration so common sources are shared and only the parameter-specific configuration is generated per variant. I’ll keep the current namespacing and ref/opt/AVX2 separation intact while checking what can safely be consolidated.

I’ll also work through the CI failures after that restructuring, since reducing the generated source layout may resolve some of them at the same time.

vacuas:
You may want to take a look at the MQOM or SNOVA source code to recuce the number of files.
SNOVA went down from some 700+ files at R2 to about 120 for R3.

RodriM11:
another question @0rlych1kk4 ! Where are the integration files for liboqs allocated (i.e. the ones that the QR-UOV people would have to maintain somewhere)?

0rlych1kk4:
Thanks @RodriM11 . I reworked the QR-UOV import to use shared common_dep source groups instead of materializing the full source tree for every parameter set and implementation.
The 30 variants are still preserved, but each per-variant implementation directory now contains only the small signature.c wrapper, with common/ref/opt/AVX2 sources shared centrally.

I revalidated the updated layout with a fresh build, all 30 test_sig runs, all 30 deterministic KATs, importer verification (395/395), and the datasheet/memory/SPDX checks.

0rlych1kk4:
@RodriM11  The liboqs-specific integration files are maintained in the QR-UOV upstream repository under integration/liboqs/.
That directory contains the per-variant metadata, META_Common.yml, namespacing, the OQS_randombytes shim, and the detached-signature wrapper used by the liboqs importer.

I also reworked the import to use shared common_dep source groups, so the full QR-UOV source tree is no longer materialized for every parameter set and implementation. The 30 variants remain, while the common/ref/opt/AVX2 sources are shared centrally.
I revalidated the updated layout with a fresh build, all 30 test_sig runs, all 30 deterministic KATs, importer verification (395/395), and the datasheet/memory/SPDX checks.

RodriM11:
> Thanks @RodriM11 . I reworked the QR-UOV import to use shared common_dep source groups instead of materializing the full source tree for every parameter set and implementation.
The 30 variants are still preserved, but each per-variant implementation directory now contains only the small signature.c wrapper, with common/ref/opt/AVX2 sources shared centrally.

Thank you a lot @0rlych1kk4 ! May I ask how have you worked around QR-UOV structure? Because they were dependent on the `qruov.h` activating a particular variant based upon compilation macros and that seems to be the case still.

> The liboqs-specific integration files are maintained in the QR-UOV upstream repository under integration/liboqs/.
That directory contains the per-variant metadata, META_Common.yml, namespacing, the OQS_randombytes shim, and the detached-signature wrapper used by the liboqs importer.

I include the link to your copy of the QR-UOV [repo](https://github.com/0rlych1kk4/round3/tree/feat/liboqs-integration/integration/liboqs) with the inclusion files just to ease the review for other members. Thanks!

paschal533:
Hi @0rlych1kk4 Great work here! I ran this on x86-64. Head 078c4d3c, Windows 11 x86-64, gcc 16.1.0, cmake 4.4.3, OQS_USE_OPENSSL=OFF, minimal build restricted to the 30 QR-UOV schemes.

Good news first. All 30 schemes pass the liboqs KAT check and test_sig on ref, opt and avx2, and all three produce identical digests. I also went past liboqs' own oracle, since kats.json here is self-generated: I reconstructed the NIST .rsp layout from kat_sig --all and hashed it against qruov/round3@3d84d0f:test/KAT_SHA256SUMS.txt. That is 100 vectors per scheme, 3000 published Round 3 vectors, all matching, on the avx2 build.

AVX2 really executes, and I checked that under gdb... the ..._avx2_crypto_sign_* breakpoints are hit and the ref ones are not, and single-stepping retires real vpcmpeqd %ymm0,%ymm0,%ymm0 inside matmulfq_vnni_4x16.

Now the part worth your attention, because I think it changes what your own 30/30 covered.

Nothing except ref can be selected through the current option surface. .CMake/alg_support.cmake declares 30 OQS_ENABLE_SIG_qruov_*_avx2 options and zero _opt, while src/sig/qruov/CMakeLists.txt refers to _opt 360 times. So the 30 opt directories and qruov_qruov_opt/ ship and are never compiled. AVX2 is separately gated to Linux|Darwin, and on Apple Silicon the inner x86-64 guard is false. Putting those together, I believe your 30/30 was necessarily ref only, and opt and avx2 had not run anywhere before this. I only reached them by forcing the 30 cache variables on the cmake line, which is a configuration override, not a source change.

One trap before you add the missing options. The wrapper tests #if _opt / #elif _avx2, so enabling both selects opt and drops AVX2. I built it that way to check: opt symbols 5, avx2 0, ref 0. Adding _opt on its own would therefore remove AVX2 from every default Linux build, and no test would notice, because the KATs pass either way. The ordering needs fixing in the same change.

tests/test_kat_all.py fails 30/30 with KeyError: 'all'. All 30 QRUOV entries in tests/KATs/sig/kats.json have only single; all 77 existing sig entries have both. I have the 30 computed all values and can post them or open a PR against your branch, whichever you prefer.

The importer points at a personal fork. copy_from_upstream.yml has git_url: https://github.com/0rlych1kk4/round3.git, pinned to a SHA that is the head of an unmerged PR in that fork. qruov/round3's default branch is main at 3d84d0f and contains neither that SHA nor the de98876 in the PR description. Every other sig family in ALGORITHMS.md points at the upstream project's own repository. I suspect this resolves itself once qruov/round3#1 lands, but as it stands liboqs would be importing from a contributor fork, so it is probably the one thing that has to change before merge.

AES does not go through liboqs in an avx2 build. Measured by call counts over a full kat_sig run: OQS_AES128_CTR_inc_stream_iv 360 on ref and 360 on opt, 0 on avx2, with 720 calls into the vendored x86aesni stream function instead. The CMakeLists passes -DPRIM_AES_BACKEND_X86AESNI=1 alongside -DQRUOV_PRIM_OQS, and prim.c:212 tests the AES-NI macro first. SHAKE goes through liboqs in all three. Since avx2 is the default on Linux x86-64, that is the shipped default path, so it is worth saying explicitly rather than leaving "liboqs primitive integration for SHAKE and AES-CTR" to be read as covering all three.

Smaller things: 72 -Wvla warnings from QR-UOV sources under -Wall -Wextra -Wpedantic -Wvla and zero without it; SPDX headers on 31 of 163 QR-UOV files against 100 percent for uov, mayo and snova; no entries in tests/constant_time/sig/{passes,issues}.json, and qruov_qruov_common/linsys.c branches on echelon->rank and on fastop_dotprod results, which look secret-derived in a signing path, so that is a question for the QR-UOV team rather than a claim from me; and ALGORITHMS.md advertises x86_64 (aes+avx2+bmi2) while a default Windows x86-64 build gets ref only.

Two offers. The KAT validation and the portable non-AVX2 path were my half of the split, so the CI to cover them is mine to write: a matrix over ref, opt and avx2 on x86-64 Linux running the QR-UOV KATs, a guard that fails when a scheme ships an implementation directory no option can enable, and a kats.json completeness check. Every finding above would have been caught by one of those three. Say the word and I will open it against your branch.

Boundaries, so none of this is read as more than it is: one machine, Windows x86-64, gcc only, no clang, no sanitizers and no valgrind, OQS_USE_OPENSSL=OFF, static minimal build. The AVX2 CPU-feature fallback was never exercised because this CPU has all three features. The gdb dispatch check was done on one scheme and the other 29 rest on identical generated wrappers plus KAT agreement. A Linux x86-64 run is still worth having, because there avx2 is reachable through your own options than my override. And I reviewed the integration, not the cryptography: nothing here speaks to QR-UOV's security or to AVX2 correctness outside the KAT set.

0rlych1kk4:
> > Thanks @RodriM11 . I reworked the QR-UOV import to use shared common_dep source groups instead of materializing the full source tree for every parameter set and implementation.
> > The 30 variants are still preserved, but each per-variant implementation directory now contains only the small signature.c wrapper, with common/ref/opt/AVX2 sources shared centrally.
> 
> Thank you a lot @0rlych1kk4 ! May I ask how have you worked around QR-UOV structure? Because they were dependent on the `qruov.h` activating a particular variant based upon compilation macros and that seems to be the case still.
> 
> > The liboqs-specific integration files are maintained in the QR-UOV upstream repository under integration/liboqs/.
> > That directory contains the per-variant metadata, META_Common.yml, namespacing, the OQS_randombytes shim, and the detached-signature wrapper used by the liboqs importer.
> 
> I include the link to your copy of the QR-UOV [repo](https://github.com/0rlych1kk4/round3/tree/feat/liboqs-integration/integration/liboqs) with the inclusion files just to ease the review for other members. Thanks!

@RodriM11  I did not remove the compile-time variant selection in qruov.h. The change is only in how the importer materializes the sources.

The shared common/ref/opt/avx2 sources are declared as include_only common dependencies, so they are stored once in the generated tree but are still compiled separately for each scheme implementation target. Each target therefore receives its own QRUOV_PARAM_*, PRG_IS_AES, namespace prefix, and implementation-specific compile options.
So, for example, 1q31L3-aes and 5q127L10-shake compile the same shared source files with different parameter macros; the source duplication is removed, but QR-UOV’s compile-time selection model is preserved.

The liboqs integration files are here in my QR-UOV branch: integration/liboqs/. This contains the per-variant metadata, META_Common.yml, namespace.h, randombytes.c, and signature.c.

0rlych1kk4:
> Hi @0rlych1kk4 Great work here! I ran this on x86-64. Head [078c4d3](https://github.com/open-quantum-safe/liboqs/commit/078c4d3c73b982de8d7a1665d68073bfdaf48147), Windows 11 x86-64, gcc 16.1.0, cmake 4.4.3, OQS_USE_OPENSSL=OFF, minimal build restricted to the 30 QR-UOV schemes.
> 
> Good news first. All 30 schemes pass the liboqs KAT check and test_sig on ref, opt and avx2, and all three produce identical digests. I also went past liboqs' own oracle, since kats.json here is self-generated: I reconstructed the NIST .rsp layout from kat_sig --all and hashed it against [qruov/round3@3d84d0f](https://github.com/qruov/round3/commit/3d84d0f):test/KAT_SHA256SUMS.txt. That is 100 vectors per scheme, 3000 published Round 3 vectors, all matching, on the avx2 build.
> 
> AVX2 really executes, and I checked that under gdb... the ..._avx2_crypto_sign_* breakpoints are hit and the ref ones are not, and single-stepping retires real vpcmpeqd %ymm0,%ymm0,%ymm0 inside matmulfq_vnni_4x16.
> 
> Now the part worth your attention, because I think it changes what your own 30/30 covered.
> 
> Nothing except ref can be selected through the current option surface. .CMake/alg_support.cmake declares 30 OQS_ENABLE_SIG_qruov_*_avx2 options and zero _opt, while src/sig/qruov/CMakeLists.txt refers to _opt 360 times. So the 30 opt directories and qruov_qruov_opt/ ship and are never compiled. AVX2 is separately gated to Linux|Darwin, and on Apple Silicon the inner x86-64 guard is false. Putting those together, I believe your 30/30 was necessarily ref only, and opt and avx2 had not run anywhere before this. I only reached them by forcing the 30 cache variables on the cmake line, which is a configuration override, not a source change.
> 
> One trap before you add the missing options. The wrapper tests #if _opt / #elif _avx2, so enabling both selects opt and drops AVX2. I built it that way to check: opt symbols 5, avx2 0, ref 0. Adding _opt on its own would therefore remove AVX2 from every default Linux build, and no test would notice, because the KATs pass either way. The ordering needs fixing in the same change.
> 
> tests/test_kat_all.py fails 30/30 with KeyError: 'all'. All 30 QRUOV entries in tests/KATs/sig/kats.json have only single; all 77 existing sig entries have both. I have the 30 computed all values and can post them or open a PR against your branch, whichever you prefer.
> 
> The importer points at a personal fork. copy_from_upstream.yml has git_url: https://github.com/0rlych1kk4/round3.git, pinned to a SHA that is the head of an unmerged PR in that fork. qruov/round3's default branch is main at 3d84d0f and contains neither that SHA nor the de98876 in the PR description. Every other sig family in ALGORITHMS.md points at the upstream project's own repository. I suspect this resolves itself once [qruov/round3#1](https://github.com/qruov/round3/pull/1) lands, but as it stands liboqs would be importing from a contributor fork, so it is probably the one thing that has to change before merge.
> 
> AES does not go through liboqs in an avx2 build. Measured by call counts over a full kat_sig run: OQS_AES128_CTR_inc_stream_iv 360 on ref and 360 on opt, 0 on avx2, with 720 calls into the vendored x86aesni stream function instead. The CMakeLists passes -DPRIM_AES_BACKEND_X86AESNI=1 alongside -DQRUOV_PRIM_OQS, and prim.c:212 tests the AES-NI macro first. SHAKE goes through liboqs in all three. Since avx2 is the default on Linux x86-64, that is the shipped default path, so it is worth saying explicitly rather than leaving "liboqs primitive integration for SHAKE and AES-CTR" to be read as covering all three.
> 
> Smaller things: 72 -Wvla warnings from QR-UOV sources under -Wall -Wextra -Wpedantic -Wvla and zero without it; SPDX headers on 31 of 163 QR-UOV files against 100 percent for uov, mayo and snova; no entries in tests/constant_time/sig/{passes,issues}.json, and qruov_qruov_common/linsys.c branches on echelon->rank and on fastop_dotprod results, which look secret-derived in a signing path, so that is a question for the QR-UOV team rather than a claim from me; and ALGORITHMS.md advertises x86_64 (aes+avx2+bmi2) while a default Windows x86-64 build gets ref only.
> 
> Two offers. The KAT validation and the portable non-AVX2 path were my half of the split, so the CI to cover them is mine to write: a matrix over ref, opt and avx2 on x86-64 Linux running the QR-UOV KATs, a guard that fails when a scheme ships an implementation directory no option can enable, and a kats.json completeness check. Every finding above would have been caught by one of those three. Say the word and I will open it against your branch.
> 
> Boundaries, so none of this is read as more than it is: one machine, Windows x86-64, gcc only, no clang, no sanitizers and no valgrind, OQS_USE_OPENSSL=OFF, static minimal build. The AVX2 CPU-feature fallback was never exercised because this CPU has all three features. The gdb dispatch check was done on one scheme and the other 29 rest on identical generated wrappers plus KAT agreement. A Linux x86-64 run is still worth having, because there avx2 is reachable through your own options than my override. And I reviewed the integration, not the cryptography: nothing here speaks to QR-UOV's security or to AVX2 correctness outside the KAT set.

Hi @paschal533  Thanks for doing the x86-64 sweep and for checking the implementation dispatch directly. This catches a few things my Apple Silicon validation could not cover.
Please go ahead and open the PR against feat/qruov-integration with the CI matrix, kats.json completeness check, and the 30 computed all KAT values.

I’ll handle the integration-side fixes for:
- generating the missing _opt enable options,
- correcting the wrapper selection order so AVX2 is not masked when opt is enabled,
- clarifying that SHAKE uses liboqs across implementations while the AVX2 AES path uses QR-UOV’s AES-NI backend,
- and switching the importer from my fork to the official QR-UOV repository once the upstream integration PR lands.

I’ll also raise the linsys.c/constant-time question with the QR-UOV upstream maintainers rather than treating it as resolved locally.

Thanks again, the independent 30/30 ref/opt/AVX2 KAT validation and the published Round 3 vector comparison are especially useful.

0rlych1kk4:
@paschal533  I followed up on the missing opt option on Apple Silicon.
Making opt the default exposed reproducible SIGBUS crashes in 8/30 variants during test_sig: 3q7L10, 5q7L10, 5q31L10, and 5q127L10, for both AES and SHAKE. The corresponding ref builds pass, including a fresh 30/30 test_sig sweep.

I reverted the default-selection experiment and kept ref as the liboqs default. The final ref-default build also passes all 30 deterministic KATs, and copy_from_upstream.py verify reports 395/395 schemes matching.

So I don't think we should expose opt through normal liboqs selection yet; its ARM64 behavior needs investigation first. AVX2 remains the x86-64 optimized path.

paschal533:
Hi  @0rlych1kk4. I opened the CI side against `feat/qruov-integration`: 0rlych1kk4/liboqs#1. It has the 30 missing `all` digests, a guard that fails when a CMakeLists tests an option nothing declares, and a ref/opt/avx2 matrix that asserts which implementation was actually selected before it builds.

All 30 digests are re-derived on a `ref` build, 30 of 30 with nothing skipped, so they do not depend on the implementation that generated them. The reason recorded for withholding opt cites your ARM64 SIGBUS result.

I am on Windows, so the workflow has not run on Linux yet. Rest is in the PR description

PR #2584, opened 2026-09-21
Do not mark the public API dllexport in Windows static builds

matlimatli:
On Windows, OQS_API expands to __declspec(dllexport) unconditionally, including when liboqs is built as a static library. This will lead to re-exporting the symbols when linking with the static library by having an /EXPORT linker directive included in the .lib file.

This means that a project that links oqs.lib into its own DLL will make the entire libOQS API visible as exported symbols from this DLL, which defeats the point of building the library statically.

Solved by only expanding OQS_API to __declspec(dllexport) when building liboqs as a shared library for Windows, otherwise leave OQS_API empty on Windows.

Verification of fix:
```
cmake -B build-shared -DBUILD_SHARED_LIBS=ON -DOQS_MINIMAL_BUILD="KEM_ml_kem_768;SIG_ml_dsa_44" .
cmake -B build-static -DOQS_MINIMAL_BUILD="KEM_ml_kem_768;SIG_ml_dsa_44" .
cmake --build build-shared --config Release
cmake --build build-static --config Release
dumpbin /directives .\build-static\lib\oqs.lib | findstr /i OQS_SIG_ml_dsa_44_sign
dumpbin /symbols .\build-static\lib\oqs.lib | findstr OQS_SIG_ml_dsa_44_sign
dumpbin /exports .\build-shared\bin\oqs.dll | findstr /i OQS_SIG_ml_dsa_44_sign
```

Without the change, the static lib shows `/EXPORT:OQS_SIG_ml_dsa_44_sign` in the `/directives` output. After the change it does not. The symbol remains in the COFF symbol tables after the change, so linking against it remains unaffected. The shared library does not change.

Both branches are already tested by the CI. `windows.yml` builds without BUILD_SHARED_LIBS, and the windows-dll job in `linux.yml` cross-compiles with BUILD_SHARED_LIBS set.

* [ ] Does this PR change the input/output behaviour of a cryptographic algorithm (i.e., does it change known answer test values)?  -- **No**, only changes symbol visibility
* [ ] Does this PR change the list of algorithms available -- either adding, removing, or renaming? Does this PR otherwise change an API? **No**, only changes symbol visibility

Generative AI reviewed the approach, suggested verification steps and helped with wording. 
I have run the verification steps as mentioned above before and after the change.


dstebila:
Thanks for the contribution, @matlimatli. I've added a unit test to check this; can you take a look and see if it makes sense to you?

matlimatli:
@dstebila Thanks for the test! It makes perfect sense. I pushed a small improvement to the test that also catches the issue when building for win32, where an underscore prefixes the symbol name (`/EXPORT:_OQS_...`). And MinGW would use `-export:` instead of `/EXPORT`.

